PR #32, opened 2017-01-09
X509 certificate support

mp-loki:


transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
```diff
@@ -88,4 +100,54 @@ public SslHandler getSslHandler() {
...
+            for (X509Certificate cert : chain) {
+                try {
+                    String strCert = SslUtil.getX509CertificateString(cert);
+                    String sha3Hash = EncryptionUtil.getSha3Hash(strCert);
+                    DeviceCredentials deviceCredentials = deviceCredentialsService.findDeviceCredentialsByCredentialsId(sha3Hash);
+                    if (deviceCredentials == null) {
+                        throw new CertificateException("Invalid Device Certificate");
+                    }
+                } catch (IOException e) {
+                    e.printStackTrace();
```
ashvayka: please log this using @Slf4j

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
```diff
@@ -88,4 +100,54 @@ public SslHandler getSslHandler() {
...
+        @Override
+        public void checkClientTrusted(X509Certificate[] chain,
+                                       String authType) throws CertificateException {
+            for (X509Certificate cert : chain) {
+                try {
+                    String strCert = SslUtil.getX509CertificateString(cert);
+                    String sha3Hash = EncryptionUtil.getSha3Hash(strCert);
+                    DeviceCredentials deviceCredentials = deviceCredentialsService.findDeviceCredentialsByCredentialsId(sha3Hash);
+                    if (deviceCredentials == null) {
+                        throw new CertificateException("Invalid Device Certificate");
```
ashvayka: What is the purpose of loop by all certificates in chain if you throw exception here?

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
```diff
@@ -88,4 +100,54 @@ public SslHandler getSslHandler() {
...
 
+    private TrustManager getX509TrustManager(TrustManagerFactory tmf) throws Exception {
+        X509TrustManager x509Tm = null;
+        for (TrustManager tm : tmf.getTrustManagers()) {
+            if (tm instanceof X509TrustManager) {
+                x509Tm = (X509TrustManager) tm;
+                break;
+            }
+        }
+        X509TrustManager x509TmWrapper = new ThingsboardMqttX509TrustManager(x509Tm, deviceCredentialsService);
```
ashvayka: return new ThingsboardMqttX509TrustManager(x509Tm, deviceCredentialsService);

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
```diff
@@ -71,13 +82,14 @@ public SslHandler getSslHandler() {
...
-            TrustManager[] tm = tmFactory.getTrustManagers();
+            TrustManager x509wrapped = getX509TrustManager(tmFactory);
+            TrustManager[] tm = {x509wrapped};
             SSLContext sslContext = SSLContext.getInstance(TLS);
             sslContext.init(km, tm, null);
             SSLEngine sslEngine = sslContext.createSSLEngine();
             sslEngine.setUseClientMode(false);
             sslEngine.setNeedClientAuth(false);
-            sslEngine.setWantClientAuth(false);
+            sslEngine.setWantClientAuth(true);
```
ashvayka: let's make this configurable. There are three options:
1) Using Access Token on unsecure connection (mqtt.ssl.key-store is not set)
2) Using Access Token over secure connection (mqtt.ssl.key-store is set, but mqtt.ssl.client-auth=false)
3) Using X.509 certificates over secure connection (mqtt.ssl.key-store is set and mqtt.ssl.client-auth=true)

mp-loki: 1 is already in place (this bean won't be instantiated). 2-3 I'll move that to the property file

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
mp-loki: Normally there will be only one certificate, then chain[0] will be enough. Is there a use case where client will send more than one cert?

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
mp-loki: sorry :(

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
mp-loki: I think it makes sense to make both WantClientAuth and NeedClientAuth configurable - 
If you want to disable AUTH_TOKEN, you would .setNeedClientAuth(true)

PR #38, opened 2017-01-21
X 509

mp-loki:


dao/src/main/java/org/thingsboard/server/dao/EncryptionUtil.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.thingsboard.server.dao;
 
+import com.google.common.base.CharMatcher;
```
ashvayka: This import is not used, right?

mp-loki: yes

PR #79, opened 2017-03-16
Change .useSanitizeValueStrategy parameter to 'sce' to solve the disp…

walter-c:
Change $useSanitizeValueStrategy parameter to 'sce' to solve the displaying issue in some languages.
The related issue:
https://github.com/angular-translate/angular-translate/issues/1101


ashvayka:
Thank you, @walter-c !

PR #81, opened 2017-03-17
add locale resource for Korean

OutOfBedlam:
Adding Locale Resource for Korean.
It still needs additional changes on "ui" that allow end-users to choose their language but guess it is meaningful as a starting point adding multi-languages.


ashvayka:
Thank you! We will also modify the UI to be able to switch languages in next releases.

huily6:
Can the device add a recent historical data to export?

PR #84, opened 2017-03-19
exception check (for IE)

OutOfBedlam:
MS Internet Explorer compatibility


ashvayka:
Thanks, we will double check and merge this week.

PR #140, opened 2017-05-18
Fixed some translation errors

junehappylove:
Fixed some translation errors


junehappylove:
Fixed some translation errors about chinese

junehappylove:
ok

PR #189, opened 2017-06-29
Added deviceType and deviceName update.

volodymyr-babak:
Merged sql.enabled and cassandra.enabled into single property - database.type.


extensions-api/src/main/java/org/thingsboard/server/extensions/api/device/DeviceNameOrTypeUpdateMsg.java
```diff
@@ -0,0 +1,44 @@
...
+
+import lombok.Getter;
+import lombok.ToString;
+import org.thingsboard.server.common.data.id.DeviceId;
+import org.thingsboard.server.common.data.id.TenantId;
+
+/**
+ * @author Andrew Shvayka
+ */
+@ToString
```
ashvayka: If you use @Data it will automatically create all args constructor

extensions-api/src/main/java/org/thingsboard/server/extensions/api/device/DeviceNameOrTypeUpdateMsg.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.thingsboard.server.extensions.api.device;
+
+import lombok.Getter;
+import lombok.ToString;
+import org.thingsboard.server.common.data.id.DeviceId;
+import org.thingsboard.server.common.data.id.TenantId;
+
+/**
+ * @author Andrew Shvayka
```
ashvayka: I am not the author of this class :-)

ashvayka:
Few minor comments, otherwise LGTM

extensions-api/src/main/java/org/thingsboard/server/extensions/api/device/DeviceNameOrTypeUpdateMsg.java
volodymyr-babak: I'll use @AllArgsConstructor, because @Data will create setters as well for immutable msg object

extensions-api/src/main/java/org/thingsboard/server/extensions/api/device/DeviceNameOrTypeUpdateMsg.java
volodymyr-babak: copy paste...

extensions-api/src/main/java/org/thingsboard/server/extensions/api/device/DeviceNameOrTypeUpdateMsg.java
ashvayka: @Setter is created only for all non-final fields

PR #217, opened 2017-07-24
Windows installer fix.

mp-loki:
Currently in the Windows installer package there is winsw config file - thingsboard.xml and conf/thingsboard.yaml file. When the ThingsboardInstallApplication is started, it passes DEFAULT_SPRING_CONFIG_PARAM to SpringApplication class. SpringApplication starts to look for the Spring configuration file named 'thingsboard' (with no extension), finds thingsboard.xml and tries to parse it as Spring xml configuration, and fails obviously. 
The fix proposed makes Spring look for configuration in conf/ directory and avoids collision


ashvayka:
I have fixed this by modifying the script: https://github.com/thingsboard/thingsboard/blob/master/application/src/main/scripts/windows/install.bat#L85
If my fix works for you I would like to leave it as-is. Please confirm

mp-loki:
Yep, your fix works. 
Closing

PR #248, opened 2017-08-16
Fix for error during customer description update

ssmaurya:
In version 1.3.0, if we update customer entity description or any other field and do a save, it gives error saying customer title already exists. This fix addresses the issue.


ikulikov:
@ssmaurya thanks for your contribution!

PR #295, opened 2017-09-16
Plugin: REST API - https:// support

azrin1972:
REST API now support https or any other protocol handler typed in the host input box


azrin1972:
not sure why it detects that I'm not a git user. Already has my email in my account but still it detects me as a non GIT user.

volodymyr-babak:
@azrin1972 thanks for your help on finding this bug and providing fix for it. 
Unfortunately we can't accept PR that are not signed properly, but we need this to be fixed to be able to send REST call to https endpoints and that's why it was implemented by other developer. 
But still we really appreciate your input on this and we hope that in the future your input is going to be growing into project.

 Thanks! Sincerely, ThingsBoard team!

azrin1972:
No worries. Just please help me to use thingsboard efficiently by answering questions in google groups or gitter. 

Thanks

PR #333, opened 2017-10-07
Centralize checking for whether customer is public

eykamp:


ashvayka:
Thank you, we will review it with @ikulikov

ashvayka:
@ikulikov LGTM,what do you think?

ikulikov:
LGTM

PR #363, opened 2017-10-21
add  "Deny device creation via gateway" option for gateway

liaow:
- gateway can not create device on connect if name not exists when "Deny device creation via gateway" option checked.
- gateway can create device on connect if name not exists for default .
- to define a gateway device, tenant user need to check "is gateway" and "Deny device creation via gateway" option.

wx20171024-110305 2x

### DIFF
--- a/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
+++ b/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
@@ -90,11 +90,8 @@ public class GatewaySessionCtx {
         if (!devices.containsKey(deviceName)) {
             Optional<Device> deviceOpt = deviceService.findDeviceByTenantIdAndName(gateway.getTenantId(), deviceName);
             Device device = deviceOpt.orElseGet(() -> {
-                JsonNode infoNode = gateway.getAdditionalInfo();
-                if (infoNode == null)
-                    return null;
-                JsonNode allowCreateDevice = infoNode.get("allow_create_device");
-                if (allowCreateDevice == null || !allowCreateDevice.asBoolean())
+                JsonNode forbidCreateDevice = gateway.getAdditionalInfo().get("forbid_create_device");
+                if (forbidCreateDevice != null && forbidCreateDevice.asBoolean())
                     return null;
                 Device newDevice = new Device();
                 newDevice.setTenantId(gateway.getTenantId());
--- a/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
+++ b/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
@@ -15,6 +15,7 @@
  */
 package org.thingsboard.server.transport.mqtt.session;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.google.gson.JsonArray;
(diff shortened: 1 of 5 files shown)


ashvayka:
@liaow Thank you. I will review this and will get back to you with some tips to improve performance and error handling. stay tuned

ssmaurya:
Few more things to be added: filter gateway devices so that we could see all the devices from particular gateway. Plus add option for allowing or blocking every gateway device.

ssmaurya:
Default for gateway devices could be allowed.

liaow:
@ssmaurya How should I deal these details?

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
```diff
@@ -89,6 +90,9 @@ private void onDeviceConnect(String deviceName, String deviceType) {
         if (!devices.containsKey(deviceName)) {
             Optional<Device> deviceOpt = deviceService.findDeviceByTenantIdAndName(gateway.getTenantId(), deviceName);
             Device device = deviceOpt.orElseGet(() -> {
+                JsonNode forbidCreateDevice = gateway.getAdditionalInfo().get("forbid_create_device");
```
ashvayka: please rename to "allowDeviceCreation" to be camel case. (The boolean condition need to be reversed).

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
```diff
@@ -89,6 +90,9 @@ private void onDeviceConnect(String deviceName, String deviceType) {
...
+                if (forbidCreateDevice != null && forbidCreateDevice.asBoolean())
```
ashvayka: please use bracers {} in if statements and in blocks

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
ashvayka: This boolean flag should be fetched. No need to get and calculate this each time the device sends a request.

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
```diff
@@ -97,6 +101,10 @@ private void onDeviceConnect(String deviceName, String deviceType) {
                 relationService.saveRelationAsync(new EntityRelation(gateway.getId(), newDevice.getId(), "Created"));
                 return newDevice;
             });
+            if (null == device) {
+                log.debug("[{}] NOT allowed add to add device [{}] to the gateway session", gatewaySessionId, deviceName);
```
ashvayka: please change this to warning. Also, the data upload request will be failed, but gateway can't know if it is doing something wrong. So, the best idea is to throw RuntimeException or AdaptorException and close gateway session in the MqttTransportHandler.handleMqttPublishMsg

transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/GatewaySessionCtx.java
liaow: sorry, it is hard to find a suitable place to set this attribute for gateway device.

Issue #394, opened 2017-11-03
DeviceActorMessageProcessor resending RPC calls on subsequent Device Connects


steves2j:
When sending an RPC call to an offline device, the Call is cached in the rpcPendingMap which has a key of integer.

When the Device comes online, the pending RPC call is successfully send, however the rpcPendingMap does not have the entry removed due to an incorrect search of the key to remove.

A interim set is used to store the successfully send id's however this is a SUID set not an integer set.

To fix, change this set to Integer and push the entry.getKey() instead of the value.getMsg().getMsg().getId() object in the processPendingRpc method.

ashvayka:
Do you want to submit PR?

ashvayka:
Thank you for the report

vparomskiy:
Issue already resolved.

PR #399, opened 2017-11-06
Fix for Defect #394

steves2j:
Changed the RPC delete set from UUID to Integer to match what is stored as a key in the RPCPending Map


ashvayka:
@steves2j please sign the CLA using the online tool above

ashvayka:
@steves2j Thank you, we will review and merge.

PR #400, opened 2017-11-06
Allow public devices on private dash

### LINKED ISSUE #278: Error viewing private dashboard with public device
I created a Customer "CUST" and a device "DEV", which I made public.  I also created a dashboard for CUST called "DASH".  

When I am logged in as the tenant administrator, I can see DEV on DASH just fine.  But when I create a user for CUST, and login as that user, I cannot get DASH to display info about DEV.  Some of the widgets are displayed with the message "Widget Error: TypeError: Cannot read property '0' of undefined".

I am guessing there is a permissions issue when users view devices assigned to the public user.

The goal I am trying to achieve is to create a series of users, each with one or more devices, each of whom can see a detailed dash with their devices.  I also want to create a publicly available map of all devices, to let all users browse through each other's data.

eykamp:
Allow devices owned by a public user be added to private dashboards.  Fixes #278.

I added an @Ignore annotation to three tests which my patch causes to fail.  I do not @Ignore tests lightly, but I have spent nearly 3 full weeks trying to resolve these errors, and I cannot afford to spend more.  I suspect they are easily fixed by someone who knows the codebase better than I do.

In order to determine if a device is owned by a public customer, it is necessary to load the customer object and look either at the customer name or the attributes.  This requires a database call, which requires that the database be initialized, and I cannot figure out how to do that in these tests. I am reasonably confident it is a test artifact, rather than a problem that would be encountered in production.

I would feel much more comfortable if someone could help me fix the broken tests and remove the @Ignore annotations before merging this PR into the main codebase.

The extra database call could be avoided if, instead of using a magic customer name ("public"), we used a magic customer UUID.  In that event, the customer could be determined to be public or not directly from the device.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/plugin/PluginProcessingContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/plugin/PluginProcessingContext.java
@@ -273,6 +273,10 @@ public final class PluginProcessingContext implements PluginContext {
 
     private boolean isPublicCustomer(CustomerId customerId) {
         Customer customer = pluginCtx.systemContext.getCustomerService().findCustomerById(customerId);
+
+        if (customer == null)
+            return false;
+
         return customer.isPublic();
     }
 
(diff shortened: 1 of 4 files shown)


ashvayka:
Closing due to inactivity and merge conflicts

PR #412, opened 2017-11-11
Fix intervals

eykamp:
Fix times to match what is documented.  Both off by a factor of 10,000.  I'm guessing this was not intended.


ashvayka:
Thank you!

PR #414, opened 2017-11-14
Witsml depth data support

Himanshu-it:
As per the present code, depth data (double) can be posted but not stored in the DB via mqtt. The control returns back at the start of dao layer i.e in function:
public void handleTelemetryUploadRequest(PluginContext ctx, TenantId tenantId, RuleId ruleId, TelemetryUploadRequestForDepthRuleToPluginMsg msg)

in file: TelemetryRuleMsgHandler.java

1. The mqtt topic name is : v1/devices/mewitsml/telemetry
2. depth data sample : [{"ds":2000.1,"values":{"temperature":21, "humidity":55.0, "active": false}}]


ashvayka:
:1st_place_medal:

PR #448, opened 2017-11-23
Multiple changes

steves2j:
1. We added caching of Shared Scoped Attributes in a similar way as RPC caches request when a device is offline, so that when the device comes online again, all the changed Shared Scoped Attributes are send to the device.

2. We've also added the feature that Server Scoped Attributes are appended to the Telemetry data so that any Telemetry subscriptions that need to be processed by a Post Function, now also have access to the Server Attributes so that additional calculations can be made to the data displayed based on the Telemetry data and server attributes.

3. We've updated the Device panel to show the device description and also make this device description available to the entity table widget.

4. We have bug fixed an issue with the Gateway server where if a message is sent in without a reply too, then the message id will be missing. So instead of raising an exception, there is a check and a log message instead.


ashvayka:
This is a significant change and we will take some time to review. Is it possible to make 3 smaller PRs (one for items (1) and (2). Second for item (3) and third for item (4))? This will speed up review and merge process.

ashvayka:
@steves2j Forgot to say thanks. This is awesome!

ashvayka:
Hi @steves2j, please provide comments

steves2j:
Doubtful, as we run two Git Repositories, The forked one from thingsboard and a private one. Most to all our commits are done on the private one and then a select few are pushed back into the forked one for merge.

So this time round I would say we will be hard pressed to make 3 smaller PR as we just don't have the time atm being End of year and mad rush for the holidays. We will endeavor to make any future changes as individual PR's.

affoltep:
It would be very useful to have this pull request integrated in V1.4.0. Especially point "3" would be great

steves2j:
Closing this pull request due to non adoption by the authors. If you want this for version 1.4 visit my git site. New pull request will be made with changes for version 2.x soon.

PR #480, opened 2017-12-14
added some funcs to rest client

dmytro-landiak:


tools/src/main/java/org/thingsboard/client/tools/RestClient.java
```diff
@@ -71,18 +78,45 @@ public void login(String username, String password) {
...
     }
 
+    public Asset createAsset(String name, String type) {
+        Asset asset = new Asset();
+        asset.setName(name);
+        asset.setType(type);
+        return restTemplate.postForEntity(baseURL + "/api/asset", asset, Asset.class).getBody();
+    }
+
+    public Alarm createAlarm(String type, AlarmSeverity severity, AlarmStatus status, EntityId originator) {
```
ashvayka: It is better to write method
 public Alarm createAlarm(Alarm alarm)

dmytro-landiak: okey!

PR #543, opened 2018-01-23
For Rules with no processor and sendEmailActions, not evaluating template

MohamedSalahSayed:
For Rules with no processor and sendEmailActions, the device meta data was sent empty. Velocity evaluation wasnot happening.
The change has very small impact on the performance as it is activated
only incase of SendEmailAction and No processor rule.


ashvayka:
We are refactoring rule engine now and hope this will not be needed any more.

ashvayka:
Closing due to inactivity and merge conflicts and it is now replaced by new rule engine.

PR #641, opened 2018-03-09
opc fix

dmytro-landiak:


ui/src/app/extension/extensions-forms/extension-form-opc.tpl.html
```diff
@@ -212,8 +212,8 @@
                                                     </md-input-container>
 
                                                     <section class="dropdown-section">
-                                                        <div class="tb-container" ng-class="{'ng-invalid':!server.keystore.file}">
-                                                            <span ng-init='fieldsToFill = {"fileName":"fileName", "file":"file"}'></span>
+                                                        <div class="tb-container" ng-class="{'ng-invalid':!server.keystore.location}">
+                                                            <span ng-init='fieldsToFill = {"fileName":"fileName", "location":"location"}'></span>
```
mp-loki: Can we keep using 'location' for the file name (as it was before) and 'fileContent' for file content?

dmytro-landiak: yeah, will be changed.

PR #693, opened 2018-03-30
Add Russian translations

asnik:


ashvayka:
please validate your doc. I think there is a simple typo somewhere.

ashvayka:
Please fix the localization changes due to new structure (we use json instead of js now)

PR #783, opened 2018-05-16
Add missed import statements!

jacklicn:


ashvayka:
Thank you! We were not noticing this but I hope it helps in some corner cases. Maybe we had node_modules cached.

PR #798, opened 2018-05-21
Add Chinese translation for audit-log.

jacklicn:


untilbit:
@jacklicn The English file has more strings if you wish to translate them.

PR #831, opened 2018-06-03
l10n improvements

iglilaw:
- Removed plugin, rule & rule plugin strings
- Added missing strings


ikulikov:
@iglilaw thanks for your contribution.

PR #834, opened 2018-06-04
Fixed couldn't get shared attributes.

lonwern:


ikulikov:
@lonwern thanks for your contribution.

PR #853, opened 2018-06-08
A device created by gateway should be auto assigned to the customer that the gateway is assigned to

lucky2you:
The device connect to the gateway will be auto created by the thingsboard. If a gateway is assigned to a customer, the auto created device should be auto assigned to the customer too.


ashvayka:
Hi, @lucky2you, this is a good idea!

PR #872, opened 2018-06-14
Changes appeared to be needed while transitioning from v1.3 to v2.0.2.

stanurkov:
Checking for null was missing in new Rule Engine - perhaps deeper investigation is also needed on why a new Root Rule was not created during transition.
 
Default constructor was causing errors when using Cassandra, so no dashboards could be assigned to customers (reading for customer's list JSON was failing with an exception).


cannium:
We also met this problem, please fix ASAP.

PR #873, opened 2018-06-14
Feature/relation server attributes inheritance

steves2j:
When an Asset or other device is a parent of a child device, the parent will have Server defined attributes that would be nice for a child to inherit.
This MR allows that,
server attributes so that when you open a device and view the server attributes you can see the inherited attributes that have been supplied from the parent via the to relation
relation to

The related attributes and other scope type attribute (Client,Server,Shared,Server Inherited) are now also shown with an idetifier in the auto complete to help identify which are which.
attribute list

Attributes are inherited from all parents and grand parents etc, up to 10 levels (Restricted for performance reasons and in case there are incestial relations) and a relation can be stopped from being inherited with the inclusion of noinherit in the additional info text area of the relation.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
+++ b/application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
@@ -319,9 +319,9 @@ public class DeviceActorMessageProcessor extends AbstractContextAwareMsgProcesso
     private ListenableFuture<List<AttributeKvEntry>> getAttributeKvEntries(DeviceId deviceId, String scope, Optional<Set<String>> names) {
         if (names.isPresent()) {
             if (!names.get().isEmpty()) {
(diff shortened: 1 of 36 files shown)


ashvayka:
HI @steves2j . This is an awesome contribution but I will need some time to review the implementation. Stay tuned.

ashvayka:
@steves2j please fix the conflicts

steves2j:
@ashvayka So I've merged changes to the latest version. However, I need to understand if this is actually going to get merged into the master or not. If it is when would the likelihood be.

I've spent a lot of time adding this feature but since nearly 3/4 of a year has passed and it has not been merged and adding a separate feature branch and merging into git takes time, I need to understand if it worth investing my time to do this or if feature will not be used. And I should just keep my version in my GIT repo. and you can pick what you need when you have time.

I'm starting on a new change to the rule-engine and rather that it be done with my merged changes then on the baseline.

Can you let me know what your intentions are please?

Steve

PR #880, opened 2018-06-19
Change RequestMapping value of getTenantDevice

lonwern:
RequestMapping value is duplicated between getTenantDevices and getTenantDevice, so swagger couldn't export both of them.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/controller/AssetController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/AssetController.java
@@ -230,7 +230,7 @@ public class AssetController extends BaseController {
     }
 
     @PreAuthorize("hasAuthority('TENANT_ADMIN')")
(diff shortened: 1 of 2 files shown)


ashvayka:
@lonwern This will probably break UI as well? Please confirm/fix

lonwern:
@ashvayka getTenantDevice and getTenantAsset has not yet been used in UI, it only use getTenantDevices and getTenantAssets for now.

https://github.com/thingsboard/thingsboard/blob/2b2cd99110e7b535e63f9f4c1d0c008690de7864/ui/src/app/api/device.service.js#L32

https://github.com/thingsboard/thingsboard/blob/2b2cd99110e7b535e63f9f4c1d0c008690de7864/ui/src/app/api/asset.service.js#L31

ashvayka:
HI @lonwern . To ensure backward compatibility, please add both request mappings to the method. We can't change the API without major change of version.

PR #881, opened 2018-06-20
Introduce partial Italian translation

mcicolella:


ikulikov:
@mcicolella Thanks for your contribution.

PR #919, opened 2018-07-06
RpcBroadcastMsg never arrives at other cluster nodes

janbols:
When firing an RpcBroadcastMsg message or a ToAllNodesMsg from one node, the other nodes never get notified because the messages doesn't have a serverAddress.
As a result firing such a method is not forwarded to the other cluster nodes.
This PR adds the serverAddress for each node that the message is sent to so the other nodes get the message and can react to component lifecycle changes.


ashvayka:
:+1: Thank you!

PR #948, opened 2018-07-19
Cache assets

janbols:
Cache assets when finding them by tenant and name just like it's done when caching of devices.
For this the method signature of findAssetByTenantIdAndName was changed. It now returns an `Asset` or `null` instead of an `Option<Asset>`.


ashvayka:
:+1:

PR #977, opened 2018-08-01
Add numeric input widget

nabil-rejeb:
hi @ashvayka 
Added a card widget to modify numeric attributes from dashboard (system-data.cql)


ashvayka:
Hi @nabil-rejeb~ The file you have used is now only for tests. You should put your widget bundle or updated widget bundle here: application/src/main/data/json/system/widget_bundles

nabil-rejeb:
Hi @ashvayka , 
Added : 

-  amchart based chart with static or attribute based range, and graph/data export,

- SVG based clock widget,

- Data input widget bundle (Email ,Numeric values, 8 digit phone number (tunisian model)) with optional data change log (key: log) as telemetry value.  

ToDo : improve dynamic font resize

nabil-rejeb:
fix issue #985
Missing telemetry or null value causes exception in "Originator attribute". posted by @mfalkvidd

ashvayka:
Hi @nabil-rejeb . Thank you for your contribution. We will review and merge in next release. We will do some cleanup as well. One thing I must say is that TbAbstractGetAttributesNode.java behaviour is intentional, because in your case there is no way to figure out if attribute was missing.

PR #982, opened 2018-08-03
Fixed java.lang.IllegalArgumentException: Property 'searchTextSource' is not writable, and Hibernate: column does not exist, When use latest spring stack and datastax driver.

jacklicn:
Fixed java.lang.IllegalArgumentException: Property 'searchTextSource' is not writable, and Hibernate: column does not exist, When use latest spring stack and datastax driver.


ashvayka:
Hi @jacklicn What version of java driver and spring you are using? Are you able to commit those changes as well?

jacklicn:
@ ashvayka ，spring boot 2.0.3，spring framework 5.0.7，datastax cassandra 3.5.1.

PR #1054, opened 2018-09-06
Add stylelint

tagliala:
Hi 👋 

This is my first PR to this project and I'm new to Thingsboard

I know that this is a huge change. You may be interested in such a feature to enforce a uniform coding style on the front-end side.

There *shouldn't* be breaking changes, the most "dangerous" change was been removing a few empty css blocks here and there

Ref: https://github.com/stylelint/stylelint


ikulikov:
@tagliala thanks for your contribution!

tagliala:
@ikulikov is there a list of the browsers supported by ThingsBoard UI?

Issue #1066, opened 2018-09-11
Browser support


tagliala:
Hi,

Sorry, I cannot find this information in the documentation

Which browsers are officially supported by the ThingsBoard front-end?

vparomskiy:
The main browser for development - Chrome. 
But in Latest Firefox, IE 11, Edge I think it also should be stable.

tagliala:
@vparomskiy thanks for the answer.

I'm the author of PR #1054. I was interested in adding autoprefixer support and drop vendor prefixes from the SCSS files, but in order to do that I need a list of supported browsers

ikulikov:
ThingsBoard Front-End is based on AngularJS v.1.5.8 and [AngularJS Material v.1.1.1](https://github.com/angular/material). 
TB Front-End is compatible with all browsers supported by AngularJS Material.

From the AngularJS Material [official documentation](https://github.com/angular/material):

> AngularJS Material is targeted for the browser versions defined in the broswerslist field of our package.json. Below is a screenshot from browserl.ist that provides a visual representation of this configuration

MarsWarrior:
Are there any plans to upgrade the AngularJS version used?

The used version has some vulnerabilities according to this site: https://snyk.io/test/npm/angular/1.5.8

We did test an upgrade, but newer Angular versions (1.6/1.7) have breaking changes...

tagliala:
@MarsWarrior 

I'm new to ThingsBoard. As far as I can tell, [some widget](https://github.com/thingsboard/thingsboard/blob/master/ui/src/app/widget/lib/timeseries-table-widget.tpl.html) [use angular](https://github.com/thingsboard/thingsboard/blob/master/ui/src/app/widget/lib/alarm-status-filter-panel.tpl.html) and I can't find tests for the front-end

So, I think that third party plugins could be affected by an Angular upgrade and I would conclude that it is not safe to upgrade.

Ultrazombie:
closed as not active

PR #1074, opened 2018-09-12
Add autoprefixer

tagliala:
Hi again! 👋 

According to https://github.com/thingsboard/thingsboard/issues/1066#issuecomment-420306111

> ThingsBoard Front-End is based on AngularJS v.1.5.8 and [AngularJS Material v.1.1.1](https://github.com/angular/material).
> TB Front-End is compatible with all browsers supported by AngularJS Material.

I've checked AngularJS Material v.1.1.1's Readme at https://github.com/angular/material/tree/v1.1.1 and I can see:

> Angular Material is targeted for all browsers with versions n-1; where n is the current browser version.

Since version 1.1.1 has been released on Sep 2016, I should have checked the latest versions of those browsers at that date and adjust autoprefixer accordingly... but I didn't

I'm using the `browserslist` from the latest version.

#### Could this PR represent a breaking change?
Yes, it could break some style on older Konqueror, Opera and browsers below < .5%

#### Will that breaking change have impact on the functionality of the front-end?
As far as I can tell, it could. There are some 

#### Does this PR fixes some issues?
Yes, it does. There were rules without vendor prefixes that wouldn't have effect on old supported versions of chrome and safari:
- https://github.com/thingsboard/thingsboard/blob/4ba34513a402b80bab6d91e4998e409a5875e331/ui/src/app/dashboard/dashboard-settings.scss#L51

If it could be useful, there is a gist with the old compiled style and the new one. Take a look at the revisions
https://gist.github.com/tagliala/309bdf275ba9da9bd09645fa9c03e162/revisions

### DIFF
--- a/ui/src/app/dashboard/dashboard-card.scss
+++ b/ui/src/app/dashboard/dashboard-card.scss
@@ -16,12 +16,12 @@
 
 .tb-dashboard-assigned-customers {
   display: block;
-  display: -webkit-box;
+  display: -webkit-box; /* stylelint-disable-line value-no-vendor-prefix */
   height: 34px;
   margin-bottom: 4px;
   overflow: hidden;
   text-overflow: ellipsis;
   -webkit-line-clamp: 2;
-  -webkit-box-orient: vertical;
+  -webkit-box-orient: vertical; /* stylelint-disable-line property-no-vendor-prefix */
 }
 
--- a/ui/.stylelintrc
+++ b/ui/.stylelintrc
@@ -251,7 +251,7 @@
       "fill",
       "stroke"
     ],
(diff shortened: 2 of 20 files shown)


ui/src/app/dashboard/dashboard-card.scss
```diff
@@ -16,12 +16,12 @@
 
 .tb-dashboard-assigned-customers {
   display: block;
-  display: -webkit-box;
+  display: -webkit-box; /* stylelint-disable-line value-no-vendor-prefix */
```
tagliala: Why is this using flexbox on webkit based browsers?

PR #1094, opened 2018-09-21
Entity view

ViktorBasanets:


dao/src/main/java/org/thingsboard/server/dao/entityview/EntityViewServiceImpl.java
```diff
@@ -102,14 +102,16 @@ public EntityView saveEntityView(EntityView entityView) {
         log.trace("Executing save entity view [{}]", entityView);
         entityViewValidator.validate(entityView);
         EntityView savedEntityView = entityViewDao.save(entityView);
-        copyAttributesFromEntityToEntityView(savedEntityView, DataConstants.CLIENT_SCOPE, savedEntityView.getKeys().getAttributes().getCs());
-        copyAttributesFromEntityToEntityView(savedEntityView, DataConstants.SERVER_SCOPE, savedEntityView.getKeys().getAttributes().getSs());
-        copyAttributesFromEntityToEntityView(savedEntityView, DataConstants.SHARED_SCOPE, savedEntityView.getKeys().getAttributes().getSh());
+        if (savedEntityView.getKeys() != null) {
+            copyAttributesFromEntityToEntityView(savedEntityView, DataConstants.CLIENT_SCOPE, savedEntityView.getKeys().getAttributes().getCs());
```
ashvayka: This is async operation and it may cause issues in the future. Please consider collecting all the futures and invoking get to make sure that the client call that creates the view will return the successful result only when all attributes are copied.

application/src/test/java/org/thingsboard/server/controller/BaseEntityViewControllerTest.java
```diff
@@ -308,17 +315,66 @@ public void testGetTenantEntityViewsByName() throws Exception {
...
+        options.setUserName(accessToken);
+        client.connect(options);
+        Thread.sleep(3000);
+
+        MqttMessage message = new MqttMessage();
+        message.setPayload(("{\"caValue1\":\"value1\", \"caValue2\":true, \"caValue3\":42.0, \"caValue4\":73}").getBytes());
+        client.publish("v1/devices/me/attributes", message);
+        Thread.sleep(1000);
+
+        List<String> actualTelemetry =
```
ashvayka: actualAttributes instead of actualTelemetry would improve the "readability" of the test

PR #1141, opened 2018-10-11
Add Multiple Customer Assignment to Asset and Device Entities

sonersoysal:
Hi @ashvayka ,

We build a PSIM product using Thingsboard and one of the requirements is to enable monitoring a device by more than one customer group. 

One possible solution would be to use rule engine to duplicate devices, attributes, telemetry data and all of the other relations with other entities. However, it would reduce the maintainability/usability of the system. So, we implemented this feature by using the exact same methods used in the DashboardInfo entities with multiple customer assignment. 

Two interfaces, HasMultipleCustomers and MultipleCustomerAssignmentEntity, provided a good abstraction and Relations are stored in the database like Dashboards with multiple customer groups. 

One problem is to integrate this feature with rule engine nodes, TbChangeOriginatorNode and TbGetCustomerAttributeNode. However, in the description of TbChangeOriginatorNode the following is stated: "_Related Entity found using configured relation direction and RelationType. If multiple Related Entities are found, only first Entity is used as new Originator, other entities are discarded_". Hence, only the first customerId element is returned. I believe this is not the best solution and it requires refactoring. 

BR,
Soner


ashvayka:
Hi @sonersoysal . This is a major change and will take some time to review. However, the similar item you have described was just recently implemented as Entity View feature in TB CE v2.1.2. This feature allows you to assign different device/asset views to different customers, and limit what telemetry customer can see over time and by keys. Please review this feature and get back to me with your feedback. I hope we can apply your efforts to improve EntityViews instead of hosting two similar features in one platform

ashvayka:
Hi @sonersoysal  I will need to close this due to inactivity from your side.

robertoenr:
Hi @sonersoysal, is there any updates about this pull? This is an interesting feature. I'm trying to implement now (TB 3.1.1) following your commits, but some changes does not match. Thanks in advance!

PR #1142, opened 2018-10-11
fix bug get shared config for gateway

Maddin-619:
If `remoteConfiguration: true` in gateway config and a extension config in thingsboard backend is pressent when gateway service starts it will request the shard config form the backend. This always fails because the request message gets not correctly parsed. The json entry key is `shared` and not `sharedKeys`.


ikulikov:
This shoud be fixed in the gateway repository: https://github.com/thingsboard/thingsboard-gateway/blob/master/src/main/java/org/thingsboard/gateway/service/gateway/MqttGatewayService.java#L665

PR #1186, opened 2018-10-25
Add ability to setup interval through env vars

satblip:
Small fix to be able to setup intervals through environment variables


ashvayka:
Thanks!

PR #1191, opened 2018-10-26
Zookeeper reconnect error

ymucahit:
Hello @ashvayka,
In Release-2.0, tb application can't understand whether its' zNode is removed somehow while the tb server is still up and running. I found out some bugs while playing with cluster setup on kubernetes. I list down here some drawbacks of current implementation;

- Sometimes zookeeper can be unreachable for some very small time. (Mostly after zookeeper cluster restarts.) In these situations, tb application creates multiple ephemeral sequential zNodes for the current server without knowing there is already a record for itself.

- If you delete a zNode manually from zk for the tb server, it simply ignores it. Newly added tb applications wouldn't be aware of the deleted tb server. With this commit it is now able to understand its' zNode is removed and insert a new one. 

- If tb is reconnected to zk server, it was trying to add itself twice because sometimes zk server throws CHILD_REMOVED event after cluster restart. After making publishCurrentServer method synchronized, listener for reconnection or the childEvent method is checking whether current server is recorded on zk.

Hope this makes sense.

Best,

Yusuf


ashvayka:
Thank you for the contribution! We plan to make this part of TB v2.2!

PR #1192, opened 2018-10-26
Subscription scope fix

ymucahit:
Finally i figured out the reason why i can see the telemetry data on latest telemetry tab but not on a widget on tb node where actor is not created, which is explained in the issue https://github.com/thingsboard/thingsboard/issues/978

Reason is that on latest telemetry tab it doesn't use datasource.service.js but telemetry-websocket.service.js. On telemetry-websocket.service.js it sets scope attribute on subscription commands, however it is forgotten on datasource.service.js. It is important because in DefaultTelemetrySubscriptionService class on line 621 it sets the scope of the subscription command to ClusterAPIProtos builder scope. This setter throws an exception which is not caught if parameter is null. 
This fix solves the problem and now telemetry data is visible on widgets on all tb servers.

Best regards,

Yusuf


ashvayka:
Hi @ymucahit . Thank you for your research. I have added protection on server side recently, so there is no need to modify the UI code. Please check the line 621 you have mentioned again to make sure it is now safe.

PR #1197, opened 2018-10-29
Fix for Entities-table-widget & Attributes-card-widget (decimals, units, p…

jktu2870:
…ost processing function)


jktu2870:
Canceled because of the fix already has been done by another contributor

PR #1272, opened 2018-11-21
Updating npm version to create package-lock.json 

gmatheu:
Updating npm version (in ui/pom.xml) to current version: [6.4.1](https://github.com/npm/cli/releases/tag/v6.4.1)

Since v5, npm generates [package-lock.json](https://docs.npmjs.com/files/package-lock.json). This file describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.

The main benefit of including this file is that **provides a single representation of a dependency tree** such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.

:warning: It includes https://github.com/thingsboard/thingsboard/pull/1271

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,0 +1,16437 @@
+{
+  "name": "thingsboard",
+  "version": "2.2.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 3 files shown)


ikulikov:
@gmatheu Thanks for your contribution!

PR #1291, opened 2018-11-29
fix bug CP-77 (telemetry delete timeseries)

yefimov-andrey:


application/src/main/java/org/thingsboard/server/controller/TelemetryController.java
```diff
@@ -279,8 +279,18 @@ public void shutdownExecutor() {
             deleteFromTs = 0L;
             deleteToTs = System.currentTimeMillis();
         } else {
-            deleteFromTs = startTs;
-            deleteToTs = endTs;
+            if(startTs == null) {
```
dmytro-landiak: I think it is better to throw some message to specify that you need to either set deleteAllDataForKeys parameter to "true" or leave both startTs and endTs parameters not empty.

dmytro-landiak: Something like:
if (startTs == null || endTs == null) {
return getImmediateDeferredResult("some_message", BAD_REQUEST)
}

ashvayka:
(it is merged in other PR)

PR #1359, opened 2018-12-26
Fixed URL templating

yefimov-andrey:


application/pom.xml
```diff
@@ -264,6 +264,10 @@
             <groupId>org.javadelight</groupId>
             <artifactId>delight-nashorn-sandbox</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.springfox.ui</groupId>
+            <artifactId>springfox-swagger-ui-rfc6570</artifactId>
```
ashvayka: No version is set. Not sure if this will compile. Version should be set in dependency management in main pom.xml

yefimov-andrey: Added version already

ashvayka: This version should be added to main pom.xml into dependencyManagement and not to this child pom.xml (application). We keep all versions in one place!

PR #1390, opened 2019-01-10
Update pom.xml

mysty-mystic:
to resolve npm install issue by changing the version


ashvayka:
Build failed and why you downgrade the npm version?

PR #1394, opened 2019-01-14
Use a fast serialization library like FST for serialization of TbActo…

janbols:
Use a fast serialization library like FST for serialization of TbActorMsg instead of the default java serialization because java serialization is notoriously slow.

This should improve throughput when serialization/deserialization is the bottleneck.

See https://github.com/RuedigerMoeller/fast-serialization

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -42,6 +42,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>de.ruedigermoeller</groupId>
+            <artifactId>fst</artifactId>
+        </dependency>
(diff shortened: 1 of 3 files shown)


ashvayka:
Have you tested this with the black-box-tests project (one of our modules) to work in a cluster mode (the only place when this is involved)?

janbols:
```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 270.268 s - in org.thingsboard.server.msa.ContainerTestSuite
```

PR #1435, opened 2019-01-30
Map update polygons

MrKartoshka:
Polygon functional and customization options


Shedaim:
care to explain how to create a polygon on the existing widget? I don't understand where to use CreatePolygon() or where to create the latLang values. (Using OpenStreet)

PR #1524, opened 2019-03-01
fix widget title longed

vvlladd28:


ashvayka:
Need to reformat code to highlight changes and simplify UI

PR #1537, opened 2019-03-07
Gateway-mqtt-api: deviceName added to attribute Attributes Request response message

gmatheu:
On Gateway-mqtt-api the current Attributes Request response payload is missing `device` attribute. [Documentation](https://thingsboard.io/docs/reference/gateway-mqtt-api/#request-attribute-values-from-the-server) shows that it should be included.

Before this PR:
```
{"id":` $request_id, "value": "value1"}
```

After this PR:
```
{"id":` $request_id, "device": "Device A", "value": "value1"}
```


ashvayka:
@gmatheu Thank You!

PR #1573, opened 2019-03-24
Add full Canadian postal code validation.

descipher:
Please review. Note the upper case change, if necessary lower case can be added back but generally upper case is used in Canadian postal codes.


ikulikov:
@descipher thanks for your contribution!

PR #1578, opened 2019-03-27
Corrected series sort return value order of tooltipIndividual mode.

descipher:


ikulikov:
@descipher. Thanks. But the issue is not in series info sorting. Incorrect series info is taken inside tooltip formatter function. I'm going to close this PR. I will provide correct fix soon.

descipher:
Odd, I did test the result of the sort change as successful. I walked the formatter but did not see an issue there. I will watch for the fix you have to understand the formatter better.

PR #1610, opened 2019-04-05
Add new map provider HERE

vvlladd28:


ui/src/app/widget/lib/openstreet-map.js
```diff
@@ -31,9 +31,14 @@ export default class TbOpenStreetMap {
 			mapProvider = "OpenStreetMap.Mapnik";
 		}
 
+		if (!mapProvider.indexOf("HERE.")) {
+			credentials.app_id = credentials.app_id || "AhM6TzD9ThyK78CT3ptx";
+			credentials.app_code = credentials.app_code || "p6NPiITB3Vv0GMUFnkLOOg";
+		}
```
ikulikov: Don't understand why we need credentials for map providers other than 'HERE'.

PR #1619, opened 2019-04-10
Fix Chinese translation problem

gemron:


ikulikov:
@gemron thanks for your contribution!

PR #1653, opened 2019-04-19
Added debug events ttl

ViktorBasanets:


application/src/main/resources/thingsboard.yml
```diff
@@ -166,6 +166,7 @@ cassandra:
     ts_key_value_partitioning: "${TS_KV_PARTITIONING:MONTHS}"
     ts_key_value_ttl: "${TS_KV_TTL:0}"
     events_ttl: "${TS_EVENTS_TTL:0}"
+    debug_events_ttl: "${DEBUG_EVENTS_TTL:WEEK}"
```
ashvayka: This should be long value, similar to ts_key_value_ttl or events_ttl. I believe Cassandra TTL is set in seconds, but I may be wrong.

dao/src/main/java/org/thingsboard/server/dao/event/CassandraBaseEventDao.java
```diff
@@ -188,11 +192,15 @@ public Event findEvent(UUID tenantId, EntityId entityId, String eventType, Strin
...
                 .value(ModelConstants.EVENT_UID_PROPERTY, entity.getEventUid())
                 .value(ModelConstants.EVENT_BODY_PROPERTY, entity.getBody());
+
         if (ifNotExists) {
             insert = insert.ifNotExists();
         }
-        if(ttl > 0){
-            insert.using(ttl(ttl));
+
+        int selectedTtl = entity.getEventType().matches(DataConstants.DEBUG_RULE_NODE) ? toMillis(debugEventsTtl) : ttl;
```
ashvayka: Why matches instead of equals?

PR #1661, opened 2019-04-23
First approach to implement integration with Amazon DynamoDB for time…

V0L0DYMYR:
… series data.

In this pull-request Aggregations other than NONE is not supported,
due to absents of aggregations in DynamoDb.


denisac77:
#1661 fix

PR #1672, opened 2019-04-27
Add rowClick and actionButtonClick action sources to timeseries_table widget…

pgrisu:
…. The actionHadler will receive selected row[] array as additionalParams


ikulikov:

PR #1698, opened 2019-05-08
add hide empty lines option to timeseries-table widget

pgrisu:
This new config option for timeseries-table widget allows removing of empty lines in the timeseries table that may appears when a processingFunction that blanks table cells on some conditions is configured on dataKeys.


ikulikov:

PR #1704, opened 2019-05-10
widget entity datakeys flexible creation order with reordering option

pgrisu:
Widget entity dataKeys may be created for timeseries and attibutes keys selecting keys in arbitrary way (no more restriction due to timeseries keys before attributes keys).
Data key type (timeseries or attributes) is now showed with icon and tooltip on detaKeys selector and dataKey chips.
Moreover it's possible to change the datKeys order by drag-and-drop dataKey chips on UI for both entity dataKeys and function dataKeys

### DIFF
--- a/ui/src/app/components/datasource-entity.directive.js
+++ b/ui/src/app/components/datasource-entity.directive.js
@@ -50,11 +50,8 @@ function DatasourceEntity($compile, $templateCache, $q, $mdDialog, $window, $doc
             scope.alarmFields.push(alarmField);
         }
 
-        scope.selectedTimeseriesDataKey = null;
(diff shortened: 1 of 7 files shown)


danibjor:
Without having tested this, when you are able to mix & match timeseries and attributes, is there a way to differentiate them if you have one of each with the same name?

pgrisu:
When you select dataKeys from the list (with autocompletion) there is a icon and description tooltip on each key, to qualify it as attribute or timeseries allowing proper key selection.
Moreover if you write a new key name (not in the BE database) you can choose to add it as timeseries or attibute key by clicking on the proper icon (with description tooltip)

danibjor:
Awesome

PR #1711, opened 2019-05-14
Chinese translation for entity-view

HollyFang:
1、Added Chinese translation for entity-view.
2、For  performance reasons, I modified the subscription.js so that the widgets do not keep updating.


ashvayka:
Hi @HollyFang Please remove ui/package-lock.json from this commit and explain changes to ui/src/app/api/subscription.js

HollyFang:
> Hi @HollyFang Please remove ui/package-lock.json from this commit and explain changes to ui/src/app/api/subscription.js

Sorry, I got confused. In my case, the maximum in the dashboard is very large, so if you keep refreshing the widget whose type is timeseries, there will be a lag. For example, It ofen broadcast "timeseries-table-data-updated" in Timeseries-table. So, I added filter codes in ui/src/app/api/subscription.js . As a novice, I think the widget whose type is latest can also use this filter code.
Hope I've made myself clear -_-|||

PR #1756, opened 2019-05-30
refactoring/improvements after review

dmytro-landiak:


application/src/main/java/org/thingsboard/server/controller/DeviceController.java
```diff
@@ -438,7 +436,8 @@ public void onFailure(Throwable t) {
     @PreAuthorize("hasAnyAuthority('TENANT_ADMIN', 'CUSTOMER_USER')")
     @RequestMapping(value = "/customer/device/{deviceId}/claim", method = RequestMethod.DELETE)
     @ResponseStatus(value = HttpStatus.OK)
-    public DeferredResult<ResponseEntity> reClaimDevice(@PathVariable(DEVICE_ID) String strDeviceId) throws ThingsboardException {
+    public DeferredResult<ResponseEntity> reClaimDevice(@PathVariable(DEVICE_ID) String strDeviceId,
+                                                        @RequestParam(required = false) String secretKey) throws ThingsboardException {
```
ashvayka: Let's make a key consist of only device id. The secret key will be used for comperison, but will be stored in the cache value and not the key.

application/src/main/java/org/thingsboard/server/controller/DeviceController.java
```diff
@@ -389,29 +391,25 @@ public Device getTenantDevice(
         }
     }
 
-    //TODO: @dlandiak deviceId -> deviceName
     @PreAuthorize("hasAnyAuthority('CUSTOMER_USER')")
-    @RequestMapping(value = "/customer/device/{deviceId}/claim", method = RequestMethod.POST)
+    @RequestMapping(value = "/customer/device/{deviceName}/claim", method = RequestMethod.POST)
     @ResponseBody
-    public DeferredResult<ResponseEntity> claimDevice(@PathVariable(DEVICE_ID) String strDeviceId,
+    public DeferredResult<ResponseEntity> claimDevice(@PathVariable(DEVICE_NAME) String deviceName,
```
ashvayka: Let's make an object instead of "String json" to simplify swagger UI.

PR #1827, opened 2019-07-02
fix entities-table widget for wrong dataKey value when same key is us…

pgrisu:
…ed on multiple columns with different processing function


ikulikov:

PR #1829, opened 2019-07-02
Add the possibility to hide single timewindow fields

mircopz:
In edit mode, dashboard or widget timewindow fields can be hidden. In this way, users could be prevented from changing some timewindow parameters, i.e. aggregation type, while changing interval period.

### DIFF
--- a/ui/src/app/api/time.service.js
+++ b/ui/src/app/api/time.service.js
@@ -234,7 +234,10 @@ function TimeService($translate, $http, $q, types) {
         var currentTime = (new Date).getTime();
         var timewindow = {
             displayValue: "",
-            selectedTab: 0,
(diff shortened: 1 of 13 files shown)


avlo:
i've discovered what appears to be a timewindow hide fields bug.

the setup requires customer dashboard containing "timeseries bars - flot" widget created by tenant with hidden "last", "aggregation" and "grouping interval".



next, to recreate bug.

from customer dashboard, open a timeseries widget & click "realtime-last day" to show fields are indeed hidden:



use mouse to visually select a custom group interval:



click "realtime - last minute".  you will now see "last", "aggregation" and "grouping interval" which were previously hidden, are no longer hidden.  this is the bug.

PR #1831, opened 2019-07-03
Create new 'array-table' widget

mircopz:
This widget represents a datakey whose value is a JSON string representation.

```javascript
{
  "key1": {
    "name": "First Key",
    "start": 1000,
    "stop": 2000
  },
  "key2": {
    "name": "Second Key",
    "start": 5000,
    "stop": 8000
  },
  "key3": {
    "name": "Third Key",
    "stop": 10000
  }
}
```

In widget configuration the same datakey is subscribed as many times as there are keys in the object to show (name, start and stop in the current example). Each datakey has a processing function which extrapolates a single value and stores it in an array. This array has many elements as there are objects in the datakey (key1, key2 and key3 in the current example).

```javascript
var currObj = JSON.parse(value);
var out = [];

for (var obj in currObj) {
  out.push(currObj[obj].name);
}

return out;
```

Resulting table shows as many rows as there are elements of this new array.

| Name | Start | Stop |
| -------- | --------| --------|
| First Key | 1000 | 5000 |
| Second Key | 2000 | 8000 |
| Third Key |  | 10000 |


mircopz:
Solved conflicts with issue #2848

PR #1833, opened 2019-07-03
Create new 'multiple-input' widget

mircopz:
This widget publishes multiple attributes (server or shared) by doing one or more POST request. Attributes values are set using several 'input' elements. These elements are represented as a matrix, whose rows are the selected datasources, and columns are the datakeys (arranged according to subscription order). Each datakey can be associated with a custom style or an icon to be placed before it.


cristianhumelnicu:
> This widget publishes multiple attributes (server or shared) by doing one or more POST request. Attributes values are set using several 'input' elements. These elements are represented as a matrix, whose rows are the selected datasources, and columns are the datakeys (arranged according to subscription order). Each datakey can be associated with a custom style or an icon to be placed before it.

Hi ,

Is this widget already integrated in TB ?

Thx

mircopz:
Not yet, you should wait for my pull request to be approved, in order to have it in the next release. Or you could merge manually my branch.

PR #1900, opened 2019-08-01
Add optional icon before widget title

mircopz:
An icon, with specified color and size selectable in widget configuration panel, can be added as a widget title prefix.


ikulikov:
@mircopz Thx for your contribution!

PR #1922, opened 2019-08-12
Feature/fetch last level only

yefimov-andrey:


application/src/main/java/org/thingsboard/server/controller/EntityRelationController.java
```diff
@@ -291,7 +291,7 @@ public EntityRelation getRelation(@RequestParam(FROM_ID) String strFromId,
         checkNotNull(query.getFilters());
         checkEntityId(query.getParameters().getEntityId(), Operation.READ);
         try {
-            return checkNotNull(relationService.findInfoByQuery(getTenantId(), query).get());
+            return checkNotNull(relationService.findInfoByQuery(getTenantId(), query, query.getParameters().isFetchLastLevelOnly()).get());
```
ashvayka: This is wrong. Why do you pass one parameter separately, if you already pass "query" as a second parameter.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/data/RelationsQuery.java
```diff
@@ -27,5 +27,5 @@
     private EntitySearchDirection direction;
     private int maxLevel = 1;
     private List<EntityTypeFilter> filters;
-
+    private boolean fetchLastLevelOnly = false;
```
ashvayka: no need to set default value to false. Default value is always false.

PR #1960, opened 2019-08-28
Add option to hide filter icon in alarm table widget

mircopz:
This feature can be useful to prevent a user from changing the type of alarm displayed when an alarm table widget should display a specific type of status.


vvlladd28:
Thank you for your pull request. But we will not merge it because it the new version we have moved this button to the widget header and added the same optional.

PR #1966, opened 2019-08-30
Add a new widget action to go back to previous dashboard state

mircopz:
In widget editor, section 'actions', it should be possible to select an action to return to previous dashboard state, alongside a new state or a new dashboard.


mircopz:
Solved conflicts with issue #2849

PR #1970, opened 2019-08-30
Does not save the dashboard after uncheking use dashboard timewindow

vvlladd28:


mircopz:
I think this commit is incorrect.
isEdit should have a two-way-binding, otherwise in timewindow panel controller its value never changes. Indeed, if you try opening the dashboard timewindow panel in edit mode, the 'hide' checkbox are not present (they are shown only if isEdit is true).
What you think?

PR #1971, opened 2019-08-30
Add the possibility to insert HTML content in the 'MsgToEmail' rule-node

mircopz:
Thingsboard email service, based on MimeMessageHelper Spring framework, can interpret any HTML content in email body message. This option, if the relevant checkbox is set, should be  as second parameter of `setText()` method: if not specified, the string used as text is displayed as a monospace font.


suess-ecbm:
This would be an excellent feature! We would be happy to use it frequently!

suess-ecbm:
Is this not relevant any more?

mircopz:
My original commit was on an old thingsboard release (2.3.1). Now I'm trying to port all these contributes on 2.5.x, but the rule-engine features requires more effort (I should align the `rulenode-core-config.js` file from another repository). I'm planning to do this in the following days, or weeks, it depends on my free time.

mircopz:
Hi @suess-ecbm 
I updated this PR to 2.5.x release, you can find it here #2879

PR #1972, opened 2019-09-02
Add the possibility to select timewindow source even for individual parameters separately

mircopz:
Currently the subscription parameters of a widget are taken either from the widget timewindow or from the general one of dashboard. It should be possible to create a mixed subscription, in which some parameters are selected locally (in the widget's timewindow), while others are taken from the global one.


mircopz:
Solve conflicts with issue #2851

PR #1976, opened 2019-09-02
Allow hiding of zero/false values from tbFlot widgets tooltips

pgrisu:


ashvayka:
Already merged as part of other PR.

PR #1977, opened 2019-09-02
Don't print label/value colon separator on tbFlot widgets tooltips when formatted value is empty

pgrisu:


ashvayka:
Please provide example of the case when this caused pain. Do you have a good use case for this?

pgrisu:
Hi @ashvayka, in my case I used flot state-chart to show system status changes during time (like entity status timeline by different colors as you can see on bottom widget of shared images on below).
In order to do it, I subscribed 'state' datakey more times (one time for each possible 'state') and defined different processing function in order to return '1' only on datakey matching current 'state' value (see the widget [state_timeline.zip](https://github.com/thingsboard/thingsboard/files/4707667/state_timeline.zip))
Hovering the graph I want see just the current state (the datakey label) in the tooltip without value (0 or 1) and ':' separator. So I enabled widget option to hide zero/false datakey values in the tooltip and I defined a tooltip formatting value function for empty value (in order to hide '1' value on current 'state' label). 
Without this PR the ':' separator stills showed after the datakey label also when formatted value is empty.
Just a couple of images to better show you the final result using this PR hoping I explained its possible usage
Tooltip-patched
Tooltip-patched2

PR #1992, opened 2019-09-09
Fix datakey function

vvlladd28:


ashvayka:
@vvlladd28 What is this fix about?

vvlladd28:
This commit fixed not save edit data key-value use data source function to a widget.

PR #2053, opened 2019-10-03
Add some datakey settings to edit legend appearance for flot widget

mircopz:
Add some options to datakey settings schema to edit legend appearance. In 'line' and 'pie' flot widget it's now possible to:
 - hide a datakey by default
 - disable data hiding by click 
 - remove a datakey from legend

### DIFF
--- a/ui/src/app/api/subscription.js
+++ b/ui/src/app/api/subscription.js
@@ -316,7 +316,8 @@ export default class Subscription {
             var datasource = this.datasources[i];
             for (var a = 0; a < datasource.dataKeys.length; a++) {
                 var dataKey = datasource.dataKeys[a];
-                dataKey.hidden = false;
+                dataKey.hidden = dataKey.settings.hideDataByDefault ? true : false;
+                dataKey.inLegend = dataKey.settings.removeFromLegend ? false : true;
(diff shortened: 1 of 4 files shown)


avlo:
is there documentation for how to use these features?

mircopz:
Sure.
First, create a "pie - flot" widget with some datakeys.

pic1

Then go to datakey advanced configuration: here there are three boolean options to specify their behavior in legend.

1. If "Data is hidden by default" is selected, this datakey is hidden, as if you had clicked on the corresponding legend key. This can be useful if there are some keys that are less important than others, so these ones are hidden and, if user wants more information, displayed only by clicking on that legend key.

pic2

2. If "Disable data hiding" is selected, clicking on the corresponding datakey has never effect, while in normal use click toggles data hiding. This can be useful to prevent some user to activate/deactive some keys that should be always active.

3. If "Remove datakey from legend" is selected, this datakey is still displayed in the pie chart, but it's not displayed on the legend.

pic3

ashvayka:
Hi @mircopz , thank you for your contribution! We have resolved conflicts and merged your code to PR https://github.com/thingsboard/thingsboard/pull/2289 and merged this PR. This will be part of v2.4.3.

PR #2055, opened 2019-10-03
Add ability to define points symbol and line width in 'line' flot charts

mircopz:
Using flot symbols plugin, it's now possible to set points marker in several ways, according to jQuery 3.1.0: 'circle' (built-in default), 'cross', 'diamond', 'square' and 'triangle'. It's also possible to define a callback function to draw other point types.


ashvayka:
Already merged as part of other PR.

PR #2056, opened 2019-10-03
Extend tooltip format function in 'line' flot charts

mircopz:
Now tooltip format function receives as parameters, in addition to value, widget context and hovered point timestamp. This can be useful to get other datakeys values at the same instant.

### DIFF
--- a/ui/src/app/widget/lib/flot-widget.js
+++ b/ui/src/app/widget/lib/flot-widget.js
@@ -54,7 +54,7 @@ export default class TbFlot {
 
         var tbFlot = this;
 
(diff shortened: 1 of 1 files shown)


ashvayka:
This is not backward compatible. If user has already created and set the tooltip function, it will be broken. Please correct this.

mircopz:
I noticed that I changed valueFormatFunction(value) to valueFormatFunction(**tbFlot.ctx, timestamp**, value): is this that is not backward compatible? Maybe I should add new parameters at bottom, i.e. valueFormatFunction(value, **tbFlot.ctx, timestamp**). And the same for other new parameters in other functions...

mircopz:
I fix it in a new PR

PR #2061, opened 2019-10-04
Fix timestamp and index search in 'line' flot charts tooltip

mircopz:
When a series is displayed by points, if mouse hover is done on a value less than datapoint X coordinate, the getHoverInfo function returns the previous index. Using datapoint.x instead of pos.x to find the index fix this problem.
Furthermore, when 'hover individual points' is selected, date displayed in tooltip is the smallest among all the series, while it should be the time of the hovered series only.


ashvayka:
Hi @mircopz , thank you for your contribution! We have resolved conflicts and merged your code to PR https://github.com/thingsboard/thingsboard/pull/2289 and merged this PR. This will be part of v2.4.3.

PR #2101, opened 2019-10-16
Redis cluster support

ShvaykaD:


ashvayka:
We need PR for documentation as well.

PR #2184, opened 2019-11-18
feature/refactored-save-attributes

YevhenBondarenko:
refactored DefaultDeviceStateService, added the field useTelemetry (need for saving attributes, if it's true, attributes will be saved like telemetry)

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
+++ b/application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
@@ -30,7 +30,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Service;
-import org.thingsboard.rule.engine.api.RpcError;
 import org.thingsboard.server.actors.service.ActorService;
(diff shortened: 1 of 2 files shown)


application/src/main/resources/thingsboard.yml
```diff
@@ -403,6 +403,7 @@ audit-log:
 state:
   defaultInactivityTimeoutInSec: "${DEFAULT_INACTIVITY_TIMEOUT:10}"
   defaultStateCheckIntervalInSec: "${DEFAULT_STATE_CHECK_INTERVAL:10}"
+  useTelemetry: "${USE_TELEMETRY:true}"
```
ashvayka: The default value should be false, for backward-compatibility. Let's rename to 
```yaml
persistToTelemetry: "${PERSIST_STATE_TO_TELEMETRY:false}"
```

ashvayka:
See build failures, probably related to previous comment about default value.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -125,9 +125,9 @@
     @Getter
     private long defaultStateCheckIntervalInSec;
 
-    @Value("${state.useTelemetry}")
+    @Value("${state.persistToTelemetry}")
```
ashvayka: add default parameter here ":false"

PR #2247, opened 2019-12-06
Set SpringResourceLoader to public

hamburml:
SpringResourceLoader needs to be public for **java -jar thingsboard-<VERSION>-boot.jar** to work.


ashvayka:
@hamburml Thank you for your contribution and extensive issue description!

PR #2261, opened 2019-12-10
set default events log TTL to 1 week

yefimov-andrey:


ashvayka:
Why should we change this?

yefimov-andrey:
To prevent storage overflow on default installations

PR #2323, opened 2020-01-09
Add a custom function, in 'multiple-input' widget settings, to be called before saving data

mircopz:
With this feature it's possible to write a custom function, in 'multiple-input' widget settings, that is called before saving data. This can be useful, for example, to perform specific checks on some values entered (i.e. edit data if a value should be greater than another one or should meet other conditions).


mircopz:
Solved conflicts with issue #2845

PR #2327, opened 2020-01-10
fix bug:Device reconnect abnormal when certificate authentication is turned on

blackstar-baba:
When **certificate authentication** is turned on, **under high concurrency**, if a large number of real devices are frequently reconnected, they may be connected to other devices on the platform, which will eventually lead to confusion in device data upload.


blackstar-baba:
@ashvayka Please check this

PR #2332, opened 2020-01-13
rough draft of alarm DB queries optimization

yefimov-andrey:


ashvayka:
Not Valid due to updates in 3.1

PR #2366, opened 2020-01-22
Table columns assignment gateway

Dmitriymush:
On this case: export/import of device in 'csv' could be applied with 'gateway' from now.


Dmitriymush:
'description' added to the list of columns assignment and could be imported/exported from now

PR #2374, opened 2020-01-27
fixed: entity list filter type exclude selected entities

Dmitriymush:
On this case: entity list filter type now working with exclusion of selected entities


Dmitriymush:
add the same for "Entity group list" on PE version

PR #2381, opened 2020-01-31
smtp tls version 

YevhenBondarenko:


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/mail/TbSendEmailNode.java
```diff
@@ -137,10 +141,13 @@ private Properties createJavaMailProperties() {
...
         javaMailProperties.put("mail.transport.protocol", protocol);
         javaMailProperties.put(MAIL_PROP + protocol + ".host", this.config.getSmtpHost());
-        javaMailProperties.put(MAIL_PROP + protocol + ".port", this.config.getSmtpPort()+"");
-        javaMailProperties.put(MAIL_PROP + protocol + ".timeout", this.config.getTimeout()+"");
+        javaMailProperties.put(MAIL_PROP + protocol + ".port", this.config.getSmtpPort() + "");
+        javaMailProperties.put(MAIL_PROP + protocol + ".timeout", this.config.getTimeout() + "");
         javaMailProperties.put(MAIL_PROP + protocol + ".auth", String.valueOf(StringUtils.isNotEmpty(this.config.getUsername())));
         javaMailProperties.put(MAIL_PROP + protocol + ".starttls.enable", Boolean.valueOf(this.config.isEnableTls()).toString());
+        if (this.config.isEnableTls()) {
+            javaMailProperties.put(MAIL_PROP + protocol + ".ssl.protocols", this.config.getTlsVersion());
```
ashvayka: What if this is an old version of the configuration (without this parameter stored). We need to check for an empty string or null here.

PR #2413, opened 2020-02-14
UI:fix. Map widget2

jktu2870:
preserves previous data on map if no data for current date range received


jktu2870:
Why still no merge? Are there any problems?

PR #2416, opened 2020-02-15
Feature/device provision

dmytro-landiak:


ashvayka:
Already merged different implementation that is based on the current one.

PR #2455, opened 2020-02-25
Add the possibility to define header cell style in table widgets

mircopz:


mircopz:
Conflicts solved with issue #2847

PR #2534, opened 2020-03-19
created Aws Sqs Queue

YevhenBondarenko:


rule-engine/rule-engine-components/pom.xml
```diff
@@ -35,7 +35,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/../..</main.dir>
-        <aws.sdk.version>1.11.323</aws.sdk.version>
+        <aws.sdk.version>1.11.747</aws.sdk.version>
```
ashvayka: This needs to be removed. We have the same in main pom.xml

vvlladd28: Please, remove from pull request file: ui/package-lock.json

application/src/main/java/org/thingsboard/server/service/queue/processing/AbstractConsumerService.java
```diff
@@ -136,6 +136,15 @@ protected void launchNotificationsConsumer() {
     @PreDestroy
     public void destroy() {
         stopped = true;
+
```
ashvayka: I have intentionally removed this. Kafka consumers will throw exceptions here (ConcurrentModificationException). Please refactor Kafka consumers to make sure they are unsubscribing in main consumer thread (however, if the executor is shut down - this is meaningless).

PR #2564, opened 2020-03-31
fix bug: Under high concurrency, Mqtt client nextMessageId exceeds the 0xffff limit

blackstar-baba:
#### description
The compareAndSet method and getAndIncrement method of the AtomicInteger class are atomic, but when these two methods are used at the same time, they are no longer atomic.


blackstar-baba:
@ikulikov PTAL

ikulikov:
LGTM. @blackstar-baba Thx for contribution!

PR #2580, opened 2020-04-06
Update Dockerfile

olirehacek:
The Cassandra KEYS document moved.
Thingsboard build has been failing if you download it using the curl.


ashvayka:
Thank you but this is no longer valid. The link contains up to date keys.

PR #2584, opened 2020-04-06
Feature/psql ts ttl

ShvaykaD:


application/src/main/java/org/thingsboard/server/service/ttl/AbstractTimeseriesCleanUpService.java
```diff
@@ -0,0 +1,89 @@
...
+    @Value("${spring.datasource.url}")
+    private String dbUrl;
+
+    @Value("${spring.datasource.username}")
+    private String dbUserName;
+
+    @Value("${spring.datasource.password}")
+    private String dbPassword;
+
+    @Scheduled(fixedDelayString = "${sql.ttl.execution_interval_ms}")
```
ashvayka, line 49: Please add initial delay. @Scheduled(fixedDelayString = "${sql.ttl.execution_interval_ms}", initialDelayString = "${sql.ttl.execution_interval_ms}"). No need to overload the system at start.

PR #2655, opened 2020-04-22
Events TTL for PostgreSQL

ShvaykaD:


ashvayka:
While merging to PE, please don't forget about 
```
    public static final String DEBUG_CONVERTER = "DEBUG_CONVERTER";
    public static final String DEBUG_INTEGRATION = "DEBUG_INTEGRATION";
```

Issue #2500, opened 2020-03-10
MQTTTransport disconnects clients due to lack of PINGREQ despite there being activity


bdrlamb:
#1665

In MQTTTransportHandler::processMqttMsg, transportService.reportActivity() is only called in response to PINGREQ messages.  When an MQTT client is regularly publishing to topics or receiving subscription messages, most MQTT clients will omit their keep-alive pings as they are unnecessary since actual traffic is being transported.

```
private void processMqttMsg(ChannelHandlerContext ctx, MqttMessage msg) {
        address = (InetSocketAddress) ctx.channel().remoteAddress();
        if (msg.fixedHeader() == null) {
            log.info("[{}:{}] Invalid message received", address.getHostName(), address.getPort());
            processDisconnect(ctx);
            return;
        }
        deviceSessionCtx.setChannel(ctx);
        switch (msg.fixedHeader().messageType()) {
            case CONNECT:
                processConnect(ctx, (MqttConnectMessage) msg);
                break;
            case PUBLISH:
                processPublish(ctx, (MqttPublishMessage) msg);
                break;
            case SUBSCRIBE:
                processSubscribe(ctx, (MqttSubscribeMessage) msg);
                break;
            case UNSUBSCRIBE:
                processUnsubscribe(ctx, (MqttUnsubscribeMessage) msg);
                break;
            case PINGREQ:
                if (checkConnected(ctx, msg)) {
                    ctx.writeAndFlush(new MqttMessage(new MqttFixedHeader(PINGRESP, false, AT_MOST_ONCE, false, 0)));
                    transportService.reportActivity(sessionInfo);
                    if (gatewaySessionHandler != null) {
                        gatewaySessionHandler.reportActivity();
                    }
                }
                break;
            case DISCONNECT:
                if (checkConnected(ctx, msg)) {
                    processDisconnect(ctx);
                }
                break;
            default:
                break;
        }

    }

```
Because of this, clients will not send any PINGREQ during extended periods with regular traffic.  AbstractTransportService::checkInactivityAndReportActivity will then disconnect the client due to the perceived inactivity timeout.

```
 private void checkInactivityAndReportActivity() {
        long expTime = System.currentTimeMillis() - sessionInactivityTimeout;
        sessions.forEach((uuid, sessionMD) -> {
            if (sessionMD.getLastActivityTime() < expTime) {
                if (log.isDebugEnabled()) {
[...]

PR #2675, opened 2020-04-27
Fix MQTT inactivity disconnects

bdrlamb:
#2500 

MQTT clients don't send PINGREQ messages if they've recently sent [UN]SUBSCRIBE/PUBLISH messages.  The MQTT server last activity timestamp is currently only updated in response to PINGREQ messages.  This causes the server to disconnect any clients that regularly send PUBLISH messages since they don't send PINGREQ. 

This change adds reportActivity calls in response to PUBLISH, SUBSCRIBE and UNSUBSCRIBE in addition to PINGREQ.

### DIFF
--- a/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
+++ b/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
@@ -138,12 +138,24 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
                 break;
             case PUBLISH:
                 processPublish(ctx, (MqttPublishMessage) msg);
+                transportService.reportActivity(sessionInfo);
+                if (gatewaySessionHandler != null) {
+                    gatewaySessionHandler.reportActivity();
+                }
                 break;
             case SUBSCRIBE:
                 processSubscribe(ctx, (MqttSubscribeMessage) msg);
+                transportService.reportActivity(sessionInfo);
+                if (gatewaySessionHandler != null) {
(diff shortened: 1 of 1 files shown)


ashvayka:
Hi @bdrlamb  Thank you for your contribution and investigation. I will merge this PR, but will also need to heavily refactor it. The one thing I don't like is:
`                
if (gatewaySessionHandler != null) {
                    gatewaySessionHandler.reportActivity();
                }
`
This is going to cause N operations for each publish if gateway has N devices connected. 
I will make it tracking the gateway session separately from device sessions.

Also, `transportService.reportActivity(sessionInfo);` is present for all publish operations. However, it is not present for gateway session. This needs to be fixed.

bdrlamb:
> Hi @bdrlamb Thank you for your contribution and investigation. I will merge this PR, but will also need to heavily refactor it. The one thing I don't like is:
> ` if (gatewaySessionHandler != null) { gatewaySessionHandler.reportActivity(); }`
> This is going to cause N operations for each publish if gateway has N devices connected.
> I will make it tracking the gateway session separately from device sessions.
> 
> Also, `transportService.reportActivity(sessionInfo);` is present for all publish operations. However, it is not present for gateway session. This needs to be fixed.

No problem!  I wasn't particularly happy with the style of the fix but I'm really not familiar enough with the code base to confidently do any kind of refactoring.  Just glad to see this get fixed :)

ayavilevich:
It says that this was merged in 2020 but when I look at this file in the master branch I see that this change is not there.
Any idea why?
https://github.com/thingsboard/thingsboard/blob/master/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java

PR #2826, opened 2020-05-24
fix bug: Resending the message causes the client to interrupt

blackstar-baba:
In the mqtt 3.1 protocol or mqtt 3.1.1 protocol, except for messages of type pubish, the isDup of other messages must be zero，Sending a subscription message multiple times causes the client to reconnect，therefore, resending the subscription message with isDup true will cause the server to think that the message is abnormal, thus interrupting the connection with the client


blackstar-baba:
@ikulikov PTAL

PR #2836, opened 2020-05-26
Follow mqtt 3.1.1 protocol number 3.3.1-2

blackstar-baba:
Follow mqtt 3.1.1 protocol number 3.3.1-2 : The DUP flag MUST be set to 0 for all QoS 0 messages. If the qos of the publish message is not 0, then set dup to true


ashvayka:
This probably fixes the build issues (black-box-tests) caused by your previous PR for dup flag. Thank you!

PR #2839, opened 2020-05-26
Refactored k8s configs (made 'basic' and 'high-availability' deployment modes)

vzikratyi-tb:


ashvayka:
What about transports? Why they have service_type setting to monolith? This is redundant. It is tb-transport by default and should not be set.

vzikratyi-tb:
Fixed
Also changed js.eval.requests topic to js_eval.requests

PR #2845, opened 2020-05-27
Add a custom function, in 'multiple-input' widget settings, to be called before saving data, and other minor features

mircopz:
With this feature it's possible to write a custom function, in 'multiple-input' widget settings, that is called before saving data. This can be useful, for example, to perform specific checks on some values entered (i.e. edit data if a value should be greater than another one or should meet other conditions).
Other minor features concern the possibility to customize single key parameters, like tooltip, text color, number range (min-max), and a new value type (select box) has been inserted, with the possibility to specify all the accettable options.


mircopz:
I can't understand what caused the error: TeamCity build says 

> 460 tests passed (all tests)

PR #2850, opened 2020-05-27
Fix misbehaviours when create devices using external provided device-id

mircopz:
Already merged in #1629 but no more present. 

> When a new device is created by external application providing a not null device-id a wrong messageType is injected into rule-engine (ENTITY-UPDATED instead of ENTITY-CREATED) and the deviceStateService doesn't handle properly active/incative state transactions for the newly created device

### DIFF
--- a/application/src/main/java/org/thingsboard/server/controller/DeviceController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/DeviceController.java
@@ -89,10 +89,11 @@ public class DeviceController extends BaseController {
     @ResponseBody
     public Device saveDevice(@RequestBody Device device,
                              @RequestParam(name = "accessToken", required = false) String accessToken) throws ThingsboardException {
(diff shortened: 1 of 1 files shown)


ashvayka:
We have corresponding PR here: https://github.com/thingsboard/thingsboard/pull/2812 This is a more generic way to fix similar issues.

mircopz:
Hi @ashvayka 
I think that #2812 doesn't solve my problem.
In my scenario, during the provisioning phase **the device is already created with an id** (time based UUID for compatibility, as you suggest) assigned by my provisioning service. So when the `saveDevice` function is called, the `device.getId() == null` control returns always false, and the creation event is managed as an update event. This cause two errors:
- the ENTITY-CREATED event is not injected into rule-engine
- the active state transactions are not handled
Is it ok to create a device with an assigned id, or is it not recommended for some reason? So far I haven't seen any contraindications, but to ensure the correct behavior of all events I think this fix is mandatory. Let me know, otherwise I should change my provisioning service (or I keep this fix locally without pulling it again).
Thank you

PR #2851, opened 2020-05-27
Add the possibility to select timewindow source even for individual parameters separately

mircopz:
Currently the subscription parameters of a widget are taken either from the widget timewindow or from the general one of dashboard. It should be possible to create a mixed subscription, in which some parameters are selected locally (in the widget's timewindow), while others are taken from the global one.


ChrisEvisi:
Hi,
Yes, I think this is an important feature.  
 
That way,  widgets can be set with their own Aggregation type and Group interval on a dashboard and some or all the widgets can be set to still use the common Dashboard Time window period (realtime or historical) and Aggregation type and group period.

PR #2856, opened 2020-05-27
Add pre/post functions in widget editor to be called before/after onInit and onDataUpdated

mircopz:
With this feature it's possible to write different custom functions in widget editor that will be called before (pre) or after (post) standard `onInit` and `onDataUpdated`. This can be useful, for example, if some data should be edited before calling the original `onDataUpdated`, or if after the original `onInit` an HTTP GET should be called to obtain a value usable in widget context. Furthermore, it's possibile to override original onInit and onDataUpdated if the 'pre' functions return **skip-default** string.

### DIFF
--- a/ui/src/app/api/subscription.js
+++ b/ui/src/app/api/subscription.js
@@ -794,14 +794,16 @@ export default class Subscription {
     }
 
     alarmsUpdated(alarms, apply) {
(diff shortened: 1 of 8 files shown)


mircopz:
I can't understand what caused the error: changes are few and not very invasive, and the build process works locally

PR #2891, opened 2020-06-01
feature - add controller and schema_update for entity profiles

karpunets:


application/src/main/data/upgrade/3.1.0/schema_update.sql
```diff
@@ -0,0 +1,27 @@
...
+DROP TABLE IF EXISTS entity_profile;
+
+CREATE TABLE IF NOT EXISTS entity_profile (
+    id              varchar(31) NOT NULL CONSTRAINT entity_profile_pkey PRIMARY KEY,
+    name            varchar(255),
+    tenant_id       varchar(31),
+    entity_type     varchar(255),
+    profile         varchar,
+    additional_info varchar,
+    CONSTRAINT name_tenant_id_type_unq_key UNIQUE (name, tenant_id, entity_type)
```
ashvayka, line 26: I think that constraint should be (tenant_id, entity_type, name). In such case the automatic index is more usable. We always know tenant_id and entity_type when executing find queries. But Name may have different conditions (IN, LIKE, etc). This needs to be changed in the main SQL file as well.

application/src/main/java/org/thingsboard/server/controller/EntityProfileController.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    @PreAuthorize("hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN')")
+    @PostMapping("/entityProfiles")
+    public EntityProfile saveEntityProfile(@RequestBody EntityProfile entityProfile) throws ThingsboardException {
+        entityProfile = entityProfile.toBuilder()
+                .tenantId(getCurrentUser().getTenantId())
+                .build();
+        Operation operation = entityProfile.getId() == null ? Operation.CREATE : Operation.WRITE;
```
ashvayka, line 70: There is a new method called checkEntity. Please use it instead of checkPermission.

application/src/test/java/org/thingsboard/server/controller/BaseEntityProfileControllerTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.thingsboard.server.common.data.page.PageLink;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.commons.lang3.RandomStringUtils.randomAlphanumeric;
+import static org.junit.Assert.*;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+public class BaseEntityProfileControllerTest extends AbstractControllerTest {
```
ashvayka, line 33: Please add tests similar to DeviceControllerTest.testUpdateDeviceFromDifferentTenant

application/src/test/java/org/thingsboard/server/controller/BaseEntityProfileControllerTest.java
```diff
@@ -0,0 +1,102 @@
...
+                saved.getId().getId().toString());
+        assertNotNull(found);
+        assertEquals(saved, found);
+        PageData<EntityProfile> page = doGetTypedWithPageLink("/api/entityProfiles?",
+                PAGE_ENTITY_PROFILE_TYPE, new PageLink(5));
+        assertFalse(page.getData().isEmpty());
+    }
+
+    @Test
+    public void testTenantAdminDeviceEntityProfileFlow() throws Exception {
```
ashvayka, line 64: Please add more tests similar to BaseDeviceControllerTest.testFindCustomerDevicesByType. We need to check that all parameters from getEntityProfiles rest api call work fine. Using testFindCustomerDevicesByType we create 100+ profiles and check pagination properly.

application/src/main/java/org/thingsboard/server/controller/EntityProfileController.java
karpunets: done, but as for me it seems not good way to check some entity by calling base class method with all logic inside
BaseController.class already has 700+ lines and this number will be grown

application/src/main/data/upgrade/3.1.0/schema_update.sql
karpunets: agree, fixed

application/src/test/java/org/thingsboard/server/controller/BaseEntityProfileControllerTest.java
karpunets: added `testUpdateEntityProfileFromDifferentTenant()`

application/src/test/java/org/thingsboard/server/controller/BaseEntityProfileControllerTest.java
karpunets: added
`testFindEntityProfileByType()`
`testFindEntityProfileByName()`

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -461,9 +455,10 @@ private void onDeviceDeleted(TenantId tenantId, DeviceId deviceId) {
             @Override
             public DeviceStateData apply(@Nullable List<T> data) {
                 try {
+                    DeviceProfile deviceProfile = entityProfileService.findProfile(device, DeviceProfile.class);
```
ashvayka, line 458: Please add TODO to replace this with cache.

PR #2942, opened 2020-06-11
Feature/entity profile

karpunets:


application/src/main/java/org/thingsboard/server/ThingsboardInstallApplication.java
```diff
@@ -29,6 +29,7 @@
 @ComponentScan({"org.thingsboard.server.install",
         "org.thingsboard.server.service.component",
         "org.thingsboard.server.service.install",
+        "org.thingsboard.server.service.entityprofile",
```
ashvayka, line 32: Let's rename package to "profile" (for simplicity)

application/src/main/java/org/thingsboard/server/service/install/SqlDatabaseUpgradeService.java
```diff
@@ -222,6 +231,25 @@ public void upgradeDatabase(String fromVersion) throws Exception {
                     schemaUpdateFile = Paths.get(installScripts.getDataDir(), "upgrade", "3.1.0", SCHEMA_UPDATE_SQL);
                     loadSql(schemaUpdateFile, conn);
                     log.info("Schema updated.");
+                    try {
+                        conn.createStatement().execute(
+                                "ALTER TABLE device ADD entity_profile_id varchar(31);"); //NOSONAR, ignoring because method used to execute thingsboard database upgrade script
```
ashvayka, line 236: Do we have a constraint for deletion of entity-profile if it is already used for some devices? We should add one.

application/src/main/java/org/thingsboard/server/service/install/SqlDatabaseUpgradeService.java
```diff
@@ -222,6 +231,25 @@ public void upgradeDatabase(String fromVersion) throws Exception {
...
+                    }
+                    transactionTemplate.execute(transactionStatus -> {
+                        deviceRepository.findAllTenantIdAndTypeWhereEmptyEntityProfileId().forEach(objects -> {
+                            String tenantIdStr = (String) objects[0];
+                            TenantId tenantId = new TenantId(UUIDConverter.fromString(tenantIdStr));
+                            String type = (String) objects[1];
+
+                            EntityProfile profile = tbEntityProfileService.createDefault(tenantId, type, DeviceProfile.class);
+                            String entityProfileStr = UUIDConverter.fromTimeUUID(profile.getId().getId());
+                            deviceRepository.updateEntityProfileId(entityProfileStr, tenantIdStr, type);
```
ashvayka, line 248: Whey do we use entityProfileStr, tenantIdStr instead of TenantId and EntityProfileId?

application/src/main/java/org/thingsboard/server/service/install/SqlDatabaseUpgradeService.java
karpunets: Maybe is it better to create foreign key for entity_profile_id after setting default profiles?

application/src/main/java/org/thingsboard/server/service/install/SqlDatabaseUpgradeService.java
karpunets: because I use directly deviceRepository and DeviceEntity has tenantId and entityProfileId as string
no needed add this methods for service layer

application/src/main/java/org/thingsboard/server/ThingsboardInstallApplication.java
karpunets: 👌

karpunets: done

PR #2943, opened 2020-06-11
Added proxy configs to rest api call rule node

YevhenBondarenko:


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -63,7 +69,25 @@
             if (config.getMaxParallelRequestsCount() > 0) {
                 pendingFutures = new ConcurrentLinkedDeque<>();
             }
-            if (config.isUseSimpleClientHttpFactory()) {
+
+            if (config.isEnableProxy()) {
+                SimpleClientHttpRequestFactory factory = new SimpleClientHttpRequestFactory();
```
ashvayka, line 74: add validation of config and meaningful exception

YevhenBondarenko: done

PR #2945, opened 2020-06-11
removed Apache Email Validator

YevhenBondarenko:


dao/pom.xml
```diff
@@ -94,11 +94,11 @@
 		<dependency>
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-lang3</artifactId>
-		</dependency>        
-		<dependency>
-			<groupId>commons-validator</groupId>
-			<artifactId>commons-validator</artifactId>
 		</dependency>
+        <dependency>
```
ashvayka, line 98: identation

dao/src/main/java/org/thingsboard/server/dao/service/DataValidator.java
```diff
@@ -64,11 +64,21 @@ protected boolean isSameData(D existentData, D actualData) {
...
             throw new DataValidationException("Invalid email address format '" + email + "'!");
         }
     }
 
+    private static boolean doValidateEmail(String email) {
+        if (email == null) {
+            return false;
+        }
+
+        Pattern emailPattern = Pattern.compile(VALID_EMAIL_REGEX, Pattern.CASE_INSENSITIVE);
```
ashvayka, line 77: to contstant

dao/pom.xml
YevhenBondarenko: done

PR #2978, opened 2020-06-18
enable default credential provider chain for aws sqs

### LINKED ISSUE #2843: AWS-SQS: Forced usage of credentials
The AWS SQS implementation forces the use of access keys provided by specific config properties and makes it impossible to fallback to the standard behavior off the AWS sdks, which includes fallback to ENV variables, the default credential profile and most importantly: the instance profile.

https://github.com/thingsboard/thingsboard/blob/b124f86a0f49c6b39b7bcce1760edb0922b26841/common/queue/src/main/java/org/thingsboard/server/queue/sqs/TbAwsSqsAdmin.java#L40

Would you accept a PR which enables the default credentials provider chain? https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-roles.html

janschumann:
fixes #2843


ashvayka:
@YevhenBondarenko please review.

ashvayka:
+ we also need to cherry-pick this to 2.5.3

YevhenBondarenko:
LGTM, 
we will merge this PR to the develop branch for now since the same changes should be added to the other AWS services in TB.

PR #2999, opened 2020-06-24
Fix Check alarm status

mp-loki:
Check the status of the alarm, fetched from the database, not the incoming message


YevhenBondarenko:
LGTM

PR #3010, opened 2020-06-26
changed aggregation interval to 1/7

yefimov-andrey:
New intervals:
30s - 5s
1m - 10s
2hours - 15m
7days - 1 day


vvlladd28:
Not correct write new intervals:
2m - 15s
15m - 2m
2h - 15m
7days - 1day

Issue #2941, opened 2020-06-11
[Bug] The Cassandra Data Center cannot be changed


yuyihan666:
The cassandra data center can only be datacenter1 and cannot be replaced, resulting in failure to start。
log:
You specified datacenter1 as the local DC, but some contact points are from a different DC: Node(endPoint=/127.0.0.1:9042, hostId=4a965dc3-8b4a-48d7-b512-be67cabca05c, hashCode=79c465ed)=dc1; please provide the correct local DC, or check your contact points。

vzikratyi-tb:
@yuyihan666 what type of installation are you using (docker, k8s, ubuntu etc)?

yuyihan666:
> @yuyihan666 what type of installation are you using (docker, k8s, ubuntu etc)?

centos 7

yuyihan666:
This problem occurs after upgrading to 3.0

vzikratyi-tb:
If you have only one cassandra datacenter, as a quick fix you can manually change its name to 'datacenter1'. 
We are aware of this problem and are planning to fix it in next releases.

TrevorMuraro:
The Release Notes page has "Set correct cassandra datacenter" in the bug fixes for 3.0.1, but the bug does not appear to be fixed (["datacenter1" is still hardcoded](https://github.com/thingsboard/thingsboard/blob/7e418337711183d5269bed140bedcb33c650aec1/common/dao-api/src/main/java/org/thingsboard/server/dao/cassandra/AbstractCassandraCluster.java#L85)).

PR #3032, opened 2020-07-01
Moved hardcoded local datacenter name to config

vzikratyi-tb:
There's an [issue](https://github.com/thingsboard/thingsboard/issues/2941) connected to this change.


common/dao-api/src/main/java/org/thingsboard/server/dao/cassandra/AbstractCassandraCluster.java
```diff
@@ -36,6 +36,8 @@
     private Boolean jmx;
     @Value("${cassandra.metrics}")
     private Boolean metrics;
+    @Value("${cassandra.local_datacenter}")
```
ashvayka, line 39: please add default value here. This will make sure that everything will work for old clients, even if they use same thingsboard.yml

PR #3046, opened 2020-07-03
Feature/UUID update

YevhenBondarenko:


ashvayka:
I think we need to drop all indexes before the update. Then do the update, Then create all indexes. 
The update will be faster and the code will be easier to review. 
We also need to execute VACUUM FULL

application/src/main/data/upgrade/3.0.1/schema_update_to_uuid.sql
```diff
@@ -21,21 +21,806 @@ DECLARE
...
+    DROP INDEX IF EXISTS idx_asset_type;
+END;
+$$;
+
+CREATE OR REPLACE PROCEDURE create_all_idx()
+    LANGUAGE plpgsql AS
+$$
+BEGIN
+    CREATE INDEX IF NOT EXISTS idx_alarm_originator_alarm_type ON alarm(originator_id, type, start_ts DESC);
+    CREATE INDEX IF NOT EXISTS idx_event_type_entity_id ON event(tenant_id, event_type, entity_type, entity_id);
```
ashvayka, line 87: Add one more index

PR #3052, opened 2020-07-06
Prometheus Metrics

vzikratyi-tb:


application/src/main/java/org/thingsboard/server/service/queue/TbRuleEngineConsumerStats.java
```diff
@@ -43,61 +43,72 @@
...
 
-    private final Map<String, AtomicInteger> counters = new HashMap<>();
+    private final List<StatsCounter> counters = new ArrayList<>();
     private final ConcurrentMap<UUID, TbTenantRuleEngineStats> tenantStats = new ConcurrentHashMap<>();
     private final ConcurrentMap<TenantId, RuleEngineException> tenantExceptions = new ConcurrentHashMap<>();
 
     private final String queueName;
 
-    public TbRuleEngineConsumerStats(String queueName) {
+    public TbRuleEngineConsumerStats(String queueName, StatsCounterFactory counterFactory) {
```
vzikratyi-tb, line 63: maybe it's possible to do this somehow better and without passing factory obj?

application/src/main/java/org/thingsboard/server/service/stats/StatsCounterFactory.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.stereotype.Service;
+import org.thingsboard.server.service.metrics.StubCounter;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+@Service
+public class StatsCounterFactory {
+    private static final String STATS_NAME_TAG = "statsName";
+
+    private static final Counter STUB_COUNTER = new StubCounter();
```
vzikratyi-tb, line 31: not sure if that's a good idea

PR #3076, opened 2020-07-10
Develop/2.5.3 batch improvements

YevhenBondarenko:


dao/src/main/java/org/thingsboard/server/dao/sql/attributes/JpaAttributeDao.java
```diff
@@ -78,7 +82,9 @@ private void init() {
                 .maxDelay(maxDelay)
                 .statsPrintIntervalMs(statsPrintIntervalMs)
                 .build();
-        queue = new TbSqlBlockingQueue<>(params);
+
+        Function<AttributeKvEntity, Integer> hashcodeFunction = entity -> entity != null ? entity.getId().getEntityId().hashCode() : 0;
```
ashvayka, line 86: put null check to the     public ListenableFuture<Void> add(E element)  method to avoid code duplication.

dao/src/main/java/org/thingsboard/server/dao/service/DataValidator.java
```diff
@@ -64,7 +64,7 @@ protected boolean isSameData(D existentData, D actualData) {
         return actualData.getId() != null && existentData.getId().equals(actualData.getId());
     }
 
-    protected static void validateEmail(String email) {
+    public static void validateEmail(String email) {
```
ashvayka, line 67: this is not relevant to the feature

dao/src/main/java/org/thingsboard/server/dao/sql/TbSqlBlockingQueueWrapper.java
```diff
@@ -0,0 +1,53 @@
...
+        for (int i = 0; i < maxThreads; i++) {
+            TbSqlBlockingQueue<E> queue = new TbSqlBlockingQueue<>(params);
+            queues.add(queue);
+            queue.init(logExecutor, saveFunction, i);
+        }
+    }
+
+    public ListenableFuture<Void> add(E element) {
+        int hash = hashCodeFunction.apply(element);
+        int queueIndex = (hash & 0x7FFFFFFF) % maxThreads;
```
ashvayka, line 46: Why do you need this operation?

YevhenBondarenko: because hash can be negative.

PR #3108, opened 2020-07-17
Feature/mqtt device creation lock upgrade

vzikratyi-tb:


ashvayka:
This need to be refactored to use constraint in SQL DB. Not the lock per Java process (might be multiple processes in cluster mode and lock will not work)

vzikratyi-tb:
But we still need to lock per Java process so that it doesn't create multiple GetOrCreateDevice futures for the same deviceName.

PR #3277, opened 2020-08-10
Display sorted metadata keys in rule nodes

mp-loki:
Display sorted metadata keys in rule node debug messages and test transformer function dialogs


ashvayka:
@vvlladd28 please review this

Issue #3317, opened 2020-08-14
[Feature Request] Alarm for originator type in rule engine


ymucahit:
**Is your feature request related to a problem? Please describe.**
I want to trigger some external rest services when some alarms occur. However, 'originator type switch' and 'originator type' nodes do not contain 'Alarm' choice. 

**Describe the solution you'd like**
As for as I understand from the codebase adding 'Alarm' to these rule nodes is enough since 'Entity created' message emerges in the rule engine for entity type 'Alarm'.

ymucahit:
> Merged and will be part of 3.1.2 and 3.2 releases -ashvayka 


https://github.com/thingsboard/thingsboard/pull/3379#issuecomment-682589548

PR #3379, opened 2020-08-27
[Feature] Alarm option for originator type in rule engine

ymucahit:
Hello,

The user wasn't able to distinguish Alarm message in rule engine. Alarm option is added to switch nodes in rule engine which is described in #3317.

Best.


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/filter/TbOriginatorTypeFilterNodeConfiguration.java
```diff
@@ -31,7 +31,7 @@
     public TbOriginatorTypeFilterNodeConfiguration defaultConfiguration() {
         TbOriginatorTypeFilterNodeConfiguration configuration = new TbOriginatorTypeFilterNodeConfiguration();
         configuration.setOriginatorTypes(Arrays.asList(
-                EntityType.DEVICE
+                EntityType.ALARM
```
ashvayka, line 34: No need to change the default configuration. Otherwise - LGTM. Please correct this and I will merge.

ashvayka (changes requested):

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/filter/TbOriginatorTypeFilterNodeConfiguration.java
ymucahit: It is done.

ashvayka:
Merged and will be part of 3.1.2 and 3.2 releases

PR #3383, opened 2020-08-28
in memory queue logs

YevhenBondarenko:


ashvayka:
Maybe it is better to add printStats to a queue interface, so any queue implementation may or may not print stats?

PR #3405, opened 2020-09-01
[2.5.5] Improvements/queue reprocessing strategy

ShvaykaD:


application/src/main/java/org/thingsboard/server/service/queue/processing/TbRuleEngineProcessingStrategyFactory.java
```diff
@@ -108,6 +114,11 @@ public TbRuleEngineProcessingDecision analyze(TbRuleEngineProcessingResult resul
                         } catch (InterruptedException e) {
                             throw new RuntimeException(e);
                         }
+                        if (multiplyPauseBetweenRetries && maxPauseBetweenRetries > 0) {
+                            if (pauseBetweenRetries != maxPauseBetweenRetries) {
```
ashvayka, line 118: you can add this to previous if or remove completely.

application/src/main/java/org/thingsboard/server/service/queue/processing/TbRuleEngineProcessingStrategyFactory.java
```diff
@@ -108,6 +114,11 @@ public TbRuleEngineProcessingDecision analyze(TbRuleEngineProcessingResult resul
...
+                                pauseBetweenRetries = Math.min(maxPauseBetweenRetries, pauseBetweenRetries * pauseBetweenRetries);
```
ashvayka, line 119: Maybe 
`pauseBetweenRetries = Math.min(maxPauseBetweenRetries, pauseBetweenRetries * 2);`
?

common/queue/src/main/java/org/thingsboard/server/queue/settings/TbRuleEngineQueueAckStrategyConfiguration.java
```diff
@@ -24,5 +24,7 @@
     private int retries;
     private double failurePercentage;
     private long pauseBetweenRetries;
+    private boolean multiplyPauseBetweenRetries;
```
ashvayka, line 27: Maybe we don't need this flag at all? Just check if `max-pause-between-retries > pause-between-retries` to calculate this.

ShvaykaD: @ashvayka I guess that someone may be no need to increase the pause, in some cases, so that's why I think that we need to add this flag.

application/src/main/java/org/thingsboard/server/service/queue/processing/TbRuleEngineProcessingStrategyFactory.java
ShvaykaD: @ashvayka sorry for this typo.

PR #3406, opened 2020-09-01
[3.2] Improvements/queue reprocessing strategy

ShvaykaD:


application/src/main/java/org/thingsboard/server/service/queue/processing/TbRuleEngineProcessingStrategyFactory.java
```diff
@@ -103,10 +116,25 @@ public TbRuleEngineProcessingDecision analyze(TbRuleEngineProcessingResult resul
...
+                                pause = maxExpPauseBetweenRetries;
+                            }
+                            try {
+                                Thread.sleep(TimeUnit.SECONDS.toMillis(
+                                        pause));
+                            } catch (InterruptedException e) {
+                                throw new RuntimeException(e);
+                            }
+                        } else {
+                            try {
```
ashvayka, line 133: First, just calculate the value of the "pause" variable and then use `try{ sleep }catch` block to avoid code duplication.

application/src/main/java/org/thingsboard/server/service/queue/processing/TbRuleEngineProcessingStrategyFactory.java
```diff
@@ -69,6 +76,12 @@ public RetryStrategy(String queueName, boolean retrySuccessful, boolean retryFai
             this.maxRetries = configuration.getRetries();
             this.maxAllowedFailurePercentage = configuration.getFailurePercentage();
             this.pauseBetweenRetries = configuration.getPauseBetweenRetries();
+            this.expPauseBetweenRetries = configuration.isExpPauseBetweenRetries();
+            if (this.expPauseBetweenRetries) {
+                this.expDegreeStep = new AtomicInteger(1);
+                this.maxExpPauseBetweenRetries = configuration.getMaxExpPauseBetweenRetries();
+                this.maxExpDegreeValue = Math.log(maxExpPauseBetweenRetries) / Math.log(pauseBetweenRetries);
```
ashvayka, line 83: This variable is redundant. You can use expDegreeStep and just compare the `maxExpPauseBetweenRetries` and `new Double(Math.pow(pauseBetweenRetries, expDegreeStep.getAndIncrement())).longValue();`

Issue #3420, opened 2020-09-03
[Feature Request] Cant use external Cassandra with Datastax


JoniSanchez:
Hello,

We need to connect thingsboard with an external Datastax with SSL connection. At least, the certificate of the server needs to be provided.

Currently, this is the code in the thingsboard/cassandra class "CassandraDriverOptions": 

```
@Value("${cassandra.ssl}")
private Boolean ssl;
@PostConstruct    
public void initLoader() {        
     ProgrammaticDriverConfigLoaderBuilder driverConfigBuilder =  DriverConfigLoader.programmaticBuilder();
         if (this.ssl) {            
             driverConfigBuilder.withString(DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS, "DefaultSslEngineFactory");        
         }
``` 
From my understanding, we need to integrate this further configuration.

```
@Value("${cassandra.ssl}")
private Boolean ssl;
@Value("${cassandra.trustStoreLocation}")
private String trustStoreLocation;
@Value("${cassandra.trustStorePassword}")
private String trustStorePassword;
@PostConstruct    
public void initLoader() {        
     ProgrammaticDriverConfigLoaderBuilder driverConfigBuilder =  DriverConfigLoader.programmaticBuilder();
         if (this.ssl) {            
             driverConfigBuilder.withString(DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS, "DefaultSslEngineFactory") 
                 .withString(DefaultDriverOption.SSL_TRUSTSTORE_PATH, this.trustStoreLocation)                    
                 .withString(DefaultDriverOption.SSL_TRUSTSTORE_PASSWORD, this.trustStorePassword); 
         }
```
It would be also nice that the keystore could be also set in the input config file.

TrevorMuraro:
Being able to set the Cassandra keystore and truststore in thingsboard.yml would definitely be preferable to the current method, but you can use Cassandra SSL with the current setup.

If the DataStax java driver isn't provided a keystore, keystore password, truststore or truststore password, it uses the JSSE system properties. That means that you can add this to thingsboard.conf:

```
export JAVA_OPTS="$JAVA_OPTS -Djavax.net.ssl.trustStore=/etc/thingsboard/conf/thingsboard_trust.p12 -Djavax.net.ssl.trustStoreType=pkcs12 -Djavax.net.ssl.trustStorePassword=the_password"
[...]

PR #3421, opened 2020-09-04
Add full DataStax Java Driver SSL configs

TrevorMuraro:
Inspired by the recommendations in [this issue](https://github.com/thingsboard/thingsboard/issues/3420#issue-692135699), added mechanisms to provide all DataStax DefautlSslEngineFactory properties through thingsboard.yml.

Tests passed with Cassandra SSL disabled, enabled with server-only authentication and enabled with two-way authentication. I also confirmed it works in my usual ThingsBoard dev setup, with two-way Cassandra SSL to a remote server.


ashvayka:
Thank you for this great contribution!

PR #3423, opened 2020-09-07
[3.2] Feature/Proto Converter

ShvaykaD:


common/data/src/main/java/org/thingsboard/server/common/data/device/profile/MqttDeviceProfileTransportConfiguration.java
```diff
@@ -16,11 +16,14 @@
 package org.thingsboard.server.common.data.device.profile;
 
 import lombok.Data;
+import org.thingsboard.server.common.data.TransportPayloadType;
 import org.thingsboard.server.common.data.DeviceTransportType;
 
 @Data
 public class MqttDeviceProfileTransportConfiguration implements DeviceProfileTransportConfiguration {
 
+    private TransportPayloadType transportPayloadType;
```
ashvayka, line 25: please assign default value.

common/queue/src/main/proto/queue.proto
```diff
@@ -66,11 +66,12 @@ enum SessionType {
 }
 
 enum KeyValueType {
-  BOOLEAN_V = 0;
-  LONG_V = 1;
-  DOUBLE_V = 2;
-  STRING_V = 3;
-  JSON_V = 4;
+  UNKNOWN = 0;
```
ashvayka, line 69: This may cause us problems during upgrade (if some data is in the queue). Why not UNKNOWN= something else?

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -216,22 +243,28 @@ private void handleGatewayPublishMsg(String topicName, int msgId, MqttPublishMes
     private void processDevicePublish(ChannelHandlerContext ctx, MqttPublishMessage mqttMsg, String topicName, int msgId) {
         try {
             if (deviceSessionCtx.isDeviceTelemetryTopic(topicName)) {
-                TransportProtos.PostTelemetryMsg postTelemetryMsg = adaptor.convertToPostTelemetry(deviceSessionCtx, mqttMsg);
+                TransportProtos.PostTelemetryMsg postTelemetryMsg = deviceSessionCtx.isJsonPayloadType() ?
```
ashvayka, line 246: var adaptor = deviceSessionCtx.getPayloadAdaptor(). Then you don't need to do a copy-paste all the time.

PR #3433, opened 2020-09-09
Bug/batch update deadlock

vzikratyi-tb:


dao/src/main/java/org/thingsboard/server/dao/sqlts/insert/latest/psql/PsqlLatestInsertTsRepository.java
```diff
@@ -151,4 +161,10 @@ public int getBatchSize() {
             }
         });
     }
+
+    private static List<TsKvLatestEntity> sortEntities(List<TsKvLatestEntity> entities) {
```
ashvayka, line 165: Insert of sorting in each saveOrUpdate, please introduce interface TbSqlQueueEntity that extends Comparable and highlight that the sorting should happen based on the primary key of SQL entity to avoid deadlocks in cluster mode.

ashvayka: or path Comparator<E> to the init function of the queue.

PR #3486, opened 2020-09-18
empty string validation fix

hardshell:
dashboard-form name validation fixed (used parent method, like in other entities), rule-chain-details name validation pattern and trim added

### DIFF
--- a/ui-ngx/src/app/modules/home/pages/dashboard/dashboard-form.component.ts
+++ b/ui-ngx/src/app/modules/home/pages/dashboard/dashboard-form.component.ts
@@ -95,11 +95,6 @@ export class DashboardFormComponent extends EntityComponent<Dashboard> {
     this.entityForm.patchValue({configuration: {description: entity.configuration ? entity.configuration.description : ''}});
   }
 
-  prepareFormValue(formValue: any): any {
-    formValue.configuration = {...(this.entity.configuration || {}), ...(formValue.configuration || {})};
-    return formValue;
-  }
-
   onPublicLinkCopied($event) {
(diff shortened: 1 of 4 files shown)


ui-ngx/src/app/modules/home/pages/dashboard/dashboard-form.component.ts
```diff
@@ -95,11 +95,6 @@ export class DashboardFormComponent extends EntityComponent<Dashboard> {
     this.entityForm.patchValue({configuration: {description: entity.configuration ? entity.configuration.description : ''}});
   }
 
-  prepareFormValue(formValue: any): any {
-    formValue.configuration = {...(this.entity.configuration || {}), ...(formValue.configuration || {})};
-    return formValue;
-  }
-
```
vvlladd28, line 102: This code broken saved dashboard. You need to use the parent class method prepareFormValue, before return value.

ui-ngx/src/app/modules/home/pages/rulechain/rulechain-page.component.ts
```diff
@@ -1550,6 +1550,11 @@ export class AddRuleNodeDialogComponent extends DialogComponent<AddRuleNodeDialo
 
   add(): void {
     this.submitted = true;
+
+    const formValue = this.ruleNodeDetailsComponent.ruleNodeFormGroup.value;
+
+    formValue.name = formValue.name.trim();
+
```
vvlladd28, line 1557: This code does nothing this function. Why added?

vvlladd28 (changes requested):

vvlladd28 (approved):

PR #3487, opened 2020-09-18
shared Attributes update improvements

YevhenBondarenko:
If the update of the shared attributes is originated by the user's REST API call - we push it to the device, if the update arrives from the device, we will not push it back to the device by default.


YevhenBondarenko:
UI PR https://github.com/thingsboard/thingsboard-rule-config-ui/pull/49

application/src/main/data/json/tenant/rule_chains/root_rule_chain.json
```diff
@@ -31,7 +31,8 @@
         "name": "Save Client Attributes",
         "debugMode": false,
         "configuration": {
-          "scope": "CLIENT_SCOPE"
+          "scope": "CLIENT_SCOPE",
+          "notifyDevice": "true"
```
ashvayka, line 35: please set this to false

application/src/main/data/json/demo/rule_chains/root_rule_chain.json
```diff
@@ -43,7 +43,8 @@
         "name": "Save Client Attributes",
         "debugMode": false,
         "configuration": {
-          "scope": "CLIENT_SCOPE"
+          "scope": "CLIENT_SCOPE",
+          "notifyDevice": "true"
```
ashvayka, line 47: please set this to false

PR #3514, opened 2020-09-24
Fix marker move and marker point

kalutkaz:


ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet-map.ts
```diff
@@ -410,10 +413,15 @@ export default abstract class LeafletMap {
     }
 
     convertToCustomFormat(position: L.LatLng): object {
-        return {
-            [this.options.latKeyName]: position.lat % 90,
-            [this.options.lngKeyName]: position.lng % 180
-        };
+      if(position.lng > 180){
```
ashvayka, line 416: where is check for lat?

kalutkaz: On maps, user can't set a point outside the map area in lat. This is probably a limitation at the leaflet level.

PR #3548, opened 2020-10-06
[2.5.5] fix enqueue methods impl in DefaultTbContext

ShvaykaD:


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rpc/TbSendRPCRequestNode.java
```diff
@@ -113,7 +113,7 @@ public void onMsg(TbContext ctx, TbMsg msg) {
             ctx.getRpcService().sendRpcRequestToDevice(request, ruleEngineDeviceRpcResponse -> {
                 if (!ruleEngineDeviceRpcResponse.getError().isPresent()) {
                     TbMsg next = ctx.newMsg(msg.getQueueName(), msg.getType(), msg.getOriginator(), msg.getMetaData(), ruleEngineDeviceRpcResponse.getResponse().orElse("{}"));
-                    ctx.enqueueForTellNext(next, TbRelationTypes.SUCCESS);
+                    ctx.enqueueForTellNext(next, next.getQueueName(), TbRelationTypes.SUCCESS, null, null);
```
ashvayka, line 116: This fixes usage of the API but not the problem with the API itself. Please update your fix with the following idea:

```java
    @Override
    public void enqueueForTellNext(TbMsg tbMsg, String relationType) {
        TopicPartitionInfo tpi = mainCtx.resolve(ServiceType.TB_RULE_ENGINE, tbMsg.getQueueName(), getTenantId(), tbMsg.getOriginator());
        enqueueForTellNext(tpi, tbMsg, Collections.singleton(relationType), null, null, null);
    }
```

ShvaykaD: Done.

PR #3550, opened 2020-10-07
[2.5.5] extracted duplicate to resolvePartition method in the enqueue method.

ShvaykaD:


application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -114,7 +116,7 @@ public void tellSelf(TbMsg msg, long delayMs) {
 
     @Override
     public void enqueue(TbMsg tbMsg, Runnable onSuccess, Consumer<Throwable> onFailure) {
-        TopicPartitionInfo tpi = mainCtx.resolve(ServiceType.TB_RULE_ENGINE, getTenantId(), tbMsg.getOriginator());
+        TopicPartitionInfo tpi = resolvePartition(tbMsg);
```
ashvayka, line 119: I beleive this is wrong. Enqueue is to put to a root rule chain. EnqueueForTellNext - to put into the same topic and same place in the rule chain.

ShvaykaD: Done

PR #3552, opened 2020-10-07
fixed sending inactivity event after device creation.

YevhenBondarenko:


application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -392,7 +393,7 @@ private void updateState() {
             if (stateData != null) {
                 DeviceState state = stateData.getState();
                 state.setActive(ts < state.getLastActivityTime() + state.getInactivityTimeout());
-                if (!state.isActive() && (state.getLastInactivityAlarmTime() == 0L || state.getLastInactivityAlarmTime() < state.getLastActivityTime())) {
+                if (!state.isActive() && (state.getLastInactivityAlarmTime() == 0L || state.getLastInactivityAlarmTime() < state.getLastActivityTime()) && UUIDs.unixTimestamp(deviceId.getId()) + state.getInactivityTimeout() < ts) {
```
ashvayka, line 396: Add device creation time to the DeviceStateData (device.getCreatedTime()) in the getOrFetchDeviceStateData

YevhenBondarenko: done

PR #3593, opened 2020-10-16
[3.3] Feature/attributes cache

vzikratyi-tb:


application/src/main/java/org/thingsboard/server/controller/AdminController.java
```diff
@@ -145,6 +149,16 @@ public void sendTestSms(@RequestBody TestSmsRequest testSmsRequest) throws Thing
         }
     }
 
+    @PreAuthorize("hasAuthority('SYS_ADMIN')")
+    @RequestMapping(value = "/cache/attributes/invalidate", method = RequestMethod.DELETE)
+    public void invalidateAttributesCache() throws ThingsboardException {
+        try {
+            clusterService.invalidateAttributesCache();
```
ashvayka, line 156: This will invalidate cache on a random server in a cluster mode. Not on all servers.

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -306,6 +322,28 @@ private void broadcast(ComponentLifecycleMsg msg) {
         }
     }
 
+    private void broadcast(AttributesCacheUpdatedMsg msg) {
+        byte[] msgBytes = encodingService.encode(msg);
```
ashvayka, line 326: Code duplication.

PR #3599, opened 2020-10-18
[3.3] Update to Latest Lombok 

Seelam-Ramesh-Reddy:
* Can start adding the Latest Lombok features.

*  Javax Dependency is needed separately since it has been removed after java-8


ikulikov:
Closing. Lombok has already been updated to the latest version in the current upstream to support JDK 11.

PR #3602, opened 2020-10-19
[3.2.1] Log telemetry update event to populate it into rule chains

vkukhtyn:


application/src/main/java/org/thingsboard/server/controller/BaseController.java
```diff
@@ -775,6 +769,13 @@ public static Exception toException(Throwable error) {
                         if (keys != null) {
                             keys.forEach(attrsArrayNode::add);
                         }
+                    } else if (actionType == ActionType.TIMESERIES_UPDATED) {
+                        List<TsKvEntry> telemetry = extractParameter(List.class, 0, additionalInfo);
+                        if (telemetry != null) {
+                            for (TsKvEntry tsKvEntry : telemetry) {
+                                addKvEntry(entityNode, tsKvEntry);
```
ashvayka, line 776: TsKvEntry also has a timestamp. The timestamp is lost in such a case. This is wrong. 
I suggest using the following format
```json
         [
            {
                "ts":1527863043000,
                "values":{
                    "battery":3.99,
                    "temperature":27.05
                }
            },
            {
                "ts":1527863044000,
                "values":{
                    "battery":3.98,
                    "temperature":27.06
                }
            }
        ]
```

vkukhtyn: In current implementation, the condition at line 772 will be true only when method _saveTelemetry()_ in _TelemetryController_ was called. The timestamp is set in the same _saveTelemetry()_ method (_TelemetryController_, line 425) and equals to current system time. 
So **_tsKvEntry_** element of the list will always have timestamp here. 
To process telemetry update according to proposed format it's necessary to add a new endpoint (to avoid API braking changes) in _TelemetryController_ in addition to existing ones: 
- POST /api/plugins/telemetry/{entityType}/{entityId}/timeseries/{scope}
- POST /api/plugins/telemetry/{entityType}/{entityId}/timeseries/{scope}/{ttl}
@ashvayka should I add a new endpoint to handling the proposed format of request body?

PR #3648, opened 2020-10-28
added default queue for device profile

YevhenBondarenko:


application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -159,6 +159,10 @@ private TbMsg transformMsg(TbMsg tbMsg, DeviceProfile deviceProfile) {
             if (targetRuleChainId != null && !targetRuleChainId.equals(tbMsg.getRuleChainId())) {
                 tbMsg = TbMsg.transformMsg(tbMsg, targetRuleChainId);
             }
+            String targetQueueName = deviceProfile.getDefaultQueueName();
```
ashvayka, line 162: Please add one more option and the following method

```
TbMsg.transformMsg(tbMsg, targetRuleChainId, targetQueueName)
```

ashvayka:
You forgot to add this to the transport layer. Please see

```
DefaultTransportService.resolveRuleChainId
```

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
YevhenBondarenko: done

Issue #3657, opened 2020-10-31
[Bug] Device credential validator throw an exception while creating a new device exception if database in short time inconsistency


chienfuchen32:
**Describe the bug**
Device credential validator throw an exception while creating a new device exception if database in short time inconsistency

**Your Server Environment**

* own setup
  * docker k8s deployment
  * ThingsBoard Version: v3.1.1
  * OS Name and Version: Ubuntu 18.04

**Your Client Environment**

**Desktop (please complete the following information):**

* OS: macOS 10.15.7
* Browser: chrome
* Version 86.0.4240.111 (x86_64)

**Your Device**

* Connectivity
  * HTTP

**To Reproduce**
1. Setup Postgresql HA from [bitnami/postgresql-ha](https://github.com/bitnami/charts/tree/master/bitnami/postgresql-ha)
2. Setup [Thingsboard k8s HA deployment](https://github.com/thingsboard/thingsboard/tree/master/k8s)
3. Try to run [Performance tests](https://github.com/thingsboard/performance-tests)
4. See error `"Can't assign device credentials to non-existent device!"`

**Expected behavior**
The performance tests program create many devices at the beginning, expected it should completed without problem.  
[Device save api](https://demo.thingsboard.io/swagger-ui.html#!/device-controller/saveDeviceUsingPOST) returns an exception "Can't assign device credentials to non-existent device!".  

Due to the database my deployment, postgresql with primary and standby (which is streaming replication) and `tb-node` connect with [Pgpool-II](https://www.pgpool.net/mediawiki/index.php/Main_Page).  

The current implementation of device credential validator does not allow any write/read data inconsistency through database.  
  
For example, tb-node write an new device record to primary database,   the validator check if the id existed in database in a short time.   
Sometimes it fail to find in slave database due to Postgresql streaming replication then cause the exception.
```java
    private DataValidator<DeviceCredentials> credentialsValidator =
            new DataValidator<DeviceCredentials>() {
                // ...
                @Override
                protected void validateDataImpl(TenantId tenantId, DeviceCredentials deviceCredentials) {
                    if (deviceCredentials.getDeviceId() == null) {
                        throw new DataValidationException("Device credentials should be assigned to device!");
                    }
                    if (deviceCredentials.getCredentialsType() == null) {
                        throw new DataValidationException("Device credentials type should be specified!");
                    }
                    if (StringUtils.isEmpty(deviceCredentials.getCredentialsId())) {
[...]

PR #3658, opened 2020-10-31
Fixing device credential validator unexpected exception while device in distributed database read/write in inconsistency.

chienfuchen32:
This is an approach to prevent database inconsistency deployment, add retry fetching device by id approach on device credential validate for both device (device credential) creating or device credential updating situation.

Thanks!

Fixing #3657

### DIFF
--- a/dao/src/main/java/org/thingsboard/server/dao/device/DeviceCredentialsServiceImpl.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/device/DeviceCredentialsServiceImpl.java
@@ -190,7 +190,11 @@ public class DeviceCredentialsServiceImpl extends AbstractEntityService implemen
                             found = true;
                         } else {
                             retryCount += 1;
-                            Thread.sleep(attemptRetryInterval);
+                            try {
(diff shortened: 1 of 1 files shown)


ashvayka:
The database should not be inconsistent. If you use postgres, it should have one master and few replicas. Is this the case?

chienfuchen32:
Hi, @ashvayka, Yes, In my case. I've setup a one master (primary) and 2 replicas (standby) on k8s which provided by [bitnami postgres ha](https://github.com/bitnami/charts/tree/master/bitnami/postgresql-ha). I think it's because the default postgres config `syncReplication` was false, which takes few time to copy data to replica node with lower consistency. At the same time It use pgpool to separate sql write/read command loading  to master/replica. It cause device create api failed. :( 

So what comes to my mind is keep retry find the device id while creating device credential in the code.

ashvayka:
I believe this should be fixed on the DB/PG Pool configuration side, not on the application layer. If we follow the same pattern, all our DB related code should be covered with constant retries :-)

PR #3617, opened 2020-10-22
[3.3] Added sending action notification to the gateway 

zbeacon:
If device created by the gateway was deleted.

PR #3675, opened 2020-11-04
[WIP][3.4] Feature/gateway payload proto

zbeacon:
Added minimized topics for the gateway (v2/g/...)
Added ability to change the gateway payload type by pushing to topic /payload
Changed payload type for ThingsBoard IoT gateway communication.

This PR based on PR #3617


common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -285,6 +298,14 @@ private void processDevicePublish(ChannelHandlerContext ctx, MqttPublishMessage
             } else if (topicName.equals(MqttTopics.DEVICE_CLAIM_TOPIC)) {
                 TransportProtos.ClaimDeviceMsg claimDeviceMsg = payloadAdaptor.convertToClaimDevice(deviceSessionCtx, mqttMsg);
                 transportService.process(deviceSessionCtx.getSessionInfo(), claimDeviceMsg, getPubAckCallback(ctx, msgId, claimDeviceMsg));
+            } else if (topicName.equals(MqttTopics.PAYLOAD_TYPE_TOPIC)) {
```
ashvayka, line 301: Why do we use this? We can take it from the device profile, right?

PR #3740, opened 2020-11-17
Support of Custom Protobuf payloads

ashvayka:


common/data/src/main/java/org/thingsboard/server/common/data/device/profile/ProtoTransportPayloadConfiguration.java
```diff
@@ -0,0 +1,200 @@
...
+        try {
+            ProtoFileElement protoFileElement = getTransportProtoSchema(protoSchema);
+            DynamicSchema dynamicSchema = getDynamicSchema(protoFileElement, schemaName);
+            String lastMsgName = getMessageTypes(protoFileElement.getTypes()).stream()
+                    .map(MessageElement::getName).reduce((previous, last) -> last).get();
+            DynamicMessage.Builder builder = dynamicSchema.newMessageBuilder(lastMsgName);
+            return builder.getDescriptorForType();
+        } catch (Exception e) {
+            log.warn("Failed to get Message Descriptor due to {}", e.getMessage());
+            return null;
```
ashvayka, line 76: throw new RuntimeException("Failed to get Message Descriptor due to :" + e.getMessage()); to be consistent with the getDynamicSchema

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/adaptors/ProtoMqttAdaptor.java
```diff
@@ -45,20 +50,24 @@
 
     @Override
     public TransportProtos.PostTelemetryMsg convertToPostTelemetry(MqttDeviceAwareSessionContext ctx, MqttPublishMessage inbound) throws AdaptorException {
+        DeviceSessionCtx deviceSessionCtx = (DeviceSessionCtx) ctx;
         byte[] bytes = toBytes(inbound.payload());
+        Descriptors.Descriptor telemetryDynamicMsgDescriptor = getDescriptor(deviceSessionCtx.getTelemetryDynamicMsgDescriptor());
         try {
-            return ProtoConverter.convertToTelemetryProto(bytes);
-        } catch (InvalidProtocolBufferException | IllegalArgumentException e) {
+            return JsonConverter.convertToTelemetryProto(new JsonParser().parse(dynamicMsgToJson(bytes, telemetryDynamicMsgDescriptor)));
```
ashvayka, line 57: Please add the TODO to improve this in the future releases. We need two types of improvements:
1) No need to convert to JSON and back. If we know the structure of the message we can dynamically create our PostTelemetryMsg without JSON
2) Need to support Protobuf for all other methods (and publish the schema somewhere)

PR #3800, opened 2020-12-02
Trip animation Widget Point colour Function

ecallawevad:
Trip animation colour point function to enable historical points to be mapped by colour according to telemetry data values.


vvlladd28:
Hi @ecallawevad,
Your commit overwrites the last change in the map widget. Please update your code or will write the reason for such changes)

PR #3840, opened 2020-12-11
[3.3] Lwm2m tests

nickAS21:
add Delayed Request


ui-ngx/angular.json
```diff
@@ -247,6 +247,7 @@
   },
   "defaultProject": "thingsboard",
   "cli": {
-    "packageManager": "yarn"
+    "packageManager": "yarn",
+    "analytics": false
```
vvlladd28, line 251: This parameter does not need to be changed

PR #3957, opened 2021-01-15
[2.6] Edge functionality

volodymyr-babak:


ui/src/app/api/entity.service.js
```diff
@@ -916,9 +962,7 @@ function EntityService($http, $q, $filter, $translate, $log, userService, device
                 entityFieldKeys.push(types.entityField.phone.keyName);
                 break;
             case types.entityType.entityView:
```
ChantsovaEkaterina, line 964: Entity view has no label field, only name and type entity fields should be available

ui/src/app/event/event-row.directive.js
```diff
@@ -93,14 +94,6 @@ export default function EventRowDirective($compile, $templateCache, $mdDialog, $
             });
         }
 
-        scope.checkTooltip = function($event) {
```
ChantsovaEkaterina, line 96: Function needs to be restored, it is used.

PR #3958, opened 2021-01-15
[WIP] [3.3] Fix for lastActivityTime in devices created by gateway

zbeacon:


YevhenBondarenko:
LGTM

ashvayka:
We need to store this "boolean" parameter only in the gateway session. No need to update all device sessions if one parameter in the gateway changed.

PR #3963, opened 2021-01-18
Change executions permissions to docker containers

gotelgest:
The default user hasn't permissions to run the *.sh script in the container

### DIFF
--- a/msa/js-executor/docker/Dockerfile
+++ b/msa/js-executor/docker/Dockerfile
@@ -28,6 +28,8 @@ RUN systemctl --no-reload disable --now ${pkg.name}.service > /dev/null 2>&1 || 
 
 RUN chmod 555 ${pkg.installFolder}/bin/${pkg.name}
 
+RUN chmod 555 /usr/bin/start-js-executor.sh
+
 USER ${pkg.user}
 
 CMD ["start-js-executor.sh"]
--- a/msa/tb-node/docker/Dockerfile
+++ b/msa/tb-node/docker/Dockerfile
@@ -30,6 +30,8 @@ RUN chown -R ${pkg.user}:${pkg.user} /tmp
 
 RUN chmod 555 ${pkg.installFolder}/bin/${pkg.name}.jar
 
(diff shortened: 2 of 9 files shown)


ashvayka:
is there any particular reason why you decided to do this?

gotelgest:
The thingsboard user haven't permissions to execute the script that init any container

ashvayka:
@volodymyr-babak  please review

ashvayka:
Please provide few examples why do you need this. I have just checked the update procedure and the following command works fine:

```bash
docker run -it -v ~/.mytb-data:/data --rm thingsboard/tb-postgres upgrade-tb.sh
```

nicolabortignon:
+1

`docker run -it -v ~/.mytb-data:/data --rm thingsboard/tb-postgres upgrade-tb.sh

mkdir: cannot create directory ‘/data/db’: Operation not permitted
The files belonging to this database system will be owned by user "thingsboard".
This user must also own the server process.

The database cluster will be initialized with locale "C.UTF-8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

creating directory /data/db ... initdb: could not create directory "/data/db": Operation not permitted`

ashvayka:
Hi @nicolabortignon , are you running this on linux or windows? In case of windows, you should manually create docker volumes:
```
docker volume create mytb-data
docker volume create mytb-logs
```
and use slightly different command:

```
docker run -it -v mytb-data:/data --rm thingsboard/tb-postgres upgrade-tb.sh
```

This is mentioned in the https://thingsboard.io/docs/user-guide/install/docker-windows/

nicolabortignon:
For posterity. 
this is running on linux/osx through toker. 

The hacky way to solve it is to chmod 777 the local folder at installation time. I'd discourage to give 777 permission to folders without a very clear reasons.

A better way would be to fix the user issues so the user that has permission to run the docker image has also access to those 2 folders.

PR #3964, opened 2021-01-18
[WIP] [3.3] Add keys to DeviceProfile

DanyloKobzar:


DanyloKobzar:
> What is this "keys" about? How d you plan to use them?

Device profile store the default telemetry keys which would be used in EntityView creation with DEVICE entity type

ashvayka:
The use case for this feature is not well described in the PR. Also, there is no UI support for the configuration of the entity keys. Also, merging the entity view keys seems redundant. The better option would be to automatically create an entity view when we create a device if the corresponding settings in the device profile are set. So, I am closing the PR as incomplete.

PR #3968, opened 2021-01-19
[WIP] [3.4] Entity Config Versioning

mp-loki:
Introducing EntityConfig object, which is saved each time when time the Dashboard or Rule Chain entity is changed.
Allows restoring the Dashboard/Rule chain configuration to previous configuration version (by creating a new entity configuration each time).
Provides API to list Entity Configurations, which will allow to embed this feature to UI.
Gracefully deletes configurations upon entity deletion.
Install & Upgrade scripts updated. Upon upgrade, for each dashboard/rule chain, the existing configuration will be saved as EntityConfig version 1.

This feature can be further enhanced by adding "save as new" endpoint, which will allow visually comparing different versions of the same object.
It can also be extrapolated to Widget Bundles and other objects that have sensitive configuration.


mp-loki:
PR updated with origin/develop/3.3 branch

application/src/main/java/org/thingsboard/server/controller/DashboardController.java
```diff
@@ -107,17 +108,61 @@ public Dashboard getDashboardById(@PathVariable(DASHBOARD_ID) String strDashboar
...
+    @RequestMapping(value = "/dashboard/{dashboardId}/entityConfig/{entityConfigId}/restore", method = RequestMethod.POST)
+    @ResponseBody
+    public Dashboard restoreDashboardConfig(@PathVariable(DASHBOARD_ID) String strDashboardId,
+                                            @PathVariable("entityConfigId") String strEntityConfigId,
+                                            @RequestParam(required = false) String comment) throws ThingsboardException {
+        checkParameter(DASHBOARD_ID, strDashboardId);
+        checkParameter("entityConfigId", strEntityConfigId);
+        try {
+            DashboardId dashboardId = new DashboardId(toUUID(strDashboardId));
+            checkDashboardId(dashboardId, Operation.READ);
```
ashvayka, line 121: So, the user with read-only permissions can restore an old copy of the dashboard?

application/src/main/java/org/thingsboard/server/controller/DashboardController.java
```diff
@@ -107,17 +108,61 @@ public Dashboard getDashboardById(@PathVariable(DASHBOARD_ID) String strDashboar
...
+    public Dashboard restoreDashboardAsNew(@PathVariable(DASHBOARD_ID) String strDashboardId,
+                                            @PathVariable("entityConfigId") String strEntityConfigId,
+                                            @RequestParam(required = false) String comment,
+                                            @RequestParam(required = false) String title) throws ThingsboardException {
+        checkParameter(DASHBOARD_ID, strDashboardId);
+        checkParameter("entityConfigId", strEntityConfigId);
+        try {
+            DashboardId dashboardId = new DashboardId(toUUID(strDashboardId));
+            Dashboard dashboard = checkDashboardId(dashboardId, Operation.WRITE);
+            checkDashboardId(dashboardId, Operation.READ);
```
ashvayka, line 141: you check for write permissions on the previous line. Why do you need another check for read?

PR #3984, opened 2021-01-24
Dynamic color for data points

sukiseka:
Trip Animation at the moment only displays a static color for data point. This change is to provide option to dynamically display color based on different parameters for each data point. Show Color Point function.


sukiseka:
Closing due to conflicts

PR #4025, opened 2021-01-29
addede current customer and current tenant dynamic source types to DeviceProfile key filter

YevhenBondarenko:


rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/profile/TbDeviceProfileNodeTest.java
```diff
@@ -188,6 +191,7 @@ private void init() throws TbNodeException {
         Mockito.when(ctx.getDeviceProfileCache()).thenReturn(cache);
         Mockito.when(ctx.getTimeseriesService()).thenReturn(timeseriesService);
         Mockito.when(ctx.getAlarmService()).thenReturn(alarmService);
+        Mockito.when(ctx.getDeviceService()).thenReturn(deviceService);
```
ashvayka, line 194: No tests for the customer and tenant dynamic values

PR #4052, opened 2021-02-03
created calculate delta node

YevhenBondarenko:


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/metadata/CalculateDeltaNode.java
```diff
@@ -0,0 +1,175 @@
...
+                        }
+                        result = Double.parseDouble(str);
+                    } catch (NumberFormatException e) {
+                        throw new IllegalArgumentException("Calculation failed. Unable to parse value [" + str + "]" +
+                                " of telemetry [" + kvEntry.getKey() + "] to Double");
+                    }
+                    break;
+            }
+        }
+        return new ValueWithTs(ts, result);
```
ashvayka, line 163: So, if there was no previous reading, we will calculate the difference between the current reading and 0. This is wrong.

PR #4011, opened 2021-01-27
Rest Client unable to get device types fix #3992

AndrewVolosytnykhThingsboard:

PR #4065, opened 2021-02-05
Update RestClient.java

ropeal:
getDeviceTypes in RestClient should use /api/device/types rest api rather than /api/devices


YevhenBondarenko:
this issue has fixed in the master branch 
#4011

PR #4092, opened 2021-02-12
Handle case when device was removed from db but message in the queue

volodymyr-babak:


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/TbDeviceProfileNode.java
```diff
@@ -134,7 +134,12 @@ public void onMsg(TbContext ctx, TbMsg msg) throws ExecutionException, Interrupt
                     if (deviceState != null) {
                         deviceState.process(ctx, msg);
                     } else {
-                        ctx.tellFailure(msg, new IllegalStateException("Device profile for device [" + deviceId + "] not found!"));
+                        if (ctx.getDeviceService().findDeviceById(ctx.getTenantId(), deviceId) != null) {
```
ashvayka, line 137: No need to execute a query to the database again. It is already executed in the 
```java
getOrCreateDeviceState(ctx, deviceId, null);
```

Please change log level from warn to info as well. Then I will merge.

ashvayka:
:+1:

PR #4116, opened 2021-02-17
Use msg queue instead of default

volodymyr-babak:


common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportService.java
```diff
@@ -764,7 +764,7 @@ protected void sendToDeviceActor(TransportProtos.SessionInfoProto sessionInfo, T
     }
 
     protected void sendToRuleEngine(TenantId tenantId, TbMsg tbMsg, TbQueueCallback callback) {
```
ashvayka, line 766: Please make this method private, to make sure no one will invoke him except 

```java
    protected void sendToRuleEngine(TenantId tenantId, DeviceId deviceId, TransportProtos.SessionInfoProto sessionInfo, JsonObject json,
                                    TbMsgMetaData metaData, SessionMsgType sessionMsgType, TbQueueCallback callback) {
```

volodymyr-babak: done

PR #4136, opened 2021-02-22
[3.2.2] added fix for partitions save

ShvaykaD:
reject old partitions save due to DataIntegrityViolationException


dao/src/main/java/org/thingsboard/server/dao/sqlts/psql/JpaPsqlTimeseriesDao.java
```diff
@@ -114,6 +116,12 @@ private void savePartition(PsqlPartition psqlPartition) {
                 partitioningRepository.save(psqlPartition);
                 log.trace("Adding partition to Set: {}", psqlPartition);
                 partitions.put(psqlPartition.getStart(), psqlPartition);
+            } catch (DataIntegrityViolationException ex) {
+                log.trace("Error occurred during partition save:", ex);
+                if (ex.getCause() instanceof ConstraintViolationException) {
+                    log.warn("Saving partition [{}] rejected. Timeseries data will save to the ts_kv_indefinite (DEFAULT) partition.", psqlPartition.getPartitionDate());
```
ShvaykaD, line 122: Removed ERROR info from WARN cause it will be duplicated by SqlExceptionHelper

PR #4164, opened 2021-02-25
[WIP]added redis stores support to lwm2m transport

YevhenBondarenko:


common/transport/lwm2m/pom.xml
```diff
@@ -66,6 +70,10 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-context</artifactId>
         </dependency>
+        <dependency>
```
YevhenBondarenko, line 73: remove redis dependency

PR #4167, opened 2021-02-26
Async timeout connector customizer. Refactoring

volodymyr-babak:


common/transport/http/src/main/java/org/thingsboard/server/transport/http/HttpTransportContext.java
```diff
@@ -37,4 +41,14 @@
     @Value("${transport.http.request_timeout}")
     private long defaultTimeout;
 
+    @Bean
+    public TomcatConnectorCustomizer tomcatAsyncTimeoutConnectorCustomizer() {
+        return connector -> {
+            ProtocolHandler handler = connector.getProtocolHandler();
+            if (handler instanceof Http11NioProtocol) {
+                log.trace("Setting async timeout {}", defaultTimeout);
+                connector.setAsyncTimeout(defaultTimeout);
```
ashvayka, line 50: This is wrong, because subscribeToCommands and subscribeToAttributes are able to overwrite the default timeout. I believe we should introduce new setting "Max Request Timeout" that will be used here and will be initialized with default value (5 min?)

PR #4185, opened 2021-03-02
[3.3] [PROD-806] Implement alarm condition metadata displaying

ViacheslavKlimov:


YevhenBondarenko (approved):

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/AlarmState.java
```diff
@@ -173,10 +174,20 @@ public void pushMsg(TbContext ctx, TbAlarmResult alarmResult) {
             relationType = "Alarm Cleared";
             metaData.putValue(DataConstants.IS_CLEARED_ALARM, Boolean.TRUE.toString());
         }
+        setAlarmConditionMetadata(ruleState, metaData);
         TbMsg newMsg = ctx.newMsg(lastMsgQueueName != null ? lastMsgQueueName : ServiceQueue.MAIN, "ALARM", originator, metaData, data);
         ctx.tellNext(newMsg, relationType);
     }
 
+    protected void setAlarmConditionMetadata(AlarmRuleState ruleState, TbMsgMetaData metaData) {
+        if (ruleState.getSpec().getType() == AlarmConditionSpecType.REPEATING) {
```
ashvayka, line 183: I believe it is better to put JSON representation of ruleState.getState() into the metadata "as-is". It will be more generic and we don't need to support this code if we change something in the PersistedAlarmRuleState

ViacheslavKlimov: We do not need to add the lastEventTs to the alarm event metadata as there is already defined field ("ts") with the same value

PR #4187, opened 2021-03-02
added ping for WS

YevhenBondarenko:


application/src/main/java/org/thingsboard/server/controller/plugin/TbWebSocketHandler.java
```diff
@@ -120,6 +134,9 @@ public void afterConnectionEstablished(WebSocketSession session) throws Exceptio
                 return;
             }
             internalSessionMap.put(internalSessionId, new SessionMetaData(session, sessionRef, maxMsgQueuePerSession));
+
+            pingExecutor.scheduleWithFixedDelay(() -> internalSessionMap.get(internalSessionId).sendPing(), 10000, 10000, TimeUnit.MILLISECONDS);
```
ashvayka, line 138: No need to schedule separate function execution per session. You should execute this function and loop through all the sessions. This code also has a problem that you do not cancel this task after session is closed. Do you wait/analyze the reply or not really? If you wait for the reply, it may be a very time-consuming task (better to send ping in the same thread pool where you send the updates over websockets if this call is blocking).

application/src/main/java/org/thingsboard/server/controller/plugin/TbWebSocketHandler.java
```diff
@@ -215,14 +206,14 @@ private TelemetryWebSocketSessionRef toRef(WebSocketSession session) throws IOEx
...
             this.sessionRef = sessionRef;
             this.msgQueue = new LinkedBlockingQueue<>(maxMsgQueuePerSession);
-            this.lastActivityTime = new AtomicLong(System.currentTimeMillis());
+            this.lastActivityTime = System.currentTimeMillis();
         }
 
         synchronized void sendPing() {
             try {
-                if (System.currentTimeMillis() - lastActivityTime.get() >= pingTimeout) {
+                if (System.currentTimeMillis() - lastActivityTime >= pingTimeout) {
```
ashvayka, line 214: please extract System.currentTimeMillis() to a variable and call it only once per all sessions and not twice per session.

application/src/main/java/org/thingsboard/server/controller/plugin/TbWebSocketHandler.java
```diff
@@ -189,13 +197,32 @@ private TelemetryWebSocketSessionRef toRef(WebSocketSession session) throws IOEx
...
             this.asyncRemote = nativeSession.getAsyncRemote();
             this.sessionRef = sessionRef;
             this.msgQueue = new LinkedBlockingQueue<>(maxMsgQueuePerSession);
+            this.lastActivityTime = System.currentTimeMillis();
+        }
+
+        synchronized void sendPing(long currentTime) {
+            try {
+                if (currentTime - lastActivityTime >= pingTimeout) {
+                    this.asyncRemote.sendPing(ByteBuffer.wrap(new byte[]{}));
```
ashvayka, line 215: Why not extract this to a constant?

PR #4209, opened 2021-03-05
[3.3] [PROD-800] Fix empty search result when query is empty

ViacheslavKlimov:


dao/src/main/java/org/thingsboard/server/dao/sql/query/EntityKeyMapping.java
```diff
@@ -540,16 +540,14 @@ private String buildStringPredicateQuery(QueryContext ctx, String field, StringF
                 stringOperationQuery = String.format("%s like :%s) or (%s is null and :%s = '%%')", operationField, paramName, operationField, paramName);
                 break;
             case CONTAINS:
-                if (value.length() > 0) {
-                    value = "%" + value + "%";
-                }
-                stringOperationQuery = String.format("%s like :%s) or (%s is null and :%s = '')", operationField, paramName, operationField, paramName);
+                value = "%" + value + "%";
+                stringOperationQuery = String.format("%s like :%s) or (%s is null and :%s = '%%%%')", operationField, paramName, operationField, paramName);
```
ashvayka, line 544: Please explain this case.

ViacheslavKlimov: If the search value is empty then it is set as "%%". In this case we will search for entities with field "LIKE '%%'" (all non-null values) and also entities with field set to null ("OR field IS NULL" when query = '%%'")

YevhenBondarenko (approved):

PR #4223, opened 2021-03-11
PROD-577 Add dynamic duration support in Device profile alarm condition 

DanyloKobzar:
Implemented only duration condition

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/AlarmRuleState.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/AlarmRuleState.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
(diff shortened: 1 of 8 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/AlarmRuleState.java
```diff
@@ -72,19 +71,13 @@
...
         }
         this.spec = getSpec(alarmRule);
-        long requiredDurationInMs = 0;
         long requiredRepeats = 0;
         switch (spec.getType()) {
-            case DURATION:
-                DurationAlarmConditionSpec duration = (DurationAlarmConditionSpec) spec;
-                requiredDurationInMs = duration.getUnit().toMillis(duration.getValue());
-                break;
             case REPEATING:
```
ashvayka, line 76: If you make this feature support for DURATION, please add the same for REPEATING.

PR #4224, opened 2021-03-11
changed msg originator to alarmId for ack and clear alarm events

YevhenBondarenko:


ashvayka:
This change introduced bug in the DeviceProfileNode

PR #4227, opened 2021-03-11
Refactoring of migration tool according to the last TB DB structure

AndrewVolosytnykhThingsboard:


tools/pom.xml
```diff
@@ -54,14 +54,17 @@
         <dependency>
             <groupId>org.apache.cassandra</groupId>
             <artifactId>cassandra-all</artifactId>
+            <version>3.11.10</version>
```
ashvayka, line 57: Why do you overwrite the dependency version that is managed in the parent pom.xml?

AndrewVolosytnykhThingsboard: Fixed in 55f5602bdf2b9c2479ffcb8114708b5166684dab commit

tools/src/main/java/org/thingsboard/client/tools/migrator/README.md
```diff
@@ -21,37 +21,41 @@ It will generate single jar file with all required dependencies inside `target d
...
-        -latestFrom ./source/ts_kv_latest.dmp
+        -telemetryFrom /home/user/dump/ts_kv_all.dmp 
+        -relatedEntities /home/user/dump/relatedEntities.dmp 
         -latestOut /home/ubunut/migration/ts_latest 
-        -tsFrom ./source/ts_kv.dmp
-        -tsOut /home/ubunut/migration/ts  
-        -partitionsOut /home/ubunut/migration/ts_partition
-        -castEnable false
+        -tsOut /home/ubunut/migration/ts 
+        -partitionsOut /home/ubunut/migration/ts_partition
```
YevhenBondarenko, line 55: replace "ubunut"

YevhenBondarenko (approved):

PR #4235, opened 2021-03-12
UI: extend table widget settings

ChantsovaEkaterina:
For entities, timeseries, alarms tables:
- row style function added;
- for cell style function added new parameters: entity / rowData / alarm and ctx;
- for cell content function added custom translations support;
- option to disable sticky header / sticky action cells;
- ability to hide columns by default;
- ability to hide columns in list Columns to Display;


hetvishah08:
I couldn't find the below feature in timeseries table

> "ability to hide columns in list Columns to Display;"

Issue #1726, opened 2019-05-18
Customer Unable to Update Attributes


JordanNel:
Hi there,

So I have been trying to get this to work for quite some time now without any success. The problem I'm having has to do with updating attributes from a Customers Dashboard. I have assigned the device who's attributes I would like to change and a Dashboard to make the changes to the Customer. 

If I login to my tenant account I am able to successfully update the attributes either through the "Device Attributes" window, or through the Dashboard I have setup with an "Update Shared String Attribute" widget. This is one of the system widgets. However when I login to the customers account, and follow the exact same process, the update fails. This is the case for any type of attribute I have tried (Integer, Boolean, String, etc.). I am using the latest version of Thingsboard at the time of posting - 2.3.1. 

I have a few images to help explain my problem:

**Customer Dashboard:**
![Capture1](https://user-images.githubusercontent.com/42270123/57966547-b5bfc080-7953-11e9-8a95-e2f33ee15bbf.PNG)

**Tenant Dashboard (Same Dashboard)**
![Capture2](https://user-images.githubusercontent.com/42270123/57966550-bf492880-7953-11e9-9c56-fa912d5f9f97.PNG)

![Capture4](https://user-images.githubusercontent.com/42270123/57966567-1222e000-7954-11e9-8a16-9be58538c82f.PNG)

**Attempt to Delete attribute using Customer Login JWT**
![image4144](https://user-images.githubusercontent.com/42270123/57966569-1a7b1b00-7954-11e9-8183-e9e10f82f51e.png)

**Requesting attributes works using customer JWT**
![image4145](https://user-images.githubusercontent.com/42270123/57966581-41395180-7954-11e9-9508-993a3222fd74.png)

**Successfully able to Delete attribute using Tenant Login JWT**
![rect4153-3-1-0-8-6](https://user-images.githubusercontent.com/42270123/57966595-878eb080-7954-11e9-9560-a3e424d19a88.png)

**The error I receive from Google Chrome console when using Customer Login**
![rect4399-0-2-5](https://user-images.githubusercontent.com/42270123/57966603-a55c1580-7954-11e9-9ecb-54a94101ec49.png)

I'm not sure if there are additional rules I need to add the rule chain to get this to work. I'm currently using the default rule chain with no changes. It seems like the customer does not have permission to change the devices attributes, but I am unsure how to go about giving the customer permission.

I would really appreciate any help with this.

Thanks

pgrisu:
[...]

Issue #4237, opened 2021-03-12
Cannot Update Shared Attribute with the assigned user credential


openedhardware:
**Description**
I have made a custom widget to update the devices' shared attributes by using the following code and added it to a dashboard:

```javascript
attributeService = $scope.$injector.get(self.ctx.servicesMap.get('attributeService'));
attributeService.saveEntityAttributes(
            self.ctx.datasources[index].entity.id,
            'SHARED_SCOPE',
            [
                {
                    key: name,
                    value: val
                }
            ]
        ).subscribe(
            function success() {
                alert('Successfully Updated!');
            },
            function fail() {
                
            }
        );
```
If I login to the dashboard with a tenant admin, I can update the shared attribute.

But if I assign the dashboard to a customer and login to the dashboard with a customer user, I cannot update the shared attribute of the device:

![image](https://user-images.githubusercontent.com/16890683/110933626-f6c66780-8367-11eb-893a-71183207402d.png)

![image](https://user-images.githubusercontent.com/16890683/110934131-97b52280-8368-11eb-808c-4585da6507f0.png)

Do I need a permit to allow customer users to modify the shared attributes of the device that is assigned to them?

Thanks! 

**Environment**
 * OS:  Ubuntu 18.04
 * ThingsBoard: version 3.2.0
 * Browser: Latest Chrome

ahmed-1a:
CE v3.2.0

I used the widget for updating a server side attribute, I also received the same error, a customer user couldn't update it.
A tenant user can without a problem. 

The use of the server-side attribute is a threshold value for an alarm related to a device.

jpeg2600:
I'm trying to figure out the solution proposed in #1726 if it can help you

Is there any plan to fix it in CE ?

ahmed-1a:
It seems the [solution](https://github.com/thingsboard/thingsboard/issues/1726#issuecomment-774536729) is the easiest but it requires rebuilding thingsboard. 

But I haven't tested it myself.

openedhardware:
Yes, built from source again and it works! :)
Here is a result deb file if you want - http://0x0.st/-ZRZ.deb

EmbeddedDevver:
> Yes, built from source again and it works! :)
> Here is a result deb file if you want - http://0x0.st/-ZRZ.deb

Can you show me the exact file changes that you have done? Or maybe send me the new file so I can compare it with the original?

[...]

PR #4246, opened 2021-03-15
Authorized USERS able to adjust Device Attributes

EmbeddedDevver:
Fixes bug #1726 and #4237 

Currently only logged in TENANTS are able to change device attributes using a dashboard.
This should also be logged in USERS as this was already possible in versions prior thingsboard 2.3.0
I think this option was removed by accident due the new security engine implementation in thingsboard.


ashvayka:
@YevhenBondarenko please check this.

ashvayka:
Thank you for reporting this. We have fixed it in a correct way in this commit: https://github.com/thingsboard/thingsboard/commit/60c141d3ca5b154fa48fd0a9c047c6ca504962a4

PR #4253, opened 2021-03-16
Fix for test script node function

zbeacon:


application/src/main/java/org/thingsboard/server/controller/RuleChainController.java
```diff
@@ -432,4 +433,22 @@ private String msgToOutput(TbMsg msg) throws Exception {
         return objectMapper.writeValueAsString(msgData);
     }
 
+    private String msgToOutput(List<TbMsg> msgs) throws Exception {
+        ArrayNode resultNode = objectMapper.createArrayNode();
```
ashvayka, line 437: No need to crate the arrayNode if msgs.size = 1.

zbeacon: Changed according to comment.

PR #4260, opened 2021-03-18
[3.2.2] [PROD-851] Fix invalid alarm clearing behaviour in the cluster mode

ViacheslavKlimov:
**The case:**
When profile of a device is changed, the event is pushed to specific partition that is responsible for this device and thus the device profile cache on this core service instance will be updated, unlike other core instances' caches. If we create new alarm it will go correctly to the new rule chain, but when we then clear this alarm from UI the request may go to the core instance with outdated device profile cache, and therefore the clear alarm event will be pushed to the wrong rule chain that is assigned to the old device profile, and сonsequently when a new alarm comes it will be processed by the new rule chain where there is no clear alarm event and so the alarm will no be created but updated.


YevhenBondarenko (approved):

ashvayka:
Great research!

Issue #2431, opened 2020-02-19
Entity assign/unassigned not triggered when claiming/reclaiming devices


Shawnshank:
Hi!

Is there a way to detect claims and reclaims in the rule engine? The most obvious is to use the "Entity Assigned" and "Entity Unassigned" message types from the Message Type Switch node, however, I noticed that this will only trigger when a tenant assigns or unassign a device from a customer.

I would suggest that claiming/reclaiming devices should trigger the same event as when a tenant assigns and unassign a device to be able to notice a claim/reclaim in the rule engine.

ashvayka:
Fixed and will be part of 3.3

Shawnshank:
Thanks for the update :+1:

PR #4286, opened 2021-03-23
[3.3] [PROD-239] Create assign/unassign device event when claiming/reclaiming device

ViacheslavKlimov:


ViacheslavKlimov:
#2431

PR #4290, opened 2021-03-23
corrected function name

armp30:
function name was incorrect due to the request url


ikulikov:
@armp30 Thx for contribution. Already fixed in master branch.

PR #4299, opened 2021-03-24
SNMP transport improvement and refactoring

ViacheslavKlimov:


common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
```diff
@@ -15,114 +15,75 @@
...
-    private void initializeSnmp() {
-        try {
-            this.snmp = new Snmp(new DefaultUdpTransportMapping());
-            this.snmp.listen();
-        } catch (IOException e) {
-            //TODO: what should be done if transport wasn't initialized?
-            log.error(e.getMessage(), e);
-        }
+    private void initializeSnmp() throws IOException {
+        snmp = new Snmp(new DefaultUdpTransportMapping());
```
ashvayka, line 80: Maybe someone would like to support SNMP over TCP? Let's add TODO to make this configurable.

common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
```diff
@@ -141,134 +102,124 @@ private void executeSnmpRequest(DeviceSessionContext sessionContext) {
...
-                } catch (IOException e) {
-                    log.error(e.getMessage(), e);
+                } catch (Exception e) {
+                    log.error("Failed to send SNMP request: {}", e.getMessage());
                 }
             });
         }
     }
 
     public void onNewDeviceResponse(ResponseEvent responseEvent, DeviceSessionContext sessionContext) {
```
ashvayka, line 114: Usually, we put the sessionContext as the first parameter.

PR #4305, opened 2021-03-25
added a fix for string value RPC params: removed redundant escaped characters

ShvaykaD:
fix needed for Protobuf, cause in the protobuf adaptors we sending directly TransportProtos.ToDeviceRpcRequestMsg bytes.


ashvayka:
The logic that is based on exceptions is definitely not the best but maybe ok in this case.

PR #4317, opened 2021-03-26
Provide additional validation for entities

ViacheslavKlimov:


ashvayka:
The antisamy library we [use](https://mvnrepository.com/artifact/org.owasp.antisamy/antisamy/1.6.2) here has compile-time dependency for LGPL 2.1 [library](https://mvnrepository.com/artifact/com.github.spotbugs/spotbugs-annotations). It is marked as optional. Please make sure we don't use it and exclude it from maven dependency.

dao/pom.xml
```diff
@@ -111,6 +111,16 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <version>${hibernate-validator.version}</version>
```
ashvayka, line 117: Move to the parent pom.

Issue #4242, opened 2021-03-13
[Feature Request] Add bulk device creation API method


viktorkho:
**Is your feature request related to a problem? Please describe.**
We are going to integrate TB with our provisioning subsystem, we want to create multiple (hundreds of) devices with single request. Now we have to wait a long time while TB creates devices, including CSV import.

**Describe the solution you'd like**
We expect TB to provide API method for bulk devices creation.

Hardikrathod01:
Hi,
I would like to start working on this feature,has anybody started working on this feature?

dharapatel1596:
Hi,
I would like to start working on this feature,has anybody started working on this feature?
Also, if nobody has picked this up yet, then how do you suggest to approach this? I believe at the moment, each device's payload is picked up and device is created. Then another payload for another device and repeat the procedure.

Instead I can think of a common approach where, CSV import sends all the payloads at one go to the API and we modify API as well as front end to support this.

PR #4322, opened 2021-03-28
[application] Added bulk devices creation api

dharapatel1596:
This api addresses the issue #4242.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/controller/DeviceController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/DeviceController.java
@@ -141,6 +141,45 @@ public class DeviceController extends BaseController {
         }
     }
 
+    @PreAuthorize("hasAnyAuthority('TENANT_ADMIN', 'CUSTOMER_USER')")
+    @RequestMapping(value = "/bulk/devices", method = RequestMethod.POST)
+    @ResponseBody
+    public List<Device> saveBulkDevices(@RequestBody List<Device> devices,
+                                        @RequestParam(name = "accessToken", required = false) String accessToken) throws ThingsboardException {
+        List<Device> savedDevices = new ArrayList<>();
+        for (Device device : devices) {
+            try {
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/thingsboard/server/controller/DeviceController.java
```diff
@@ -141,6 +141,45 @@ public Device saveDevice(@RequestBody Device device,
...
+    public List<Device> saveBulkDevices(@RequestBody List<Device> devices,
+                                        @RequestParam(name = "accessToken", required = false) String accessToken) throws ThingsboardException {
+        List<Device> savedDevices = new ArrayList<>();
+        for (Device device : devices) {
+            try {
+                device.setTenantId(getCurrentUser().getTenantId());
+
+                checkEntity(device.getId(), device, Resource.DEVICE);
+
+                Device savedDevice = checkNotNull(deviceService.saveDeviceWithAccessToken(device, accessToken));
```
ashvayka, line 156: This is nonsence. There is a unique key for access token. You can't save multiple devices with the same token.

application/src/main/java/org/thingsboard/server/controller/DeviceController.java
```diff
@@ -141,6 +141,45 @@ public Device saveDevice(@RequestBody Device device,
...
+    @RequestMapping(value = "/bulk/devices", method = RequestMethod.POST)
+    @ResponseBody
+    public List<Device> saveBulkDevices(@RequestBody List<Device> devices,
+                                        @RequestParam(name = "accessToken", required = false) String accessToken) throws ThingsboardException {
+        List<Device> savedDevices = new ArrayList<>();
+        for (Device device : devices) {
+            try {
+                device.setTenantId(getCurrentUser().getTenantId());
+
+                checkEntity(device.getId(), device, Resource.DEVICE);
```
ashvayka, line 154: Device id will be null. This can't work.

PR #4346, opened 2021-03-31
Rule node events filter

AndrewVolosytnykhThingsboard:


ashvayka:
There were so many errors when i have tried to use this PR, so I fixed them and added the necessary performance improvement. See https://github.com/thingsboard/thingsboard/pull/4413

PR #4350, opened 2021-04-01
Fixed title ex when make rule chain root in chinese

blackstar-baba:
when make rule chain root in chinese，title display expression like '{{RuleChainName}}'


blackstar-baba:
@vvlladd28 PTAL

vvlladd28 (approved):

blackstar-baba:
pelease merge it

PR #4361, opened 2021-04-03
Styling/dark mode | Displays a dark colour scheme depending on the preference selected in the Browser.

TTvanWillegen:
Enables the dark mode for the users.

- Changes the load screen to use the preferred mode.
- Changes the colour scheme from the previously implemented to a more standard dark-mode.

Dark mode changes depending on the user's preference, as specified in it's browser. 
Browsers, as well as OS's have the ability where a user can select their preference, using the CSS `@media (prefers-color-scheme: light|dark)` we can find out what that preference is, and update the UI accordingly.

image

### DIFF
--- a/ui-ngx/src/app/modules/home/components/attribute/attribute-table.component.scss
+++ b/ui-ngx/src/app/modules/home/components/attribute/attribute-table.component.scss
@@ -22,7 +22,6 @@
     .tb-entity-table-content {
       width: 100%;
       height: 100%;
-      background: #fff;
 
       .mat-toolbar-tools{
         min-height: auto;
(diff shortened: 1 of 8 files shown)


GrayCygnus:
any closure on this topic guys?

Dids:
It's been well over a year, but it's not really clear why this wasn't merged, or even reviewed?

TTvanWillegen:
> It's been well over a year, but it's not really clear why this wasn't merged, or even reviewed?

Good question, I have no idea either. However, now that the PR has been open this long, I don't really feel like fixing the (since occurred) conflicts has any use.

ashvayka:
After discussion with the UI team I may comment that this PR covers about 5% of the required code changes. It does not cover widgets at all. That is why this PR is constantly postponed.

PR #4393, opened 2021-04-09
[PROD-881] Dynamic alarm severity in Create Alarm Rule Node

AndrewVolosytnykhThingsboard:


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbCreateAlarmNode.java
```diff
@@ -67,10 +69,19 @@ protected TbCreateAlarmNodeConfiguration loadAlarmNodeConfig(TbNodeConfiguration
...
 
         if (!config.isUseMessageAlarmData()) {
             alarmType = TbNodeUtils.processPattern(this.config.getAlarmType(), msg);
+            if(this.config.isDynamicSeverity()) {
+                alarmSeverity = TbNodeUtils.processPattern(this.config.getSeverity(), msg);
+                if(!EnumUtils.isValidEnum(AlarmSeverity.class, alarmSeverity)) {
+                    ctx.tellFailure(msg, new TbNodeException("Used incorrect pattern or Alarm Severity not included in message: " + alarmSeverity));
+                    return null;
+                }
+                this.config.setSeverity(alarmSeverity);
```
ashvayka, line 83: You are overwriting the template with the calculated value.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbCreateAlarmNode.java
```diff
@@ -67,10 +69,19 @@ protected TbCreateAlarmNodeConfiguration loadAlarmNodeConfig(TbNodeConfiguration
     @Override
     protected ListenableFuture<TbAlarmResult> processAlarm(TbContext ctx, TbMsg msg) {
         String alarmType;
+        String alarmSeverity;
```
ashvayka, line 72: Store this as an enumeration. not a String.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbCreateAlarmNode.java
AndrewVolosytnykhThingsboard: @ashvayka Refactored

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbCreateAlarmNode.java
AndrewVolosytnykhThingsboard: @ashvayka Refactored

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbCreateAlarmNode.java
```diff
@@ -160,4 +162,22 @@ private Alarm buildAlarm(TbMsg msg, JsonNode details, TenantId tenantId) {
                 .build();
     }
 
+    private AlarmSeverity processAlarmSeverity(TbMsg msg) {
+        AlarmSeverity alarmSeverity;
+        if(this.config.isDynamicSeverity()) {
+            String processPattern = TbNodeUtils.processPattern(this.config.getSeverity(), msg);
+            if(!EnumUtils.isValidEnum(AlarmSeverity.class, processPattern)) {
```
ashvayka, line 169: This check is redundant. The EnumUtils.getEnum will simply return null if the value is not correct.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbCreateAlarmNode.java
```diff
@@ -160,4 +162,22 @@ private Alarm buildAlarm(TbMsg msg, JsonNode details, TenantId tenantId) {
...
+                alarmSeverity = null;
+            } else {
+                alarmSeverity = EnumUtils.getEnum(AlarmSeverity.class, processPattern);
+            }
+        } else {
+            alarmSeverity = EnumUtils.getEnum(AlarmSeverity.class, this.config.getSeverity());
```
ashvayka, line 175: No need to execute  EnumUtils.getEnum(AlarmSeverity.class, this.config.getSeverity()) if you are not using the dynamicSeverity. Execute this once in the init method and throw an exception if the value is invalid!

PR #4427, opened 2021-04-16
alarms count: performance fix based on pg_stat analysis. Counters que…

smatvienko-tb:
…ries refactored as two subqueries: inner join (count with relations) + left 'anti' join (count without relations). AlarmServiceSqlTest passed. CE


smatvienko-tb:
image

smatvienko-tb:
image
fixed long query with a very high query rate

smatvienko-tb:
query plan before:
image
after :
image

PR #4429, opened 2021-04-16
Merge Feature/html emails to CE

AndrewVolosytnykhThingsboard:
Merged from https://github.com/thingsboard/thingsboard-pe/pull/366


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/mail/TbMsgToEmailNode.java
```diff
@@ -50,6 +53,12 @@
 
     private static final ObjectMapper MAPPER = new ObjectMapper();
 
+    private static final String ATTACHMENTS = "attachments";
```
YevhenBondarenko, line 56: ATTACHMENTS and EMAIL_TIMEZONE not used

YevhenBondarenko (approved):

PR #4433, opened 2021-04-16
dependency upgrade: grpc.version up to 1.36.0. explicitly added proto…

smatvienko-tb:
…buf-java-util due to grpc not depends on it anymore


smatvienko-tb:
grpc 1.36.0 runs nominally about one month

PR #4434, opened 2021-04-16
Rule chain crud transactional

smatvienko-tb:


smatvienko-tb:
The motivation to write a rule chain in a transactional way is consistency! 
The problem may appear for a huge rule chain with a busy database. Another case - users hit save concurrently or the same user save twice from different tabs.
Running perfectly for about one month.

PR #4444, opened 2021-04-19
device count performance for 100k+ devices per tenant. Added addition…

smatvienko-tb:
…al tests for the DeviceService


smatvienko-tb:
countByTenantId added to the DeviceService to able test counter from service test

smatvienko-tb:
query plan before:
image
query plan after:
image

PR #4448, opened 2021-04-20
Customer level API Usage stats

ashvayka:


application/src/main/java/org/thingsboard/server/service/apiusage/DefaultTbApiUsageStateService.java
```diff
@@ -106,9 +112,9 @@ public void onFailure(Throwable t) {
     private InternalTelemetryService tsWsService;
 
     // Tenants that should be processed on this server
-    private final Map<TenantId, TenantApiUsageState> myTenantStates = new ConcurrentHashMap<>();
+    private final Map<EntityId, BaseApiUsageState> myUsageStates = new ConcurrentHashMap<>();
```
ashvayka, line 115: Maybe Map<EntityId, ApiUsageState> ?

PR #4466, opened 2021-04-22
Handle tenant and customer deletion for TbApiUsageStateService

ViacheslavKlimov:


application/src/main/java/org/thingsboard/server/service/apiusage/DefaultTbApiUsageStateService.java
```diff
@@ -111,11 +111,13 @@ public void onFailure(Throwable t) {
...
     private InternalTelemetryService tsWsService;
 
-    // Tenants that should be processed on this server
+    // Entities that should be processed on this server
     private final Map<EntityId, BaseApiUsageState> myUsageStates = new ConcurrentHashMap<>();
-    // Tenants that should be processed on other servers
+    // Entities that should be processed on other servers
     private final Map<EntityId, ApiUsageState> otherUsageStates = new ConcurrentHashMap<>();
 
+    private final Set<EntityId> deletedEntities = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
ViacheslavKlimov, line 119: This is needed in case if TbApiUsageClient will try to report left stats for deleted entity, and thus already deleted ApiUsageState for this entity would be recreated and added back to myUsageStates.
Also, the reason for this structure not being cleared is that if telemetry is still being sent from device so that we do not persist new ApiUsageState.

ViacheslavKlimov, line 119: This is needed in case if TbApiUsageClient will try to report left stats for deleted entity, and thus already deleted ApiUsageState for this entity would be recreated and added back to myUsageStates.
Also, the reason for this structure not being cleared is that if telemetry is still being sent from device so that we do not persist new ApiUsageState.

PR #4468, opened 2021-04-22
Js executor slow query log, kafka producer latency log, slow query detection, acks, gzip, request timeout, client_id

smatvienko-tb:


smatvienko-tb:
added kafka client_id to clarify which js-executor has a lag
/opt/kafka/bin/kafka-consumer-groups.sh --bootstrap-server tb-kafka:9092 --describe --group js-executor-group | sort
image

smatvienko-tb:
Slow query log. Kafka producer latency
image

smatvienko-tb:
SLOW_QUERY_LOG_BODY to find the script
image

smatvienko-tb:
added parameter TB_KAFKA_ACKS: "1" by default to improve performance. The previous version uses default ALL for producer

smatvienko-tb:
added parameter TB_QUEUE_KAFKA_REQUEST_TIMEOUT_MS to enable to admin to play around with timeouts

smatvienko-tb:
added TB_QUEUE_KAFKA_COMPRESSION="**gzip**" by default for the producer! The consumer still consumes gzip or uncompressed messages as well.
This improvement is valuable for massive functions and/or messages. It reduces disk space on kafka_log, reduces kafka memory consumption, reduces network traffic and latency as well.

smatvienko-tb:
Print config added as info log. The purpose is to identify fast what values are applied  
===CONFIG BEGIN===
.....
===CONFIG END===

PR #4470, opened 2021-04-22
QueueController has return queue list from yaml config (TbQueueRuleEn…

smatvienko-tb:
…gineSettings). Tip: queue list consumes UI on device profile details
image


smatvienko-tb:
working fine about month

PR #4478, opened 2021-04-23
kafka: added TB_KAFKA_COMPRESSION_TYPE for producer: none (default), …

smatvienko-tb:
…gzip.


smatvienko-tb:
https://kafka.apache.org/documentation/#producerconfigs_compression.type

smatvienko-tb:
Here an example how compressed Kafka logs look like:
image

ashvayka:
Is this battle-tested for our setup? :-)

smatvienko-tb:
Just applied to the stage. Its working fine now. Anyway, it is just a
standard Kafka option. Default compression is none. Gzip is optional.

PR #4479, opened 2021-04-23
feature/firmware

YevhenBondarenko:


dao/src/main/java/org/thingsboard/server/dao/firmware/BaseFirmwareService.java
```diff
@@ -316,4 +324,12 @@ protected void removeEntity(TenantId tenantId, FirmwareInfo entity) {
             return Optional.empty();
         }
     }
+
+    private static List<FirmwareId> toFirmwareInfoKey(FirmwareId firmwareId) {
+        return Collections.singletonList(firmwareId);
+    }
+
+    private static String toFirmwareDataKey(FirmwareId firmwareId) {
+        return firmwareId.toString();
```
ashvayka, line 333: The logic of clearing the cache is not obvious. It depends on the implementation of AbstractRedisFirmwareCache.toFirmwareCacheKey? This is not obvious.

dao/src/main/java/org/thingsboard/server/dao/firmware/BaseFirmwareService.java
```diff
@@ -88,7 +92,9 @@ public Firmware saveFirmware(Firmware firmware) {
         try {
             FirmwareId firmwareId = firmware.getId();
             if (firmwareId != null) {
-                cacheManager.getCache(FIRMWARE_CACHE).evict(firmwareId.toString());
+                Cache cache = cacheManager.getCache(FIRMWARE_CACHE);
+                cache.evict(toFirmwareInfoKey(firmwareId));
+                cache.evict(toFirmwareDataKey(firmwareId));
```
ashvayka, line 97: Where do yu push this data into cache (I mean toFirmwareDataKey(firmwareId))?

PR #4484, opened 2021-04-26
Removed redundant lock on device creating for provisioning feature, because of it may cause stucking on high loaded systems

zbeacon:


application/src/main/java/org/thingsboard/server/service/device/DeviceProvisionServiceImpl.java
```diff
@@ -191,28 +183,22 @@ private void notify(Device device, ProvisionRequest provisionRequest, String typ
...
-                return new ProvisionResponse(getDeviceCredentials(savedDevice), ProvisionResponseStatus.SUCCESS);
-            } else {
-                log.warn("[{}] The device is already provisioned!", device.getName());
-                notify(device, provisionRequest, DataConstants.PROVISION_FAILURE, false);
-                throw new ProvisionFailedException(ProvisionResponseStatus.FAILURE.name());
-            }
-        } catch (InterruptedException | ExecutionException e) {
+            return new ProvisionResponse(getDeviceCredentials(savedDevice), ProvisionResponseStatus.SUCCESS);
+        } catch (Exception e) {
+            log.warn("[{}] Error during device creation from provision request: [{}]", provisionRequest.getDeviceName(), provisionRequest, e);
```
ashvayka, line 201: Missing 

```java
notify(device, provisionRequest, DataConstants.PROVISION_FAILURE, false);
```

after refactoring. Probably, you should analyze the Exception.

PR #4492, opened 2021-04-26
RPC functionality refactoring for SNMP

ViacheslavKlimov:


common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
```diff
@@ -183,19 +183,29 @@ public void onToDeviceRpcRequest(DeviceSessionContext sessionContext, TransportP
...
         if (value == null && snmpMethod == SnmpMethod.SET) {
             throw new IllegalArgumentException("Value must be specified for SNMP method 'SET'");
         }
 
-        PDU request = pduService.createSingleVariablePdu(sessionContext, snmpMethod, oid, value, dataType);
-        sendRequest(sessionContext, request, new RequestInfo(toDeviceRpcRequestMsg.getRequestId(), SnmpCommunicationSpec.TO_DEVICE_RPC_REQUEST));
+        sessionContext.getProfileTransportConfiguration().getCommunicationConfigs().stream()
+                .filter(communicationConfig -> communicationConfig.getSpec() == SnmpCommunicationSpec.TO_DEVICE_RPC_REQUEST)
+                .findFirst()
+                .ifPresent(communicationConfig -> {
```
ashvayka, line 196: What if it is not present? Then we should reply with the error and not simply ignore the request!

PR #4496, opened 2021-04-27
Request SNMP devices with pagination

ViacheslavKlimov:


ashvayka:
I am merging this, but the SnmpTransportContext.initDevicesSessions seems to be quite a poor implementation and needs to be improved. We fetch the devices one by one (not in batch, not multithreaded).

PR #4503, opened 2021-04-28
Added ability to autogenerate checksum for firmware

zbeacon:


application/src/main/java/org/thingsboard/server/controller/FirmwareController.java
```diff
@@ -129,18 +129,17 @@ public Firmware saveFirmwareData(@PathVariable(FIRMWARE_ID) String strFirmwareId
...
             }
 
             firmware.setChecksumAlgorithm(checksumAlgorithm);
             firmware.setChecksum(checksum);
             firmware.setFileName(file.getOriginalFilename());
             firmware.setContentType(file.getContentType());
-            firmware.setData(ByteBuffer.wrap(data));
-            firmware.setDataSize((long) data.length);
+            firmware.setData(data);
+            firmware.setDataSize((long) data.array().length);
```
ashvayka, line 142: This is wrong. Also it will work in this case, it may stop working in the future, if someone will manipulate with the ByteBuffer. The .array() is internal data structure.

dao/src/main/java/org/thingsboard/server/dao/firmware/BaseFirmwareService.java
```diff
@@ -101,6 +101,32 @@ public Firmware saveFirmware(Firmware firmware) {
...
+    @Override
+    public String generateChecksum(String checksumAlgorithm, ByteBuffer data) {
+
+        if (data == null || !data.hasArray() || data.array().length == 0) {
+            throw new DataValidationException("Firmware data should be specified!");
+        }
+
+        HashFunction hashFunction;
+        switch (checksumAlgorithm) {
+            case "sha256":
```
ashvayka, line 113: please add more functions if present in the Hashing

common/dao-api/src/main/java/org/thingsboard/server/dao/firmware/FirmwareService.java
```diff
@@ -18,17 +18,25 @@
...
 
+import java.nio.ByteBuffer;
+
 public interface FirmwareService {
 
     FirmwareInfo saveFirmwareInfo(FirmwareInfo firmwareInfo);
 
     Firmware saveFirmware(Firmware firmware);
 
+    PageData<String> getSupportedChecksumAlgorithms(PageLink pageLink);
```
ashvayka, line 36: Why this is pageData and pageLink? More then 100 algorithms supported? UI will use paginated requests?

PR #4506, opened 2021-04-28
Find correct dashboard in case multiple dashboard starts with the same name

volodymyr-babak:
…e name

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/security/auth/oauth2/AbstractOAuth2ClientMapper.java
+++ b/application/src/main/java/org/thingsboard/server/service/security/auth/oauth2/AbstractOAuth2ClientMapper.java
@@ -206,9 +206,10 @@ public abstract class AbstractOAuth2ClientMapper {
     }
 
     private Optional<DashboardId> getDashboardId(TenantId tenantId, String dashboardName) {
-        PageLink searchTextLink = new PageLink(1, 0, dashboardName);
+        PageLink searchTextLink = new PageLink(Integer.MAX_VALUE, 0, dashboardName);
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/thingsboard/server/service/security/auth/oauth2/AbstractOAuth2ClientMapper.java
```diff
@@ -206,9 +206,10 @@ private CustomerId getCustomerId(TenantId tenantId, String customerName) {
     }
 
     private Optional<DashboardId> getDashboardId(TenantId tenantId, String dashboardName) {
-        PageLink searchTextLink = new PageLink(1, 0, dashboardName);
+        PageLink searchTextLink = new PageLink(Integer.MAX_VALUE, 0, dashboardName);
```
ashvayka, line 209: This fix is far from the best possible solution. Please add one more method to the DashboardInfoRepository to find the first (or all )dashboards with the exact name instead of search text.

volodymyr-babak: In the PE version, it's going to be a code duplicate due to the method that searches by groupsIds.
Can we consider refactoring this code snippet to correct pagination by 100 dashboards? 
PageLink searchTextLink = new PageLink(100, 0, dashboardName);
do {
...
} while()

ashvayka:
@AndrewVolosytnykhThingsboard please create a method "findDashboardsByTenantIdAndName". Since the name is not unique, we can't return only one dashboard here.

AndrewVolosytnykhThingsboard:
@ashvayka method implemented in https://github.com/thingsboard/thingsboard/pull/4665
Does it fit?

PR #4515, opened 2021-04-29
Fix entities filtering by kv when search value is empty

ViacheslavKlimov:


dao/src/main/java/org/thingsboard/server/dao/sql/query/EntityKeyMapping.java
```diff
@@ -281,15 +279,22 @@ public String toLatestJoin(QueryContext ctx, EntityFilter entityFilter, EntityTy
         }
     }
 
+    private boolean hasFilterValues(QueryContext ctx) {
+        return Arrays.stream(ctx.getParameterNames()).anyMatch(parameterName -> {
+            return !parameterName.equals(alias + "_key_id") && parameterName.startsWith(alias);
```
ashvayka, line 284: "_key_id" is a hard code. Move to constant (or method) to make sure that if someone changes "_key_id" in the other place, we will not be affected. Then  I will merge.

PR #4517, opened 2021-04-29
Fix for getting user method when additionalInfo is null and refactoring

zbeacon:


application/src/main/java/org/thingsboard/server/controller/UserController.java
```diff
@@ -89,28 +90,21 @@ public User getUserById(@PathVariable(USER_ID) String strUserId) throws Thingsbo
...
-                processDashboardIdFromAdditionalInfo((ObjectNode) user.getAdditionalInfo(), HOME_DASHBOARD);
-            }
-            UserCredentials userCredentials = userService.findUserCredentialsByUserId(user.getTenantId(), user.getId());
-            if(userCredentials.isEnabled()) {
-                addUserCredentialsEnabled((ObjectNode) user.getAdditionalInfo());
+            if(user.getAdditionalInfo().isObject()) {
+                ObjectNode additionalInfo = (ObjectNode) user.getAdditionalInfo();
+                processDashboardIdFromAdditionalInfo(additionalInfo, DEFAULT_DASHBOARD);
+                processDashboardIdFromAdditionalInfo(additionalInfo, HOME_DASHBOARD);
+                UserCredentials userCredentials = userService.findUserCredentialsByUserId(user.getTenantId(), user.getId());
```
ashvayka, line 97: You have moved this part of the code inside the "if" statement. previously, it was outside. This is wrong, right?

zbeacon: Yes, I moved, but it is not a wrong, because of previously it was doing the same check. Also we use the userCredentials only in one place.

PR #4523, opened 2021-04-29
Make remote js processing async

volodymyr-babak:


application/src/main/resources/thingsboard.yml
```diff
@@ -19,6 +19,9 @@ server:
   address: "${HTTP_BIND_ADDRESS:0.0.0.0}"
   # Server bind port
   port: "${HTTP_BIND_PORT:8080}"
+  tomcat:
+    # Maximum size of data that could be send over HTTP form POST request
```
ashvayka, line 23: How is this related to the PR? What is the reason for adding this?

PR #4550, opened 2021-05-06
Fix for org.apache.kafka.common.KafkaException: javax.security.auth.l…

volodymyr-babak:
…ogin.LoginException: unable to find LoginModule class


icebob:
When will it be released?

PR #4577, opened 2021-05-13
enhance widget extension loading

vparomskiy:
1. add HttpCleint and DrugDrop accessible for widget extensions
2. add HttpCleint and DrugDrop accessible for widget extensions

### DIFF
--- a/ui-ngx/src/app/core/services/resources.service.ts
+++ b/ui-ngx/src/app/core/services/resources.service.ts
@@ -134,6 +134,7 @@ export class ResourcesService {
           try {
             modules = this.extractNgModules(module);
           } catch (e) {
+            console.error(e);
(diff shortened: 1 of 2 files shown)


ui-ngx/src/app/core/services/resources.service.ts
```diff
@@ -172,22 +173,44 @@ export class ResourcesService {
...
+    } catch(e) {
+      console.log('Could not load NgModule', e);
     }
     return modules;
   }
 
+  private isPrimitive(test) {
+    return test !== Object(test);
+  }
+
```
vvlladd28, line 213: Use function isObject from utils.ts

vvlladd28 (approved):

PR #4601, opened 2021-05-19
Fixed error message in Led Indicator

kalutkaz:
Jira (PROD-907)


ui-ngx/src/app/shared/components/toast.directive.ts
```diff
@@ -356,7 +356,8 @@ export class TbSnackBarComponent implements AfterViewInit, OnDestroy {
     }
   }
 
-  action(): void {
+  action(event): void {
```
vvlladd28, line 359: What type of variable 'event'?

vvlladd28 (approved):

PR #4617, opened 2021-05-24
postgresql.driver.version upgrade to 42.2.20 to fix connection issue …

smatvienko-tb:
…with Postgres 11

Caused by: java.io.EOFException: null
at org.postgresql.core.PGStream.receiveChar(PGStream.java:372)


smatvienko-tb:
We faced the problem with Postgres 11 while migrating to the latest 3.2. The previous driver works fine with Postgres 12. But failed with the 11. Upgrade the driver has fixed the problem.

PR #4629, opened 2021-05-26
Removed Service id from kafka consumer groups 

akei1984:
Removed Service-ID from kafka-consumer groups to prevent multiple data and alarm processing in a microservice deployment with kafka-queue as backend.


ashvayka:
I can't understand the changes you propose. As you may notice, we also define the topic name using the serviced. See the following example:

```java
consumerBuilder.topic(partitionService.getNotificationsTopic(ServiceType.TB_RULE_ENGINE, serviceInfoProvider.getServiceId()).getFullTopicName());
```

So, remove of the service id from the consumer group is not needed. Each consumer will only consume the topic that belongs to him. The main queues (Rule Engine) are not using the same approach, they share the load. 

Are you sure you have unique Service-ID assigned to each service in your k8s scripts? This may be the root cause of the data duplication.

ashvayka:
Closing this due to my comments above. Will reopen if see valid arguments.

ashvayka:
The reason for this is that we try to support not only Kafka but also other queues and that is why we use the same approach for all queue vendors.

PR #4644, opened 2021-05-27
Coap transport improvements

ShvaykaD:
added a fix for APIs with observing options(bug with session close due to last activity time, https://thingsboard-portal.atlassian.net/browse/PROD-518). 
updated tests. 
added fix to send acknowledgment responses on confirmable requests.


common/queue/src/main/proto/queue.proto
```diff
@@ -137,13 +137,15 @@ message GetAttributeRequestMsg {
   int32 requestId = 1;
   repeated string clientAttributeNames = 2;
   repeated string sharedAttributeNames = 3;
+  bool onlyShared = 4;
```
ashvayka, line 140: bool allShared = 4;
bool allClient = 5;

PR #4665, opened 2021-06-01
Implement of method to find dashboard by name and tenantId

AndrewVolosytnykhThingsboard:
find correct dashboard in case multiple dashboard start with the same name.


dao/src/main/java/org/thingsboard/server/dao/sql/dashboard/DashboardInfoRepository.java
```diff
@@ -29,6 +29,9 @@
  */
 public interface DashboardInfoRepository extends PagingAndSortingRepository<DashboardInfoEntity, UUID> {
 
+    DashboardInfoEntity findFirstByTenantIdAndTitle(@Param("tenantId") UUID tenantId,
```
ashvayka, line 32: Redundant Param annotation

PR #4667, opened 2021-06-01
Js executors kafka batches

smatvienko-tb:
to run from command line in developer mode:
~/thingsboard/msa/js-executor$ KAFKA_CLIENT_ID="js-executor-000" SLOW_QUERY_LOG_MS="100.000000" SLOW_QUERY_LOG_BODY="false" LOGGER_LEVEL="info" MAX_ACTIVE_SCRIPTS="1000" SCRIPT_STAT_PRINT_FREQUENCY="100" mvn install -P yarn-start

Kafka & zookeeper in Docker:

mkdir kafka
cd kafka
nano docker-compose.yml
------------------------------------------------------------------------

```
version: '2.2'
services:
  zookeeper:
    restart: always
    image: zookeeper
    ports:
      - "2181:2181"
      - "8111:8080" #admin http://localhost:8111/commands
    environment:
      ZOO_MY_ID: 1
  kafka:
    restart: always
    image: wurstmeister/kafka
    depends_on:
      - zookeeper
    ports:
      - "9092:9092"
    environment:
      KAFKA_BROKER_ID: 1
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_LISTENERS: INSIDE://:9093,OUTSIDE://:9092
      KAFKA_ADVERTISED_LISTENERS: INSIDE://:9093,OUTSIDE://kafka:9092
      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: INSIDE:PLAINTEXT,OUTSIDE:PLAINTEXT
      KAFKA_INTER_BROKER_LISTENER_NAME: INSIDE
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
```

------------------------------------------------------------------------------------

docker-compose up -d
docker ps
docker logs kafka_zookeeper_1 --tail 99 --follow
^C
docker logs kafka_kafka_1 --tail 99 --follow
^C

Then add kafka to the /etc/hosts
sudo nano /etc/hosts
127.0.0.1 localhost
#add this line after localhost
127.0.0.1 kafka

---------------------------------------------------
Run tb-node with ENVs 
JS_EVALUATOR="remote" TB_QUEUE_TYPE="kafka"


smatvienko-tb:
@vvlladd28 Could you help to implement the **liveness probe for the Kubernetes** pod?

smatvienko-tb:
image

smatvienko-tb:
Recommended resources for Kubernetes pod:
```
          resources:
            requests:
              cpu: 1
              memory: 512Mi
            limits:
              cpu: 1
              memory: 512Mi
          env:
            - name: KAFKA_CLIENT_ID
              valueFrom:
                fieldRef:
                  fieldPath: metadata.name
```

msa/js-executor/api/jsInvokeMessageProcessor.js
```diff
@@ -115,15 +116,20 @@ JsInvokeMessageProcessor.prototype.processCompileRequest = function(requestId, r
...
     logger.debug('[%s] Processing invoke request, scriptId: [%s]', requestId, scriptId);
     this.executedScriptsCounter++;
-    if ( this.executedScriptsCounter >= scriptBodyTraceFrequency ) {
-        this.executedScriptsCounter = 0;
-        if (logger.levels[logger.level] >= logger.levels['debug']) {
-            logger.debug('[%s] Executing script body: [%s]', scriptId, invokeRequest.scriptBody);
-        }
+    if (this.executedScriptsCounter % statFrequency == 0) {
+        var nowMs = performance.now();
+        var msSinceLastStat = nowMs - this.lastStatTime;
```
ashvayka, line 125: please put Math.max(nowMs - this.lastStatTime, 1) just in case..

msa/js-executor/api/jsInvokeMessageProcessor.js
```diff
@@ -186,11 +193,11 @@ JsInvokeMessageProcessor.prototype.sendResponse = function (requestId, responseT
...
 }
 
-JsInvokeMessageProcessor.prototype.getOrCompileScript = function(scriptId, scriptBody) {
+JsInvokeMessageProcessor.prototype.getOrCompileScript = function (scriptId, scriptBody) {
     var self = this;
-    return new Promise(function(resolve, reject) {
-        if (self.scriptMap[scriptId]) {
-            resolve(self.scriptMap[scriptId]);
+    return new Promise(function (resolve, reject) {
+        if (self.scriptMap.has(scriptId)) {
```
ashvayka, line 199: You are searching script in map two times. One for "has" and one for "get" method call. This  can be optimized to one "get" and analyze the result.

msa/js-executor/queue/kafkaTemplate.js
```diff
@@ -197,6 +284,10 @@ async function disconnectProducer() {
         var _producer = producer;
         producer = null;
         try {
+            logger.info('Stopping loop...');
+            //TODO: send handle msg
```
ashvayka, line 288: What is this about?

msa/js-executor/queue/kafkaTemplate.js
```diff
@@ -33,21 +35,75 @@ let producer;
...
+    const promise = new Promise((resolve, reject) => {
+        resolver = resolve;
+    });
+    batchMessages.push(message);
+    batchResolvers.push(resolver);
+    if (batchMessages.length >= maxBatchSize) {
+        sendMessagesAsBatch();
+        sendLoopWithLinger(); //reset loop function and reschedule new linger
+    }
+    return promise;
```
ashvayka, line 69: Does the previous implementation also return promise? Does the consumer "wait" for the messages to be published before consuming the incoming message pack? I am not saying this need to be fix immediately, but let's add a "TODO".

ashvayka (changes requested):
Please correct some minor comments

msa/js-executor/queue/kafkaTemplate.js
```diff
@@ -33,21 +35,75 @@ let producer;
...
+        return pushMessageToSendLater(message);
+    }
+}
+
+function pushMessageToSendLater(message) {
+    let resolver;
+    const promise = new Promise((resolve, reject) => {
+        resolver = resolve;
+    });
+    batchMessages.push(message);
```
ashvayka, line 63: using both batchMessages and batchResolvers is not clear. Are you trying to save the amount of objects you create? Why not adding one object (wrapper) that contain both message and promise(resolver) and use an array of wrappers? This will make code cleaner.

smatvienko-tb: yes, it will be more clear, but this requires unwrapping for each message to produce the final array to send. so this approach is much lighter for the CPU.

msa/js-executor/queue/kafkaTemplate.js
smatvienko-tb: The previous implementation was async. The actual implementation is async too. We don't wait on any locks.
The only sync job is to prepare a batch array (if maxBatchSize) and send async as a batch.
Another event is a timer (controlled by linger parameter), async too.

msa/js-executor/queue/kafkaTemplate.js
smatvienko-tb: we have to await sendMessagesAsBatch to send all ready-to-send messages before close gracefully. 
In actual implementation, the worst scenario that can happen on tb-node is a timeout and resend the JS message to process again.
I'll consult with @vvlladd28 on how to resolve this TODO.

msa/js-executor/queue/kafkaTemplate.js
ashvayka: Ok

msa/js-executor/api/jsInvokeMessageProcessor.js
smatvienko-tb: fixed with zero check

msa/js-executor/api/jsInvokeMessageProcessor.js
smatvienko-tb: > You are searching script in map two times. One for "has" and one for "get" method call. This can be optimized to one "get" and analyze the result.

fixed

smatvienko-tb:
I have marked the config parameter partitions_consumed_concurrently as EXPERIMENTAL.
It derives from "kafka-js" library, but we have not enough pros or cons by using it in production.
So I left the default value 1 as used on previous versions for a long time (processing partition by partition).
The best low latency gained when used 1 partition per js-executor. The drawback is the scale down, for example, 10 down to 9. In that case, one of the js-executor will get a double load.

smatvienko-tb:
The current version with kafka batches are effectively utilized the whole CPU. So the optimal config is a smaller amount of js-executors (5-10) with 1 CPU limit and 256-512M memory.
The logs in this version are very useful to monitor the actual performance

PR #4668, opened 2021-06-01
No body for REST API Call Node on GET method

AndrewVolosytnykhThingsboard:


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -174,7 +174,13 @@ public void processMessage(TbContext ctx, TbMsg msg) {
         String endpointUrl = TbNodeUtils.processPattern(config.getRestEndpointUrlPattern(), msg);
         HttpHeaders headers = prepareHeaders(msg);
         HttpMethod method = HttpMethod.valueOf(config.getRequestMethod());
-        HttpEntity<String> entity = new HttpEntity<>(msg.getData(), headers);
+        HttpEntity<String> entity;
+        if(HttpMethod.GET.equals(method) || HttpMethod.HEAD.equals(method) ||
```
ashvayka, line 178: add remaining ifs for Connect, etc.

AndrewVolosytnykhThingsboard: We use HttpMethod.java to parse request method but it does not contain CONNECT method

PR #4679, opened 2021-06-03
Fix wrong configuration key for Compression Type

### LINKED ISSUE #4678: Wrong configuration key for Compression Type in JS Executor
The value of `compressionType` is currently reading from the wrong key `kafka.requestTimeout`.

https://github.com/thingsboard/thingsboard/blob/b978f0f7abe01dbb7fa79e301cab735029242f07/msa/js-executor/queue/kafkaTemplate.js#L27

kien-truong:
Fix #4678


ashvayka:
Thank you for reporting and fixing this!

smatvienko-tb:
the fix is working fine! thank you!
before
image
after
image

PR #4693, opened 2021-06-06
Added prompt login parameter to oauth2 authorize request fixing oauth2 sign out bug

Ayomi-gh:
Added prompt login parameter to oauth2 authorize request forcing user to login after being logged out fixing the current oauth2 sign out bug.


ashvayka:
This should be part of the OAuth parameters that sysadmin uses to provision OAuth providers. We can't hardcode this.

PR #4758, opened 2021-06-17
Clean exception message without class name and path

AndrewVolosytnykhThingsboard:


application/src/main/java/org/thingsboard/server/service/resource/DefaultTbResourceService.java
```diff
@@ -86,7 +86,9 @@ public TbResource saveResource(TbResource resource) throws ThingsboardException
                 } else {
                     throw new DataValidationException(String.format("Could not parse the XML of objectModel with name %s", resource.getSearchText()));
                 }
-            } catch (InvalidDDFFileException | IOException e) {
+            } catch(InvalidDDFFileException e) {
+                throw new DataValidationException("Failed to parse XML file!");
```
ashvayka, line 90: Add the name of the file to the error message. When we upload 100 files at once via UI it is not clear what file causes the error.

PR #4819, opened 2021-06-29
device cache by id (CE)

smatvienko-tb:


smatvienko-tb:
it's all about performance. Additional way to reduce query rate on PostgreSQL

PR #4878, opened 2021-07-08
Timeout for all SQL queries to survive enormous heavy query

smatvienko-tb:
SQL timeout implemented (30 seconds)
This prevents running any long queries.
Any long query is subject to refactor or make query filters much precise.
You can check out the previous commit and play with a dedicated long query to 
The motivation was the issue when long queries block the entire SQL pool and prevent the system to run normally.

In case you are going to run a very long query by design, you have setTimeout directly on you query like
`.setTimeout(600)`
`.setHint("javax.persistence.query.timeout", 50)`
`.setHint("org.hibernate.timeout", 90)`
`@Transactional(timeout = 300, readOnly = true)`
`stmt.setQueryTimeout((int) TimeUnit.HOURS.toSeconds(1));`
Here a good examples: https://vladmihalcea.com/query-timeout-jpa-hibernate/

Tip: you can find long queries using global search by "stmt.execute()"

**Works fine on prod** about two months. The slowest query detected  is about **15 sec**. Events cleanup performs stable and takes about 5 minutes (see stmt.setQueryTimeout((int) TimeUnit.HOURS.toSeconds(1)); )

**Important note**: the timeout approach helps us very well when some eager dashboard produced enormous heavy queries with a huge filter list (Postgres fires temp tables to disk) and makes the entire system is hardly responsive.


ashvayka:
Are you sure the Events/Telemetry cleanup queries will survive these changes?

smatvienko-tb:
> Are you sure the Events/Telemetry cleanup queries will survive these changes?

@ashvayka The event TTL cleanup with a new indexes takes a few seconds.
In case you are going to run a very long query by design, you have setTimeout directly on you query like
`.setTimeout(600)`
`.setHint("javax.persistence.query.timeout", 50)`
`.setHint("org.hibernate.timeout", 1)`
`@Transactional(timeout = 300, readOnly = true)`
`
Here a good examples: https://vladmihalcea.com/query-timeout-jpa-hibernate/

smatvienko-tb:
@ashvayka The pg_stats shows that event cleanup TTL takes about 2 seconds
image
with 12 million event table size

ashvayka:
But what about time-series data cleanup? It is probably working much longer. Please add necessary code changes for time-series data cleanup (overwrite the one you have configured and set to 1 hour or similar).

dmytro-landiak:
I am implementing the feature for time-series TTL where there could be a set of keys that won't be expired. The procedure for removing partitions will be turned off in the case of Postgres DB. This would lead to more time for the execution of those queries. I will get back here with more info after further investigations.

smatvienko-tb:
enabled long query for cleanup services stmt.setQueryTimeout((int) TimeUnit.HOURS.toSeconds(1));

smatvienko-tb:
> I am implementing the feature for time-series TTL where there could be a set of keys that won't be expired. The procedure for removing partitions will be turned off in the case of Postgres DB. This would lead to more time for the execution of those queries. I will get back here with more info after further investigations.

@dmytro-landiak The aim of the SQL timeout is to unblock the shared SQL pool. 
Any long-running queries have to be split into smaller pieces. 
Any long-running stored procedures are bad for the entire system. Because stored procedures in Postgres always running as a SINGLE TRANSACTION (waits, locks, deadlocks, and headache of course :) )
Cleanup is a maintenance and have to perform smoothly and gently :).
Anyway, if the timeout is an issue for any reason - you can simply set JAVAX_PERSISTENCE_QUERY_TIMEOUT (in ms).

PR #4891, opened 2021-07-12
[WIP] impl of ttl logic with keys that are not removed, move ttl cleanup logic from sql procedure to dao layer

dmytro-landiak:
… client and alarms cleanup service


dmytro-landiak:
PR contains the time-series updated TTL logic to be able to remove all keys except for the specified ones.
exclusive_keys: "${SQL_TTL_TS_EXCLUSIVE_KEYS:}"
This is needed for the case when some analytics time-series keys are being populated based on the incoming raw data and should be persisted all the time for future processing, investigations, and analysis.

In case the above param is set, the partitions for DATABASE_TS_TYPE=sql won't be removed.
The queries are set with the overridden timeout to 1h, in the same way as for the default TTL procedures.

In addition, there is a small adjustment of "getTimeseries" method in Rest Client including the "limit" param in the query since the default value of 100 was used all the time. Removed unnecessary call of 
relationService.deleteEntityRelations(tenantId, alarmId);
in the AlarmsCleanUpService since the alarmService.deleteAlarm() method is calling this inside it.

ashvayka:
Please rename "exclusive" to "excluded" everywhere you use "exclusive". The naming is simply wrong. The logic is ok. I will merge once this is done.

dmytro-landiak:
@ashvayka done

PR #4894, opened 2021-07-13
[WIP][3.3.0] Kafka max in flight requests per connection,  max.request.size

smatvienko-tb:
Added parameters for Kafka producer
The motivation is to provide a flexible Kafka producer config in Kubernetes with env vars.

Here an example of the Kafka producer latency issue under the high load (JMX, Kafka MBean, VisualVM tool):
So increasing/decreasing max.inflight.requests are really useful to play with limits in such kind of issue.
But use it carefully as well!
The most convenient way to reduce the actual max.inflight.requests is to increase linger.ms (up to 5 ms). This will help the producer to accumulate much bigger matches and reduce tiny requests to the Kafka broker.
`TB_KAFKA_MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION: "10"`
`TB_KAFKA_LINGER_MS: "5"`
image

Kafka other properties example added:
```
queue:
  kafka:
    other: # In this section you can specify custom parameters for Kafka consumer/producer and expose the env variables to configure outside
      - key: "request.timeout.ms" # refer to https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#producerconfigs_request.timeout.ms
        value: "${TB_QUEUE_KAFKA_REQUEST_TIMEOUT_MS:30000}" # (30 seconds)
      - key: "session.timeout.ms" # refer to https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#consumerconfigs_session.timeout.ms
        value: "${TB_QUEUE_KAFKA_SESSION_TIMEOUT_MS:10000}" # (10 seconds)
```

### DIFF
--- a/application/src/main/resources/thingsboard.yml
+++ b/application/src/main/resources/thingsboard.yml
@@ -757,7 +757,11 @@ queue:
       tb_ota_package:
         - key: max.poll.records
           value: 10
-    other:
+    other: # In this section you can specify custom parameters for Kafka consumer/producer and expose the env variables to configure outside
+      - key: "request.timeout.ms" # refer to https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#producerconfigs_request.timeout.ms
+        value: "${TB_QUEUE_KAFKA_REQUEST_TIMEOUT_MS:30000}" # (30 seconds)
+      - key: "session.timeout.ms" # refer to https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#consumerconfigs_session.timeout.ms
+        value: "${TB_QUEUE_KAFKA_SESSION_TIMEOUT_MS:10000}" # (10 seconds)
     topic-properties:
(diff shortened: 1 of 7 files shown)


application/src/main/resources/thingsboard.yml
```diff
@@ -739,6 +739,8 @@ queue:
     compression.type: "${TB_KAFKA_COMPRESSION_TYPE:none}" # none or gzip
     batch.size: "${TB_KAFKA_BATCH_SIZE:16384}"
     linger.ms: "${TB_KAFKA_LINGER_MS:1}"
+    max.request.size: "${TB_KAFKA_MAX_REQUEST_SIZE:1048576}"
```
ashvayka, line 742: These parameters should be added to the tb-*-transport.yml files

smatvienko-tb: Files tb-*-transport.yml updated. 
Kafka other properties example added (with default values and references).

PR #4939, opened 2021-07-21
[3.3.0] Fixed Failed to claim the device when use redis as cache

youngSpring:
when use redis as the cache, this interface
http(s)://host:port/api/v1/$ACCESS_TOKEN/claim  throw 
a
SerializationFailedException but ignored by AbstractTransformFuture
,and when do  Device Claiming API
Request(http(s)://host:port/api/customer/device/$DEVICE_NAME/claim),the
system show :Failed to claim the device

show like this


ikulikov:
@youngSpring Thx for your contribution!

PR #4983, opened 2021-07-27
[3.3.0] Added logging of error msg to easily identify root cause of failed logins

volodymyr-babak:
Added logging of error msg to easily identify root cause of failed logins


application/src/main/java/org/thingsboard/server/service/security/auth/oauth2/Oauth2AuthenticationSuccessHandler.java
```diff
@@ -99,6 +101,8 @@ public void onAuthenticationSuccess(HttpServletRequest request,
             clearAuthenticationAttributes(request, response);
             getRedirectStrategy().sendRedirect(request, response, baseUrl + "/?accessToken=" + accessToken.getToken() + "&refreshToken=" + refreshToken.getToken());
         } catch (Exception e) {
+            log.error("Error occurred during processing authentication success result. " +
```
ashvayka, line 104: Let's make it debug. This is sensitive info that we can't log to the file by default.  You can always change the log level for the class in logback.xml without restart. Also, you can add a template to logback.xml to change the log level with appropriate comment.

PR #4985, opened 2021-07-28
Include the limit value in the getTimeseries REST call.

dajtxx:
Fixes CP-4165.


ikulikov:
@dajtxx Thx for your contribution!

PR #4997, opened 2021-07-30
[3.3.0] Partitions should not be removed by custom TTL

volodymyr-babak:
DO NOT apply custom to partition, otherwise, short TTL will remove partition too early
partitions must remain in the DB forever or be removed only by systemTtl
removal of empty partition is too expensive (we need to scan all data keys for these partitions with ALLOW FILTERING)


ashvayka:
We need to think about removing the partitions completely. This will reduce the number of inserts.

PR #5000, opened 2021-07-30
Fix User TTL Override

mp-loki:
Currently the user is able to set TTL larger than it is defined in the Tenant profile, which should not be the case


ashvayka:
According to the documentation, the tenant profile contains "default" and not a "maximum" TTL value. The "maximum" TTL value is set in thingsboard.yml and is enforced on the low level (Cassandra DAO). See docs here: https://thingsboard.io/products/paas/subscription/ 
"Default value of the "time to live" parameter that is used to store time-series data.
You may overwrite the default value in the "Save Timeseries" rule node or using "TTL" metadata field of your message.
This allows you to optimize storage consumption. The maximum allowed value of TTL is 5 years.
For example, you may store "raw" data for 3 month and aggregated data for 3 years."

PR #5001, opened 2021-07-30
[3.3.0] MQTT transport accumulate msg before device is connected state

smatvienko-tb:
This is a fix for the clients that do not await the connected event.
Messages will accumulate in the device context queue until connected, out of queue limit, or channel close.
Here is a log example that motivates to make this pool request.

image

### DIFF
--- a/transport/mqtt/src/main/resources/tb-mqtt-transport.yml
+++ b/transport/mqtt/src/main/resources/tb-mqtt-transport.yml
@@ -89,7 +89,7 @@ transport:
     bind_address: "${MQTT_BIND_ADDRESS:0.0.0.0}"
     bind_port: "${MQTT_BIND_PORT:1883}"
     timeout: "${MQTT_TIMEOUT:10000}"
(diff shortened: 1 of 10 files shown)


transport/mqtt/src/main/resources/tb-mqtt-transport.yml
```diff
@@ -89,6 +89,7 @@ transport:
     bind_address: "${MQTT_BIND_ADDRESS:0.0.0.0}"
     bind_port: "${MQTT_BIND_PORT:1883}"
     timeout: "${MQTT_TIMEOUT:10000}"
+    msg_queue_size_per_device_limit: "${MQTT_MSG_QUEUE_SIZE_PER_DEVICE_LIMIT:10000}" # messages await in the queue before device connected state. This limit works on low level before TenantProfileLimits mechanism
```
ashvayka, line 92: Add same to thingsboard.yml (monolith mode). Please change default value to 100

smatvienko-tb: It's done.

PR #5032, opened 2021-08-04
Fix spring security issue

blackstar-baba:
Spring security has issue in 5.4.1: Changing SecurityContext More Than Once in Single Request Can Fail to Save.
Look here: https://tanzu.vmware.com/security/cve-2021-22112.


blackstar-baba:
PTAL

PR #5068, opened 2021-08-11
[WIP][3.3.2] Rulechain test fix

smatvienko-tb:
The first cause was fixed by introducing transaction aware cache (https://github.com/thingsboard/thingsboard/pull/5064)
The second cause was fixed by async testing refactor
The third cause is in progress...
Notes:
- added some builders
- refactored logback.xml to see the useful logs from the test (WIP)
- slf4j upgraded

Working notes:
- Cache is not a problem, test without cached BaseRelationSrvice still failure sometimes
image
image


ikulikov:
Outdated

PR #5077, opened 2021-08-12
[WIP] cache caffeine zero size-fix

smatvienko-tb:


ashvayka:
This may cause OOM? For example, caching 1M devices will require ~1Gb of RAM or more.. Correct? If it was disabled by default, maybe let's put something smaller? 10K?

smatvienko-tb:
> This may cause OOM? For example, caching 1M devices will require ~1Gb of RAM or more.. Correct? If it was disabled by default, maybe let's put something smaller? 10K?

Adjusted to 10k

PR #5122, opened 2021-08-21
Fix: resources-library page can't be redirected

blackstar-baba:
In SysAdmin HomePage, "resources-library" page can't be redirected


blackstar-baba:
PTAL

ikulikov:
@blackstar-baba Thx for your contribution!

PR #5166, opened 2021-08-30
[3.3.2] Fields length validation

ViacheslavKlimov:


common/data/src/main/java/org/thingsboard/server/common/data/widget/WidgetsBundle.java
```diff
@@ -31,6 +30,7 @@
     @NoXss
     private String alias;
```
ViacheslavKlimov, line 31: Alias length should also be less than 255

common/data/src/main/java/org/thingsboard/server/common/data/widget/WidgetsBundle.java
```diff
@@ -31,6 +30,7 @@
...
     @NoXss
+    @Length(fieldName = "title")
     private String title;
     private String image;
     @NoXss
```
ViacheslavKlimov, line 36: Same as for alias

common/data/src/main/java/org/thingsboard/server/common/data/User.java
```diff
@@ -36,8 +36,10 @@
     private String email;
     private Authority authority;
     @NoXss
+    @Length(fieldName = "firs name")
```
ViacheslavKlimov, line 39: Typo

dao/src/main/java/org/thingsboard/server/dao/service/StringLengthValidator.java
```diff
@@ -0,0 +1,41 @@
...
+@Slf4j
+public class StringLengthValidator implements ConstraintValidator<Length, String> {
+    private int max;
+
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (StringUtils.isEmpty(value)) {
+            return true;
+        }
+        return value.trim().length() <= max;
```
ViacheslavKlimov, line 34: We need to check the length of untrimmed value here as we will then save it as is and sql error will occur

application/src/main/java/org/thingsboard/server/exception/ThingsboardErrorResponseHandler.java
```diff
@@ -71,6 +74,17 @@ public void handle(Exception exception, HttpServletResponse response) {
...
+                String cause = "";
+                if (exception.getCause() != null) {
+                    cause = exception.getCause().getClass().getCanonicalName();
+                }
+                if (exception instanceof IllegalArgumentException || exception instanceof IncorrectParameterException
+                        || exception instanceof DataValidationException || cause.contains("IncorrectParameterException")) {
+                    exception = new ThingsboardException(exception.getMessage(), ThingsboardErrorCode.BAD_REQUEST_PARAMS);
+                } else if (exception instanceof MessagingException) {
+                    exception = new ThingsboardException("Unable to send mail: " + exception.getMessage(), ThingsboardErrorCode.GENERAL);
+                }
```
ViacheslavKlimov, line 86: Exceptions from controllers will not be logged according to BaseController.handleThingsboardException logic
```
if (logException && logControllerErrorStackTrace) {
    log.error("Error [{}]", exception.getMessage(), exception);
}
```

vvlladd28 (approved):

PR #5200, opened 2021-09-06
[3.3.3] Add operation IN and NOT_IN for String value type to key filter

van-vanich:
Now, for checking present or don't present element is a set of elements we need to use a complex filter. But, if we have many elements is too long. I think, the issue is to add operation IN and NOT_IN.
Operation IN and NOT_IN get a set of elements separated by a comma. Also, the user can take items to " " or ' '.
Example image data :
1. device1, device2
2. 'device1', 'device2'
3. "device1", "device2"
Old realization:
image
Now:
image


dao/src/main/java/org/thingsboard/server/dao/sql/query/EntityKeyMapping.java
```diff
@@ -560,8 +561,24 @@ private String buildStringPredicateQuery(QueryContext ctx, String field, StringF
...
                 break;
+            case IN:
+                value = value.replaceAll("'","").replaceAll("\"", "");
+                values = List.of(value.trim().split("\\s*,\\s*"));
+                stringOperationQuery = String.format("%s in (:%s))", operationField, paramName);
+                break;
+            case NOT_IN:
+                value = value.replaceAll("'","").replaceAll("\"", "");
+                values = List.of(value.trim().split("\\s*,\\s*"));
+                stringOperationQuery = String.format("%s not in (:%s))", operationField, paramName);
```
volodymyr-babak, line 572: break is missing

dao/src/main/java/org/thingsboard/server/dao/sql/query/EntityKeyMapping.java
```diff
@@ -537,6 +537,7 @@ private String buildStringPredicateQuery(QueryContext ctx, String field, StringF
             value = value.toLowerCase();
             operationField = String.format("lower(%s)", operationField);
         }
+        List<String> values = new ArrayList<>();
```
volodymyr-babak, line 540: let's remove this array that is not related to all cases

volodymyr-babak: we can create this array in second switch statement

volodymyr-babak: case IN:
            case NOT_IN:
                ctx.addStringListParameter(paramName, List.of(value.trim().split("\\s*,\\s*")));
                break;

dao/src/main/java/org/thingsboard/server/dao/sql/query/EntityKeyMapping.java
```diff
@@ -560,8 +560,23 @@ private String buildStringPredicateQuery(QueryContext ctx, String field, StringF
                 value = "%" + value + "%";
                 stringOperationQuery = String.format("%s not like :%s or %s is null)", operationField, paramName, operationField);
                 break;
+            case IN:
+                value = value.replaceAll("'","").replaceAll("\"", "");
```
ashvayka, line 564: Move 
```java
value.replaceAll("'","").replaceAll("\"", "")
```
to a method with a meaningful name. Also add 
```java
List.of(value.trim().split("\\s*,\\s*"))
```
there to avoid extra switch on line 572

PR #5218, opened 2021-09-09
[3.3.2] fixed additional info parsing for RPC entity

ShvaykaD:


ashvayka:
What was fixed? How the issue is reproduced?

ShvaykaD:
@ashvayka In the current release version, additional info of RPC entity always stored as a string independently of what data type of input passed - JSON object or string field or null.

"null" - instead of null
"{}" - instead of {}
"\"\\\"test\\\"\"" - instead of "test"

samples:

1. RPC:

{"method": "test", "params": {}, "persistent": true}

rule node debug event:

image

2. RPC:

{"method": "test", "params": {}, "persistent": true, "additionalInfo": "test"}

rule node debug event:

image

3. RPC:

{"method": "test", "params": {}, "persistent": true, "additionalInfo": {}}

rule node debug event:

image

ashvayka:
question - should we update somehow already affected additional info for stored RPCs?
- Is it possible to parse them with new code without updating them?

ShvaykaD:
> question - should we update somehow already affected additional info for stored RPCs?
> 
> * Is it possible to parse them with new code without updating them?

Yes. Verified this in the script node:

// for "null" value 

var parsedAdditionalInfo = JSON.parse(msg.additionalInfo);

// "\"\\\"test\\\"\"" - str value & "\"{}\"" - object value

var parsedAdditionalInfo = JSON.parse(JSON.parse(msg.additionalInfo));

PR #5219, opened 2021-09-09
[3.3.2] NullPointerException after upgrading device profiles from 3.2.2 to 3.3.0

Swoq:
NullPointerException was caused by wrong upgrading of Device Profile and its profile_data json from 3.2.2 to 3.3.0. Only the first alarm was affected, the others received null as value of "predicate" property.
In the first case `boolean isUpdated = false;` initialized before the cycle and changed inside it by the next equation:
`isUpdated = isUpdated || convertDeviceProfileAlarmRulesForVersion330(createRules.get(severity.name()).get("condition").get("spec"));`
variable `isUpdated` always remained true after the first execution of `convertDeviceProfileAlarmRulesForVersion330(...)` method and this method wouldn't be executed again.


ViacheslavKlimov (approved):

ashvayka:
Please explain what caused the null pointer and how it is fixed. You have extracted the variable, but how this affects NPE?

ashvayka:
Cherry-picked to master.

PR #5224, opened 2021-09-10
Fix org.thingsboard.rule.engine.api.msg.DeviceAttributes#remove

formerly11:
Fix org.thingsboard.rule.engine.api.msg.DeviceAttributes#remove method, change the key of map when removed


ikulikov:
@formerly11 Thx for your contribution.

PR #5230, opened 2021-09-10
[3.3.2] fixed import/export rule chains methods

dmytro-landiak:
Replaced ruleChainDao.save() with a service saveRuleChain method in order to validate exported rule chains with ruleChainValidator before the import. This fixes old exported rule chains without a "type" field to be imported correctly.

Removed RuleChainType.CORE param since exportRuleChains method finds and exports all rule chains including Edge related so that the importRuleChains method can import them as well.


dmytro-landiak:
Included a commit to fix an issue in the Rest Client to be able to parse JSON array values for attributes and time-series.
Currently, it throws a RuntimeException(Can't parse value...)

PR #5260, opened 2021-09-21
changed key-alg from RSA to EC

ShvaykaD:


ShvaykaD:
Changed key-alg due to https://github.com/eclipse/californium/issues/517
https://datatracker.ietf.org/doc/html/rfc7252#section-9.1.3.2

ashvayka:
Why keysize option is deleted?

ShvaykaD:
> Why keysize option is deleted?

https://bugs.openjdk.java.net/browse/JDK-8252647

PR #5304, opened 2021-09-30
Add option for HTTP client rule node to not create any message body

msqr:
This PR adds an option to the HTTP client rule node to **not** create any request message body or related headers. This can be useful, for example, on a `DELETE` request where technically a body is allowed but the receiving server disallows any body content.


msqr:
The thingsboard/thingsboard-rule-config-ui-ngx#50 PR includes the corresponding UI update to show a checkbox for this new option:

Screen shot of the new HTTP Client rule node ignoreRequestBody option

PR #5330, opened 2021-10-07
[3.3.2] UI:Fixed tables pagination after updated

ArtemDzhereleiko:
Fixed tables:
- attribute table;
- entities table;
- relation table;
- timeseries table widget;
- alarm table widget;

Screenshot from 2021-10-07 13-02-51
Screenshot from 2021-10-07 13-03-44
Screenshot from 2021-10-07 13-05-02
Screenshot from 2021-10-07 13-05-47
Fix: after table updating set pagination to the first page.

### DIFF
--- a/ui-ngx/src/app/core/api/widget-api.models.ts
+++ b/ui-ngx/src/app/core/api/widget-api.models.ts
@@ -14,7 +14,7 @@
 /// limitations under the License.
 ///
 
-import { Observable, Subject } from 'rxjs';
+import { Observable } from 'rxjs';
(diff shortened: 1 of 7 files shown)


glancia:
Hello, I found a case that is not being considered in this pull request, as below.
[bug Engecomp.pptx](https://github.com/thingsboard/thingsboard/files/10582944/bug.Engecomp.pptx)

As far as I understood, the entities table (which is the case in the file) has its pagination reset only upon Time Window change.

PR #5353, opened 2021-10-11
Cassandra buffered rate executor implemented separate queues for read and write queries

smatvienko-tb:
Motivation is shown in the picture below. The write queries are much faster in Cassandra than reads. 
Read query might be expensive and reasonably slow.
The write queries have a greater rate than reading.
Under the heavy load write queries starving when shared connection pool with reading query.
This works fine under the heavy load on prod for couple of days.
image


ashvayka:
Hi @smatvienko-tb , your idea is great. We thank you for submitting this PR. It is extremely valuable to improve the performance. Unfortunately, I must submit a comment about the implementation. I believe it is a bit redundant. Too many changes seem to be code duplication and not OOP style.  The 'AbstractBufferedRateExecutor' is a class that may be reused in other components. It should not know anything about "read" and "writes". I suggest to simply crate two instances of CassandraBufferedRateExecutor (you may extend it and create two CassandraBufferedRateReadExecutor and CassandraBufferedRateWriteExecutor). This will minimize code changes.

PR #5360, opened 2021-10-12
Clear alarm node default script changed to avoid infinite metadata grow

smatvienko-tb:
The previous script does not decouple actual metadata and previous data.
Previous data added internally inside the rule node as a regular metadata field with key "**prevAlarmDetails**"
Users may believe that metadata is the same as shown in debug messages.
When the user saves metadata as details, it brings all previous details and made inbounded details grow on each alarm.
Users will end up with the permanent exception like message size exceeds the limit (when sending to js-executor) and Postgres, pgpool cpu/memory overhead when saving huge alarm details inside alarm entity.
The script is clarify and decouples prevAlarmDetails and actual metadata.

See changed files for details

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbClearAlarmNodeConfiguration.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbClearAlarmNodeConfiguration.java
@@ -31,7 +31,7 @@ public class TbClearAlarmNodeConfiguration extends TbAbstractAlarmNodeConfigurat
                 "if (metadata.prevAlarmDetails) {\n" +
                 "    details = JSON.parse(metadata.prevAlarmDetails);\n" +
                 "    //remove prevAlarmDetails from metadata\n" +
-                "    metadata.delete('prevAlarmDetails')\n" +
+                "    delete metadata.prevAlarmDetails;\n" +
                 "    //now metadata is the same as it comes IN this rule node" +
                 "}\n" +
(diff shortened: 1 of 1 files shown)


ashvayka:
Two issues here: 
1. This affects only TbClearAlarmNodeConfiguration and not TbCreateAlarmNodeConfiguration;
2. Example of good configuration does not fix possible issues on existing environments.
I strongly believe that the right place to fix this is in TbAbstractAlarmNode.buildAlarmDetails method. Existing implementation:

```java
    protected ListenableFuture<JsonNode> buildAlarmDetails(TbContext ctx, TbMsg msg, JsonNode previousDetails) {
        try {
            TbMsg dummyMsg = msg;
            if (previousDetails != null) {
                TbMsgMetaData metaData = msg.getMetaData().copy();
                metaData.putValue(PREV_ALARM_DETAILS, mapper.writeValueAsString(previousDetails));
                dummyMsg = ctx.transformMsg(msg, msg.getType(), msg.getOriginator(), metaData, msg.getData());
            }
            return buildDetailsJsEngine.executeJsonAsync(dummyMsg);
        } catch (Exception e) {
            return Futures.immediateFailedFuture(e);
        }
    }
```

So, we already have the *previousDetails*  as JSON node here. We can simply remove it from the previous details:

```java
                if (previousDetails instanceof ObjectNode) {
                    ((ObjectNode) previousDetails).remove(PREV_ALARM_DETAILS);
                }
```

More advanced implementation may be to check the size of previous details and remove the nested previousDetails only if the side exceeds certain threshold.

PR #5361, opened 2021-10-12
Delay node is deprecated

smatvienko-tb:
The Delay node does not guarantee that message will be processed because of ack.
It accumulates fast msg in memory up to "Max limit of pending messages reached".
After that produced artificial failures that may retry with retry strategy up to infinite times.
Obvious it's a bad practice to use delay nodes in the production environment.
Deprecated to get prevent users to use that node in future rule chains


smatvienko-tb:
> please change the nodeDetails to the following "Delays messages for a configurable period. Please note, this node acknowledges the message from the current queue (message will be removed from queue). Deprecated because the acknowledged message still stays in memory (to be delayed) and this does not guarantee that message will be processed even if the "retry failures and timeouts" processing strategy will be chosen.

@ashvayka , nodeDetails updated

PR #5383, opened 2021-10-15
[3.3.3] dao sql tests: run on postgresql container with in-memory disk using testcontainers jdbc (jdbc:tc:postgresql:12.8)

smatvienko-tb:
The motivation is to maintain 100% compatibility and performance with the real PostgreSQL instance.
In-memory disk used as a storage for performance boost

Init DB is maintained by PostgreSqlInitializer class and linked from jdbc url.

You can easily verify compatibility with Postgres 13 or Postgres 11 with one line change in `sql-test.properties`:
`spring.datasource.url=jdbc:tc:postgresql:12.8:///thingsboard?TC_DAEMON=true&TC_TMPFS=/testtmpfs:rw&?TC_INITFUNCTION=org.thingsboard.server.dao.PostgreSqlInitializer::initDb`

image


smatvienko-tb:
Documentation https://github.com/thingsboard/thingsboard.github.io/pull/778

PR #5393, opened 2021-10-18
Lwm2m fix bug awake after unReg

nickAS21:
bug awake after unReg


YevhenBondarenko:
change dest branch from master to develop/3.3.2

PR #5470, opened 2021-11-01
[3.3.2] Fixed analyze reprocessing decision in TbRuleEngineProcessingStrategyFactory

dmytro-landiak:
Fixing the issue for the unnecessary consumer waiting for messages reprocessing when no messages are needed to be reprocessed.

Example of the current logic:
processing-strategy.type = RETRY_TIMED_OUT
10 messages are sent for processing in a pack. Result: 9 successful, 1 is failed.
if (retryFailed) { // false
    result.getFailedMap().forEach(toReprocess::put);
}
if (retryTimeout) { // true
    result.getPendingMap().forEach(toReprocess::put);  -> will not add anthing to the map since there are 0 timedOut messages
}

then
return new TbRuleEngineProcessingDecision(false, toReprocess); // empty map

So, no messages will be sent for the processing, but the consumer will wait approximately:
X retries * pack-processing-timeout + X retries * pause-between-retries (plus max-pause-between-retries could be set)
until the decision will be to commit the current pack and next messages will be consumed for processing.

If I am right, this could happen in 2 cases. When processing-strategy.type = RETRY_TIMED_OUT and we have failed messages in the pack and when processing-strategy.type = RETRY_FAILED and we have timedOut messages in the pack.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/queue/processing/TbRuleEngineProcessingStrategyFactory.java
+++ b/application/src/main/java/org/thingsboard/server/service/queue/processing/TbRuleEngineProcessingStrategyFactory.java
@@ -17,6 +17,7 @@ package org.thingsboard.server.service.queue.processing;
 
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
 import org.thingsboard.server.common.msg.TbMsg;
 import org.thingsboard.server.common.msg.queue.TbMsgCallback;
 import org.thingsboard.server.gen.transport.TransportProtos;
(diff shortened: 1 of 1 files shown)


application/src/main/java/org/thingsboard/server/service/queue/processing/TbRuleEngineProcessingStrategyFactory.java
```diff
@@ -101,6 +104,15 @@ public TbRuleEngineProcessingDecision analyze(TbRuleEngineProcessingResult resul
                     if (retrySuccessful) {
                         result.getSuccessMap().forEach(toReprocess::put);
                     }
+                    if (CollectionUtils.isEmpty(toReprocess)) {
+                        log.debug("[{}] Stopping the reprocessing logic due to reprocessing map is empty", queueName);
+                        if (retryFailed) {
```
ashvayka, line 109: There are very few use cases when software engineer should use 'probably' in their code or logs :-)

Please replace with something like this:
```
 if (retryTimeout) {
    result.getPendingMap().forEach(toReprocess::put);
} else if (log.isDebugEnabled() && !result.getFailedMap().isEmpty()) {
  log.debug("[{}] Skipped {} failed messages due to the processing strategy configuration", queueName, result.getFailedMap().size());
}

```

dmytro-landiak: Agree! Updated.

PR #5474, opened 2021-11-01
[3.3.3] Entities search within all tenants

ViacheslavKlimov:


ashvayka:
We also need to make sure that "Search devices" component on the Devices page (https://demo.thingsboard.io/devices) uses all visible columns (except created time) to search instead of "searchText" field. This may be a separate PR for 3.3.3.

PR #5477, opened 2021-11-02
[3.3.2] CoAP Efento WATER_METER sensor support

ShvaykaD:


ShvaykaD:
@ashvayka please review the Efento callback implementation. Maybe we need to update the default CoapOkCallback as well.
Efento engineers asked me to remove the responses at all in case of receiving not confirmable requests.

PR #5478, opened 2021-11-02
ui: event table: default interval is 5 minutes

smatvienko-tb:
To open events on the rule node when you have a few million since the last 24h. (even index did not help)
image
@ViacheslavKlimov FYI


ashvayka:
This needs to be fixed in a different way. By optimizing the request and the way we perform the request.

PR #5500, opened 2021-11-05
[3.3.2] Create relation rule node fix and tests

dmytro-landiak:
Improved create relation rule node logic related to the "Remove current relations" (hint on the UI - Removes current relations from the originator of the incoming message based on direction and type.) checkbox.

Currently, if it is selected, it is removing the relations only if the new one that should be created from the originator to the target configured entity is not present yet. That could cause many discrepancies in the hierarchy of the relationships and this logic does not reflect what should be and how the hint is described. This causes confusion.

Improved to delete the current relations if selected in any case and refactored the code. Added tests.


dmytro-landiak:
image

dmytro-landiak:
You can see the error in the screenshot. More investigation and fixes will be made.

PR #5505, opened 2021-11-05
[3.3.2] fixed concurrency exception when deleting relation

dmytro-landiak:
image

Fixing the issue on the screenshot. Added test.


dmytro-landiak:
The issue occurs when two threads come to the "deleteRelationIfExists" method. If both of them get "true" for relationRepository.existsById(key), one of them will not be able to delete the relation and the exception will be thrown.

PR #5529, opened 2021-11-10
fix(chinese translate): fix some display problem

youngSpring:
In Api Usage module,when use chinese ,there are some display problem
because of the lose of the translate items
image


vvlladd28:
Hi @youngSpring,
Thank you for your PR. To show the new translations correctly, you need to clear the browser cache.

PR #5554, opened 2021-11-12
Stats: do not persist empty stats (reducing event table size and disk IOPS)

smatvienko-tb:
We found that about 9 million from total 11 million events are about empty stats;

```
delete from event 
where body like '%,"messagesProcessed":0,"errorsOccurred":0}%'
```

image

At this moment there are no parameters available to write or not zero values.
I found no motivation to add an optional parameter to bloat the event table, but if it is necessary for some users, just let me know.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/stats/StatsActor.java
+++ b/application/src/main/java/org/thingsboard/server/actors/stats/StatsActor.java
@@ -51,6 +51,9 @@ public class StatsActor extends ContextAwareActor {
     }
 
     public void onStatsPersistMsg(StatsPersistMsg msg) {
+        if (msg.isEmpty()) {
+            return;
+        }
         Event event = new Event();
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/thingsboard/server/actors/stats/StatsActor.java
```diff
@@ -51,6 +51,9 @@ protected boolean doProcess(TbActorMsg msg) {
     }
 
     public void onStatsPersistMsg(StatsPersistMsg msg) {
+        if (msg.isEmpty()) {
+            return;
```
ssua91, line 55: '%,"messagesProcessed":0,"errorsOccurred":0

ssua91 (approved):

ssua91 (approved):
'%,"messagesProcessed":0,"errorsOccurred":0

PR #5577, opened 2021-11-17
[3.3.3] added ability to configure network configuration for lwm2m

YevhenBondarenko:


application/src/main/resources/thingsboard.yml
```diff
@@ -819,6 +819,9 @@ transport:
     log_max_length: "${LWM2M_LOG_MAX_LENGTH:1024}"
     # Use redis for Security and Registration stores
     redis.enabled: "${LWM2M_REDIS_ENABLED:false}"
+    network_config:
+      - key: "PROTOCOL_STAGE_THREAD_COUNT"
+        value: "${LWM2M_PROTOCOL_STAGE_THREAD_COUNT:12}"
```
ashvayka, line 824: What is the default value? I don't believe we need 12 threads for LwM2M by default

PR #5583, opened 2021-11-18
Develop 3.3.3 to Master

ashvayka:


application/src/main/java/org/thingsboard/server/controller/AssetController.java
```diff
@@ -379,9 +379,9 @@ public Asset assignAssetToPublicCustomer(@ApiParam(value = ASSET_ID_PARAM_DESCRI
             notes = "Requested asset must be owned by tenant that the user belongs to. " +
                     "Asset name is an unique property of asset. So it can be used to identify the asset." + TENANT_AUTHORITY_PARAGRAPH, produces = MediaType.APPLICATION_JSON_VALUE)
     @PreAuthorize("hasAuthority('TENANT_ADMIN')")
-    @RequestMapping(value = "/tenant/assets", params = {"assetName"}, method = RequestMethod.GET)
+    @RequestMapping(value = "/tenant/assets/name", method = RequestMethod.GET)
```
ashvayka, line 382: We can't simply change the API. We should use keep backward compatibility. You may add a new method if needed.

PR #5644, opened 2021-11-30
[3.3.3] fixed lwm2m device or profile update notifications

YevhenBondarenko:


common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/uplink/DefaultLwM2MUplinkMsgHandler.java
```diff
@@ -392,6 +396,7 @@ public void onDeviceUpdate(SessionInfoProto sessionInfo, Device device, Optional
         try {
             LwM2mClient client = clientContext.getClientByDeviceId(device.getUuidId());
             if (client != null) {
+                this.securityStore.remove(client.getEndpoint(), client.getRegistration().getId());
```
ashvayka, line 399: So, if I update the device name, the credentials will be removed from the security store? This looks like a hack, not a fix.

YevhenBondarenko: Yes, we have to remove the security if a new device profile has been set.

Issue #5603, opened 2021-11-22
[Bug] CASSANDRA_KEYSPACE_NAME ignored during install


resmo:
**Describe the bug**
We'd like to run 2 thingsboard setups (dev, test) using the same cassandra cluster (and postgresql db) and that is why we wanted to set cassandra keyspaces and database names. 

But we were not able to set a different cassandra keyspace for each environment (postgres db works fine). We can not even "rename" the keyspaces 

`CASSANDRA_KEYSPACE_NAME` as described in https://thingsboard.io/docs/pe/user-guide/install/config/#thingsboard-core-settings is ignored during install process.

It seems the keyspace `thingsboard` is hardcoded in https://github.com/thingsboard/thingsboard/blob/master/dao/src/main/resources/cassandra/schema-ts.cql#L17 

**Your Server Environment**
own setup:
  * docker (nomad cluster) + cassandra 3.11 cluster in VMs
  * ThingsBoard Version  3.3.2

**To Reproduce**
1. setup cassandra in a container:
2. run install with env vars:
```
CASSANDRA_CLUSTER_NAME="example-cluster"
CASSANDRA_KEYSPACE_NAME="thingsboard_dev"
CASSANDRA_USE_CREDENTIALS=false
CASSANDRA_LOCAL_DATACENTER="datacenter1"
CASSANDRA_URL="cassandra:9042"
```

**Expected behavior**
cassandra namespace used as configured in CASSANDRA_KEYSPACE_NAME in setup instead of  `thingsboard`

**additional**
```
 ===================================================
 :: ThingsBoard Professional Edition ::       (v3.3.2PE)
 ===================================================

Starting ThingsBoard Installation...
Installing DataBase schema for entities...
Installing SQL DataBase schema part: schema-entities.sql
Installing SQL DataBase schema indexes part: schema-entities-idx.sql
Installing SQL DataBase schema PostgreSQL specific indexes part: schema-entities-idx-psql-addon.sql
Installing DataBase schema for timeseries...
Installing Cassandra DataBase schema part: schema-ts.cql
Loading system data...
Installation finished successfully!
```

smatvienko-tb:
Hi, @resmo.
Thank you for your feedback.
The thingsboard keyspace name is included in schema-ts.cql and not overridden during install. 

The simplest solution is to replace keyspace in the installed folder (useful in the service model)
`nano /usr/share/thingsboard/data/cassandra/schema-ts.cql`

[...]

PR #5672, opened 2021-12-02
[3.3.3] CASSANDRA_KEYSPACE_NAME ignored during install

van-vanich:
We can't set up the CASSANDRA_KEYSPACE_NAME name during install. But we say we can do this in the documentation about installing TB.
More about this issue: https://github.com/thingsboard/thingsboard/issues/5603


application/src/main/java/org/thingsboard/server/service/install/CassandraAbstractDatabaseSchemaService.java
```diff
@@ -37,6 +39,9 @@
     @Autowired
     private InstallScripts installScripts;
 
+    @Value("${cassandra.keyspace_name}")
```
ashvayka, line 42: Use 
```
@Value("${cassandra.keyspace_name:thingsboard}")
```

and remove 'getCassandraKeyspaceName' method

PR #5673, opened 2021-12-02
[3.3.3] Nested Rule Chains support

ashvayka:


ashvayka:
@volodymyr-babak, please review the migration script once you are back from :sailboat:.

PR #5699, opened 2021-12-08
[3.3.3] Allow to login with only whitespaces in password

desoliture1:


ViacheslavKlimov (approved):

ashvayka:
Are you sure this is valid and the best practice to support whitespace in passwords?

PR #5715, opened 2021-12-13
Explicitly upgrade log4j-api, log4j-core, log4j-to-slf4j  to 2.15 on dependency management. NOTE: log4j-core is not used!

smatvienko-tb:
This is the improvement to bump log4j-api to 2.15 regarding https://github.com/advisories/GHSA-jfh8-c2jp-5v3q
Note that we don't use log4j-core. We use logback and slf4j instead.

That is why we are **NOT** affected by [CVE-2021-44228](https://nvd.nist.gov/vuln/detail/CVE-2021-44228). See additional comments from [Spring](https://spring.io/blog/2021/12/10/log4j2-vulnerability-and-spring-boot) and [Slf4j](http://slf4j.org/log4shell.html).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,6 +47,7 @@
         <jedis.version>3.3.0</jedis.version>
         <jjwt.version>0.7.0</jjwt.version>
         <slf4j.version>1.7.32</slf4j.version>
+        <log4j.version>2.15.0</log4j.version>
         <logback.version>1.2.6</logback.version>
         <rat.version>0.10</rat.version>
         <cassandra.version>4.10.0</cassandra.version>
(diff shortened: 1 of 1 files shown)


ashvayka:
The description of the PR is somehow misleading. We use [logback](https://github.com/qos-ch/logback) and we exclude log4j-core dependency. See slf4j comments here: http://slf4j.org/log4shell.html. I have checked the output of "mvn dependency:tree" and there is no log4j-core in both CE and PE.  

We can't have both logback and log4j core in the same project, because then the slf4j will not be able to start. I will rename the PR to be more informative and merge just to avoid panic. But it looks to me that the hotfix release is not necessary, since there is no log4j-core in dependencies.

ashvayka:
Additional note. For everyone who is worried about the [CVE-2021-44228](https://nvd.nist.gov/vuln/detail/CVE-2021-44228) , see [Log4j security page](https://logging.apache.org/log4j/2.x/security.html#Fixed_in_Log4j_2.15.0). Only users of log4j-core are affected and we don't use it in ThingsBoard.

PR #5726, opened 2021-12-15
[WIP] Added nofitications to gateway if device created through it was removed or renamed

zbeacon:
In order to improve user experience with the ThingsBoard IoT gateway we should notify the gateway about device removal on ThingsBoard or renaming. It is necessary to handle these notifications. 
After device removal on ThingsBoard, the gateway knows nothing about this and continue sending telemetry/attributes etc. But without connect message from the gateway - device won't be created on ThingsBoard again and all this data will be lost.

Possible issues:
1. When the gateway is not connected, but device name on ThingsBoard was changed or device was removed - the gateway doesn't receive an update.
2. When a device is not connected to the gateway, but device name on ThingsBoard was changed or device was removed - the gateway doesn't receive an update.

Possible solutions:
1. These issue may be solved by using persistent RPC's to gateway device, but it looks for me like not the best solution.
2. Creating a hashmap with device id's as keys, the gateway device id as a value in DefaultTransportService and send the notification to the gateway session (in this case possible issue 1 is still not solved)

### DIFF
--- a/common/cluster-api/src/main/proto/queue.proto
+++ b/common/cluster-api/src/main/proto/queue.proto
@@ -405,6 +405,15 @@ message ProvisionDeviceResponseMsg {
   string credentialsValue = 3;
 }
 
+message GatewayDeviceUpdated {
+  string oldDeviceName = 1;
+  string newDeviceName = 2;
+}
+
+message GatewayDeviceDeleted {
+  string deviceName = 1;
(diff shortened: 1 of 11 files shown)


zbeacon:
According to possible issue we decided to use shared attributes instead of adding specific topic to API.

PR #5730, opened 2021-12-15
Upgrade log4j artifacts to 2.16

megla-tlanghorst:
Just to be sure (even if it is not used, better safe than sorry). 2.15 is flawed because of [CVE-2021-45046](https://logging.apache.org/log4j/2.x/security.html#CVE-2021-45046).


smatvienko-tb (approved):
The PR looks fine and apply the https://nvd.nist.gov/vuln/detail/CVE-2021-45046

PR #5739, opened 2021-12-17
[WIP][3.3.X] Fix number of digits after floating point

desoliture1:


desoliture1:
Not valid PR, cause need to approve logic with frontend team

PR #5740, opened 2021-12-17
[3.3.3] Notification to gateway about device was renamed or removed

zbeacon:
In this PR:
1. On every related to gateway device renaming will be sent a new persistent RPC, with old device name data and new device name data

2.  On device removal will be sent a new persistent RPC request to the gateway with device name inside.

3. Added relation to from device to the gateway device with type "lastConnectedGateway", to give ability get information in rule chain what is the last connected gateway.


YevhenBondarenko (approved):

ashvayka:
Please add a link to the corresponding PR for documentation. This is a new feature and we need to document that the device renaming results in persistent RPC calls.

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -304,6 +304,17 @@ private BasicCredentialsValidationResult validateMqttCredentials(TransportProtos
                     TbMsg tbMsg = TbMsg.newMsg(DataConstants.ENTITY_CREATED, deviceId, customerId, metaData, TbMsgDataType.JSON, mapper.writeValueAsString(entityNode));
                     tbClusterService.pushMsgToRuleEngine(tenantId, deviceId, tbMsg, null);
                 }
+
+                List<EntityRelation> currentLastConnectedGatewayRelationList = relationService.findByFromAndType(TenantId.SYS_TENANT_ID, device.getId(), DataConstants.LAST_CONNECTED_GATEWAY, RelationTypeGroup.COMMON);
```
ashvayka, line 308: Are you 100% sure this logic covers common cases? Maybe customers want to send data to all gateways? 
I believe a much easier option is to create a relation for each gateway. Maybe we do need a configuration parameter that forces the creation of these relations. Because by default, single relation "Created" is sufficient.

application/src/main/java/org/thingsboard/server/service/gateway_device/DefaultGatewayDeviceStateService.java
```diff
@@ -0,0 +1,115 @@
...
+                null,
+                null
+        );
+    }
+
+    private Device findGatewayDeviceByRelationFromDevice(Device device) {
+        List<EntityRelation> relationToGatewayList = relationService.findByFromAndType(TenantId.SYS_TENANT_ID, device.getId(), DataConstants.LAST_CONNECTED_GATEWAY, RelationTypeGroup.COMMON);
+        if (!relationToGatewayList.isEmpty()) {
+            EntityRelation relationToGateway = relationToGatewayList.get(0);
+            return deviceService.findDeviceById(device.getTenantId(), (DeviceId) relationToGateway.getTo());
```
ashvayka, line 111: You don't need the device itself. you only need the device id. The Tenant id is always the same.

ashvayka:
We suggest adding the gateway id to the additional information of the device. Or at least add the flag that the device is managed by the gateway. This will allow us to minimize the number of DB queries when we update/delete the device

application/src/main/java/org/thingsboard/server/service/gateway_device/DefaultGatewayDeviceStateService.java
```diff
@@ -0,0 +1,106 @@
...
+                null,
+                null
+        );
+    }
+
+    private Device findGatewayDeviceByAdditionalInfoInDevice(TenantId tenantId, JsonNode deviceAdditionalInfo) {
+        if (deviceAdditionalInfo != null && deviceAdditionalInfo.has(DataConstants.LAST_CONNECTED_GATEWAY)) {
+            JsonNode lastConnectedGatewayIdNode = deviceAdditionalInfo.get(DataConstants.LAST_CONNECTED_GATEWAY);
+            DeviceId gatewayId = new DeviceId(UUID.fromString(lastConnectedGatewayIdNode.asText()));
+            return deviceService.findDeviceById(tenantId, gatewayId);
```
ashvayka, line 102: Please, just return the gatewayId. No need to fetch the gateway and check that it exists. We will simply create an RPC call for missing gateway and it will be removed by TTL function.

application/src/main/java/org/thingsboard/server/service/gateway_device/DefaultGatewayDeviceStateService.java
```diff
@@ -0,0 +1,106 @@
...
+            deviceRpcService.processRestApiRpcRequest(rpcRequest, fromDeviceRpcResponse -> {
+                log.trace("Device deleted RPC with id: [{}] processed to gateway device with id: [{}], deleted device name: [{}]",
+                        rpcRequest.getId(), gatewayDevice.getId(), device.getName());
+            }, null);
+        }
+    }
+
+    private ToDeviceRpcRequest formDeviceToGatewayRPCRequest(Device gatewayDevice, JsonNode deviceDataNode, String method) {
+        ToDeviceRpcRequestBody body = new ToDeviceRpcRequestBody(method, JacksonUtil.toString(deviceDataNode));
+        long expTime = System.currentTimeMillis() + Math.max(minTimeout, defaultTimeout);
```
ashvayka, line 84: Please explain this logic: Math.max(minTimeout, defaultTimeout)? Both parameters are from yml file, right? And they are constant. Why not to calculate this in the init() method? I believe that the timeout value of 10 seconds is too low for the persistent RPC query. Put 1 day or similar.

PR #5749, opened 2021-12-20
[3.3.3] Queue factories improvements

dmytro-landiak:


dmytro-landiak:
Fixed wrong queue admins usages and topic names. Added destroying of admin objects where needed.

PR #5762, opened 2021-12-21
[3.3.3] Mqtt rule node update

dmytro-landiak:
Added a boolean param to be able to append suffix for "clientId" param in the MQTT rule node.
This is helpful for the cluster mode when the client wants to specify "clientId" (not to have a random id) and does not face the issue with the broker that does not allow client connections with the same clientId.


dmytro-landiak:
Note, UI coming soon.

PR #5771, opened 2021-12-23
[WIP][3.4] ensure postgres database is always started in thinsgsboard/tb-postgres

andre-aktivconsultancy:
The Postgres database would only get started on first launch. When starting a second time e.g. after a `docker-compose down` and `docker-compose up`, the database was not started resulting in a failure when ThingsBoard is starting.

### DIFF
--- a/msa/tb/docker-postgres/start-db.sh
+++ b/msa/tb/docker-postgres/start-db.sh
@@ -22,14 +22,22 @@ PG_CTL=$(find /usr/lib/postgresql/ -name pg_ctl)
 if [ ! -d ${PGDATA} ]; then
     mkdir -p ${PGDATA}
     ${PG_CTL} initdb
+else
+    ${PG_CTL} start
 fi
 
 exec setsid nohup postgres >> ${PGLOG}/postgres.log 2>&1 &
 
 if [ ! -f ${firstlaunch} ]; then
     sleep 2
-    while ! psql -U ${pkg.user} -d postgres -c "CREATE DATABASE thingsboard"
+    while ! psql -U thingsboard -d postgres -c "CREATE DATABASE thingsboard"
     do
       sleep 1
+    done
+else
+    until pg_isready --dbname thingsboard --quiet
+    do
+        sleep 1
+        echo "Waiting for db"
     done
 fi


msa/tb/docker-postgres/start-db.sh
```diff
@@ -22,14 +22,22 @@ PG_CTL=$(find /usr/lib/postgresql/ -name pg_ctl)
 if [ ! -d ${PGDATA} ]; then
     mkdir -p ${PGDATA}
     ${PG_CTL} initdb
+else
+    ${PG_CTL} start
```
smatvienko-tb, line 26: I'm not sure why we need to do `${PG_CTL} start`, because the Postgres service starts automatically with no problems. Maybe there is another reason, so please, share your experience regarding that line.

msa/tb/docker-postgres/start-db.sh
```diff
@@ -22,14 +22,22 @@ PG_CTL=$(find /usr/lib/postgresql/ -name pg_ctl)
...
     sleep 2
-    while ! psql -U ${pkg.user} -d postgres -c "CREATE DATABASE thingsboard"
+    while ! psql -U thingsboard -d postgres -c "CREATE DATABASE thingsboard"
     do
       sleep 1
     done
-fi
\ No newline at end of file
+else
+    until pg_isready --dbname thingsboard --quiet
```
smatvienko-tb, line 38: Infinite awaiting is quite an unusual solution. Maybe some retry count is a better option to await the Postgres.

Here some example how to avoid infinite loop:

```
RETRIES=30
until psql -U ${pkg.user} -d postgres -c "select 1" > /dev/null 2>&1 || [ $RETRIES -eq 0 ]; do
  echo "Connecting to Postgres , $((RETRIES--)) attempts left..."
  sleep 1
done
```

msa/tb/docker-postgres/start-db.sh
```diff
@@ -22,14 +22,22 @@ PG_CTL=$(find /usr/lib/postgresql/ -name pg_ctl)
...
+else
+    ${PG_CTL} start
 fi
 
 exec setsid nohup postgres >> ${PGLOG}/postgres.log 2>&1 &
 
 if [ ! -f ${firstlaunch} ]; then
     sleep 2
-    while ! psql -U ${pkg.user} -d postgres -c "CREATE DATABASE thingsboard"
+    while ! psql -U thingsboard -d postgres -c "CREATE DATABASE thingsboard"
```
smatvienko-tb, line 33: why is it necessary to hardcode value instead of the variable?

smatvienko-tb (changes requested):
Hey, @andre-aktivconsultancy !
Great job! I am also facing this issue and have done some research on that.
The issue was reproduced on the test database about 50Gb size and unclean shutdown.
On the next start, Postgres run integrity checks and replay logs. That may take a while. So Thingsboad has to await while Postgres gets ready.

There is another file related to this issue. If you be so kind, please, add your changes to the https://github.com/thingsboard/thingsboard/blob/master/msa/tb/docker-cassandra/start-db.sh

Thank you a lot!

smatvienko-tb:
@ashvayka , the new PR is here https://github.com/thingsboard/thingsboard/pull/6719
The commit @andre-aktivconsultancy was cherry-picked to save the initial work.
Please, close this PR.

PR #5793, opened 2021-12-29
Added functionality to handle pong responses on web sockets

volodymyr-babak:


application/src/main/java/org/thingsboard/server/controller/plugin/TbWebSocketHandler.java
```diff
@@ -213,19 +230,29 @@ private TelemetryWebSocketSessionRef toRef(WebSocketSession session) throws IOEx
         synchronized void sendPing(long currentTime) {
             try {
                 if (currentTime - lastActivityTime >= pingTimeout) {
```
ashvayka, line 232: var timeSinceLastActivity = currentTime - lastActivityTime;
if(timeSinceLastActivity >= pingTimeout){
       log.warn("[{}] Closing session due to ping timeout", session.getId());
       closeSession(CloseStatus.SESSION_NOT_RELIABLE);
} else if (timeSinceLastActivity >= pingTimeout/3){
          this.asyncRemote.sendPing(PING_MSG);
}

ashvayka: where 3 should be a constant, because you use it in the scheduler

PR #5795, opened 2021-12-29
[3.3.3] UI: Disable column sort if enable post-processing function

ArtemDzhereleiko:
Disable column sort for that column where the post-processing function is enabled. Only for entities widget table and timeseries widget table.
Screenshot from 2021-12-29 13-02-19
Screenshot from 2021-12-29 13-02-54


vvlladd28 (approved):

X1folta:
May I inquire to why this was disabled? On the surface it seems like a very useful thing to have and something we notice missing here and there with certain values.

PR #5821, opened 2022-01-05
[3.3.3] Add ignoreMetadataTs to TbMsgTimeseriesNode

van-vanich:
Add ignoreMetadataTs to ignore the timestamp that arrives from message metadata.
Useful for all sorts of sequential processing if you merge messages from multiple sources (devices, assets, etc).

For example, we generate ts for devices and save this to devices and common asset:
image

And we can see, we don't save to asset last ts: 
image

So ignoreMetadataTs fix this issue.

**Need add UI for rule node**

### DIFF
--- a/application/src/test/java/org/thingsboard/server/service/ServiceSqlTestSuite.java
+++ b/application/src/test/java/org/thingsboard/server/service/ServiceSqlTestSuite.java
@@ -23,6 +23,7 @@ import org.thingsboard.server.queue.memory.InMemoryStorage;
 @RunWith(ClasspathSuite.class)
 @ClasspathSuite.ClassnameFilters({
         "org.thingsboard.server.service.resource.sql.*Test",
+        "org.thingsboard.server.service.sql.*Test"
 })
 public class ServiceSqlTestSuite {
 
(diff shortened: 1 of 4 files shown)


ashvayka:
The PR for Rule Node UI is missing.

ashvayka:
The original method implementation uses timestamp from metadata and fallback to time of message creation `msg.getTs())`

```java
    public static long getTs(TbMsg msg) {
        long ts = -1;
        String tsStr = msg.getMetaData().getValue("ts");
        if (!StringUtils.isEmpty(tsStr)) {
            try {
                ts = Long.parseLong(tsStr);
            } catch (NumberFormatException ignored) {
            }
        } else {
            ts = msg.getTs();
        }
        return ts;
    }
```

So, the parameter should be called `useServerTs` and the hint on UI should be the following:

Hint for Rule Node UI and Rule Node description: 
Enable this setting to use the timestamp of the message processing instead of the timestamp from the message. Useful for all sorts of sequential processing if you merge messages from multiple sources (devices, assets, etc). 

In the case of sequential processing, the platform guarantees that the messages are processed in the order of their submission to the queue. However, the timestamp of the messages originated by multiple devices/servers may be unsynchronized long before they are pushed to the queue. The DB layer has certain optimizations to ignore the updates of the "attributes" and "latest values" tables if the new record has a timestamp that is older than the previous record. So, to make sure that all the messages will be processed correctly, one should enable this parameter for sequential message processing scenarios.

PR #5843, opened 2022-01-10
[3.3.3] Device actor session inactivity performance + heap workout

smatvienko-tb:
Based on heap dump analysis when 1M devices are connected by MQTT to a single Thingsboard node

device check session inactivity refactored to reduce memory usage.

TenantId - factory + weap map
image

MqttTransportHandler stored an InetSocketAddress instances that is quite heavy for the heap for 1M connection
image

Live instance count reduced with a static ObjectMapper on DeviceActorMessageProcessor(AbstractContextAwareMsgProcessor)
image

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
+++ b/application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
@@ -966,21 +966,14 @@ class DeviceActorMessageProcessor extends AbstractContextAwareMsgProcessor {
         final long expTime = System.currentTimeMillis() - systemContext.getSessionInactivityTimeout();
         List<UUID> expiredIds = null;
 
-        try {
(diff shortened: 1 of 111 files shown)


application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
```diff
@@ -961,19 +963,42 @@ void init(TbActorCtx ctx) {
...
+                if (kv.getValue().getLastActivityTime() < expTime) {
+                    final UUID id = kv.getKey();
+                    if (expiredIds == null) {
+                        expiredIds = new ArrayList<>(1); //most of the expired sessions is a single event
+                    }
+                    expiredIds.add(id);
+
+                }
+            }
+        } catch (ConcurrentModificationException ignored) {
```
ashvayka, line 980: Please remove this since we are inside an Actor and all messages are processed sequentially.

PR #5857, opened 2022-01-11
[3.3.3] lwm2m fix bug update release

nickAS21:
- add loadSql for schema_update_lwm2m_bootstrap.sql
- if securityMode==null in profile? securityMode="NO_SEC"


application/src/main/java/org/thingsboard/server/service/install/SqlDatabaseUpgradeService.java
```diff
@@ -1,12 +1,12 @@
 /**
  * Copyright © 2016-2021 The Thingsboard Authors
- *
+ * <p>
```
YevhenBondarenko, line 3: wrong license

nickAS21: ok

PR #5889, opened 2022-01-14
[3.3.3] Reduce number of scheduled messages for session timeout

ashvayka:
Now we have one scheduled message per tenant and not one per device.


application/src/main/java/org/thingsboard/server/actors/tenant/TenantActor.java
```diff
@@ -84,6 +85,8 @@ public void init(TbActorCtx ctx) throws TbActorException {
                 cantFindTenant = true;
                 log.info("[{}] Started tenant actor for missing tenant.", tenantId);
             } else {
+                systemContext.schedulePeriodicMsgWithDelay(ctx, SessionTimeoutCheckMsg.instance(), systemContext.getSessionReportTimeout(), systemContext.getSessionReportTimeout());
```
smatvienko-tb, line 88: I can't find where is the scheduled task going down by the end of the lifetime of the particular TenantActor. For example, repartition event

PR #5915, opened 2022-01-18
[3.3.3] fixed NPE in AlarmRuleState

YevhenBondarenko:
If we set for the device the profile with rules which use duration or repeating (dynamic value) and the attribute in the source is not int, we will get NPE.

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/AlarmRuleState.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/AlarmRuleState.java
@@ -229,7 +229,7 @@ class AlarmRuleState {
         long repeatingTimes = 0;
         AlarmConditionSpec alarmConditionSpec = getSpec();
         AlarmConditionSpecType specType = alarmConditionSpec.getType();
-        if(specType.equals(AlarmConditionSpecType.REPEATING)) {
+        if (specType.equals(AlarmConditionSpecType.REPEATING)) {
             RepeatingAlarmConditionSpec repeating = (RepeatingAlarmConditionSpec) spec;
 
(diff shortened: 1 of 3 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/AlarmRuleState.java
```diff
@@ -238,7 +238,13 @@ private long resolveRequiredRepeats(DataSnapshot data) {
                     repeating.getPredicate().getDynamicValue().getSourceAttribute() != null) {
                 EntityKeyValue repeatingKeyValue = getDynamicPredicateValue(data, repeating.getPredicate().getDynamicValue());
                 if (repeatingKeyValue != null) {
-                    repeatingTimes = repeatingKeyValue.getLngValue();
+                    var longValue = getLongValue(repeatingKeyValue);
```
ashvayka, line 241: Code duplication

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/profile/AlarmRuleState.java
```diff
@@ -259,7 +265,13 @@ private long resolveRequiredDurationInMs(DataSnapshot data) {
                     duration.getPredicate().getDynamicValue().getSourceAttribute() != null) {
                 EntityKeyValue durationKeyValue = getDynamicPredicateValue(data, duration.getPredicate().getDynamicValue());
                 if (durationKeyValue != null) {
-                    durationTimeInMs = timeUnit.toMillis(durationKeyValue.getLngValue());
+                    var longValue = getLongValue(durationKeyValue);
```
ashvayka, line 268: Code duplication

PR #5940, opened 2022-01-20
[3.x] Caffeine cache stats 

smatvienko-tb:
Caffeine cache stats for general and Cassandra ts partition cache

This feature is intended to able tune-up caches based on the hit ratio.

image

for the Cassandra

image

more about Caffeine stats - https://github.com/ben-manes/caffeine/wiki/Statistics


common/cache/src/main/java/org/thingsboard/server/cache/CaffeineCacheConfiguration.java
```diff
@@ -72,13 +95,41 @@ public CacheManager cacheManager() {
         return new TransactionAwareCacheManagerProxy(manager);
     }
 
+    @PostConstruct
+    public void init() {
+        if (cacheStatsEnabled) {
+            log.debug("initializing cache stats scheduled job");
+            scheduler = Executors.newSingleThreadScheduledExecutor(ThingsBoardThreadFactory.forName("cache-stats"));
```
ashvayka, line 102: We need to extract the ScheduledExecutor to a component and use this component for all sorts of periodic logging/stats tasks. This will save many threads.

PR #5945, opened 2022-01-21
[3.3.3]Fix for PROD-1337 (Device session doesn't close on device removal)

zbeacon:
This PR contains changes to close existing session between device and ThingsBoard if device was removed on UI.

### DIFF
--- a/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
+++ b/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
@@ -48,6 +48,7 @@ import org.thingsboard.server.common.data.DeviceProfile;
 import org.thingsboard.server.common.data.DeviceTransportType;
 import org.thingsboard.server.common.data.TransportPayloadType;
 import org.thingsboard.server.common.data.device.profile.MqttTopics;
+import org.thingsboard.server.common.data.id.DeviceId;
 import org.thingsboard.server.common.data.id.OtaPackageId;
 import org.thingsboard.server.common.data.ota.OtaPackageType;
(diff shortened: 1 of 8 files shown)


common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -928,6 +929,9 @@ public void doDisconnect() {
                 gatewaySessionHandler.onGatewayDisconnect();
             }
             deviceSessionCtx.setDisconnected();
+            context.onAuthFailure(address);
```
ashvayka, line 932: Looks like you should add these three lines to the "onDeviceDeleted" method instead of calling this every time the "doDisconnect" is executed.

zbeacon: Moved

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -1073,6 +1070,9 @@ public void onDeviceUpdate(TransportProtos.SessionInfoProto sessionInfo, Device
     @Override
     public void onDeviceDeleted(DeviceId deviceId) {
         doDisconnect();
```
ashvayka, line 1072: I believe, the doDisconnect is redundant here. The ctx.close will invoke "doDisconnect", because it will trigger "operationComplete". Please check this and improve the code.

zbeacon: You are right, I will remove this call.

PR #5966, opened 2022-01-26
[3.4] Prohibit deletion of device profiles that are referenced by OTA update packages

ViacheslavKlimov:
Before, after deleting device profile that is used by OTA update package, the following error was occurring when opening OTA update details:
image

The deletion was being allowed due to no foreign key constraint in table ota_package for device_profile_id field in DB.


ashvayka:
Why don't we add a constraint for the database table? I believe we can drop the profiles that do not target any device profile since they will be still useless.

PR #5977, opened 2022-01-28
[3.3.3] LwM2mClient is no more serializable

YevhenBondarenko:


application/src/main/java/org/thingsboard/server/service/install/update/DefaultCacheCleanupService.java
```diff
@@ -64,6 +64,10 @@ public void clearCache(String fromVersion) throws Exception {
                 log.info("Clear cache to upgrade from version 3.3.2 to 3.3.3 ...");
                 clearAll();
                 break;
+            case "3.3.3":
+                log.info("Clear cache to upgrade from version 3.3.3 to 3.4 ...");
```
ashvayka, line 68: Please change the 3.4 to 3.3.4

ashvayka:
We will merge this to 3.3.4 since it is blocking bug fixes that are going to be part of 3.3.4

Issue #4666, opened 2021-06-01
[Bug] Rule Chain rest api call node - curly brackets { } in URL params error


mihaelking:
Hello,
I am having problems with Rest api call node in Rule chains. I am trying to create a POST request to server that requires data in JSON format to be in URL params. The problem is with { } curly brackets as Thingsboard thinks it's a variable. 
If I understand correctly variable in URL should only be recognised if there is a dolar sign in front of curly brackets - ${metaKeyName}. 
When I perform POST request to this example URL:
`"https://hidden_url.com/test?data=[{\"test\":\"test\"}]"`
I get the following error:
![image](https://user-images.githubusercontent.com/80819736/120334020-8ee14680-c2f0-11eb-8266-6ab46404dd18.png)

I am using script node to insert the data in URL and then I save it in metadata and forward it to rest api call node. 
Rest api call node looks as the picture below:
![image](https://user-images.githubusercontent.com/80819736/120330889-8a675e80-c2ed-11eb-8328-3be442bf5017.png)

I tried to insert data directly in URL in  with a form like example below, but got the same result:
![image](https://user-images.githubusercontent.com/80819736/120334295-d071f180-c2f0-11eb-99d4-96196f1aeeec.png)

We are using Thingsboard PE version 3.2.1

Any advice how to solve this problem?

Best regards,
Miha

ashvayka:
See https://www.w3schools.com/tags/ref_urlencode.ASP. URLs can only be sent over the Internet using the ASCII character-set.

The 
[{\"test\":\"test\"}]
Is converted to 
%5B%7B%5C%22test%5C%22%3A%5C%22test%5C%22%7D%5D

mihaelking:
> See https://www.w3schools.com/tags/ref_urlencode.ASP. URLs can only be sent over the Internet using the ASCII character-set.
> 
> The
> [{"test":"test"}]
> Is converted to
> %5B%7B%5C%22test%5C%22%3A%5C%22test%5C%22%7D%5D

Thank you for fast response. I am fully aware of url encoding and i am sorry I didn't mention I already tried that. With the help of Postman Mock serverb I found out that Thingsboard already coverts ASCII to HEX. 
If I enter an URL like this: 
`https://postman_url.com/test?data=%5B%7B%5C%22test%5C%22%3A%5C%22test%5C%22%7D%5D`
Thingsboard will convert % sign to %25, you can see on the bottom picture. This is causing parsing error on the REST server.
[...]

PR #6262, opened 2022-03-15
[3.4] Rest API Call node Json as URL param fix

desoliture1:
## Pull Request description

Issue: [#4666](https://github.com/thingsboard/thingsboard/issues/4666)

The main reason is that the URL is not encoded and as a result some characters are not valid or treated by Spring as URL variables that it should handle.

Instead of passing raw string data to `HttpClient.exchange` method, firstly building URI object from that string, then encoding it and after then passing to _exchange_ method

## General checklist

## Back-End feature checklist


ViacheslavKlimov (approved):

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/rest/TbHttpClientTest.java
```diff
@@ -58,4 +71,23 @@ public void givenNull_whenGetEventLoop_ThenReturnShared() {
...
+        var httpClient = new TbHttpClient(config, eventLoop);
+        var ctx = mock(TbContext.class);
+        var msg = TbMsg.newMsg(
+                "Main", "GET", new DeviceId(EntityId.NULL_UUID),
+                TbMsgMetaData.EMPTY, ""
+        );
+
+
+        httpClient.processMessage(ctx, msg);
+    }
```
smatvienko-tb, line 92: Test without assertion looks not really cool

smatvienko-tb (commented):
Hey, I found your new test. I'm happy to see you are making the tests. Please, find my comment below.

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/rest/TbHttpClientTest.java
```diff
@@ -58,4 +82,63 @@ public void givenNull_whenGetEventLoop_ThenReturnShared() {
...
+        verify(ctx, times(1)).transformMsg(
+                eq(msg), eq(msg.getType()),
+                eq(msg.getOriginator()),
+                eq(msg.getMetaData()),
+                eq(msg.getData())
+        );
+        verify(ctx, times(1))
+                .tellSuccess(eq(successMsg));
+
+        URI uri = UriComponentsBuilder.fromUriString(ENDPOINT_URL).build().encode().toUri();
```
ashvayka, line 141: I don't think this is a good test. It is better to create some kind of a dummy server that listens to all requests and echoes the URI in response. Otherwise, you are not testing the real behavior of the component. Also, this test does not check behavior for "normal" URLs

ashvayka: Other option is to replace the dummy server with the:
1) extract your changes ( " URI uri = UriComponentsBuilder.fromUriString(endpointUrl).build().encode().toUri();") to the new method
2) cover the method with unit tests with all sorts of input URL templates
3) make sure your method was executed in the main message processing flow.

rule-engine/rule-engine-components/pom.xml
```diff
@@ -136,6 +136,17 @@
...
+        <dependency>
+            <groupId>org.mock-server</groupId>
+            <artifactId>mockserver-netty</artifactId>
+            <version>5.13.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.mock-server</groupId>
+            <artifactId>mockserver-client-java</artifactId>
+            <version>5.13.1</version>
+        </dependency>
```
ViacheslavKlimov, line 148: Please move versions to dependency management in the parent pom

PR #6522, opened 2022-05-04
[3.4] Piggybacked CoAP responses

### LINKED ISSUE #1450: Piggybacked CoAP responses
From my observations it looks like Thingsboard currently never piggybacks CoAP responses, it always goes the "async" route of request, ACK, response, ACK.

I know this is absolutely allowed by the protocol, but the specifications do encourage the usage of piggybacked responses if possible. The response can then be carried directly within the ACK packet. Especially on the client this has multiple advantages, e.g. reduced complexity, network-traffic and ultimately battery-power spent.

> https://tools.ietf.org/html/rfc7252#section-5.2.1
> Implementation Note:  The protocol leaves the decision whether to
      piggyback a response or not (i.e., send a separate response) to
      the server.  The client MUST be prepared to receive either.  On
      the quality-of-implementation level, there is a strong expectation
      that servers will implement code to piggyback whenever possible --
      saving resources in the network and both at the client and at the
      server.

Is it possible for Thingsboard to use piggybacked responses?

JcBernack:
## Pull Request description

- Use coap piggybacked responses if possible
- Introduces a new `piggyback_timeout` config parameter
- Implements #1450

Please let me know if you see any issues with the way this is implemented.

## General checklist

- [ ] Similar PR is opened for PE version to simplify merge. Required for internal contributors only.
  
## Back-End feature checklist

### DIFF
--- a/common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportResource.java
+++ b/common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportResource.java
@@ -355,7 +355,11 @@ public class CoapTransportResource extends AbstractCoapTransportResource {
      * Essentially this allows the use of piggybacked responses.
      */
     private void deferAccept(CoapExchange exchange) {
-        transportContext.getScheduler().schedule(exchange::accept, piggybackTimeout, TimeUnit.MILLISECONDS);
+        if (piggybackTimeout > 0) {
+            transportContext.getScheduler().schedule(exchange::accept, piggybackTimeout, TimeUnit.MILLISECONDS);
+        } else {
+            exchange.accept();
+        }
     }
 
     private UUID toSessionId(TransportProtos.SessionInfoProto sessionInfoProto) {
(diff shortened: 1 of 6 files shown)


mikelohmann:
+1

common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportResource.java
```diff
@@ -346,6 +348,16 @@ private void handleToServerRpcRequest(TbCoapClientState clientState, CoapExchang
...
     }
 
+    /**
+     * Send an empty ACK if we are unable to send the full response within the timeout.
+     * If the full response is transmitted before the timeout this will not do anything.
+     * If this is triggered the full response will be sent in a separate CON/NON message.
+     * Essentially this allows the use of piggybacked responses.
+     */
+    private void deferAccept(CoapExchange exchange) {
+        transportContext.getScheduler().schedule(exchange::accept, piggybackTimeout, TimeUnit.MILLISECONDS);
```
ashvayka, line 358: This is a valid feature and we will merge it, but please, add the check for piggybackTimeout. Something like this:
```
if (piggybackTimeout > 0){
scheduleAccept
} else {
exchange.accept();
}
```
This is required for backward compatibility and 0 performance degradation in case the user will set this setting to 0.

JcBernack: Thanks for the review, added the check.

PR #6635, opened 2022-06-03
[3.4] Add resource service to widget context.

devaskim:
## Pull Request description

This PR is not very useful with the current list of resource types. As for 3.3.4.1 these are LWM2M model and security key storages. But if add to this list, for example, JSON resource type, these files may be used as extendable configuration for various widgets.

At the moment we are using such approach. There we store calibration tables for various device types and use that values as default configuration. Tenant admins have dashboard where they can edit such files. Then widgets on user`s dashboards load that value as defaults.

If this is interesting for TB team and community, I will add another one PR for JSON resource type and extending customer user permissions to upload only such resource type.


ikulikov:
@devaskim Thx for your contribution.

PR #6655, opened 2022-06-07
[3.4] Send attribute updates event when device state has been changed

devaskim:
As of 3.3.4.1 each device state change triggers one of 3 Rule Engine events (`ACTIVITY_EVENT`, `INACTIVITY_EVENT`,`DISCONNECT_EVENT`) and update of one of 5 state parameters (`active`, `lastConnectTime`, `lastDisconnectTime`, `lastActivityTime`, `inactivityAlarmTime`). This PR is to trigger either `ATTRIBUTES_UPDATED`, or `TIMESERIES_UPDATED` event when after state parameters described above has been updated.

Such *additional* event is useful when one redirects incoming telemetry/attributes to some asset using *Change originator*.


ashvayka:
This may cause too many events to RE. You may just fetch the necessary attributes when you receive the original event (ACTIVITY_EVENT/INACTIVITY_EVENT/DISCONNECT_EVENT)

PR #6703, opened 2022-06-13
[WIP][3.4.x] User activation timeout fix

desoliture1:
## Pull Request description

Issue: #6710 

The cause of the problem is that `timeout` parameter was not used at all, when we use invalid data (host, port, etc.) - mail sender throwing exception immediately, but if the host and port are correct but something else (login or password) is invalid - we just waiting for default timeout (around 2 mins) and timeout from sysadmin configs is not being used

Add *submitWithTimeout* method to `DonAsynchron` class
Add support for timeout in sending messages
Add async sending the 'account activated' email
Add corresponding tests

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/mail/DefaultMailService.java
+++ b/application/src/main/java/org/thingsboard/server/service/mail/DefaultMailService.java
@@ -16,6 +16,8 @@
 package org.thingsboard.server.service.mail;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
 import freemarker.template.Configuration;
 import freemarker.template.Template;
 import lombok.extern.slf4j.Slf4j;
@@ -25,6 +27,7 @@ import org.springframework.context.MessageSource;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.core.NestedRuntimeException;
 import org.springframework.core.io.InputStreamSource;
+import org.springframework.mail.MailSendException;
 import org.springframework.mail.javamail.JavaMailSender;
 import org.springframework.mail.javamail.JavaMailSenderImpl;
 import org.springframework.mail.javamail.MimeMessageHelper;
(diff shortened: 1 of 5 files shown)


application/src/main/java/org/thingsboard/server/service/mail/DefaultMailService.java
```diff
@@ -77,10 +86,14 @@ public class DefaultMailService implements MailService {
     @Autowired
     private MailExecutorService mailExecutorService;
 
+    private ScheduledExecutorService timeoutExecutor = Executors.newSingleThreadScheduledExecutor();
```
ashvayka, line 89: This executor is not destroyed in the PreDestroy of the Service

common/util/src/main/java/org/thingsboard/common/util/DonAsynchron.java
```diff
@@ -65,4 +67,32 @@ public static <T> ListenableFuture<T> submit(Callable<T> task, Consumer<T> onSuc
...
     }
 
+    public static <T> ListenableFuture<T> submitWithTimeout(
+            Callable<T> task,
+            long timeout, TimeUnit timeUnit,
+            Executor executor,
+            ScheduledExecutorService timeoutExecutor
+    ) {
+        var future = Futures.submit(task, executor);
+        Futures.withTimeout(future, timeout, timeUnit, timeoutExecutor);
```
ashvayka, line 77: Completely wrong. You should return the result of Futures.withTimeout. Also, this method is completely redundant

common/util/src/main/java/org/thingsboard/common/util/DonAsynchron.java
```diff
@@ -65,4 +67,32 @@ public static <T> ListenableFuture<T> submit(Callable<T> task, Consumer<T> onSuc
...
+        return future;
+    }
+
+    public static <T> ListenableFuture<T> submitWithTimeout(
```
ashvayka, line 81: remove changes to this file. they are not needed.

common/util/src/main/java/org/thingsboard/common/util/DonAsynchron.java
desoliture1: There is no difference what we'll return - Futures.withTimeout will cancel the future-delegate, so we will getting the exception either on 'future' or Futures.withTimeout

desoliture1:
#6862

Issue #6778, opened 2022-06-21
[Question] oAuth2 with Cas Server problem


S0UL4:
**Component**

 * UI
 * Generic

**Description**
Hi, i tried to do an oAuth2 with a private CAS server , i always get this error : [invalid_id_token] Missing (required) ID Token in Token Response for Client Registration: 5036b830-f16f-11ec-a1f8-bb3bf059bf30

Any idea please ?

PS: i tried an oAuth2 with Google and it worked very well ! 

this is the returned request from my private CAS server ( got it with python ) : 
{
  "attributes": {}, 
  "client_id": "client", 
  "id": "soula", 
  "service": "http://domain_here/authorized"
}
**Environment**

 * OS:  docker container
 * ThingsBoard: latest
 * Browser: Chrome

devaskim:
Could you provide response from Google oAuth2 ?

S0UL4:
Hi Thanks for your response @devaskim ! 

The reponse of the python code on /token (post request ) on my Private CAS server is :
{
  "access_token": "xxxx", 
  "expires_in": 28800, 
  "scope": "profile", 
  "token_type": "bearer"
}

The reponse of the python code on /token (post request ) on Google api is : 
 {
  "access_token": "xxxxxxx",
  "expires_in": 3599,
  "scope": "openid https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/userinfo.profile",
  "token_type": "Bearer",
  "id_token": "xxxxxxx"
}.

S0UL4:
@devaskim  i got this error when i changed the mapping to custom : 
[
![image](https://user-images.githubusercontent.com/62818840/175103060-f9cbe2fb-1305-4249-b703-679dddefb38f.png)
](url)

S0UL4:
@devaskim  there is actually some progress.. this now is my response from the CAS : 

{
  "attributes": {
    "mail": "xxxx", 
    "memberof": [
      "CN=xx,OU=xx,OU=xx,OU=xx,DC=xx,DC=xx", 
      "CN=xx,OU=xx,OU=xx,DC=xx,DC=xx"
    ], 
    "oauthClientId": "client"
  }, 
  "client_id": "client", 
  "id": "soula", 
  "service": "http://domain_here:8080/authorized"
}

any idea of how can i add to the email attribute in the basic mapper the values under "attributes"["mail"] ?

devaskim:
Basic mapper doesn't support nested fields.

[...]

PR #6815, opened 2022-06-23
[3.4.x OAuth] Add support of JSON path in OAuth mapper.

devaskim:
## Pull Request description

This PR let admins to map nested field in OAuth response, for example, `mail` in `attributes`.
```json
{
"attributes": {
    "mail": "xxxx",
    "oauthClientId": "client"
},
"client_id": "client",
"id": "soula",
"service": "http://domain_here:8080/authorized"
}
```

## General checklist

Related [discussion](https://github.com/thingsboard/thingsboard/issues/6778#issuecomment-1163468471).

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/security/auth/oauth2/BasicMapperUtils.java
+++ b/application/src/main/java/org/thingsboard/server/service/security/auth/oauth2/BasicMapperUtils.java
@@ -15,9 +15,13 @@
  */
 package org.thingsboard.server.service.security.auth.oauth2;
 
+import com.jayway.jsonpath.Configuration;
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.spi.json.JacksonJsonProvider;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/thingsboard/server/service/security/auth/oauth2/BasicMapperUtils.java
```diff
@@ -69,7 +73,12 @@ public static String getTenantName(String email, Map<String, Object> attributes,
     public static String getStringAttributeByKey(Map<String, Object> attributes, String key) {
         String result = null;
         try {
-            result = (String) attributes.get(key);
+            if (key.startsWith("$")) {
+                Object json = Configuration.defaultConfiguration().jsonProvider().parse(JacksonUtil.toString(attributes));
```
ashvayka, line 77: It is not a good idea to parse attributes every time we need to extract something. It is better to create a new method that will accept JSON object and key. Also, what if JsonPath.read will fail? There might be a key that starts with $ but is not a JSON expression, right?

Issue #6865, opened 2022-07-02
[Feature] UI: Enhance widget no-data settings section


devaskim:
**Is your feature request related to a problem? Please describe.**
As of version 3.3.4.1 Thingsboard lets user to specify text for case when there is no data to display.
![data_settings](https://user-images.githubusercontent.com/86909129/176986635-654b1fb1-033a-4aaa-b794-7e6ffa53286d.jpg)

**Describe the solution you'd like**
[Min] As user I would like to have HTML/CSS template to display customized view.
[Min] As custom widget developer I would like to have widget configuration option (True/False) to skip 'No data' section and handle such case in custom widget code.
[Max] As custom widget developer I would like to have 'no-data' widget action, for example, to redirect user to another dashboard or its state.

**Describe alternatives you've considered**
No

PR #6876, opened 2022-07-04
Let custom widget to process no-data case

devaskim:
This PR adds configuration option to widget data settings to let a (custom) widget to process by its own the case there is no subscription data. As of version 3.3.4.1 widget shows predefined message in case of no data.

This is simplified version of suggested improvements in #6865

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/widget-config.component.html
+++ b/ui-ngx/src/app/modules/home/components/widget/widget-config.component.html
@@ -305,11 +305,14 @@
                 <input matInput formControlName="decimals" type="number" min="0" max="15" step="1">
               </mat-form-field>
             </div>
-            <div fxLayout="row">
-              <mat-form-field fxFlex>
+            <div fxLayout.xs="column" fxLayout="row" fxLayoutGap="8px">
+              <mat-form-field fxFlex="grow">
                 <mat-label translate>widget-config.no-data-display-message</mat-label>
                 <input matInput formControlName="noDataDisplayMessage">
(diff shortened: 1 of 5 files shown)


vvlladd28:
Hi @devaskim,
Thanks for the interesting improvement to the platform. Perhaps it is better not to show this checkbox to all users but to include it as a widget configuration parameter in function code typeParameters ([models](https://github.com/thingsboard/thingsboard/blob/a7dc2272a444e1ecea8898fc15b6c69cc739454f/ui-ngx/src/app/shared/models/widget.models.ts#L151) for these parameters), because you indicated that it is only for custom widgets.

devaskim:
@vvlladd28 I like your idea and already did new PR [#6917](https://github.com/thingsboard/thingsboard/pull/6917).

Also @vvlladd28 could you give some tips how to add [this dialog reference](https://github.com/thingsboard/thingsboard/blob/master/ui-ngx/src/app/modules/home/components/widget/dialog/embed-dashboard-dialog.component.ts#L76) to [widget context](https://github.com/thingsboard/thingsboard/blob/master/ui-ngx/src/app/modules/home/models/widget-component.models.ts#L109) as it is [already done](https://github.com/thingsboard/thingsboard/blob/master/ui-ngx/src/app/modules/home/components/widget/dialog/custom-dialog-container.component.ts#L72) for widget's custom action with HTML template?

vvlladd28:
Hi @devaskim, 
Let's move the discussion to new issues because we are closing this PR.

PR #6943, opened 2022-07-13
UI: Add rule chain selector to toolbar of rule chain page.

devaskim:
Such selector saves time when you are debugging several rule chains at time.

I copied code from *DashboardSelectComponent* but without *DashboardSelectPanelComponent* because rule chain selector looks good both on the desktop and mobile devices. Screenshots attached.

rule-chain-selector-desktop-closed
rule-chain-selector-desktop-opened
rule-chain-selector-mobile


ashvayka:
Valid request and changes. We will review and merge in 3.4.1.

devaskim:
@ashvayka Finally I fixed and improved this PR.

Now rule chain selector is disabled when the page is in the *dirty* state (this is for #7190) and is hidden when new chain is being imported and yet not saved.

ashvayka:
Thank you for improving the PR. Unfortunately, we did not have time to test it yet. Forced to move to 3.5.1

vvlladd28:
Thanks for the improvement.
We took your code and implemented it in https://github.com/thingsboard/thingsboard/pull/8638

vvlladd28:
If we have missed something, we will be glad to improve it for you

Issue #83, opened 2017-03-17
See what version of server is running


nikil511:
After deploying a few times from source, we are already confused on which version we are actually running.
Is there a way to know ? maybe have a API call which returns some system info or somewhere in the admin user backend ?

ashvayka:
Hi! The version is always present in pom.xml that you use for the source code build. However, your idea is very valid and I've created [issue](https://thingsboard.atlassian.net/browse/TB-54) for this.

ashvayka:
This is fixed now in v1.2.3. Available in user dashboards on the bottom-right corner of the screen.

PR #6955, opened 2022-07-15
[3.4.2] Add option to either put or not RPC request to response metadata.

devaskim:
As of 3.3.4.1 version Thingsboard server puts in Rule Chain message data of RPC response either empty object (`{}`) or response from device. It is not enough at least in case of oneway RPC without response because there is no information in  Rule Chain message what RPC was processed.

Link to rule-config-ui PR: [#83](https://github.com/thingsboard/thingsboard-rule-config-ui-ngx/pull/83)


devaskim:
@ashvayka @ikulikov Why you close this PR without any comments?

Issue #6867, opened 2022-07-02
[Question] How to programmatically close dashboard state opened as dialog


devaskim:
**Component**
 * UI

**Description**
When I trigger widget's custom action with template, I have access to `dialogRef` and its method `close()` from `vm` variable. So I can programmatically close the dialog.

I want to have or do the same but with dashboard state opened as dialog. Is it possible as of 3.3.4.1? I couldn't find any `dialogRef` in widget context and `$scope` container.

**My case is the following:**
Dashboard state includes single custom widget, it's a creation form. I want to click *Save* button on this form, save what I need and then close dashboard state dialog.

**Environment**

 * ThingsBoard: 3.3.4.1

devaskim:
@vvlladd28 Could you suggest something?

devaskim:
need help - help yourself(

vvlladd28:
Hi @devaskim,
As an option, I can suggest saving in the widget context in the dashboard key to minimize the number of translations linked to dialogRef. Because if you save the link in the root of widget context, it will need to be copied to all state widgets

vvlladd28:
Perhaps the best option would be to make a link in the dashboard in stateController.

devaskim:
Thanks @vvlladd28 for idea, but how would I pass dialog link from [EmbedDashboardDialogComponent](https://github.com/thingsboard/thingsboard/blob/master/ui-ngx/src/app/modules/home/components/widget/dialog/embed-dashboard-dialog.component.ts#L62) to stateController?

devaskim:
Or do I need to set the link in (WidgetComponent [https://github.com/thingsboard/thingsboard/blob/master/ui-ngx/src/app/modules/home/components/widget/widget.component.ts#L1434] after dialog creation?

vvlladd28:
It is better to set the link after dialog creation

devaskim:
> Perhaps the best option would be to make a link in the dashboard in stateController.

@vvlladd28 Unfortunately code structure is unclear for me, for example:
`widgetContext -> stateController -> dashboardCtrl -> dashboardCtx -> !!! AGAIN !!!  stateController `

I tried to add _dialogRef_ to _dashboardCtx _ and then set it after dialog creation. But for other widgets _dialogRef_ stayed _null_

So I suggest to approve this PR [#6991](https://github.com/thingsboard/thingsboard/pull/6991) :) and close this ticket.

PR #6991, opened 2022-07-25
[3.4.1] UI: Add id to closing button of dashboard state dialog.

devaskim:
I wish to close dashboard state opened in dialog from my custom widget. I could not find a way to pass `MatDialogRef` to each widget context and no one (unfortunately, as usual) helped me (see [#6867](https://github.com/thingsboard/thingsboard/issues/6867)) in finding the right way.

So I added id to closing button which (id) is a current dashboard state. So now I can trigger button click from my custom widget and close the dashboard state dialog. Like this:
```javascript
$(`button[id$='${self.ctx.stateController.getStateId()}']`).click();
```


vvlladd28:
Hi @devaskim, you need to save the link to the dialog because this method is unreliable.

PR #7014, opened 2022-07-29
[3.4.1] UI: No black background behind snack bar.

devaskim:
**Before**
before

**After**

after


vvlladd28:
Hi @devaskim,
After testing the latest version of the master branch product, it was impossible to reproduce the described problem. Instead, we can assume you may have incorrectly edits to the source code from the attached images.

Issue #7258, opened 2022-09-10
[Question] Dynamic subscription to data keys


devaskim:
**Component**
 * UI

**Description**
Out of the box Thingsboard lets user to configure widget *"statically"*, i.e. this is done on creation or while updating widget configuration. Does custom widget API provide any way to dynamically subscribing to set of keys?

*My case is*:
Each device has its own unique set of keys. And I want to display that data in widget something like already existing *Entities table* or *Timeseries table*. Additionally devices periodically send their vocabulary where each data key is described so I know the whole set of keys.  

**Environment**
 * ThingsBoard: 3.4.1

devaskim:
Almost as usual I helped to myself and suggest PR #7275. With this PR I would receive continuous updates  in my custom widgets from all timeseries data keys without explicitly specifying them in widget configuration:
```
self.onInit = function() {
    let utils = self.ctx.$scope.$injector.get(self.ctx.servicesMap.get('utils'));

    utils.subscribeToEntityTelemetry(self.ctx).subscribe(
        data => console.log('Data: ', data),
        error => console.log('Error: ', error));
}

self.onDestroy = function() {
    if (self.ctx.telemetrySubscriber) {
        self.ctx.telemetrySubscriber.unsubscribe();
    }
}
```

PR #7275, opened 2022-09-14
[3.4.2] UI: Add API to create telemetry websocket updater in custom widgets.

devaskim:
Description of problem is in #7258

With this PR I would receive continuous updates  in my custom widgets from all timeseries data keys without explicitly specifying them in widget configuration:
```
self.onInit = function() {
    let utils = self.ctx.$scope.$injector.get(self.ctx.servicesMap.get('utils'));

    utils.subscribeToEntityTelemetry(self.ctx).subscribe(
        data => console.log('Data: ', data),
        error => console.log('Error: ', error));
}

self.onDestroy = function() {
    if (self.ctx.telemetrySubscriber) {
        self.ctx.telemetrySubscriber.unsubscribe();
    }
}
```

### DIFF
--- a/ui-ngx/src/app/core/services/utils.service.ts
+++ b/ui-ngx/src/app/core/services/utils.service.ts
@@ -43,6 +43,15 @@ import { WidgetInfo } from '@home/models/widget-component.models';
 import jsonSchemaDefaults from 'json-schema-defaults';
 import materialIconsCodepoints from '!raw-loader!./material-icons-codepoints.raw';
 import { Observable, of, ReplaySubject } from 'rxjs';
+import { publishReplay, refCount } from 'rxjs/operators';
+import { WidgetContext } from '@app/modules/home/models/widget-component.models';
+import {
+  AttributeData,
+  TelemetrySubscriber,
+  TelemetryType,
+  LatestTelemetry
+} from '@shared/models/telemetry/telemetry.models';
+import { EntityId } from '@shared/models/id/entity-id';
 
 const i18nRegExp = new RegExp(`{${i18nPrefix}:[^{}]+}`, 'g');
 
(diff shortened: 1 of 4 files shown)


ui-ngx/src/app/core/services/utils.service.ts
```diff
@@ -479,4 +488,30 @@ export class UtilsService {
...
+    }
+    return TelemetrySubscriber.createEntityAttributesSubscription(ctx.telemetryWsService, entityId, type, ctx.ngZone, keys);
+  }
+
+  public subscribeToEntityTelemetry(ctx: WidgetContext,
+                                    entityId?: EntityId,
+                                    type: TelemetryType = LatestTelemetry.LATEST_TELEMETRY,
+                                    keys: string[] = null): Observable<Array<AttributeData>> {
+    ctx.telemetrySubscriber = this.createTelemetrySubscriber(ctx, entityId, type, keys);
+    ctx.telemetrySubscriber.subscribe();
```
vvlladd28, line 511: This code can cause a memory leak:
1. If the subscription is stored in the context, the user may not know exactly where to unsubscribe. As a solution, you need to automatically unsubscribe when the widget component is destroyed.
2. The user can call this function several times, and the previous subscription will not be cleared. As a solution, you can store all subscriptions in an array and clear them when the widget is destroyed or return an observable to the user

ui-ngx/src/app/modules/home/components/widget/dynamic-widget.component.ts
```diff
@@ -80,6 +81,7 @@ export class DynamicWidgetComponent extends PageComponent implements IDynamicWid
     this.ctx.dialogs = $injector.get(DialogService);
     this.ctx.customDialog = $injector.get(CustomDialogService);
     this.ctx.resourceService = $injector.get(ResourceService);
+    this.ctx.telemetryWsService = $injector.get(TelemetryWebsocketService);
```
vvlladd28, line 84: What is the need to add the context if it has already been added to the ServicesMap?

ui-ngx/src/app/core/services/utils.service.ts
```diff
@@ -479,4 +488,30 @@ export class UtilsService {
       return defaultValue;
     }
   }
+
+  public getEntityIdFromDatasource(dataSource: Datasource) {
```
vvlladd28, line 492: Please add return type

vvlladd28 (changes requested):

Issue #7322, opened 2022-09-23
[Question] [Issue] How to speed up loading


psxbox:
**Component**

 * UI 

**Description**
As we can see in the picture, the UI part is loading very slowly on 3G or slow networks. The size of the downloaded resources is large (14.8Mb). Download 1.2 minutes. How can I reduce the load?
![image](https://user-images.githubusercontent.com/87794982/191941745-88539862-f859-4d51-b020-b1f8842d3b9e.png)

**Environment**

 * OS:  Sentos Steam 8
 * ThingsBoard: 3.4.1 PE
 * Browser: Chrome

devaskim:
If you are using haproxy, see [my solution](https://github.com/thingsboard/thingsboard/issues/7203#issuecomment-1236388994)

psxbox:
I don't use a proxy

psxbox:
resources are not being cached.

devaskim:
Finally I [implemented this feature](https://github.com/thingsboard/thingsboard/pull/7333). Waiting for review from core team...

psxbox:
It would be better if they released it sooner. For example, version 3.4.1.1. Because there is some problem with clients.

devaskim:
If I were in your shoes, I'd not wait for release but install proxy on machine where Thingsboard is running. Besides there is already [official doc](https://thingsboard.io/docs/user-guide/install/pe/add-haproxy-rhel/) fro CentOS from core team.

psxbox:
> Если бы я был на вашем месте, я бы не стал ждать релиза, а установил прокси на машине, где работает Thingsboard. Кроме того, уже есть [официальный док](https://thingsboard.io/docs/user-guide/install/pe/add-haproxy-rhel/) для CentOS от основной команды.

Thanks for the reply. I will consider.

psxbox:
> If I were in your shoes, I'd not wait for release but install proxy on machine where Thingsboard is running. Besides there is already [official doc](https://thingsboard.io/docs/user-guide/install/pe/add-haproxy-rhel/) fro CentOS from core team.

Thanks for the reply. I will consider.

albertorivelli:
Hi, is there any solution to this? I don't want to install HAProxy and I don't want to recompile either by adding the patch by @devaskim (pull request #7333 has not been merged in the official thignsboard release).
JS application files are quite big and the android app get stuck many times. I don't update the server side very often so it'd be much better to cache all those js files that are around 20MB.

Please help
Thanks

Issue #7206, opened 2022-09-05
[Question] How to enable Spring cache control policy


devaskim:
It is continuation of issue [#7203](https://github.com/thingsboard/thingsboard/issues/7203). I would like to enable *cachecontrol* policy for static TB resources.

Google says that minimum solution is to add `spring.resources.cache.cachecontrol.max-age=365d` to *application.properties*. In case of TB it is *thingsboard.yml*, **right**? So I pulled *tb-posgress* docker image, ran it, got bash access and modified *thingsboard.yml* by adding `spring.resources.cache.cachecontrol.max-age: "365d"` at the end. Then I committed changes and ran modified image. But HTTP response headers left the same for JS static resources (see *Cache-Control* header and *Expires*):
![tb-noo-cache](https://user-images.githubusercontent.com/86909129/188363668-1a2df4ec-0d9e-496d-9b6f-92e03660acdf.jpg)

PR #7333, opened 2022-09-25
Implemented caching static resources.

devaskim:
The discussion: #7322 and #7206
Added the following section to server configuration:
```
# caching configuration of static web resources (1 year by default)
cache.ui:
  enabled: "${SPRING_MVC_CACHE_ENABLED:true}"
  public: "${SPRING_MVC_CACHE_PUBLIC:true}"
  period: "${SPRING_MVC_CACHE_PERIOD:31536000}"
```

### DIFF
--- a/application/src/main/java/org/thingsboard/server/config/MvcConfig.java
+++ b/application/src/main/java/org/thingsboard/server/config/MvcConfig.java
@@ -1,0 +1,55 @@
+/**
+ * Copyright © 2016-2022 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.config;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.CacheControl;
+import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+import java.util.concurrent.TimeUnit;
+
+@Configuration
+public class MvcConfig implements WebMvcConfigurer {
+
+    @Value("${spring.mvc.cache.enabled}")
+    private Boolean cacheEnabled;
+    @Value("${spring.mvc.cache.public}")
+    private Boolean publicCache;
+    @Value("${spring.mvc.cache.period}")
+    private Integer period;
+
+    private static final String[] CLASSPATH_RESOURCE_LOCATIONS = {
+        "classpath:/static/",
+        "classpath:/public/"
+    };
+
+    @Override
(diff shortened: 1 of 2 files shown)


ashvayka:
Although this is a valid change, I would like to request a few edits before we merge.

application/src/main/java/org/thingsboard/server/config/MvcConfig.java
```diff
@@ -0,0 +1,55 @@
...
+    private Integer period;
+
+    private static final String[] CLASSPATH_RESOURCE_LOCATIONS = {
+        "classpath:/static/",
+        "classpath:/public/"
+    };
+
+    @Override
+    public void addResourceHandlers(ResourceHandlerRegistry registry) {
+        if (cacheEnabled) {
```
ashvayka, line 43: Please disable the caching logic by default.

application/src/main/resources/thingsboard.yml
```diff
@@ -544,6 +544,12 @@ spring:
     hikari:
       maximumPoolSize: "${SPRING_DATASOURCE_MAXIMUM_POOL_SIZE:16}"
 
+# spring configuration for caching static resources (1 year by default)
+spring.mvc.cache:
```
ashvayka, line 548: Please move this to the existing 'cache' section. The sub-section may be called 'ui' (similar to 'attributes' and 'CACHE_ATTRIBUTES_ENABLED'). The property naming that you suggest is a bit misleading. Developers may think that this is regular Spring Boot configs but this is a completely custom config parameters that are specific to our implementation.

application/src/main/resources/thingsboard.yml
```diff
@@ -544,6 +544,12 @@ spring:
...
+  enabled: "${SPRING_MVC_CACHE_ENABLED:true}"
+  public: "${SPRING_MVC_CACHE_PUBLIC:true}"
+  period: "${SPRING_MVC_CACHE_PERIOD:31536000}"
```
ashvayka, line 551: please name the env variables: CACHE_STATIC_UI_RESOURCE_ENABLED,  CACHE_STATIC_UI_RESOURCE_PUBLIC, CACHE_STATIC_UI_RESOURCE_PERIOD_IN_SEC

application/src/main/resources/thingsboard.yml
```diff
@@ -544,6 +544,12 @@ spring:
     hikari:
       maximumPoolSize: "${SPRING_DATASOURCE_MAXIMUM_POOL_SIZE:16}"
 
+# spring configuration for caching static resources (1 year by default)
+spring.mvc.cache:
+  enabled: "${SPRING_MVC_CACHE_ENABLED:true}"
```
ashvayka, line 549: disable by default for backward-compatibility and minimum risk of issues.

application/src/main/java/org/thingsboard/server/config/MvcConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.http.CacheControl;
+import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+import java.util.concurrent.TimeUnit;
+
+@Configuration
+public class MvcConfig implements WebMvcConfigurer {
+
+    @Value("${spring.mvc.cache.enabled}")
```
ashvayka, line 29: add default values to these parameters (similar to yml). This is needed for customers that skip the upgrade of the yml file and use old yml. Example:

```
@Value("${spring.mvc.cache.enabled:false}")
```

devaskim:
@ashvayka I moved caching configuration to `cache` section and disable caching by default

ashvayka:
This is quite valuable PR, but we don't have time to fix it before 3.5. Forced to move to 3.5.1. Implementation is OK.

Issue #7349, opened 2022-09-29
[Bug] VC: Cannot selectively export widget bundles


devaskim:
**Describe the bug**
No widget bundles are shown in the autocomplete list of *Create entities version* dialog.

**Your Server Environment**
* own setup
  * docker deployment
  * ThingsBoard 3.4.1
  * Ubuntu 22

**To Reproduce**

1. Create widget bundle with name *Test bundle*
2. Go to *Version control* section
3. Click on *Create entities version* button
4. Scroll down to *Widget bundle* section
5. Switch off *All entities filter*
6. Click on *Entity list* and (**optionally**) type *Test bundle*

**Expected behavior**
At least *Test bundle* widget bundle should appears in a list.

**Screenshots**
![Скриншот 29-09-2022 171341](https://user-images.githubusercontent.com/86909129/193029604-e4cefd1b-22a8-4a5d-a592-9b2d67fb1970.jpg)

PR #7360, opened 2022-10-02
[3.4.2] UI: Let user to selective export widget bundles to version control.

devaskim:
Bug: #7349

bundle_export


ashvayka:
Nice work! :+1:

PR #7375, opened 2022-10-04
[3.4.2] UI: Convert subscription boolean-in-string data to boolean type.

devaskim:
## Pull Request description

This PR (a little) will simplify new widget development and configuration existing ones because now you don't need to additionally check on `true/false`, e.g. `if (entity['MY_BOOLEAN_ATTRIBUTE'] === 'true') {}` instead just use it as it is `if (entity['MY_BOOLEAN_ATTRIBUTE'])`


devaskim:
@ashvayka I think [server side could also send already type-aware data](https://github.com/thingsboard/thingsboard/blob/master/application/src/main/java/org/thingsboard/server/service/telemetry/sub/TelemetrySubscriptionUpdate.java#L43) instead of string, couldn't it?

ashvayka:
Hi @devaskim , this issue is a bad design choice we made 6 years ago. We are going to move to the type-aware data, but this will require breaking API changes. This may be done only via new API or in ThingsBoard 4.0. I am closing this PR because we can't risk and break things for users with existing/custom widgets and actions.

PR #7386, opened 2022-10-06
[3.4.2] UI: Add widget context API call to create alarm queries.

devaskim:
## Pull Request description

The widget context already provides access to `alarmService`, but to find specific alarms we need to create instance of `AlarmQuery`. So this PR is to add API call that helps in creating such data objects.

*In general*: to avoid implementing such constructor proxies I would like to suggest core team to move up `toQuery()`-like methods from such data classes as `AlarmQuery` and others to corresponding service or additional layer for converting purpose.


ashvayka:
:+1:

PR #7387, opened 2022-10-06
[3.4.2] UI: Added support of custom table header title in addition to data key label.

devaskim:
Here is single table column of *Entity table* widget. With this PR user would specify custom column header title (see *Power / Capacity* on the screenshot) meanwhile cell content function will operate with data key's label as it was before. For example,
```
return `${entity['Power]} / ${entity['Capacity]}` ;
```
single_table_column


ashvayka:
@vvlladd28 , these changes look promising. Please approve and I will merge.

PR #7399, opened 2022-10-10
fix(sec): upgrade org.apache.cassandra:cassandra-all to 3.11.12

zhoumengyks:
### What happened？
There are 1 security vulnerabilities found in org.apache.cassandra:cassandra-all 3.11.10
- [CVE-2021-44521](https://www.oscs1024.com/hd/CVE-2021-44521)

### What did I do？
Upgrade org.apache.cassandra:cassandra-all from 3.11.10 to 3.11.12 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <!--
 
     Copyright © 2016-2022 The Thingsboard Authors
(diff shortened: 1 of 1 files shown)


smatvienko-tb:
Cassandra-all package is used for the test scope, and the cassandra-migrator tool is once in a lifecycle and quite rare to use at all.

I suggest excluding the `tool` package and the test scope from the check. because those are not in production.

Here is an example of how I made the latest check:
```
mvn org.owasp:dependency-check-maven:aggregate -Dformat=CSV  -DskipTestScope -DnodeAuditSkipDevDependencies -DnodePackageSkipDevDependencies -DassemblyAnalyzerEnabled=false --projects=-tools
```

ashvayka:
Thank you for your efforts. We have analyzed your PR and bumped the version in this commit: https://github.com/thingsboard/thingsboard/commit/08b226c18f3dc0c9c5559f437d463e6894b2b2e6

PR #7475, opened 2022-10-23
[3.4.2] UI: Form field appearance in timezone and entity autocomplete components

devaskim:


ashvayka:
Please provide more info on why you need this change.

devaskim:
We use `outline` appearance in our widgets (creation forms) and re-use TB core components, to look in same manner it would useful to have such option

Issue #6334, opened 2022-03-29
[Feature Request] Latest values support for Timeseries widgets


ikulikov:
### General description

Add ability to specify additional latest values data keys to timeseries widgets. In the widget data settings add additional latest data keys field to datasource. Latest values data is optional and used depending on implementation of each timeseries widget type.

**Chart widgets**

Includes *Timeseries Line Chart* / *Timeseries Bar Chart* / *State Chart* widgets. Latest data keys used to configure chart thresholds and used in data key label patterns.

**Cards**

*Timeseries table* widget. Use latest data keys to show data as separate column, or can be hidden and used in cell content or cell style functions of other data keys.

**Maps**

Includes all *Route Map* widgets and *Trip Animation* widget. Latest data keys generally used to display/edit polygon and circle shapes.

PR #7501, opened 2022-10-27
[3.4.2] UI: Add post function to chart threshold keys.

devaskim:
General settings for data key already supports post processing function. This PR introduces the same functionality for threshold keys.


ashvayka:
Starting ThingsBoard 3.4, you may use the latest values in the time-series widgets (See https://github.com/thingsboard/thingsboard/issues/6334). You may use the latest value (which supports the post-processing function) as a threshold on the line chart.

PR #7503, opened 2022-10-28
[3.4.2] UI: Chart thresholds now support array of values.

devaskim:
With the help of this PR chart would able to draw not only single threshold lines but set of lines based on array of values provided in entity' attribute.


ashvayka:
See my comment to https://github.com/thingsboard/thingsboard/pull/7501#issuecomment-1298309561. You may use multiple latest values to create multiple thresholds. So far, it is not clear why this specific PR is needed. It makes Data Key UI more complicated and usage not obvious.

devaskim:
Our case is a calibration table. It is JSON object with not fixed number of items in single attribute. The keys are what I need to display on chart. So I get JSON value and convert it to array of keys, thus I need post function in threshold keys

PR #7546, opened 2022-11-04
[3.4.2] Fix for PROD-1652 cassandra save

imbeacon:
## Pull Request description

According to [PROD-1652](https://thingsboard-portal.atlassian.net/browse/PROD-1652) - Device states weren't processed after server restart in case of saving activity to telemetry (PERSIST_STATE_TO_TELEMETRY parameter) and DATABASE_TS_TYPE is Cassandra.

The problem was that when 2 threads tried to save data at the same time at the application launch stage. 
The first thread could invoke a class to create statements for data types. But the other, checked only that the variable saveStmts is not null. So it returned the saveStmts[dataType.ordinal()] value, which could still be null at that moment.

This is what provoked such behavior. It also provoked the fact that these devices were not added to the partitionedEntities variable and they were not checked later on whether they were active or not.

## General checklist

## Back-End feature checklist


YevhenBondarenko (approved):

smatvienko-tb:
I guess we worked on the same topic. I suggest to consider my fix https://github.com/thingsboard/thingsboard/pull/7552

PR #7680, opened 2022-11-24
[WIP] Add default state for default user dashboard.

devaskim:
TB let to set user's default dashboard. This PR is to set dashboard's state of default dashboard.

I intentionally didn't make UI changes, because there are several places (widget actions, dashboard state widget) where implemented dashboard state autocomplete. It would be better to create new component for that purpose and re-use it. After that we could add state autocomplete component to user's details tab to set not only default dashboard but also its state.


ashvayka:
Changed to 3.5 because of the code-freeze.

vvlladd28 (approved):

vvlladd28:
Hi @devaskim,
We are added a new component 'tb-dashboard-state-autocomplete'. Now you can use it to complete the PR)

PR #7705, opened 2022-11-30
[3.4.2] Hotfix of some UI component. Extra file extensions removed in UI components

sefaun:


sefaun:
Extra file extensions removed in UI components

PR #7738, opened 2022-12-05
Added redis sentinel support

laarchenko:
Added configuration class to resolve redis sentinel cluster, also refactored alike redis config classes


volodymyr-babak:
1. Why PR is not created by template? 
2. thingsboard.yml file must be updated as well - I don't see these changes.
3. Additionally, documentation must be updated - please created PR for docs as well.
https://thingsboard.io/docs/user-guide/install/config/#redis-connection-parameters

Issue #7757, opened 2022-12-08
[Feature Request] Core: Provision gateway device


devaskim:
**Is your feature request related to a problem? Please describe.**
We are developing LoRa gateway that utilizes TB MQTT gateway API. Also we would like to [provision](https://thingsboard.io/docs/user-guide/device-provisioning/#mqtt-device-apis) such devices. But current implementation (v 3.4.2) doesn't support such functionality.

**Describe the solution you'd like**
Add to provision request boolean option called `gateway` to create either raw device or gateway entity.

**Describe alternatives you've considered**
Implement rule chain to listen for new devices, filter them by provision-enabled device profile and mark `gateway` flag manually via REST node.

ashvayka:
Hi @devaskim , the gateway flag in the request from each device is redundant and, most importantly - insecure.
Please add this flag to the provisioning strategy configuration. It may be located below the provision device key. We should use separate profiles for the gateway and for the device(not the gateway).

PR #7759, opened 2022-12-08
Support of provisioning gateway.

devaskim:
## Pull Request description
Discussion #7757

https://user-images.githubusercontent.com/86909129/206439970-629b5dc8-fe17-4a4d-9daa-236a5bb52671.mp4


ashvayka:
Hi @devaskim , the *gateway* flag in the request from each device is redundant and, most importantly - insecure. 
Please add this flag to the provisioning strategy configuration. It may be located below the provision device key. We should use separate profiles for the gateway and for the device(not the gateway).

ashvayka:
@imbeacon please review once the implementation is updated

imbeacon:
> @imbeacon please review once the implementation is updated

Ok

PR #7773, opened 2022-12-09
[3.4.3][Bug] Getting a lot of updates over WS in case of aggregation query, even if no real updates happened

volodymyr-babak:
## Pull Request description

PR fixes issue with redundant updates over the websockets:
https://github.com/thingsboard/thingsboard/issues/7774

Let's have next timeseries data at 7th of November for next ts: **04:34**, **11:28**, **14:55** and at **20:37**.
We request aggregate by **SUM** of timeseries data for the 7th of November.
Grouping interval is set to 1 hour. 
Cassandra will provide 4 results at ts of these results are: **04:30**, **11:30**, **14:30** and at **20:30**.
**20:30** is used as the latest TS for key states of subscription context, and we are going to request all missed updates starting from **20:30**. But because of this, the timeseries data at **20:37** is going to be considered as missed updates. But this value was already used in **SUM** calculation for **20:00-21:00** period.

Scope of changes:
Added check for aggregation of the query, and in case not **NONE** is used - set ts of the keystate to the end interval of the query, and no the ts of the aggregated interval. This change will fix the issue, that timeseries data, that was already used in the aggregation calculation, is going to be pushed again over the websockets as missed updates.

## General checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/subscription/TbEntityDataSubCtx.java
+++ b/application/src/main/java/org/thingsboard/server/service/subscription/TbEntityDataSubCtx.java
@@ -89,7 +89,7 @@ public class TbEntityDataSubCtx extends TbAbstractDataSubCtx<EntityDataQuery> {
 
     @Override
     protected Aggregation getCurrentAggregation() {
-        return this.curTsCmd.getAgg();
+        return this.curTsCmd.getAgg() != null ? this.curTsCmd.getAgg() : Aggregation.NONE;
     }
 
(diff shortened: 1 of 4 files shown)


application/src/main/java/org/thingsboard/server/service/subscription/TbEntityDataSubCtx.java
```diff
@@ -86,6 +87,11 @@ protected void sendWsMsg(String sessionId, TelemetrySubscriptionUpdate subscript
         }
     }
 
+    @Override
+    protected Aggregation getCurrentAggregation() {
+        return this.curTsCmd.getAgg();
```
ashvayka, line 92: please return Aggregation.NONE in case curTsCmd is null. Just in case :-)

PR #7775, opened 2022-12-09
[3.5] base entity service improvement

ShvaykaD:
## Pull Request description

Original PR: https://github.com/thingsboard/thingsboard/pull/7722   

## General checklist

- [ ] Similar PR is opened for PE version to simplify merge. Required for internal contributors only.
  
## Back-End feature checklist

### DIFF
--- a/common/dao-api/src/main/java/org/thingsboard/server/dao/alarm/AlarmService.java
+++ b/common/dao-api/src/main/java/org/thingsboard/server/dao/alarm/AlarmService.java
@@ -30,13 +30,14 @@ import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.common.data.page.PageData;
 import org.thingsboard.server.common.data.query.AlarmData;
 import org.thingsboard.server.common.data.query.AlarmDataQuery;
+import org.thingsboard.server.dao.entity.SimpleEntityService;
 
 import java.util.Collection;
 
 /**
  * Created by ashvayka on 11.05.17.
  */
-public interface AlarmService {
+public interface AlarmService extends SimpleEntityService {
 
     AlarmOperationResult createOrUpdateAlarm(Alarm alarm);
 
--- a/common/dao-api/src/main/java/org/thingsboard/server/dao/alarm/AlarmService.java
+++ b/common/dao-api/src/main/java/org/thingsboard/server/dao/alarm/AlarmService.java
@@ -30,14 +30,14 @@ import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.common.data.page.PageData;
 import org.thingsboard.server.common.data.query.AlarmData;
 import org.thingsboard.server.common.data.query.AlarmDataQuery;
-import org.thingsboard.server.dao.entity.SimpleEntityService;
+import org.thingsboard.server.dao.entity.TbEntityService;
 
 import java.util.Collection;
 
 /**
  * Created by ashvayka on 11.05.17.
  */
-public interface AlarmService extends SimpleEntityService {
+public interface AlarmService extends TbEntityService {
 
     AlarmOperationResult createOrUpdateAlarm(Alarm alarm);
 
--- a/common/dao-api/src/main/java/org/thingsboard/server/dao/entity/EntityServiceBeanFactory.java
+++ b/common/dao-api/src/main/java/org/thingsboard/server/dao/entity/EntityServiceBeanFactory.java
@@ -1,0 +1,43 @@
+/**
+ * Copyright © 2016-2022 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.dao.entity;
+
+import org.apache.commons.text.CaseUtils;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.stereotype.Service;
+import org.thingsboard.server.common.data.EntityType;
+
+@Service
+public class EntityServiceBeanFactory {
+
+    private static final String SERVICE_SUFFIX = "DaoService";
+
+    private final BeanFactory beanFactory;
+
+    public EntityServiceBeanFactory(BeanFactory beanFactory) {
+        this.beanFactory = beanFactory;
+    }
+
+    public SimpleEntityService getServiceByEntityType(EntityType entityType) {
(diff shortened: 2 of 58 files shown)


common/dao-api/src/main/java/org/thingsboard/server/dao/entity/TbEntityService.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static org.thingsboard.server.common.data.id.EntityId.NULL_UUID;
+
+public interface TbEntityService {
+
+    CustomerId NULL_CUSTOMER_ID = new CustomerId(NULL_UUID);
+
+    Optional<HasId<?>> fetchEntity(TenantId tenantId, EntityId entityId);
+
+    default CustomerId getCustomerId(TenantId tenantId, EntityId entityId) {
```
imbeacon, line 33: What about checking for HasCustomerId interface for found entity via fetchEntity and just return entity.getCustomerId?

common/dao-api/src/main/java/org/thingsboard/server/dao/alarm/AlarmService.java
```diff
@@ -30,13 +30,14 @@
...
 import org.thingsboard.server.common.data.query.AlarmDataQuery;
+import org.thingsboard.server.dao.entity.TbEntityService;
 
 import java.util.Collection;
 
 /**
  * Created by ashvayka on 11.05.17.
  */
-public interface AlarmService {
+public interface AlarmService extends TbEntityService {
```
ashvayka, line 40: Rename TbEntityService to EntityDaoService

common/dao-api/src/main/java/org/thingsboard/server/dao/entity/TbEntityService.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.thingsboard.server.common.data.id.EntityId;
+import org.thingsboard.server.common.data.id.HasId;
+import org.thingsboard.server.common.data.id.TenantId;
+
+import java.util.Optional;
+
+public interface TbEntityService {
+
+    Optional<HasId<?>> fetchEntity(TenantId tenantId, EntityId entityId);
```
ashvayka, line 26: fetchEntity -> findEntity

common/dao-api/src/main/java/org/thingsboard/server/dao/entity/EntityServiceBeanFactory.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final BeanFactory beanFactory;
+
+    public EntityServiceBeanFactory(BeanFactory beanFactory) {
+        this.beanFactory = beanFactory;
+    }
+
+    public TbEntityService getServiceByEntityType(EntityType entityType) {
+        String beanName = EntityType.RULE_NODE.equals(entityType) ? getBeanName(EntityType.RULE_CHAIN) : getBeanName(entityType);
+        return beanFactory.getBean(beanName, TbEntityService.class);
```
ashvayka, line 36: It seems that this method is a bit heavy. You may use it but in post-construct. Loop through all entityTypes and cache results to Map<EntityType, TbEntityService>. This may be additional validation that the method will fail on start and not during runtime.

dao/src/main/java/org/thingsboard/server/dao/entity/BaseEntityService.java
```diff
@@ -125,100 +74,34 @@ public PageData<EntityData> findEntityDataByQuery(TenantId tenantId, CustomerId
...
+                return Optional.ofNullable(hasName.getName());
+            }
         }
-        entityName = Futures.transform(hasName, (Function<HasName, String>) hasName1 -> hasName1 != null ? hasName1.getName() : null, MoreExecutors.directExecutor());
-        return entityName;
+        return Optional.empty();
     }
 
     @Override
     public CustomerId fetchEntityCustomerId(TenantId tenantId, EntityId entityId) {
```
ashvayka, line 93: This method should also return the Optional<CustomerId> to be consistent with the `fetchEntityName`

dao/src/main/java/org/thingsboard/server/dao/util/AbstractBufferedRateExecutor.java
```diff
@@ -327,7 +328,8 @@ public void printStats() {
                     if (printTenantNames) {
                         String name = tenantNamesCache.computeIfAbsent(tenantId, tId -> {
                             try {
-                                return entityService.fetchEntityNameAsync(TenantId.SYS_TENANT_ID, tenantId).get();
+                                Optional<String> entityNameOpt = entityService.fetchEntityName(TenantId.SYS_TENANT_ID, tenantId);
```
ashvayka, line 331:
```
return entityService.fetchEntityName(TenantId.SYS_TENANT_ID, tenantId).orElse("N/A");
```

dao/src/test/java/org/thingsboard/server/dao/service/BaseEntityServiceBeanFactoryTest.java
```diff
@@ -0,0 +1,46 @@
...
+@Slf4j
+public abstract class BaseEntityServiceBeanFactoryTest extends AbstractServiceTest {
+
+    @Autowired
+    private EntityServiceBeanFactory entityServiceBeanFactory;
+
+    @Test
+    public void givenAllEntityTypes_whenGetServiceByEntityTypeCalled_thenNoExceptionsThrows() {
+        for (EntityType entityType : EntityType.values()) {
+            Assertions.assertThatCode(() -> entityServiceBeanFactory.getServiceByEntityType(entityType))
```
ashvayka, line 36: We should check that the result of the method call is not null.

common/dao-api/src/main/java/org/thingsboard/server/dao/entity/EntityServiceBeanFactory.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.thingsboard.server.dao.entity;
+
+import org.apache.commons.text.CaseUtils;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.stereotype.Service;
+import org.thingsboard.server.common.data.EntityType;
+
+@Service
+public class EntityServiceBeanFactory {
```
ashvayka, line 24: Create an interface in the `dao-api` called `EntityServiceRegistry`. Rename this class to Move this class to `DefaultEntityServiceRegistry` and move it to the `dao` module. We must avoid adding new dependencies to the 'dao-api' module, so move the `commons-text` dependency to the dao module.

common/dao-api/src/main/java/org/thingsboard/server/dao/entity/EntityServiceBeanFactory.java
ashvayka: Actually, I believe you should use:

```java
ApplicationContext.getBeansOfType(TbEntityService.class)
```

which already provides you with the map. Then you don't need the complex `CaseUtils.toCamelCase`. You may just replace the '_' with empty string and compare using simple String  "toLowerCase".

dao/src/test/java/org/thingsboard/server/dao/service/BaseEntityServiceBeanFactoryTest.java
ShvaykaD: It's throwing an exception if the bean is not found instead of returning null. That's actually why I have added a check for exception instead of a check for null

Issue #7724, opened 2022-12-02
[Feature Request] Entity Alias for device list with *multivalue* filter by device type


smatvienko-tb:
## Entity Alias (edit dashboard) for a device list with *multivalue* filter by device type

At this moment, only a single device type can be shown by entity alias (see screenshot)
![image](https://user-images.githubusercontent.com/79898499/205278591-f053768c-4912-47af-8372-6310a681f0fd.png)

To override this, the user needs to use additional filters that adds overhead for the high-load databases

**Describe the solution you'd like**
The desired filter may look similar to the "Device search query" that uses a relation query with a multivalue filter by device type (see screenshot)

![image](https://user-images.githubusercontent.com/79898499/205280269-8c6d5920-304d-41b0-8965-340fe12fc425.png)

**Describe alternatives you've considered**
The alternatives already exist, but they bring too much overhead to the database for high-load projects.

**Additional context**
The performance will be skyrocketing!

volodymyr-babak:
Related PR has been successfully merged, and the feature is included in the 3.5 Release. Closing this issue.

PR #8026, opened 2023-02-01
Filter type alias - add possibility to specify multiple types, instead of single one

volodymyr-babak:
## Pull Request description

Added possibility to select multiple device, asset, edge or entity view types in respective alias filters.
Related issue:
https://github.com/thingsboard/thingsboard/issues/7724

Scope of changes:
1. Updated respective filters on java side
2. Updated DAO layer to handle properly multiple filter types
3. Reuse multiple types components on front-end side

Before:

After:

Possible minor changes in screenshots in the docs https://thingsboard.io/docs/user-guide/ui/aliases/ 

Upgrade script to update dashboards provided.

Added functionality into dashboard service to update filters if there are singular filter types present.

Deprecate and not removed singular filter type - to provide possibility for custom widgets that uses WS subscriptions with singular filter type.

## General checklist

- [x ] Changes are backward compatible or upgrade script is provided.
  
## Back-End feature checklist


application/src/main/java/org/thingsboard/server/service/install/update/DefaultDataUpdateService.java
```diff
@@ -681,4 +691,69 @@ private boolean getEnv(String name, boolean defaultValue) {
...
+
+                @Override
+                protected void updateEntity(Tenant tenant) {
+                    updateTenantDashboardsFilters(tenant.getId());
+                }
+            };
+
+    private void updateTenantDashboardsFilters(TenantId tenantId) {
+        PageLink pageLink = new PageLink(100);
+        PageData<DashboardInfo> pageData = dashboardService.findDashboardsByTenantId(tenantId, pageLink);
```
ashvayka, line 720: This approach will be extremely slow in production. You should search the dashboards that have the corresponding filter (based on the full text search, etc) and update only the ones that are needed. We have 50K tenants and maybe 1K dashboards that use this feature on demo.thingsboard.io.

dao/src/main/java/org/thingsboard/server/dao/dashboard/DashboardServiceImpl.java
```diff
@@ -103,13 +108,39 @@ public Dashboard saveDashboard(Dashboard dashboard) {
...
+                updateDashboardFilterIfRequired(entityAlias);
+            }
             return dashboardDao.save(dashboard.getTenantId(), dashboard);
         } catch (Exception e) {
             checkConstraintViolation(e, "dashboard_external_id_unq_key", "Dashboard with such external id already exists!");
             throw e;
         }
     }
 
+    private static void updateDashboardFilterIfRequired(JsonNode entityAlias) {
```
ashvayka, line 121: I just wanted to let you know that this is not needed.

vvlladd28 (approved):

common/data/src/main/java/org/thingsboard/server/common/data/query/AssetTypeFilter.java
```diff
@@ -15,18 +15,41 @@
...
+    /**
+     * Replaced by {@link AssetTypeFilter#getAssetTypes()} instead.
+     */
+    @Deprecated(since = "3.5", forRemoval = true)
+    private String assetType;
+
+    private List<String> assetTypes;
+
+    public List<String> getAssetTypes() {
+        if (assetType != null) {
```
ashvayka, line 41: This check is wrong from a forward-compatibility point of view. You should do something like this:

```
return CollectionUtils.isNotEmpty(assetTypes) ? assetTypes : Collections.singletonList(assetType);
```

My logic is the following:
if new variable assetTypes is set, we should always use it. If not, we should use old variable. Otherwise, if someone update the TB version and uses new UI to add multiple asset types, they will not be able to use it if ```assetType``` is stored in the json file of the dashboard alias.

common/data/src/main/java/org/thingsboard/server/common/data/query/DeviceTypeFilter.java
```diff
@@ -16,21 +16,48 @@
...
+    /**
+     * Replaced by {@link DeviceTypeFilter#getDeviceTypes()} instead.
+     */
+    @Deprecated(since = "3.5", forRemoval = true)
     private String deviceType;
 
+    private List<String> deviceTypes;
+
+    public List<String> getDeviceTypes() {
+        if (deviceType != null) {
```
ashvayka, line 45: See my comment about [assets](https://github.com/thingsboard/thingsboard/pull/8026/files#r1161838850)

common/data/src/main/java/org/thingsboard/server/common/data/query/EntityViewTypeFilter.java
```diff
@@ -15,18 +15,41 @@
...
+    /**
+     * Replaced by {@link EntityViewTypeFilter#getEntityViewTypes()} instead.
+     */
+    @Deprecated(since = "3.5", forRemoval = true)
+    private String entityViewType;
+
+    private List<String> entityViewTypes;
+
+    public List<String> getEntityViewTypes() {
+        if (entityViewType != null) {
```
ashvayka, line 41: See my comment about [assets](https://github.com/thingsboard/thingsboard/pull/8026/files#r1161838850)

PR #8044, opened 2023-02-06
[3.5] fix: remove the mandatory sequence of data binding for labelText and requiredText properties

rjarman:
## Pull Request description

Fix: remove the mandatory sequence of data binding for `labelText` and `requiredText` properties

Example:
This sample code will not work because `requiredText` and `labelText` property are used after `entityType` property. So this pull request will remove this mandatory sequence.
```html
<tb-entity-autocomplete
    fxFlex
    [required]="required"
    (entityChanged)="entityLoaded($event)"
    [entityType]="EntityType.DEVICE "
    [requiredText]="requiredText"
    [labelText]="labelText"
    formControlName="entityId">
</tb-entity-autocomplete>
```

### DIFF
--- a/ui-ngx/src/app/shared/components/entity/entity-autocomplete.component.ts
+++ b/ui-ngx/src/app/shared/components/entity/entity-autocomplete.component.ts
@@ -92,10 +92,18 @@ export class EntityAutocompleteComponent implements ControlValueAccessor, OnInit
   excludeEntityIds: Array<string>;
 
   @Input()
-  labelText: string;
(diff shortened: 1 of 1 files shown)


kalutkaz:
Hello @rjarman you can see the fix for this issue here: https://github.com/thingsboard/thingsboard/pull/8264.

rjarman:
> Thanks for reporting the problem, but the fix is incomplete because the previous version of the code stops working

It should not because it has some basic changes but it's okay @kalutkaz resolved it anyways😅

PR #8061, opened 2023-02-08
Hidden password history

dashevchenko:
## Pull Request description

Added logic that remove password history from User response.

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/controller/UserController.java
```diff
@@ -434,4 +451,12 @@ public void deleteUserSettings(@ApiParam(value = PATHS)
         SecurityUser currentUser = getCurrentUser();
         userSettingsService.deleteUserSettings(currentUser.getTenantId(), currentUser.getId(), Arrays.asList(paths.split(",")));
     }
+    private void hidePasswordHistory(PageData<User> userPageData) {
+        userPageData.getData().forEach(user -> {
+            if (user.getAdditionalInfo().isObject()) {
```
ashvayka, line 456: use the method: `hidePasswordHistory(User)` here to avoid code duplication

ashvayka:
it is clear that the `userPasswordHistory` that is stored in the `additionalInfo` is a bad practice. Instead of supporting bad practise, we should create a separate table where we store the password history and avoid this issue once and forever.

The approach you suggest is ok if we have a deadline for tomorrow. But since this will be in 3.5, you have time to fix it correctly. Don't forget about the upgrade script :-)

PR #8067, opened 2023-02-09
[3.5] UI: Use entity's label (if set) in select entity dialog.

devaskim:
In some cases entity name is not human readable (serial number or UUID), so it would be useful to display entity's label if it is set.


vvlladd28:
Perhaps it is better to write the entity label in parentheses if it exists?

devaskim:
@vvlladd28 Do you think it would be risky to display only label?
Here is how it looks like in our case, where name is UUID
select_entity

How about displaying name and label on different lines? But somewhere in the map implementation label becomes the same as name if label doesn't set.
my

vvlladd28:
Hi @devaskim,
Entity labels are not unique in the platform, so name duplications are possible.
I liked the design option of placing the entity name and entity label in two lines. A workaround you can check that the label and the name have different text

devaskim:
@vvlladd28 I added configuration option in the marker section to switch between entity's name and label in drag-n-drop operations.

vvlladd28:
Thank you for the proposed improvement. We implemented it in PR #8298

PR #8072, opened 2023-02-10
Optionally update dashboard state when using navigate back API.

devaskim:
This PR a bit simplifies dashboard state management in case of "dynamic navigation". 

Our approach in dashboard development is to have single dashboard what covers all type of devices. But user/customer should see only that types what he/she wants or buys. We called it "dynamic navigation". So to have smooth navigation experience, we need to pass additional information in dashboard state.

### DIFF
--- a/ui-ngx/src/app/modules/home/components/dashboard-page/states/default-state-controller.component.ts
+++ b/ui-ngx/src/app/modules/home/components/dashboard-page/states/default-state-controller.component.ts
@@ -168,10 +168,15 @@ export class DefaultStateControllerComponent extends StateControllerComponent im
     }
   }
 
-  public navigatePrevState(index: number): void {
-    if (index < this.stateObject.length - 1) {
-      this.stateObject.splice(index + 1, this.stateObject.length - index - 1);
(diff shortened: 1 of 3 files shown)


ui-ngx/src/app/modules/home/components/dashboard-page/states/default-state-controller.component.ts
```diff
@@ -168,10 +168,15 @@ export class DefaultStateControllerComponent extends StateControllerComponent im
...
   }
 
-  public navigatePrevState(index: number): void {
-    if (index < this.stateObject.length - 1) {
-      this.stateObject.splice(index + 1, this.stateObject.length - index - 1);
-      this.gotoState(this.stateObject[this.stateObject.length - 1].id, true);
+  public navigatePrevState(index: number, params?: StateParams): void {
+    let lastIndex: number = this.stateObject.length - 1;
+    if (index < lastIndex) {
+      this.stateObject.splice(index + 1, lastIndex - 1);
```
vvlladd28, line 174: Why does the **lastIndex** parameter change again?

vvlladd28 (changes requested):

vvlladd28 (approved):

Issue #7701, opened 2022-11-29
[Bug] Websocket get/find alarms API doesn't use customerId


devaskim:
[Proof](https://github.com/thingsboard/thingsboard/blob/master/application/src/main/java/org/thingsboard/server/service/subscription/TbAlarmDataSubCtx.java#L111)

In consequence of that customer user could access all previous alarms (what not belonged to this customer).

devaskim:
Sorry, wrong link. This happens in [TbAlarmDataSubCtx](https://github.com/thingsboard/thingsboard/blob/master/application/src/main/java/org/thingsboard/server/service/subscription/TbAlarmDataSubCtx.java#L111)

PR #8137, opened 2023-02-27
Core: Filter alarms by customer ownership.

devaskim:
I am not sure, that my PR covers all cases. At least, I put several _FIXME_ in code. So I am opened for a discussion.
Here is a [bug](https://github.com/thingsboard/thingsboard/issues/7701)

Steps to reproduce:
1. Assign a device to first customer.
2. Generate alarm.
3. Ensure alarm is visible for the first customer.
4. Re-assign the device to second customer.
5. The generated alarm is visible for the second customer.

*Expected behaviour*
The second customer should see only that alarms that happened during his/her ownership.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
+++ b/application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
@@ -181,6 +181,7 @@ public class DefaultEntityQueryService implements EntityQueryService {
 
     @Override
     public PageData<AlarmData> findAlarmDataByQuery(SecurityUser securityUser, AlarmDataQuery query) {
+        query.setCustomerId(securityUser.getCustomerId());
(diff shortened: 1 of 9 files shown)


ashvayka:
I believe the scope of changes is valid. We need to add ability to filter by CustomerId. Of course, it will require more work on PE, but I will be able to merge the PR to PE as well. Minor commends about the code style below:

Please add the `CustomerId` to the `AlarmDataQuery` instead of separate parameter. This will minimize the required code changes and you will not need to change all the tests. Also, please add a new test validating that the bug you described is now fixed.

devaskim:
@ashvayka Improved PR according to your comments.

ashvayka:
PR looks good. I will merge to 3.5.1. I will try to merge to 3.5 if we will have time.

devaskim:
@ashvayka Why bug fixing is so slow?

PR #8312, opened 2023-04-04
Revert "[Fix bug][3.5] tbel json decode id"

ashvayka:
Reverts thingsboard/thingsboard#8273


ashvayka:
Revert because the artifact is not released yet.

PR #8751, opened 2023-06-13
move data folder under application

emrekardaslar:
## Pull Request description

I moved the data folder under application since the install script needs it. It is easy to do that but it may be better for new developers to just run the script without any file operations.


ashvayka:
Please reopen if you still think this is a required change.

PR #8953, opened 2023-07-18
remove useless setPassword

SecureDong:
## Pull Request description

Remove useless setPassword. The code in question serves no purpose.

## General checklist


ashvayka:
What is the reason for removing those lines? Please provide a detailed explanation.

SecureDong:
> What is the reason for removing those lines? Please provide a detailed explanation.

if (StringUtils.isNotEmpty(mqttCredentials.getPassword())) {
        mqttCredentials.setPassword(mqttCredentials.getPassword());
}

This seems somewhat redundant because it's essentially setting the password to its own existing value.
Firstly, it uses a conditional statement to check if the password in the MQTT credentials (mqttCredentials) is not empty.
If the password is not empty, it performs an operation: it sets the password of the mqttCredentials object to its own current password value.

ashvayka:
Oh, I am sorry... Somehow, I missed that this was the same object. I thought we set the value to a different object. Of course, you are right! So embarrassing..

Issue #7954, opened 2023-01-16
develop new rule node 'device state'


YuriyLytvynchuk:
develop new rule node 'device state'

PR #9030, opened 2023-08-02
Device state rule node; device state service improvements

dskarzh:
## Pull Request description

Case: data must be pulled from external REST API or new data is computed within the rule chain, device is being created inside the rule chain using relation node, for example etc.

Problem: device state service doesn’t know about artificially created device and considers it inactive, when in fact it is active.

Workaround: using integrations or custom-made rule chain just to handle this case is messy and less performant

Solution: add rule node for this case that will directly interact with device state service

Scope of changes: added device state rule node and tests for it; device state service: fixed concurrency bug, minor refactoring of some parts of service code, added tests to cover affected logic

[PE PR](https://github.com/thingsboard/thingsboard-pe/pull/2382)
[UI changes PR](https://github.com/thingsboard/thingsboard-rule-config-ui-ngx/pull/127)
[Documentation PR](https://github.com/thingsboard/thingsboard.github.io/pull/1202) WIP

Please close outdated [PR](https://github.com/thingsboard/thingsboard/pull/7955) and [issue](https://github.com/thingsboard/thingsboard/issues/7954) for this feature by @YuriyLytvynchuk

## General checklist

## Back-End feature checklist


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+@Slf4j
+@RuleNode(
+        type = ComponentType.ACTION,
+        name = "device state",
+        nodeDescription = "Triggers device connectivity events",
+        nodeDetails = "If incoming message originator is a device," +
```
ShvaykaD, line 39: We need to say that rule node registers configured connectivity event in the device state service.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,102 @@
...
+                    .onDeviceActivity(ctx.getTenantId(), (DeviceId) msg.getOriginator(), System.currentTimeMillis()),
+            DataConstants.INACTIVITY_EVENT,
+            (ctx, msg) -> ctx.getRuleEngineDeviceStateService()
+                    .onDeviceInactivity(ctx.getTenantId(), (DeviceId) msg.getOriginator()),
+            DataConstants.DISCONNECT_EVENT,
+            (ctx, msg) -> ctx.getRuleEngineDeviceStateService()
+                    .onDeviceDisconnect(ctx.getTenantId(), (DeviceId) msg.getOriginator())
+    );
+
+    @FunctionalInterface
```
ShvaykaD, line 75: move to the top of the class

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    String event;
+
+    @Override
+    public void init(TbContext ctx, TbNodeConfiguration configuration) throws TbNodeException {
+        event = TbNodeUtils.convert(configuration, TbDeviceStateNodeConfiguration.class).getEvent();
+        if (event == null) {
+            throw new TbNodeException("Event cannot be null!");
+        } else if (!EVENT_ACTIONS.containsKey(event)) {
```
ShvaykaD, line 89: Remove redundant else

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -502,17 +507,21 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
...
     }
 
     @Nonnull
     DeviceStateData getOrFetchDeviceStateData(DeviceId deviceId) {
-        DeviceStateData deviceStateData = deviceStates.get(deviceId);
-        if (deviceStateData != null) {
-            return deviceStateData;
-        }
-        return fetchDeviceStateDataUsingEntityDataQuery(deviceId);
+        return deviceStates.computeIfAbsent(deviceId, this::fetchDeviceStateDataUsingEntityDataQuery);
```
ShvaykaD, line 524: remove deviceStates.putIfAbsent(deviceId, deviceStateData); inside fetchDeviceStateDataUsingEntityDataQuery

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -62,22 +85,163 @@ public class DefaultDeviceStateServiceTest {
...
+        assertThat(actualTrigger.getDeviceId(), is(deviceId));
+        assertThat(actualTrigger.isActive(), is(false));
+    }
+
+    @Test
+    public void givenActivityTimeoutState_whenUpdateInactivityState_thenReportInactivityInvoked() {
+        // GIVEN
+        long currentTime = System.currentTimeMillis();
+        DeviceState deviceState = DeviceState.builder()
+                .active(false)
```
ShvaykaD, line 178: true

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
ShvaykaD: also please rename this method

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -62,22 +85,163 @@ public class DefaultDeviceStateServiceTest {
...
+                try {
+                    allThreadsReadyLatch.await();
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
+                service.getOrFetchDeviceStateData(deviceId);
+            });
+        }
+
+        allThreadsReadyLatch.await();
```
ShvaykaD, line 232: remove latch

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,102 @@
...
+    );
+
+    @FunctionalInterface
+    private interface ConnectivityAction {
+
+        void execute(TbContext ctx, TbMsg msg);
+
+    }
+
+    String event;
```
ShvaykaD, line 82: @ashvayka is it ok to leave the event as a package-private field in order to simplify tests? Or should we refactor this and rewrite the tests?

Same question regarding the couple of methods in the device state service. The reason why we are asking this is because we already have the package-private fields and methods there.

ShvaykaD: I vote to create a separate ticket for test updates and update any package-private fields and methods to remain private.

ashvayka:
The idea is great, but the implementation will not work in a microservices mode. Please note that ```DefaultDeviceStateService``` is marked with the ```@TbCoreComponent```. It is unavailable in the rule engine when launched as a separate microservice. Therefore, you should use ```TbClusterService.pushMsgToCore``` or a similar method to deliver the activity event to the corresponding DeviceStateService.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public void init(TbContext ctx, TbNodeConfiguration configuration) throws TbNodeException {
+        event = TbNodeUtils.convert(configuration, TbDeviceStateNodeConfiguration.class).getEvent();
+        if (event == null) {
+            throw new TbNodeException("Event cannot be null!");
+        }
+        if (!EVENT_ACTIONS.containsKey(event)) {
+            throw new TbNodeException("Unsupported event: " + event);
```
ashvayka, line 92: Use ```throw new TbNodeException("Unsupported event: " + event, true);```. This will mark the error as unrecoverable and rule engine will skip re-init attempts for this node.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
ashvayka: I have checked the tests, and I believe the event may be private. You may execute ```node.init(config with your event)``` to replace your logic.

ShvaykaD: @dskarzh please update the tests.

dskarzh:
Close by accident

ashvayka:
@dskarzh, great contribution! LGTM. @smatvienko-tb , please take a look and approve. If the concurrency bug fix is important, let's merge it to 3.6?

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,187 @@
...
+                "</ul>" +
+                "This node is particularly useful when device isn't using transports to receive data," +
+                " such as when fetching data from external API or computing new data within the rule chain.",
+        configClazz = TbDeviceStateNodeConfiguration.class,
+        uiResources = {"static/rulenode/rulenode-core-config.js"},
+        configDirective = "tbActionNodeDeviceStateConfig"
+)
+public class TbDeviceStateNode implements TbNode {
+
+    private final Map<TbMsgType, ConnectivityEvent> SUPPORTED_EVENTS = Map.of(
```
smatvienko-tb, line 62: List.of

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,187 @@
...
+
+    private void sendDeviceActivityMsg(TbContext ctx, TbMsg msg) {
+        var tenantUuid = ctx.getTenantId().getId();
+        var deviceUuid = msg.getOriginator().getId();
+        var deviceActivityMsg = TransportProtos.DeviceActivityProto.newBuilder()
+                .setTenantIdMSB(tenantUuid.getMostSignificantBits())
+                .setTenantIdLSB(tenantUuid.getLeastSignificantBits())
+                .setDeviceIdMSB(deviceUuid.getMostSignificantBits())
+                .setDeviceIdLSB(deviceUuid.getLeastSignificantBits())
+                .setLastActivityTime(System.currentTimeMillis())
```
smatvienko-tb, line 129: Why the current wall clock? The activity event might come from the queue after a while

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private void sendDeviceDisconnectMsg(TbContext ctx, TbMsg msg) {
+        var tenantUuid = ctx.getTenantId().getId();
+        var deviceUuid = msg.getOriginator().getId();
+        var deviceDisconnectMsg = TransportProtos.DeviceDisconnectProto.newBuilder()
+                .setTenantIdMSB(tenantUuid.getMostSignificantBits())
+                .setTenantIdLSB(tenantUuid.getLeastSignificantBits())
+                .setDeviceIdMSB(deviceUuid.getMostSignificantBits())
+                .setDeviceIdLSB(deviceUuid.getLeastSignificantBits())
```
smatvienko-tb, line 146: when this event happens? See proto comments

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private void sendDeviceConnectMsg(TbContext ctx, TbMsg msg) {
+        var tenantUuid = ctx.getTenantId().getId();
+        var deviceUuid = msg.getOriginator().getId();
+        var deviceConnectMsg = TransportProtos.DeviceConnectProto.newBuilder()
+                .setTenantIdMSB(tenantUuid.getMostSignificantBits())
+                .setTenantIdLSB(tenantUuid.getLeastSignificantBits())
+                .setDeviceIdMSB(deviceUuid.getMostSignificantBits())
+                .setDeviceIdLSB(deviceUuid.getLeastSignificantBits())
```
smatvienko-tb, line 111: when this event happens? See proto comments

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private void sendDeviceInactivityMsg(TbContext ctx, TbMsg msg) {
+        var tenantUuid = ctx.getTenantId().getId();
+        var deviceUuid = msg.getOriginator().getId();
+        var deviceInactivityMsg = TransportProtos.DeviceInactivityProto.newBuilder()
+                .setTenantIdMSB(tenantUuid.getMostSignificantBits())
+                .setTenantIdLSB(tenantUuid.getLeastSignificantBits())
+                .setDeviceIdMSB(deviceUuid.getMostSignificantBits())
+                .setDeviceIdLSB(deviceUuid.getLeastSignificantBits())
```
smatvienko-tb, line 163: when this event happens? See proto comments

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,187 @@
...
+                .build();
+        var toCoreMsg = TransportProtos.ToCoreMsg.newBuilder()
+                .setDeviceInactivityMsg(deviceInactivityMsg)
+                .build();
+        ctx.getClusterService().pushMsgToCore(
+                ctx.getTenantId(), msg.getOriginator(), toCoreMsg, getMsgProcessedCallback(ctx, msg)
+        );
+    }
+
+    private TbQueueCallback getMsgProcessedCallback(TbContext ctx, TbMsg msg) {
```
smatvienko-tb, line 173: I found 27 implementations of this callback.
Do we need another one?
If so, please create a new inner or outer class extending TbQueueCallback.
If you analyze the heap dump, you can count and filter exactly your object types. 
image

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/action/TbDeviceStateNodeTest.java
```diff
@@ -0,0 +1,321 @@
...
+    private static TbMsg DUMMY_MSG;
+    private static DeviceId DUMMY_MSG_ORIGINATOR;
+    @Mock
+    private TbContext ctxMock;
+    @Mock
+    private TbClusterService tbClusterServiceMock;
+    private TbDeviceStateNode node;
+    private TbDeviceStateNodeConfiguration config;
+
+    @BeforeAll
```
smatvienko-tb, line 64: I see no reason to use static content for this test. The static will live forever until JVM shuts down. So static is unwanted.
The data should be isolated between tests

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/action/TbDeviceStateNodeTest.java
```diff
@@ -0,0 +1,321 @@
...
+        var nodeConfig = new TbNodeConfiguration(JacksonUtil.valueToTree(config));
+
+        // WHEN-THEN
+        assertThatThrownBy(() -> node.init(ctxMock, nodeConfig))
+                .isInstanceOf(TbNodeException.class)
+                .hasMessage("Event cannot be null!")
+                .matches(e -> ((TbNodeException) e).isUnrecoverable());
+    }
+
+    @Test
```
smatvienko-tb, line 114: ParametrizedTest with EnumSource.

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/action/TbDeviceStateNodeTest.java
```diff
@@ -0,0 +1,321 @@
...
+        assertThatThrownBy(() -> node.init(ctxMock, nodeConfig))
+                .isInstanceOf(TbNodeException.class)
+                .hasMessage("Unsupported event: " + unsupportedEvent)
+                .matches(e -> ((TbNodeException) e).isUnrecoverable());
+    }
+
+    @Test
+    public void givenNonDeviceOriginator_whenOnMsg_thenTellsSuccessAndNoActivityActionsTriggered() {
+        // GIVEN
+        var originator = new AssetId(UUID.randomUUID());
```
smatvienko-tb, line 131: var asset

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/action/TbDeviceStateNodeTest.java
```diff
@@ -0,0 +1,321 @@
...
+        TransportProtos.ToCoreMsg expectedToCoreMsg = TransportProtos.ToCoreMsg.newBuilder()
+                .setDeviceDisconnectMsg(expectedDeviceDisconnectMsg)
+                .build();
+        assertThat(expectedToCoreMsg).isEqualTo(protoCaptor.getValue());
+
+        then(ctxMock).should(times(1)).tellSuccess(DUMMY_MSG);
+        then(tbClusterServiceMock).shouldHaveNoMoreInteractions();
+        then(ctxMock).shouldHaveNoMoreInteractions();
+    }
+
```
smatvienko-tb, line 320: Please, assert that SUPPORTED_EVENTS.size() == 4
this will fail tests if someone adds supported types in the rule-node and skip the test writing :)

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -503,29 +511,31 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
...
         final Device device = deviceService.findDeviceById(TenantId.SYS_TENANT_ID, deviceId);
         if (device == null) {
             log.warn("[{}] Failed to fetch device by Id!", deviceId);
             throw new RuntimeException("Failed to fetch device by Id " + deviceId);
         }
         try {
-            DeviceStateData deviceStateData = fetchDeviceState(device).get();
-            deviceStates.putIfAbsent(deviceId, deviceStateData);
-            return deviceStateData;
+            return fetchDeviceState(device).get();
```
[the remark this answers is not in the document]
dskarzh: Yes, it is

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -49,14 +53,26 @@
...
+import static org.mockito.Mockito.eq;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
+import static org.thingsboard.server.service.state.DefaultDeviceStateService.ACTIVITY_STATE;
+import static org.thingsboard.server.service.state.DefaultDeviceStateService.INACTIVITY_ALARM_TIME;
 import static org.thingsboard.server.service.state.DefaultDeviceStateService.INACTIVITY_TIMEOUT;
 
 @RunWith(MockitoJUnitRunner.class)
```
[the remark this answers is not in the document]
dskarzh: I have migrated this test to JUnit 5 in a different PR. [Link](https://github.com/thingsboard/thingsboard/pull/9155)

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+        this.event = event;
+    }
+
+    @Override
+    public void onMsg(TbContext ctx, TbMsg msg) {
+        var originator = msg.getOriginator();
+        if (!ctx.isLocalEntity(originator)) {
+            log.warn("[{}][device-state-node] Received message from non-local entity [{}]!", ctx.getSelfId(), originator);
+            return;
```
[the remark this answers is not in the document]
dskarzh: I have added `ctx.ack(msg)`. My logic is that receiving msg non-local entity is not an success or failure at processing it. We just acknowledge it and log that this happened without doing meaningful business logic.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: Set seems better to me

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,187 @@
...
+public class TbDeviceStateNode implements TbNode {
+
+    private final Map<TbMsgType, ConnectivityEvent> SUPPORTED_EVENTS = Map.of(
+            TbMsgType.CONNECT_EVENT, this::sendDeviceConnectMsg,
+            TbMsgType.ACTIVITY_EVENT, this::sendDeviceActivityMsg,
+            TbMsgType.DISCONNECT_EVENT, this::sendDeviceDisconnectMsg,
+            TbMsgType.INACTIVITY_EVENT, this::sendDeviceInactivityMsg
+    );
+
+    private interface ConnectivityEvent {
```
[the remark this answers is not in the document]
dskarzh: Because of changing an implementation to use switch clause this interface is no longer used and thus removed from code

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -341,4 +459,37 @@ private void activityVerify(boolean isActive) {
...
+                try {
+                    allThreadsReadyLatch.await();
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
+                service.getOrFetchDeviceStateData(deviceId);
+            });
+        }
+
+        executor.shutdown();
```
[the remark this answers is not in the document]
dskarzh: In this test, executor is local to this method and because of that I have used try-finally block to ensure that executor will be closed in case something unexpected happens

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -102,21 +120,121 @@ public void setUp() {
...
+
+        then(telemetrySubscriptionService).should().saveAttrAndNotify(
+                any(), eq(deviceId), any(), eq(ACTIVITY_STATE), eq(false), any()
+        );
+
+        var msgCaptor = ArgumentCaptor.forClass(TbMsg.class);
+        then(clusterService).should().pushMsgToRuleEngine(eq(tenantId), eq(deviceId), msgCaptor.capture(), any());
+        var actualMsg = msgCaptor.getValue();
+        assertThat(actualMsg.getType()).isEqualTo(TbMsgType.INACTIVITY_EVENT.name());
+        assertThat(actualMsg.getOriginator()).isEqualTo(deviceId);
```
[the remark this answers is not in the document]
dskarzh: I verify tenant id here `then(clusterService).should().pushMsgToRuleEngine(eq(tenantId), eq(deviceId), msgCaptor.capture(), any());`. If you want me to verify tenant id on the message itself, then I am not really sure how to do that

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -256,14 +256,23 @@ protected void launchMainConsumers() {
                                     log.trace("[{}] Forwarding message to device actor {}", id, toCoreMsg.getToDeviceActorMsg());
                                     forwardToDeviceActor(toCoreMsg.getToDeviceActorMsg(), callback);
                                 } else if (toCoreMsg.hasDeviceStateServiceMsg()) {
-                                    log.trace("[{}] Forwarding message to state service {}", id, toCoreMsg.getDeviceStateServiceMsg());
+                                    log.trace("[{}] Forwarding message to device state service {}", id, toCoreMsg.getDeviceStateServiceMsg());
                                     forwardToStateService(toCoreMsg.getDeviceStateServiceMsg(), callback);
                                 } else if (toCoreMsg.hasEdgeNotificationMsg()) {
                                     log.trace("[{}] Forwarding message to edge service {}", id, toCoreMsg.getEdgeNotificationMsg());
                                     forwardToEdgeNotificationService(toCoreMsg.getEdgeNotificationMsg(), callback);
+                                } else if (toCoreMsg.hasDeviceConnectMsg()) {
```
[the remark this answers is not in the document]
dskarzh: I will not do that. Every engineer has access to searching functionality in just about every text editor imaginable.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -253,16 +253,17 @@ void updateActivityState(DeviceId deviceId, DeviceStateData stateData, long last
                 onDeviceActivityStatusChange(deviceId, true, stateData);
             }
         } else {
-            log.debug("updateActivityState - fetched state IN NULL for device {}, lastReportedActivity {}", deviceId, lastReportedActivity);
+            log.debug("updateActivityState - fetched state IS NULL for device {}, lastReportedActivity {}", deviceId, lastReportedActivity);
```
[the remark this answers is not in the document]
dskarzh: I do not have it available here, because the only place where tenant id is available is state data and this else branch covers the case when mentioned state data is null

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -102,21 +120,121 @@ public void setUp() {
                 .thenReturn(new PageData<>(List.of(deviceIdInfo), 0, 1, false));
     }
 
+    @Test
+    public void givenDeviceBelongsToExternalPartition_whenOnDeviceInactivity_thenCleansStateAndDoesNotReportInactivity() {
+        // GIVEN
+        service.deviceStates.put(deviceId, DeviceStateData.builder().build());
+        given(partitionService.resolve(ServiceType.TB_CORE, tenantId, deviceId))
```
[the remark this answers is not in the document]
dskarzh: To do that I would need to make this method visible inside test class. I try not to do that in my tests. I think that mocking calls to external systems is perfectly fine

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -102,21 +120,121 @@ public void setUp() {
...
+                .willReturn(TopicPartitionInfo.builder().build());
+
+        // WHEN
+        service.onDeviceInactivity(tenantId, deviceId);
+
+        // THEN
+        assertThat(service.deviceStates).isEmpty();
```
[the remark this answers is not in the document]
dskarzh: See my other comment about visibility

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -102,21 +120,121 @@ public void setUp() {
...
+        // THEN
+        then(telemetrySubscriptionService).should().saveAttrAndNotify(
+                any(), eq(deviceId), any(), eq(INACTIVITY_ALARM_TIME), anyLong(), any()
+        );
+        then(telemetrySubscriptionService).should().saveAttrAndNotify(
+                any(), eq(deviceId), any(), eq(ACTIVITY_STATE), eq(false), any()
+        );
+
+        var msgCaptor = ArgumentCaptor.forClass(TbMsg.class);
+        then(clusterService).should().pushMsgToRuleEngine(eq(tenantId), eq(deviceId), msgCaptor.capture(), any());
```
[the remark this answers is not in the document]
dskarzh: I am not sure I understood what you meant here

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: Not sure what are you trying to say

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -503,29 +511,31 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
         }
     }
 
+    private void reportInactivity(long ts, DeviceId deviceId, DeviceStateData stateData) {
+        DeviceState state = stateData.getState();
+        state.setActive(false);
+        state.setLastInactivityAlarmTime(ts);
+        save(deviceId, INACTIVITY_ALARM_TIME, ts);
+        onDeviceActivityStatusChange(deviceId, false, stateData);
```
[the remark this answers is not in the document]
dskarzh: Can you clarify about what you mean here? As far as I know, this piece of code functions in the same fashion as any other part of this class.

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/action/TbDeviceStateNodeTest.java
dskarzh: Please see my other reply about visibility

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/action/TbDeviceStateNodeTest.java
dskarzh: I agree with you. I tried to get rid of static, but later I needed it to be static to use them in parameterized test. I have added `@AfterAll` to perform small cleanup.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: I made one implementation public and used it in my code. Also, it required small changes in place that it was used before. Please see `SimpleTbQueueCallback` and `DefaultTbContext` classes

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -592,17 +601,54 @@ private void forwardToStateService(DeviceStateServiceMsgProto deviceStateService
...
         }
-        TenantId tenantId = toTenantId(deviceActivityMsg.getTenantIdMSB(), deviceActivityMsg.getTenantIdLSB());
-        DeviceId deviceId = new DeviceId(new UUID(deviceActivityMsg.getDeviceIdMSB(), deviceActivityMsg.getDeviceIdLSB()));
+        var tenantId = toTenantId(deviceActivityMsg.getTenantIdMSB(), deviceActivityMsg.getTenantIdLSB());
+        var deviceId = new DeviceId(new UUID(deviceActivityMsg.getDeviceIdMSB(), deviceActivityMsg.getDeviceIdLSB()));
         try {
             stateService.onDeviceActivity(tenantId, deviceId, deviceActivityMsg.getLastActivityTime());
             callback.onSuccess();
         } catch (Exception e) {
-            callback.onFailure(new RuntimeException("Failed update device activity for device [" + deviceId.getId() + "]!", e));
```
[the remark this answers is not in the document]
dskarzh: I think passing caught exception to the the callback is fine, I do not see anything useful that we can add to message on this stage. Also, I improved exception messages in state service so they have useful info in them

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -592,17 +601,54 @@ private void forwardToStateService(DeviceStateServiceMsgProto deviceStateService
         stateService.onQueueMsg(deviceStateServiceMsg, callback);
     }
 
+    private void forwardToStateService(TransportProtos.DeviceConnectProto deviceConnectMsg, TbCallback callback) {
```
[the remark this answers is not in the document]
dskarzh: I fully agree with you that test is needed and I would appreciate some assistance with this test because I do not fully understand what kind of test do you want. Please message me with more explanation or we can schedule a small meeting. Thanks!

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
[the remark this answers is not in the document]
dskarzh: Reverted removal of `RuntimeException`

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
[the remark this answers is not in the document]
ashvayka: ctx.ack is ok here.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,194 @@
...
+            case DISCONNECT_EVENT: {
+                sendDeviceDisconnectMsg(ctx, msg);
+                break;
+            }
+            case INACTIVITY_EVENT: {
+                sendDeviceInactivityMsg(ctx, msg);
+                break;
+            }
+            default: {
+                ctx.tellFailure(msg, new IllegalStateException("Configured event [" + event + "] is not supported!"));
```
ashvayka, line 113: ctx.tellSuccess is missing for the "normal" cases. This will cause processing timeouts. Please fix this.

application/src/main/java/org/thingsboard/server/service/state/DeviceStateService.java
```diff
@@ -27,11 +27,13 @@
  */
 public interface DeviceStateService extends ApplicationListener<PartitionChangeEvent> {
 
-    void onDeviceConnect(TenantId tenantId, DeviceId deviceId);
+    void onDeviceConnect(TenantId tenantId, DeviceId deviceId, long lastConnectTime);
```
ashvayka, line 30: default void onDeviceConnect(TenantId tenantId, DeviceId deviceId, long lastConnectTime){
   onDeviceConnect(tenantId, deviceId, System.currentTimeMillis());
};

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -507,32 +514,34 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
...
     }
 
     @Nonnull
     DeviceStateData getOrFetchDeviceStateData(DeviceId deviceId) {
-        DeviceStateData deviceStateData = deviceStates.get(deviceId);
-        if (deviceStateData != null) {
-            return deviceStateData;
-        }
-        return fetchDeviceStateDataUsingEntityDataQuery(deviceId);
+        return deviceStates.computeIfAbsent(deviceId, this::fetchDeviceStateDataUsingSeparateRequests);
```
ashvayka, line 531: :+1:

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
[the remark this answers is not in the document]
ashvayka: replace

```
        save(deviceId, INACTIVITY_ALARM_TIME, ts);
        onDeviceActivityStatusChange(deviceId, false, stateData);
```

with

```
        onDeviceActivityStatusChange(deviceId, false, stateData);
        save(deviceId, INACTIVITY_ALARM_TIME, ts);
```

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,194 @@
...
+                .build();
+        var toCoreMsg = TransportProtos.ToCoreMsg.newBuilder()
+                .setDeviceActivityMsg(deviceActivityMsg)
+                .build();
+        ctx.getClusterService().pushMsgToCore(
+                ctx.getTenantId(), msg.getOriginator(), toCoreMsg, getMsgEnqueuedCallback(ctx, msg)
+        );
+    }
+
+    private void sendDeviceDisconnectMsg(TbContext ctx, TbMsg msg) {
```
ashvayka, line 154: Please move this code to the TbClusterService, as it may be used by other nodes as well. Also, we need a performance improvement to avoid pushing via the Kafka, if we run in a monolith mode.

ashvayka: See DefaultSubscriptionManagerService <-> DefaultTbLocalSubscriptionService for inspiration. Maybe TbClusterService is not the best place. You may create separate service if needed that will be working in both TbCore, TbRuleEngine, and Monolith.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -507,32 +514,34 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
         }
     }
 
+    private void reportInactivity(long ts, DeviceId deviceId, DeviceStateData stateData) {
+        DeviceState state = stateData.getState();
+        state.setActive(false);
+        state.setLastInactivityAlarmTime(ts);
```
ashvayka, line 520: We need to make sure that we do not overwrite device state with the stale data. What if the inactivityTs < lastActivityTs?

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,194 @@
...
+@Slf4j
+@RuleNode(
+        type = ComponentType.ACTION,
+        name = "device state",
+        nodeDescription = "Triggers device connectivity events",
+        nodeDetails = "If incoming message originator is a device," +
+                " registers configured event for that device in the Device State Service," +
+                " which sends appropriate message to the Rule Engine." +
+                " If metadata <code>ts</code> property is present, it will be used as event timestamp." +
+                " Incoming message is forwarded using the <code>Success</code> chain," +
```
smatvienko-tb, line 45: Otherwise, the message ts will be used.
see `msg.getMetaDataTs()`

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/action/TbDeviceStateNodeTest.java
```diff
@@ -0,0 +1,246 @@
...
+        // THEN
+        then(ctxMock).should(times(1)).isLocalEntity(DEVICE_ID);
+        then(ctxMock).should().getTenantId();
+        then(ctxMock).should().getSelfId();
+        then(ctxMock).should(times(1)).ack(msg);
+        then(ctxMock).shouldHaveNoMoreInteractions();
+    }
+
+    @ParameterizedTest
```
smatvienko-tb, line 168: :+1:

smatvienko-tb (commented):
Hi! It's been a long time since You requested my review.
You did complicated stuff that involved cluster architecture.
Most of the PR looks fine.
Please find my comments below. Some of them are open for discussion. Just reach out if you have any concerns.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
dskarzh: Fixed in https://github.com/thingsboard/thingsboard/pull/9030/commits/e275678c22e4ba89eaca353117c140150fbdb016

application/src/main/java/org/thingsboard/server/service/state/DeviceStateService.java
dskarzh: Fixed in https://github.com/thingsboard/thingsboard/pull/9030/commits/1822703730ee2bace9aed2ebc907df8e79dea4c2

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/action/TbDeviceStateNodeTest.java
```diff
@@ -0,0 +1,246 @@
...
+        metaData.putValue("ts", String.valueOf(METADATA_TS));
+        var data = JacksonUtil.newObjectNode();
+        data.put("humidity", 58.3);
+        msg = TbMsg.newMsg(TbMsgType.POST_TELEMETRY_REQUEST, device.getId(), metaData, JacksonUtil.toString(data));
+    }
+
+    @BeforeEach
+    public void setUp() {
+        node = new TbDeviceStateNode();
+        config = new TbDeviceStateNodeConfiguration();
```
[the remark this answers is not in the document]
dskarzh: Fixed in https://github.com/thingsboard/thingsboard/pull/9030/commits/a1876977652a99b13570403364d5e70115cf454d

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,194 @@
...
+                    ctx.getTenantId().getId(), ctx.getSelfId().getId(), originator.getId());
+            ctx.ack(msg);
+            return;
+        }
+        if (!EntityType.DEVICE.equals(originator.getEntityType())) {
+            ctx.tellSuccess(msg);
+            return;
+        }
+        switch (event) {
+            case CONNECT_EVENT: {
```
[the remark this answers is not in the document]
dskarzh: Fixed in https://github.com/thingsboard/thingsboard/pull/9030/commits/2edd29a3971fe88fdce39a50bcda979c310b2b56

common/queue/src/main/java/org/thingsboard/server/queue/common/SimpleTbQueueCallback.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public SimpleTbQueueCallback(Runnable onSuccess, Consumer<Throwable> onFailure) {
+        this.onSuccess = onSuccess;
+        this.onFailure = onFailure;
+    }
+
+    @Override
+    public void onSuccess(TbQueueMsgMetadata metadata) {
+        if (onSuccess != null) {
+            onSuccess.run();
```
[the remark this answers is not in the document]
dskarzh: I guess you are hinting at `Consumer` functional interface that is able to accept `TbQueueMsgMetadata` instance
Changed `Runnable` to `Consumer<TbQueueMsgMetadata>` in https://github.com/thingsboard/thingsboard/pull/9030/commits/5e846116a885c655bb471776153f843c7f419a34

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -256,14 +256,23 @@ protected void launchMainConsumers() {
...
                                 } else if (toCoreMsg.hasEdgeNotificationMsg()) {
                                     log.trace("[{}] Forwarding message to edge service {}", id, toCoreMsg.getEdgeNotificationMsg());
                                     forwardToEdgeNotificationService(toCoreMsg.getEdgeNotificationMsg(), callback);
+                                } else if (toCoreMsg.hasDeviceConnectMsg()) {
+                                    log.trace("[{}] Forwarding message to device state service {}", id, toCoreMsg.getDeviceConnectMsg());
+                                    forwardToStateService(toCoreMsg.getDeviceConnectMsg(), callback);
                                 } else if (toCoreMsg.hasDeviceActivityMsg()) {
                                     log.trace("[{}] Forwarding message to device state service {}", id, toCoreMsg.getDeviceActivityMsg());
                                     forwardToStateService(toCoreMsg.getDeviceActivityMsg(), callback);
+                                } else if (toCoreMsg.hasDeviceDisconnectMsg()) {
```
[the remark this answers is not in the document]
dskarzh: Added in https://github.com/thingsboard/thingsboard/pull/9030/commits/d7a24648ff7ad463461552e317f67305c1460224

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: Please review my implementation of this in https://github.com/thingsboard/thingsboard/pull/9030/commits/eadcf916f5c57b938776a971797f6c0d416040c8

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -592,17 +601,54 @@ private void forwardToStateService(DeviceStateServiceMsgProto deviceStateService
         stateService.onQueueMsg(deviceStateServiceMsg, callback);
     }
 
-    private void forwardToStateService(TransportProtos.DeviceActivityProto deviceActivityMsg, TbCallback callback) {
+    void forwardToStateService(TransportProtos.DeviceConnectProto deviceConnectMsg, TbCallback callback) {
+        var tenantId = toTenantId(deviceConnectMsg.getTenantIdMSB(), deviceConnectMsg.getTenantIdLSB());
+        var deviceId = new DeviceId(new UUID(deviceConnectMsg.getDeviceIdMSB(), deviceConnectMsg.getDeviceIdLSB()));
+        try {
+            stateService.onDeviceConnect(tenantId, deviceId, deviceConnectMsg.getLastConnectTime());
```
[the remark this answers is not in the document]
dskarzh: Please see if got your point correctly in https://github.com/thingsboard/thingsboard/pull/9030/commits/b1b0e44c7fda9118cea95b6715430cf1641943bc

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: Fixed in https://github.com/thingsboard/thingsboard/pull/9030/commits/b19d7f1406775e0afd511e46ecd08a4594907064

dskarzh:
@ashvayka @smatvienko-tb Thanks for all the comments. I made according changes. You can find specific commits with changes in my replies for your convenience.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
ShvaykaD: Seems that the issue is not fixed yet. I don't see a check for the condition: lastInactivityTime is greater than lastActivityTime.

ShvaykaD: It needs to be greater than current lastActivityTime and current lastInactivityTime

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,198 @@
...
+
+            return now;
+        });
+    }
+
+    private void scheduleCleanupMsg(TbContext ctx) {
+        TbMsg cleanupMsg = ctx.newMsg(
+                null, TbMsgType.DEVICE_STATE_STALE_ENTRIES_CLEANUP_MSG, ctx.getSelfId(), TbMsgMetaData.EMPTY, TbMsg.EMPTY_STRING
+        );
+        ctx.tellSelf(cleanupMsg, Duration.ofHours(1L).toMillis());
```
ShvaykaD, line 132: move `Duration.ofHours(1L).toMillis()` to constant.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,198 @@
...
+        stopwatch = Stopwatch.createStarted();
+        scheduleCleanupMsg(ctx);
+    }
+
+    @Override
+    public void onMsg(TbContext ctx, TbMsg msg) {
+        if (msg.isTypeOf(TbMsgType.DEVICE_STATE_STALE_ENTRIES_CLEANUP_MSG)) {
+            removeStaleEntries();
+            scheduleCleanupMsg(ctx);
+            ctx.ack(msg);
```
ShvaykaD, line 97: this message is not persisted in the queue so no needs to ack it.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,198 @@
...
+            return;
+        }
+
+        DeviceId originator = new DeviceId(msg.getOriginator().getId());
+
+        lastActivityEventTimestamps.compute(originator, (__, lastEventTs) -> {
+            Duration now = stopwatch.elapsed();
+
+            if (lastEventTs == null) {
+                sendEvent(ctx, originator, msg);
```
ShvaykaD, line 112: please rename to sendEventAndTell

ShvaykaD, line 112:
```suggestion
                sendEventAndTell(ctx, originator, msg);
```

common/data/src/main/java/org/thingsboard/server/common/data/msg/TbMsgType.java
```diff
@@ -76,6 +76,7 @@ public enum TbMsgType {
     DEDUPLICATION_TIMEOUT_SELF_MSG(null, true),
     DELAY_TIMEOUT_SELF_MSG(null, true),
     MSG_COUNT_SELF_MSG(null, true),
+    DEVICE_STATE_STALE_ENTRIES_CLEANUP_MSG(null, true),
```
ShvaykaD, line 79: Maybe add SELF:

```suggestion
    DEVICE_STATE_STALE_ENTRIES_CLEANUP_SELF_MSG(null, true),
```

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -195,6 +199,7 @@ public void init() {
         this.consumersExecutor = Executors.newCachedThreadPool(ThingsBoardThreadFactory.forName("tb-core-consumer"));
         this.usageStatsExecutor = Executors.newSingleThreadExecutor(ThingsBoardThreadFactory.forName("tb-core-usage-stats-consumer"));
         this.firmwareStatesExecutor = Executors.newSingleThreadExecutor(ThingsBoardThreadFactory.forName("tb-core-firmware-notifications-consumer"));
+        this.deviceActivityEventsExecutor = MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor(ThingsBoardThreadFactory.forName("tb-core-device-activity-events-consumer")));
```
ShvaykaD, line 202: this executor used only to forward events to device state service. So consider to replace consumer with something else, or remove it at all.

dskarzh: Renamed in https://github.com/thingsboard/thingsboard/pull/9030/commits/691a6b5d6a1cc0acd300eca5d9396e296207bf82

common/data/src/main/java/org/thingsboard/server/common/data/msg/TbMsgType.java
dskarzh: Renamed in https://github.com/thingsboard/thingsboard/pull/9030/commits/691a6b5d6a1cc0acd300eca5d9396e296207bf82

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: Renamed in https://github.com/thingsboard/thingsboard/pull/9030/commits/691a6b5d6a1cc0acd300eca5d9396e296207bf82

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: Fixed in https://github.com/thingsboard/thingsboard/pull/9030/commits/691a6b5d6a1cc0acd300eca5d9396e296207bf82

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: Moved in https://github.com/thingsboard/thingsboard/pull/9030/commits/691a6b5d6a1cc0acd300eca5d9396e296207bf82

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
dskarzh: Good point, fixed in https://github.com/thingsboard/thingsboard/pull/9030/commits/46dca971e119c031f63cd5fab40390d6cf9576a0

application/src/main/java/org/thingsboard/server/service/state/LocalRuleEngineDeviceStateManager.java
```diff
@@ -0,0 +1,85 @@
...
+import org.thingsboard.rule.engine.api.RuleEngineDeviceStateManager;
+import org.thingsboard.server.common.data.id.DeviceId;
+import org.thingsboard.server.common.data.id.TenantId;
+import org.thingsboard.server.common.msg.queue.TbCallback;
+import org.thingsboard.server.queue.util.TbCoreComponent;
+
+@Slf4j
+@Service
+@Primary
+@TbCoreComponent
```
ashvayka, line 31: use partitionService.isMyPartition to detect routing strategy

ashvayka: Use 
```
public void pushMsgToCore(TopicPartitionInfo tpi, UUID msgId, ToCoreMsg msg, TbQueueCallback callback)
```

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,197 @@
...
+    @Override
+    public void onMsg(TbContext ctx, TbMsg msg) {
+        if (msg.isTypeOf(TbMsgType.DEVICE_STATE_STALE_ENTRIES_CLEANUP_SELF_MSG)) {
+            removeStaleEntries();
+            scheduleCleanupMsg(ctx);
+            return;
+        }
+
+        if (!EntityType.DEVICE.equals(msg.getOriginator().getEntityType())) {
+            ctx.tellSuccess(msg);
```
ashvayka, line 102: tellFailure. We can't use DeviceStateNode with Assets, etc.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,197 @@
...
+            Duration now = stopwatch.elapsed();
+
+            if (lastEventTs == null) {
+                sendEventAndTell(ctx, originator, msg);
+                return now;
+            }
+
+            Duration elapsedSinceLastEventSent = now.minus(lastEventTs);
+            if (elapsedSinceLastEventSent.compareTo(ONE_SECOND) < 0) {
+                ctx.tellSuccess(msg);
```
ashvayka, line 118: rateLimit + add to annotation the corresponding relation types

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
```diff
@@ -0,0 +1,197 @@
...
+
+    private static final Set<TbMsgType> SUPPORTED_EVENTS = EnumSet.of(
+            TbMsgType.CONNECT_EVENT, TbMsgType.ACTIVITY_EVENT, TbMsgType.DISCONNECT_EVENT, TbMsgType.INACTIVITY_EVENT
+    );
+    private static final Duration ONE_SECOND = Duration.ofSeconds(1L);
+    private static final Duration ENTRY_EXPIRATION_TIME = Duration.ofDays(1L);
+    private static final Duration ENTRY_CLEANUP_PERIOD = Duration.ofHours(1L);
+
+    private Stopwatch stopwatch;
+    private ConcurrentMap<DeviceId, Duration> lastActivityEventTimestamps;
```
ashvayka, line 75: ConcurrentReferenceHashMap<DeviceId, TbRateLimits>  with SOFT
Make the rate limit configuration an env parameter. Default value: 1:1,30:60,60:3600.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: https://github.com/thingsboard/thingsboard/pull/9030/commits/094cff6174039febbf3fc3a4b64bc99085f79942

application/src/main/java/org/thingsboard/server/service/state/LocalRuleEngineDeviceStateManager.java
dskarzh: Please see https://github.com/thingsboard/thingsboard/pull/9030/commits/c5c18202c21dfb7714b67efb8ce501e50773bdaf

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbDeviceStateNode.java
dskarzh: Please see https://github.com/thingsboard/thingsboard/pull/9030/commits/e55d14325a7658193eca00115c7f918098a8406c

application/src/main/java/org/thingsboard/server/service/state/DefaultRuleEngineDeviceStateManager.java
```diff
@@ -0,0 +1,194 @@
...
+import org.thingsboard.server.queue.common.SimpleTbQueueCallback;
+import org.thingsboard.server.queue.discovery.PartitionService;
+import org.thingsboard.server.queue.discovery.TbServiceInfoProvider;
+
+import java.util.Optional;
+import java.util.UUID;
+
+@Slf4j
+@Service
+public class DefaultRuleEngineDeviceStateManager implements RuleEngineDeviceStateManager {
```
ashvayka, line 38: Please mark this with the @TbRuleEngineComponent

application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
```diff
@@ -203,6 +204,10 @@ public ConcurrentMap<TenantId, DebugTbRateLimits> getDebugPerTenantLimits() {
     @Getter
     private DeviceCredentialsService deviceCredentialsService;
 
+    @Autowired
```
ashvayka, line 207: @Autowired(required = false)

PR #9062, opened 2023-08-10
Feature/Edge: tenant and tenant profile sync up

AndriiLandiak:
## Pull Request description

Sync up tenant and tenant profile info on edge, so if any configuration for specific tenant profile has changed, edge will get notification about this with new configuration.

## General checklist

## Back-End feature checklist


dao/src/test/java/org/thingsboard/server/dao/service/AbstractServiceTest.java
```diff
@@ -107,29 +108,10 @@ protected RuleNodeDebugEvent generateEvent(TenantId tenantId, EntityId entityId)
                 .data(JacksonUtil.toString(readFromResource("TestJsonData.json")))
                 .build();
     }
-//
```
ashvayka, line 110: Thx

common/data/src/main/java/org/thingsboard/server/common/data/EdgeUtils.java
```diff
@@ -25,60 +25,35 @@
...
+    private static final int STACK_TRACE_LIMIT = 10;
+
+    private EdgeUtils() {}
+
     public static int nextPositiveInt() {
         return ThreadLocalRandom.current().nextInt(0, Integer.MAX_VALUE);
     }
 
+    public static EdgeEventType getEdgeEventTypeByEntityType(EntityType entityType) {
+        return entityTypeEdgeEventTypeEnumMap.get(entityType);
```
ashvayka, line 54: You have missed the warning after refactoring. If map does not have the value, you should print the same warnings.
See:

```
log.warn("Unsupported entity type [{}]", entityType);
```

AndriiLandiak: Thanks for noticing, but I removed this warning on purpose from being duplicated. At the method's invocation points, there were (and still are) null checkers, which log if the type is not supported. If this warning needs to be returned as well, I will provide it. 

For example, call points
image
image

PR #9155, opened 2023-08-30
[3.6] Device state service: fix device not being timeouted bug, fix flaky tests, add more test coverage

dskarzh:
## Pull Request description

[Bug](https://github.com/thingsboard/thingsboard/issues/9156) - wrong timeout expiry check

No documentation changes required

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -486,7 +486,7 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
         if (stateData != null) {
             DeviceState state = stateData.getState();
             if (!isActive(ts, state)
-                    && (state.getLastInactivityAlarmTime() == 0L || state.getLastInactivityAlarmTime() < state.getLastActivityTime())
+                    && (state.getLastInactivityAlarmTime() == 0L || state.getLastInactivityAlarmTime() <= state.getLastActivityTime())
```
smatvienko-tb, line 489: please, add parametrized tests to cover all corner cases

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -294,6 +294,7 @@ public void decreaseInactivityForActiveDeviceTest() throws Exception {
         Mockito.verify(telemetrySubscriptionService, Mockito.never()).saveAttrAndNotify(Mockito.any(), Mockito.eq(deviceId), Mockito.any(), Mockito.eq("active"), Mockito.any(), Mockito.any());
 
         long newTimeout = 1;
+        Thread.sleep(newTimeout);
```
smatvienko-tb, line 297: This test fails 100% in the same millisecond. As we saw when we mocked the system time wrapper.

I agree with your suggestion to add another test. But in the parametrized way.
AtomicLong atomicTs = new AtomicLong(); // in the Setup section

Lets mock `willAnswer (() -> atomicTs.addAndGet(msFromParameter)).given(service).currentTimeMs()`
 msFromParameter=0 - you got a frozen time
 msFromParameter=1 - you got controlled time each call will return the next ms (this is default mock)

smatvienko-tb (changes requested):
please, find my comments below

dskarzh:
What has changed:
- Migrated to JUnit 5 for better support for parameterized tests
- Minor code style changes like adding static imports to avoid repeating Mockito each time, replaced literals with constants etc..
- Removed redundant `Thread.sleep()` from `decreaseInactivityForInactiveDeviceTest()`. Reason why it initially was added turned out to affect only `decreaseInactivityForActiveDeviceTest()` 
- Added test to with controlled time for decreasing inactivity timeout use case when device is active: now correctly checks that if timeout is decreased to minimum value (1) and it has not yet passed then device should stay active
- Added test to cover normal use cases and corner cases of `updateInactivityStateIfExpired()`. This test is not in any way exhaustive and was designed to check only realistic conditions. If we account for user having full control over some attributes it becomes very hard to test all cases because time can be negative etc. (Maybe it is a good idea to restrict user from changing attributes used to represent device state and giving him alternative ways to solve use cases which previously were solved by directly manipulating device state attributes. Other side of this is that restricting direct access to these attributes is a breaking change, because a lot of people rely on it)
- Made `stateData.getDeviceCreationTime() + state.getInactivityTimeout() = ts` check to be constistent with `isActive()`.

dskarzh:
image

PR #9390, opened 2023-10-10
[DO NOT MERGE] added fst stats

YevhenBondarenko:
`<logger name="org.thingsboard.server.common.data.FSTUtils" level="DEBUG" />`

Screenshot from 2023-10-10 16-21-06


ashvayka:
We will not merge this to the master since FST is removed in 3.7. But we have cherry-picked this to demo

PR #9416, opened 2023-10-15
Add entity service to actor context.

devaskim:
**The case**

Periodically our system needs to select devices based on several attributes and telemetry keys, then send to each of them RPC command. For these purpose we need custom rule node that will be able to get devices with predefined filter.

**Drawback of implementation**
Dynamic filters are not supported, because this functionality is in `application` module.

**Partial workaround**
1. Implement log in procedure.
2. Call Entities Query REST API
3. Convert response to separate messages per each device.
4. !!! **Change originator to device entity based on its ID located in message** !!! (doesn't support by TB yet)

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
@@ -66,6 +66,7 @@ import org.thingsboard.server.dao.device.DeviceProfileService;
 import org.thingsboard.server.dao.device.DeviceService;
 import org.thingsboard.server.dao.edge.EdgeEventService;
 import org.thingsboard.server.dao.edge.EdgeService;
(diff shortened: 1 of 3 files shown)


devaskim:
@NistantriTech-Dhaval Yes, with custom rule nodes we can implement a lot of useful things. I meant there is not support out of the box by existing `Change Originator` node.

devaskim:
@ashvayka Are there any plans to review and hopefully merge the pr or can I just close it for the lack of necessity? I understand that the dev team can implement some/all features in more reliable way than such contributors as me. Just no silence, please, it is really frustrating.

ashvayka:
So sorry about the frustration. I have merged this, but I must warn everyone who is going to use this (and other DAO layer services) - the DB queries, especially the ones that are not cached, are expensive. You should use this rule node only in special cases and avoid invocation of the query on every message.

PR #9534, opened 2023-10-31
Fix trim double quotes for RestApiCallNode

AndriiLandiak:
## Pull Request description

_Issue:_ Currently, the implementation is not allowed to send a correct **plain/text** payload. The RestApiCallNode features a configuration field labeled "trimDoubleQuotes.", however, this approach has proven problematic. It does not adequately handle situations where specific symbols like newline characters \n, tabs \t, curly braces {}, square brackets [], and others are present in the serialized message.

_Fix:_ To address this issue, we should implement a more robust solution. Specifically, we should perform deserialization after the message has been serialized within the unbindMsg() method of the RuleNodeScriptEngine (JacksonUtil.toString()). 
This deserialization process should make use of JacksonUtil.fromString(). To ensure the correctness of this approach, comprehensive tests have been provided to cover scenarios involving these special symbols.

Rule node UI: https://github.com/thingsboard/thingsboard-rule-config-ui-ngx/pull/147

## General checklist

## Back-End feature checklist

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
@@ -247,7 +247,7 @@ public class TbHttpClient {
     }
 
     protected String parseJsonStringToPlainText(String data, boolean parseToJson) {
-        if (data.startsWith("\"") && data.endsWith("\"") && data.length() >= 2) {
+        if (parseToJson && data.startsWith("\"") && data.endsWith("\"") && data.length() >= 2) {
             final String dataBefore = data;
             try {
                 data = JacksonUtil.fromString(data, String.class);
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
@@ -194,7 +194,7 @@ public class TbHttpClient {
                 config.isIgnoreRequestBody()) {
             entity = new HttpEntity<>(headers);
         } else {
-            entity = new HttpEntity<>(getData(msg), headers);
+            entity = new HttpEntity<>(getData(msg, config.isIgnoreRequestBody(), config.isParseToPlainText()), headers);
         }
 
         URI uri = buildEncodedUri(endpointUrl);
@@ -242,12 +242,15 @@ public class TbHttpClient {
         return uri;
     }
 
-    private String getData(TbMsg tbMsg) {
-        return parseJsonStringToPlainText(tbMsg.getData(), config.isParseToPlainText());
+    private String getData(TbMsg tbMsg, boolean ignoreBody, boolean parseToPlainText) {
+        if (!ignoreBody && parseToPlainText) {
+            return parseJsonStringToPlainText(tbMsg.getData());
+        }
+        return tbMsg.getData();
     }
 
-    protected String parseJsonStringToPlainText(String data, boolean parseToJson) {
-        if (parseToJson && data.startsWith("\"") && data.endsWith("\"") && data.length() >= 2) {
+    protected String parseJsonStringToPlainText(String data) {
+        if (data.startsWith("\"") && data.endsWith("\"") && data.length() >= 2) {
             final String dataBefore = data;
             try {
                 data = JacksonUtil.fromString(data, String.class);
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
@@ -243,7 +243,7 @@ public class TbHttpClient {
     }
 
     private String getData(TbMsg tbMsg) {
-        return parseJsonStringToPlainText(tbMsg.getData(), config.isTrimDoubleQuotes());
+        return parseJsonStringToPlainText(tbMsg.getData(), config.isParseToPlainText());
     }
 
     protected String parseJsonStringToPlainText(String data, boolean parseToJson) {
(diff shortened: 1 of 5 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -242,12 +242,18 @@ public URI buildEncodedUri(String endpointUrl) {
...
     }
 
-    private String getData(TbMsg msg) {
-        String data = msg.getData();
+    private String getData(TbMsg tbMsg) {
+        return trimDoubleQuotesIfRequired(tbMsg.getData());
+    }
 
-        if (config.isTrimDoubleQuotes()) {
+    protected String trimDoubleQuotesIfRequired(String data) {
```
smatvienko-tb, line 249: 1. The feature has to be renamed to make more sense. as it acts as JSON.parse to make plain/text from JSON string.
2. UI and description and documentation have to be adjusted.
3. the method getData should get `getConfig().isTrimDoubleQuotes()`as a boolean parameter. This make your code better and much easier to test without excessive mocking

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -242,12 +242,18 @@ public URI buildEncodedUri(String endpointUrl) {
         return uri;
     }
 
-    private String getData(TbMsg msg) {
-        String data = msg.getData();
+    private String getData(TbMsg tbMsg) {
```
smatvienko-tb, line 245: the method seems useless

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/rest/TbHttpClientTest.java
```diff
@@ -60,10 +61,12 @@ public class TbHttpClientTest {
 
     EventLoopGroup eventLoop;
     TbHttpClient client;
+    TbRestApiCallNodeConfiguration config;
 
     @Before
     public void setUp() throws Exception {
         client = mock(TbHttpClient.class);
+        config = mock(TbRestApiCallNodeConfiguration.class);
```
smatvienko-tb, line 69: I guess this is not necessary, just make a refactoring of the subject method

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/rest/TbHttpClientTest.java
```diff
@@ -224,4 +227,55 @@ public void testHeadersToMetaData() {
         Assertions.assertThat(data.get("Set-Cookie")).isEqualTo("[\"sap-context=sap-client=075; path=/\",\"sap-token=sap-client=075; path=/\"]");
     }
 
-}
\ No newline at end of file
+    @Test
```
smatvienko-tb, line 230: Please, make a parametrized test

smatvienko-tb (changes requested):
Hey!
This fix makes perfect sense, as we discussed in the morning. 

But for the product, it requires more effort from You.
Please find my comments below.

smatvienko-tb:
as a reference:

JSON.stringify()
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

JSON.parse()
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse

smatvienko-tb:
please, add in the description that the current implementation are not allowed to send a correct **plain/text** payload

ashvayka:
Let's analyze the headers configured by the user. If we are sending the JSON ("Content-type" = "application/json;charset=UTF-8"), we should create an HTTPEntity with correct type and use it. We can also apply validation that we have a valid JSON before posting it. IF we are sending the request as plain/text then we should convert it to String. 
See more here: https://www.baeldung.com/spring-resttemplate-post-json#2-handling-character-encoding

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/rest/TbHttpClientTest.java
```diff
@@ -219,9 +225,41 @@ public void testHeadersToMetaData() {
...
-        Assertions.assertThat(data.get("Set-Cookie")).isEqualTo("[\"sap-context=sap-client=075; path=/\",\"sap-token=sap-client=075; path=/\"]");
+        Assertions.assertEquals(2, data.size());
+        Assertions.assertEquals(data.get("Content-Type"), "binary");
+        Assertions.assertEquals(data.get("Set-Cookie"), "[\"sap-context=sap-client=075; path=/\",\"sap-token=sap-client=075; path=/\"]");
     }
 
-}
\ No newline at end of file
+    @ParameterizedTest
+    @MethodSource("provideParameters")
```
smatvienko-tb, line 234: Please, use
`@ValueSource(strings = {"", "  "})`
for primitive types

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -242,12 +242,16 @@ public URI buildEncodedUri(String endpointUrl) {
...
+        return parseJsonStringToPlainText(tbMsg.getData(), config.isParseToPlainText());
+    }
 
-        if (config.isTrimDoubleQuotes()) {
+    protected String parseJsonStringToPlainText(String data, boolean parseToJson) {
```
smatvienko-tb, line 249: it looks like the method doing something only for parseToJson == true
It makes sense to move such a case upper and not fire the method if parseToJson == false

smatvienko-tb, line 249: I see there is "no body" configuration present. This means parsing data makes no sense at all when "no body" == true

smatvienko-tb (commented):
Please, find some minor notes below

PR #9600, opened 2023-11-10
Improved afterTest method

dashevchenko:
## Pull Request description

Improved test cleanup method to ensure tenant deletion.

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
+++ b/application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
@@ -29,7 +29,6 @@ import org.apache.http.HttpStatus;
 import org.awaitility.Awaitility;
 import org.hamcrest.Matcher;
 import org.hibernate.exception.ConstraintViolationException;
-import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
@@ -359,6 +358,10 @@ public abstract class AbstractWebTest extends AbstractInMemoryStorageTest {
             try {
                 status = doDelete("/api/tenant/" + tenantId.getId().toString())
                         .andReturn().getResponse().getStatus();
+                if (status != HttpStatus.SC_OK) {
+                    log.warn("Tenant deletion failed, tenantId: {}", tenantId.getId().toString());
+                    Thread.sleep(1000L);
+                }
             } catch (Exception e) {
                 throw new RuntimeException(e);
             }
--- a/application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
+++ b/application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
@@ -333,6 +333,8 @@ public abstract class AbstractWebTest extends AbstractInMemoryStorageTest {
         log.debug("Executing web test teardown");
 
         loginSysAdmin();
+        deleteTenant(tenantId);
+        deleteDifferentTenant();
         verifyNoTenantsLeft();
 
         tenantProfileService.deleteTenantProfiles(TenantId.SYS_TENANT_ID);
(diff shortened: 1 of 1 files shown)


application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
```diff
@@ -330,18 +333,39 @@ public void teardownWebTest() throws Exception {
...
         tenantProfileService.deleteTenantProfiles(TenantId.SYS_TENANT_ID);
 
         log.info("Executed web test teardown");
     }
 
-    void verifyNoTenantsLeft() throws Exception {
+    private void verifyNoTenantsLeft() throws Exception {
+        List<Tenant> loadedTenants = getAllTenants();
+        if (!loadedTenants.isEmpty()) {
+            loadedTenants.forEach(tenant -> deleteTenant(tenant.getId()));
```
smatvienko-tb, line 348: I suggest you log a warning if loadedTenants is not empty. this may help to find the root cause test (first match).

application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
```diff
@@ -330,18 +333,39 @@ public void teardownWebTest() throws Exception {
...
+            loadedTenants = getAllTenants();
+        }
+        assertThat(loadedTenants).as("All tenants expected to be deleted, but some tenants left in the database").isEmpty();
+    }
+
+    private void deleteTenant(TenantId tenantId) {
+        int status = 0;
+        int retries = 0;
+        while (status != HttpStatus.SC_OK && retries < CLEANUP_TENANT_RETRIES_COUNT) {
```
smatvienko-tb, line 357: In general, the problem is that the tenant can not be deleted is something permanent. like some related entity having a foreign key, etc.
It might be helpful for some cases, when the test fails and not delete additional tenants. Will see...

application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
```diff
@@ -25,9 +25,11 @@
 import io.jsonwebtoken.Jwt;
 import io.jsonwebtoken.Jwts;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.http.HttpStatus;
 import org.awaitility.Awaitility;
 import org.hamcrest.Matcher;
 import org.hibernate.exception.ConstraintViolationException;
+import org.jetbrains.annotations.NotNull;
```
smatvienko-tb, line 32: It seems unused

smatvienko-tb (commented):
Hey! Seeing your effort to improve the tests for the whole team is nice!

I added the tenant's presence check to prevent unclear errors like 'can't create default tenant profile' raised at the beginning of some innocent tests.

The fix was helpful and forced me to fix some particular tests instead of making a global workaround.

The only concern for this moment is whether we should continue testing at all if the state of the DB is unclear.

The plan is to use a HouseKeeper async service for a lengthy operation like a delete tenant.

We will need to call it and wait until such a service does all the cleanup after the test.

application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
dashevchenko: Hey, thank you for the review! This fix is dedicated to the race condition when some unprocessed messages fail tenant deletion. The plan is to create test tenants only once before all tests and not do clean-up (it could save some build time)

application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
ashvayka: @smatvienko-tb , please create an automatic check to fail the build if the code has "org.jetbrains.annotations".

application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java
ashvayka: We should add waring every time we can't delete the tenant and also add a sleep for 1 sec between retries

PR #9617, opened 2023-11-13
Edge - JSON converter for proto

AndriiLandiak:
## Pull Request description

**Issue**:
Currently, maintaining backward compatibility with Edge when new fields are added to entities is cumbersome. It involves updating proto configurations and extensive testing after any Tb entity updates.

**Solution**:
To simplify this process, this PR uses serialization and deserialization for sending entities. By using Object Mapper serialization and setting FAIL_ON_UNKNOWN_PROPERTIES to **false** during deserialization, we can seamlessly handle different entity versions between our Edge and TB.

## General checklist

## Back-End feature checklist


volodymyr-babak (approved):

volodymyr-babak:
Edge processors versioning is planned as well, but as a separate pull request.

common/edge-api/src/main/proto/edge.proto
```diff
@@ -138,25 +139,28 @@ message RuleChainUpdateMsg {
...
-  bool root = 7;
-  bool debugMode = 8;
-  string configuration = 9;
+  string name = 4 [deprecated = true];
+  optional int64 firstRuleNodeIdMSB = 5 [deprecated = true];
+  optional int64 firstRuleNodeIdLSB = 6 [deprecated = true];
+  bool root = 7 [deprecated = true];
+  bool debugMode = 8 [deprecated = true];
+  string configuration = 9 [deprecated = true];
+  string entity = 10;
```
smatvienko-tb, line 148: I am on a similar topic where a Proto version is needed for different versions.

For your case, the edge.proto is a speaking language between endpoints. The endpoints may have different versions, but the language (proto) should be **the same** or at least **fully compatible**.

For example, you add a mandatory field `string entity = 10;` that is not optional, so it is not compatible with the old proto version.

I assume that you have no fully integrated test of the ThingsBoard/master  + Edges with all compatible versions. May be I wrong.

As far as I see, you are trying to maintain compatibility with many older versions and the way to do this solid is to retain all the proto versions you need. When you establish the gRPC session, you can negotiate the proto version and than proceed with a compatible serializer/deserializer.

One of the ways to do this is to use the version number in the package.

For now, you are on v1 despite making major changes to the speaking language.


- Your messaging is not persistent, and it is much simple to establish the compatible proto version and process it accordingly.

On the next major version, you can cut off some compatibility and remove unsupported proto from the project



```
option java_package = "org.thingsboard.server.gen.edge.v1";
```

smatvienko-tb (commented):
I've included my comment on the proto-versioning strategy. 
I hope it will help you to make a rock-stable compatible release.

Here is some good reading on gRPC to my taste  https://learn.microsoft.com/en-us/aspnet/core/grpc/versioning?view=aspnetcore-8.0

common/edge-api/src/main/proto/edge.proto
AndriiLandiak: Thank you for your insights. We will carefully consider the most appropriate approach for handling deprecated fields, and it seems that opting for a major version cut-off might be the most suitable solution.

According to compatibility with the previous Edge, I should admit that backward compatibility works fine when introducing new fields without the 'optional' keyword, as exemplified by our usage of string entity = 10. However, it's worth noting that the Protocol Buffers documentation recommends adding new fields as either 'optional' or 'repeated'.

_`Make all fields optional or repeated. You never know how long a message type is going to last and whether someone will be forced to fill in your required field with an empty string or zero in four years when it’s no longer logically required but the proto still says it is.`_  [Proto best practices](https://protobuf.dev/programming-guides/dos-donts/)

AndriiLandiak: Thank you for your insights. We will carefully consider the most appropriate approach for handling deprecated fields, and it seems that opting for a major version cut-off might be the most suitable solution.

> For example, you add a mandatory field string entity = 10; that is not optional, so it is not compatible with the old proto version.

According to compatibility with the previous Edge, I should admit that backward compatibility works fine when introducing new fields without the 'optional' keyword, as it was by our usage of string entity = 10. However, it's worth noting that the Protocol Buffers documentation recommends adding new fields as either 'optional' or 'repeated'.

Make all fields optional or repeated. You never know how long a message type is going to last and whether someone will be forced to fill in your required field with an empty string or zero in four years when it’s no longer logically required but the proto still says it is. [Proto best practices](https://protobuf.dev/programming-guides/dos-donts/)

ashvayka: As a suggestion:

```
message EntityUpdateMsg {
  UpdateMsgType msgType = 1;
  string entityType = 2
  int64 idMSB = 3;
  int64 idLSB = 4;
  string entity = 5;
}
```

PR #10063, opened 2024-01-25
Support DTLS Connection ID with configuration

### LINKED ISSUE #10061: [Bug] LwM2M Transport Service does not support DTLS CID
**LwM2M Transport Service does not support DTLS CID**

When a LwM2M client indicates (in the dtls "Client Hello" message) it wants to use the connection_id extension, its request is ignored by the server (the "Server Hello, Server Hello Done" message does not contain the connection_id extension). As a result, a client that has requested the extension and changes ip address or port (for instance due to a longer sleep period) cannot reuse the dtls session. 

Depending on the client implementation this results in:
* a new dtls handshake before sending data (proper implementation) 
* first sending data using the invalid dtls session and only after a timeout starting a new dtls handshake (less proper implementation)
* continue to send data using the invalid dtls session (poor implementation)

For low power devices, even the proper implementation where a new dtls handshake is performed is not desired, as it is a waist of ernergy that can be avoided.

**Server Environment**
* Local docker cluster deployment, build from source
* ThingsBoard Version: 3.6.3-SNAPSHOT
* OS Name and Version: Ubuntu 22.04

**Your Device**
* nRF9160 using custom firmware
* LwM2M v1.1

**To Reproduce**
1. Configure a LwM2M profile and device appropriate for the device
2. Start a network packet dump on the server network interface
3. Let the device connect to the server
4. Notice the Client Hello request contains the "connection_id" extension
5. Notice the Server Hello Done response does not contain the "connection_id" extension
6. Notice packets send by the client from a different port as used by the handhake are not responded to

**Expected behavior**
* The server responds with the "connection_id" extension, with a connection_id set.
* Packets from the client using the given connection_id are accepted and responded to by the server

**Additional context**
The leshan demo server (which uses the same components as used by the Thingsboard LwM2M transport service) does support the dtls CID extension properly.

Rhyaldir:
## Pull Request description

This pullrequest configures the LwM2M Transport and LwM2M Transport Bootstrap Service to support the DTLS Connection ID and fixes #10061.

It adds the `transport.lwm2m.dtls.connection_id_length` configuration option, which allows DTLS CID to be turned off or on with a specific connection id length. 
As DTLS CID is not used for the connection if the client does not request it, I have chosen to turn it on by default as it gives the best user experience. The default  is a 6 byte connection id as is the Leshan demo server default and it should fit almost all use cases.

The configuration option is used to configure the scandium DtlsConnector, so it will create the ConnectionIdGenerator; the requirement for CID support.

As now, an existing dtls connection can be made from a different address or port, those parameters should not be part of the identity used to generate a key, as the lookup will fail. This has been fixed in newer versions of the leshan library but an upgrade is not easily done. Therefore this pull requests uses a new serializer for `Identity` that does not serialize the address and port for secure Identities (as is done with newer versions of the Leshan libraries).

Creating tests was not easy as there are no black box tests for LwM2M and the modified code is quite hard to test using unit tests. A constructor was added to `TbLwM2mRedisRegistrationStore` that allow the injection of the `RedisLockRegistry` used, to allow some tests. Also a dependency to `mockito-inline` was added to allow mocking final classes (and mocking statics).

As a configuration option was added, the `LWM2M server parameters` documentation should be updated.

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/resources/thingsboard.yml
+++ b/application/src/main/resources/thingsboard.yml
@@ -1044,6 +1044,8 @@ transport:
     dtls:
       # RFC7925_RETRANSMISSION_TIMEOUT_IN_MILLISECONDS = 9000
       retransmission_timeout: "${LWM2M_DTLS_RETRANSMISSION_TIMEOUT_MS:9000}"
+      # "" disables connection id support, 0 enables support but not for incoming traffic, any value greater than 0 set the connection id size in bytes
+      connection_id_length: "${LWM2M_DTLS_CONNECTION_ID_LENGTH:6}"
     server:
       # LwM2M Server ID
       id: "${LWM2M_SERVER_ID:123}"
--- a/common/transport/lwm2m/pom.xml
+++ b/common/transport/lwm2m/pom.xml
@@ -95,6 +95,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.awaitility</groupId>
             <artifactId>awaitility</artifactId>
             <scope>test</scope>
--- a/common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/bootstrap/LwM2MTransportBootstrapService.java
+++ b/common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/bootstrap/LwM2MTransportBootstrapService.java
@@ -38,6 +38,7 @@ import javax.annotation.PreDestroy;
 import java.security.cert.X509Certificate;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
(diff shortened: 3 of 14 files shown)


Rhyaldir:
As mentioned in #10061 I've added some code to prevent a `4.04 not Found` response when CID is enabled and the client changes port or IP address. I encountered this issue if the client would do a `SEND`. This response was due to the fact that `SendResource.handlePOST` did not get a registration returned from `registrationStore.getRegistrationByIdentity(sender)`.

It turns out, the (Redis) registration store creates a key with includes the address and port. For secure connections this is not required as the secure identities are unique. This has been addressed in newer versions of the Leshan libraries, but an upgrade was way to much work. Instead, I created a serializer for the identity that does not include the address and port, and use this serializer for generating the key. The serializer is based on the Identity serializer used in newer versions of the Leshan libraries.

With these changes a registration is found for the identity even if it has changed address or port as is the intention of the CID.

I have changed this pull requests description to reflect these changes.

ashvayka:
Thank you!

PR #10083, opened 2024-01-29
Performance improvements for entities saving

ViacheslavKlimov:
## Pull Request description

When using `JpaRepository.save`, it executes additional select by id in case the entity has preset id (thus for all our entities that use JpaAbstractDao).
Used `EntityManager` directly in the `JpaAbstractDao`.

Default JpaRepository implementation:
```
if (entityInformation.isNew(entity)) {
    em.persist(entity);
    return entity;
} else {
    return em.merge(entity);
}
```

Before:
image

After:
image

Also, normalized SQL-related error messages.
Before:
image

After:
image

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/controller/BaseController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/BaseController.java
@@ -18,10 +18,7 @@ package org.thingsboard.server.controller;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.util.concurrent.ListenableFuture;
 import lombok.Getter;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
-import org.postgresql.util.PSQLException;
-import org.postgresql.util.ServerErrorMessage;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
@@ -48,6 +45,7 @@ import org.thingsboard.server.common.data.HasName;
 import org.thingsboard.server.common.data.HasTenantId;
 import org.thingsboard.server.common.data.OtaPackage;
 import org.thingsboard.server.common.data.OtaPackageInfo;
+import org.thingsboard.server.common.data.StringUtils;
 import org.thingsboard.server.common.data.TbResource;
 import org.thingsboard.server.common.data.TbResourceInfo;
 import org.thingsboard.server.common.data.Tenant;
@@ -171,7 +169,7 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+import static org.thingsboard.server.common.data.StringUtils.isNotEmpty;
 import static org.thingsboard.server.common.data.query.EntityKeyType.ENTITY_FIELD;
 import static org.thingsboard.server.controller.UserController.YOU_DON_T_HAVE_PERMISSION_TO_PERFORM_THIS_OPERATION;
 import static org.thingsboard.server.dao.service.Validator.validateId;
@@ -378,15 +376,11 @@ public abstract class BaseController {
         } else if (exception instanceof AsyncRequestTimeoutException) {
             return new ThingsboardException("Request timeout", ThingsboardErrorCode.GENERAL);
         } else if (exception instanceof DataAccessException) {
-            Throwable rootCause = ExceptionUtils.getRootCause(exception);
-            if (rootCause instanceof PSQLException) {
-                String sqlError = Optional.ofNullable(((PSQLException) rootCause).getServerErrorMessage())
-                        .map(ServerErrorMessage::getMessage).orElse(null);
-                if (isNotEmpty(sqlError)) {
-                    sqlError = StringUtils.capitalize(sqlError);
-                    return new ThingsboardException(sqlError, ThingsboardErrorCode.GENERAL);
-                }
+            String errorType = exception.getClass().getSimpleName();
+            if (!logControllerErrorStackTrace) { // not to log the error twice
+                log.warn("Database error: {} - {}", errorType, ExceptionUtils.getRootCauseMessage(exception));
             }
+            return new ThingsboardException("Database error: " + errorType, ThingsboardErrorCode.GENERAL);
         }
         return new ThingsboardException(exception.getMessage(), exception, ThingsboardErrorCode.GENERAL);
     }
--- a/application/src/main/java/org/thingsboard/server/controller/BaseController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/BaseController.java
@@ -380,7 +380,7 @@ public abstract class BaseController {
             if (!logControllerErrorStackTrace) { // not to log the error twice
                 log.warn("Database error: {} - {}", errorType, ExceptionUtils.getRootCauseMessage(exception));
             }
-            return new ThingsboardException("Database error: " + errorType, ThingsboardErrorCode.GENERAL);
+            return new ThingsboardException("Database error", ThingsboardErrorCode.GENERAL);
         }
         return new ThingsboardException(exception.getMessage(), exception, ThingsboardErrorCode.GENERAL);
     }
--- a/application/src/main/java/org/thingsboard/server/controller/BaseController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/BaseController.java
@@ -16,14 +16,16 @@
 package org.thingsboard.server.controller;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.MoreExecutors;
 import lombok.Getter;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.postgresql.util.PSQLException;
+import org.postgresql.util.ServerErrorMessage;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.dao.DataAccessException;
 import org.springframework.http.MediaType;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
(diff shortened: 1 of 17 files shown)


dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
```diff
@@ -51,6 +62,33 @@ protected void setSearchText(E entity) {
...
+    @Override
+    @Transactional
+    public D saveAndFlush(TenantId tenantId, D domain) {
+        D d = save(tenantId, domain);
+        getRepository().flush();
+        return d;
+    }
+
+    @Override
+    public D create(TenantId tenantId, D domain) {
```
smatvienko-tb, line 79: looks kind of complicated. Is it not enough just `@Transactional`?

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
```diff
@@ -51,6 +62,33 @@ protected void setSearchText(E entity) {
...
+        E entity = prepare(domain);
```
smatvienko-tb, line 80: On this stage we have a marker do we need to create or save:
```
if (entity.getUuid() == null) {
```
So if we have a new entity with null id - it should create. otherwise - save by default.
The interface might remain the same. Services will ask to save and we will find out how to handle in the internal implementation depending on the new or existing entity. 
It seems good for me as we use random UUID as an ID (effectively unique)

application/src/main/java/org/thingsboard/server/controller/BaseController.java
```diff
@@ -376,9 +377,18 @@ private ThingsboardException handleException(Exception exception, boolean logExc
             return new ThingsboardException("Unable to send mail: " + exception.getMessage(), ThingsboardErrorCode.GENERAL);
         } else if (exception instanceof AsyncRequestTimeoutException) {
             return new ThingsboardException("Request timeout", ThingsboardErrorCode.GENERAL);
-        } else {
-            return new ThingsboardException(exception.getMessage(), exception, ThingsboardErrorCode.GENERAL);
+        } else if (exception instanceof DataAccessException) {
```
smatvienko-tb, line 380: :+1:

smatvienko-tb (changes requested):
Great job! Couple comments below:

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
ViacheslavKlimov: It would require autowiring `self`, because `create` is called directly from the dao and `@Transactional` won't make a difference.

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
ViacheslavKlimov: For partitioned entities we preset the id sooner, and here we have the entity already with id.

ViacheslavKlimov: But I agree it would be good to refactor this for non-partitioned entities too (in the `save` call create in case the id is null).

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
```diff
@@ -60,21 +85,30 @@ public D save(TenantId tenantId, D domain) {
...
+        boolean isNew = entity.getUuid() == null;
+        if (isNew) {
             UUID uuid = Uuids.timeBased();
             entity.setUuid(uuid);
             entity.setCreatedTime(Uuids.unixTimestamp(uuid));
         }
-        entity = getRepository().save(entity);
-        return DaoUtil.getData(entity);
+
+        if (preSaveAction != null) {
```
smatvienko-tb, line 95: You don't need another method.
just use an ordinary save

partitioned stuff is not related to eny antity, but for the whole dao (e.g. auditLogDao)
in JpaAbstractDao
this.isPartitioned() -- default false
this.createPartitionIfExists() -- default empty method

Override it in  JpaAuditLogDao

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
```diff
@@ -60,21 +85,30 @@ public D save(TenantId tenantId, D domain) {
...
-    @Transactional
-    public D saveAndFlush(TenantId tenantId, D domain) {
-        D d = save(tenantId, domain);
-        getRepository().flush();
-        return d;
+    private D doSave(E entity, boolean isNew) {
+        if (isNew) {
+            entityManager.persist(entity);
+        } else {
+            entity = entityManager.merge(entity);
```
smatvienko-tb, line 109: as i understand, this is a replacement for repository.save()

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
```diff
@@ -60,21 +85,30 @@ public D save(TenantId tenantId, D domain) {
...
             entity.setUuid(uuid);
             entity.setCreatedTime(Uuids.unixTimestamp(uuid));
         }
-        entity = getRepository().save(entity);
-        return DaoUtil.getData(entity);
+
+        if (preSaveAction != null) {
+            preSaveAction.accept(entity);
+        }
+        if (TransactionSynchronizationManager.isActualTransactionActive()) {
```
smatvienko-tb, line 98: As we are going to use save method only, the @Transactional will always present on top

smatvienko-tb (changes requested):
Great progress!
You can find my comments around.
Do we need to adjust event table save in the same way or there is batch implementation already in place?

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
ViacheslavKlimov: I prefer current approach

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
ViacheslavKlimov: Yes

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
ViacheslavKlimov: The @Transactional is present, but if someone decides to call the save method directly from the dao (like it is now for audit logs), Hibernate will throw an exception, because we cannot use EntityManager without a transaction.

smatvienko-tb: I see no problem with annotating 4 methods override
image

At this point. Some transactions should be set or supported. Otherwise - exception - it is desired result

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
smatvienko-tb: People will asking why we replace the working repo.save() with something else

dao/src/main/java/org/thingsboard/server/dao/sql/JpaAbstractDao.java
smatvienko-tb: Anyway, passing an effective constant as a parameter is not good.
I am afraid we all will be forced to use more OOP and less functions :)

application/src/main/java/org/thingsboard/server/controller/BaseController.java
```diff
@@ -170,7 +171,7 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import static org.thingsboard.server.common.data.StringUtils.isNotEmpty;
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
```
ashvayka, line 174: ?

application/src/main/java/org/thingsboard/server/controller/BaseController.java
ashvayka: We can't publish SQL errors to user's browser. Please log it with the corresponding log level but hide it from the user.

ashvayka:
This PR does not pass the "risk assessment". You can't change how we save all entities and expect us to merge this to the hotfix release. Please create a separate JpaPartitionedEntityDao and add your suggested logic.
Also, You have removed the following lines, so I think your PR does not work. 
```
        if (auditLog.getId() == null) {
            UUID uuid = Uuids.timeBased();
            auditLog.setId(new AuditLogId(uuid));
            auditLog.setCreatedTime(Uuids.unixTimestamp(uuid));
        }
```

application/src/main/java/org/thingsboard/server/controller/BaseController.java
```diff
@@ -378,15 +376,11 @@ private ThingsboardException handleException(Exception exception, boolean logExc
...
-                        .map(ServerErrorMessage::getMessage).orElse(null);
-                if (isNotEmpty(sqlError)) {
-                    sqlError = StringUtils.capitalize(sqlError);
-                    return new ThingsboardException(sqlError, ThingsboardErrorCode.GENERAL);
-                }
+            String errorType = exception.getClass().getSimpleName();
+            if (!logControllerErrorStackTrace) { // not to log the error twice
+                log.warn("Database error: {} - {}", errorType, ExceptionUtils.getRootCauseMessage(exception));
             }
+            return new ThingsboardException("Database error: " + errorType, ThingsboardErrorCode.GENERAL);
```
smatvienko-tb, line 383: Reporting an error type may also explain how an attacker query affects the DB. e.g., looking for a security flow expecting something like SQLSyntaxException

smatvienko-tb (changes requested):

dao/src/main/java/org/thingsboard/server/dao/sql/JpaPartitionedAbstractDao.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.persistence.EntityManager;
+import javax.persistence.PersistenceContext;
+
+@SqlDao
+public abstract class JpaPartitionedAbstractDao<E extends BaseEntity<D>, D> extends JpaAbstractDao<E, D> {
+
+    @PersistenceContext
+    private EntityManager entityManager;
+
+    @Override
```
smatvienko-tb, line 30: Please add a detailed Java doc to provide a new person with a shortcut to get the idea ASAP.
Something like that
```
/**
* The initial implementation is taken from repository.save(). The only difference is that id is generated by us, and we provide the flag isNew for new entities instead of null id.
**/
```

smatvienko-tb (approved):
Please add a JavaDoc, as I suggested in the comment below.

PR #10089, opened 2024-01-30
Updated CassandraBaseTimeseriesDao to do one db insert instead of 5 when setNullValuesEnabled=true

dashevchenko:
## Pull Request description

Previously, Cassandra saved value only for 1 type. If we had to save null values for the other types, we used param setNullValuesEnabled and did 5 inserts to the DB. Now, we do one insert to DB when setNullValuesEnabled=true.

## General checklist

## Back-End feature checklist

### DIFF
--- a/dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
@@ -86,6 +86,18 @@ public class CassandraBaseTimeseriesDao extends AbstractCassandraBaseTimeseriesD
     public static final String ASC_ORDER = "ASC";
     public static final long SECONDS_IN_DAY = TimeUnit.DAYS.toSeconds(1);
     protected static final List<Long> FIXED_PARTITION = List.of(0L);
+    private static String INSERT_WITH_NULL = INSERT_INTO + ModelConstants.TS_KV_CF +
+            "(" + ModelConstants.ENTITY_TYPE_COLUMN +
+            "," + ModelConstants.ENTITY_ID_COLUMN +
+            "," + ModelConstants.KEY_COLUMN +
+            "," + ModelConstants.PARTITION_COLUMN +
+            "," + ModelConstants.TS_COLUMN +
+            "," + ModelConstants.BOOLEAN_VALUE_COLUMN +
+            "," + ModelConstants.STRING_VALUE_COLUMN +
+            "," + ModelConstants.LONG_VALUE_COLUMN +
+            "," + ModelConstants.DOUBLE_VALUE_COLUMN +
+            "," + ModelConstants.JSON_VALUE_COLUMN + ")" +
+            " VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?)";
 
     private CassandraTsPartitionsCache cassandraTsPartitionsCache;
 
@@ -566,18 +578,7 @@ public class CassandraBaseTimeseriesDao extends AbstractCassandraBaseTimeseriesD
             stmtCreationLock.lock();
             try {
                 if (saveWithNullStmt == null) {
-                    saveWithNullStmt = prepare(INSERT_INTO + ModelConstants.TS_KV_CF +
-                            "(" + ModelConstants.ENTITY_TYPE_COLUMN +
-                            "," + ModelConstants.ENTITY_ID_COLUMN +
-                            "," + ModelConstants.KEY_COLUMN +
-                            "," + ModelConstants.PARTITION_COLUMN +
-                            "," + ModelConstants.TS_COLUMN +
-                            "," + ModelConstants.BOOLEAN_VALUE_COLUMN +
-                            "," + ModelConstants.STRING_VALUE_COLUMN +
-                            "," + ModelConstants.LONG_VALUE_COLUMN +
-                            "," + ModelConstants.DOUBLE_VALUE_COLUMN +
-                            "," + ModelConstants.JSON_VALUE_COLUMN + ")" +
-                            " VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?)");
+                    saveWithNullStmt = prepare(INSERT_WITH_NULL);
                 }
             } finally {
                 stmtCreationLock.unlock();
@@ -591,18 +592,7 @@ public class CassandraBaseTimeseriesDao extends AbstractCassandraBaseTimeseriesD
             stmtCreationLock.lock();
             try {
                 if (saveWithNullWithTtlStmt == null) {
-                    saveWithNullWithTtlStmt = prepare(INSERT_INTO + ModelConstants.TS_KV_CF +
-                            "(" + ModelConstants.ENTITY_TYPE_COLUMN +
-                            "," + ModelConstants.ENTITY_ID_COLUMN +
-                            "," + ModelConstants.KEY_COLUMN +
-                            "," + ModelConstants.PARTITION_COLUMN +
-                            "," + ModelConstants.TS_COLUMN +
-                            "," + ModelConstants.BOOLEAN_VALUE_COLUMN +
-                            "," + ModelConstants.STRING_VALUE_COLUMN +
-                            "," + ModelConstants.LONG_VALUE_COLUMN +
-                            "," + ModelConstants.DOUBLE_VALUE_COLUMN +
-                            "," + ModelConstants.JSON_VALUE_COLUMN + ")" +
-                            " VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?) USING TTL ?");
+                    saveWithNullWithTtlStmt = prepare(INSERT_WITH_NULL + " USING TTL ?");
                 }
             } finally {
                 stmtCreationLock.unlock();
--- a/dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
@@ -65,6 +65,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -117,6 +118,8 @@ public class CassandraBaseTimeseriesDao extends AbstractCassandraBaseTimeseriesD
     private PreparedStatement[] fetchStmtsAsc;
     private PreparedStatement[] fetchStmtsDesc;
     private PreparedStatement deleteStmt;
+    private PreparedStatement saveWithNullStmt;
+    private PreparedStatement saveWithNullWithTtlStmt;
     private final Lock stmtCreationLock = new ReentrantLock();
 
     private boolean isInstall() {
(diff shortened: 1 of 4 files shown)


smatvienko-tb (changes requested):
I'm afraid writing all 5 columns every time is too expensive by resources.
Cassandra does not store the data by rows. When we are not writing a column - it literally does not exist. Writing the `null` means that we will write the x5 data columns.

If you want to overwrite the data to another column, you can delete previous data by primary key and, next moment, write the column you need.

If you end up with 2 columns filled with data, you can compare the TTL for each column (yes. each cell has its own TTL)  in case you are using TTL. Most people use TTL.

Overwriting the TS data is quite rare. So I am concerned about this PR. I you really need to go straight this way, please provide the numbers on how it impacts the storage and read/write/compaction performance

YevhenBondarenko (approved):

dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
```diff
@@ -159,19 +174,36 @@ public ListenableFuture<Integer> save(TenantId tenantId, EntityId entityId, TsKv
...
-        addValue(tsKvEntry, stmtBuilder, 5);
-        if (ttl > 0) {
-            stmtBuilder.setInt(6, (int) ttl);
+            Boolean booleanValue = tsKvEntry.getBooleanValue().orElse(null);
+            String strValue = tsKvEntry.getStrValue().orElse(null);
+            Long longValue = tsKvEntry.getLongValue().orElse(null);
+            Double doubleValue = tsKvEntry.getDoubleValue().orElse(null);
+            String jsonValue = tsKvEntry.getJsonValue().orElse(null);
+            if (ttl == 0) {
+                stmtBuilder = new BoundStatementBuilder(prepare(INSERT_QUERY_WITH_NULL)
```
smatvienko-tb, line 190: I would suggest you to follow the pattern as before.
I see the `prepare()` here. Before it was a dedicated method that used locks and stored the statement into a volatile variable.
The difference is getting from a variable is much cheaper than searching in a hashmap with computeIfAbsent.
This is a very hot piece of code.
Please, see examples `getSaveStmt(type) getSaveTtlStmt()`

dao/src/test/java/org/thingsboard/server/dao/service/timeseries/nosql/TimeseriesServiceNoSqlSetNullEnabledTest.java
```diff
@@ -0,0 +1,63 @@
...
+@TestPropertySource(properties = {
+        "cassandra.query.set_null_values_enabled=true",
+})
+public class TimeseriesServiceNoSqlSetNullEnabledTest extends BaseTimeseriesServiceTest {
+
+    @Test
+    public void shouldSetNullValuesForNonTargetDataTypeColumn() throws ExecutionException, InterruptedException, TimeoutException {
+        long ts = 5000;
+        TsKvEntry entry = new BasicTsKvEntry(ts, new LongDataEntry("temp", 0L));
+        TsKvEntry entry2 = new BasicTsKvEntry(ts, new DoubleDataEntry("temp", 20.6));
```
smatvienko-tb, line 52: entry2 is a bad naming. Try to use words instead of numbers. Like entryDouble, entryLong.

dao/src/test/java/org/thingsboard/server/dao/service/timeseries/nosql/TimeseriesServiceNoSqlSetNullEnabledTest.java
```diff
@@ -0,0 +1,63 @@
...
+        DeviceId deviceId = new DeviceId(Uuids.timeBased());
+        tsService.save(tenantId, deviceId, entry).get(MAX_TIMEOUT, TimeUnit.SECONDS);
+        tsService.save(tenantId, deviceId, entry2).get(MAX_TIMEOUT, TimeUnit.SECONDS);
+
+        List<TsKvEntry> list = tsService.findAll(tenantId, deviceId, Collections.singletonList(new BaseReadTsKvQuery("temp", 0L,
```
smatvienko-tb, line 57: It would be nice to have a test with no aggregation and ensure that the entity is exactly ONE and equals the latest one.
It also would be nice to add another data point to have more than 1 point in time-series (ensure that ts is working as expected)

dao/src/test/java/org/thingsboard/server/dao/service/timeseries/nosql/TimeseriesServiceNoSqlSetNullEnabledTest.java
```diff
@@ -0,0 +1,63 @@
...
+                8000000L, 200000, 3, Aggregation.AVG))).get(MAX_TIMEOUT, TimeUnit.SECONDS);
```
smatvienko-tb, line 58: I assume that 8 million is a max ts. Please use something human-readable like TimeUnit.DAYS.to ...
Or get ts from ISO date string, which is also very good for human perception.

dao/src/test/java/org/thingsboard/server/dao/service/timeseries/nosql/TimeseriesServiceNoSqlTest.java
```diff
@@ -15,9 +15,44 @@
...
+    @Test
+    public void shouldUpdateOnlyTargetDataTypeColumn() throws ExecutionException, InterruptedException, TimeoutException {
+        long ts = 5000;
+        TsKvEntry entry = new BasicTsKvEntry(ts, new LongDataEntry("temp", 0L));
+        TsKvEntry entry2 = new BasicTsKvEntry(ts, new DoubleDataEntry("temp", 20.6));
+        DeviceId deviceId = new DeviceId(Uuids.timeBased());
+        tsService.save(tenantId, deviceId, entry).get(MAX_TIMEOUT, TimeUnit.SECONDS);
+        tsService.save(tenantId, deviceId, entry2).get(MAX_TIMEOUT, TimeUnit.SECONDS);
+
+        List<TsKvEntry> list = tsService.findAll(tenantId, deviceId, Collections.singletonList(new BaseReadTsKvQuery("temp", 0L,
```
smatvienko-tb, line 52: it would be nice to have an understanding what the result with no aggregation

dao/src/test/java/org/thingsboard/server/dao/service/timeseries/nosql/TimeseriesServiceNoSqlSetNullEnabledTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.junit.Assert.assertTrue;
+
+@DaoNoSqlTest
+@TestPropertySource(properties = {
+        "cassandra.query.set_null_values_enabled=true",
+})
+public class TimeseriesServiceNoSqlSetNullEnabledTest extends BaseTimeseriesServiceTest {
+
+    @Test
+    public void shouldSetNullValuesForNonTargetDataTypeColumn() throws ExecutionException, InterruptedException, TimeoutException {
```
smatvienko-tb, line 49: For Cassandra, please add a TTL test.
Options:
- You should validate that the data is saved and after TTL you should read again and the data will disappear.
- You can run a custom Cassandra query to fetch the actual TTL(double_v) along with the data. see https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/useTTL.html

dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
```diff
@@ -85,6 +86,20 @@ public class CassandraBaseTimeseriesDao extends AbstractCassandraBaseTimeseriesD
     public static final String ASC_ORDER = "ASC";
     public static final long SECONDS_IN_DAY = TimeUnit.DAYS.toSeconds(1);
     protected static final List<Long> FIXED_PARTITION = List.of(0L);
+    private static final String INSERT_QUERY_WITH_NULL = INSERT_INTO + ModelConstants.TS_KV_CF +
```
smatvienko-tb, line 89: Please make a parametrized test that will save each data type. The test should fail if you swap columns, for example, swap here string and JSON. or long with TTL.

dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
```diff
@@ -85,6 +86,20 @@ public class CassandraBaseTimeseriesDao extends AbstractCassandraBaseTimeseriesD
...
+            "," + ModelConstants.PARTITION_COLUMN +
+            "," + ModelConstants.TS_COLUMN +
+            "," + ModelConstants.BOOLEAN_VALUE_COLUMN +
+            "," + ModelConstants.STRING_VALUE_COLUMN +
+            "," + ModelConstants.LONG_VALUE_COLUMN +
+            "," + ModelConstants.DOUBLE_VALUE_COLUMN +
+            "," + ModelConstants.JSON_VALUE_COLUMN + ")" +
+            " VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?)";
+
+    private static final String INSERT_QUERY_WITH_NULL_USING_TTL = INSERT_QUERY_WITH_NULL + " USING TTL ?";
```
smatvienko-tb, line 102: Please, have a look at where other queries are located in this class. try to follow a similar pattern

smatvienko-tb (changes requested):
Good progress!
Please, take a good look at the performance comment.
Other comments are related to tests and style

smatvienko-tb (approved):
Great improvement job!
I am approving now.

Please find my comments on the latest commit:
https://github.com/dashevchenko/thingsboard/commit/53b57b1351143ed6f23422e7f983e563b3860fb1#comments

dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
```diff
@@ -591,6 +561,56 @@ private PreparedStatement getDeleteStmt() {
         return deleteStmt;
     }
 
+    private PreparedStatement getSaveWithNullStmt() {
+        if (saveWithNullStmt == null) {
+            stmtCreationLock.lock();
+            try {
+                if (saveWithNullStmt == null) {
+                    saveWithNullStmt = prepare(INSERT_INTO + ModelConstants.TS_KV_CF +
```
ashvayka, line 569: Please extract this to a separate constant (C) and use it in the next method (C+" USING TTL ?")

dao/src/main/java/org/thingsboard/server/dao/timeseries/CassandraBaseTimeseriesDao.java
```diff
@@ -159,19 +162,36 @@ public ListenableFuture<Integer> save(TenantId tenantId, EntityId entityId, TsKv
...
-        stmtBuilder.setString(0, entityId.getEntityType().name())
-                .setUuid(1, entityId.getId())
-                .setString(2, tsKvEntry.getKey())
-                .setLong(3, partition)
-                .setLong(4, tsKvEntry.getTs());
-        addValue(tsKvEntry, stmtBuilder, 5);
-        if (ttl > 0) {
-            stmtBuilder.setInt(6, (int) ttl);
+            Boolean booleanValue = tsKvEntry.getBooleanValue().orElse(null);
+            String strValue = tsKvEntry.getStrValue().orElse(null);
```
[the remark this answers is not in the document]
smatvienko-tb: This will instantiate at least two anonymous objects in the heap memory. As lambdas references outside variable `bb`.
The lambda style code is not optimal there, I believe. We should use the stack and old plain Java as simply as possible.

Issue #10169, opened 2024-02-12
TB_QUEUE_PREFIX is not applied to call consumer-groups in Kafka


akei1984:
**Describe the bug**
The Global Prefix TB_QUEUE_PREFIX is not applied correctly to rule-engine-, js-executor and tb-rule-engine-notifications consumer groups. 

This blocks applying prefix-based ACLs on Kafka and running multiple instances of things board on the same cluster the consumer interfere would interfere with each other. 

**Your Server Environment**
* own setup
  * Deployment: both
  * Deployment type: docker-compose with kafka-queue
  * ThingsBoard Version 3.6.2 PE
  * OS Name  MacOS 14.1

**To Reproduce**
1. Configure Thingsboard to use Kafka Queueing,  TB_QUEUE_PREFIX as test
2. Run Docker-compose with kafka
3. Check Kafka Consumer Groups in Kadeck or other kafka-ui
4. Prefix test is not being applied to consumer-groups for to rule-engine-, js-executor and tb-rule-engine-notifications

**Expected behavior**
prefix test should be applied to all consumer groups

akei1984:
Added a PR that fixes the issue https://github.com/thingsboard/thingsboard/pull/10170/commits/0a66ea13e26f2301e5afbbde644de9e4d83b1f6f

dashevchenko:
@akei1984 
thanks for reporting! We are going to pick up your fix for the upcoming release.

akei1984:
Thats great to hear! Looking forward to it.

dashevchenko:
Merged https://github.com/thingsboard/thingsboard/pull/10185.

AndriichnekoDm:
Hi @akei1984!
I will close this issue. 
If you have any questions or suggestions, please create new tickets. Thanks!

PR #10170, opened 2024-02-12
 TB_QUEUE_PREFIX is not applied to call consumer-groups in Kafka #10169 

akei1984:
## Pull Request description
Small fix by adding the Prefix for Kafka js-executor,rule-notifications-node- and rule-node topics. No tests are affected.


ashvayka:
We have merged this to main branch and it will be part of 3.6.3. Basically your commit. Will propagate to 3.7 automatically once we merge the branches

ashvayka:
Thank you!

PR #10202, opened 2024-02-15
Mapping of command id to unique sequence number per subscription id

ashvayka:
This PR fixes quite a rare bug in our WebSocket API. All new widgets use the new WebSocket commands based on entity data queries and use separate atomic sequence numbers to generate internal subscription commands. Our old widgets used slightly different APIs that used cmdId from the incoming web socket message to generate internal subscription commands. In a rare case, when the dashboard has many widgets of both types, the race condition occurs, and both messages generate the same internal subscription commands. eventually, one of them was ignored.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
+++ b/application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
@@ -259,13 +259,13 @@ public class DefaultWebSocketService implements WebSocketService {
     }
 
     @Override
-    public void sendUpdate(String sessionId, TelemetrySubscriptionUpdate update) {
-        sendUpdate(sessionId, update.getSubscriptionId(), update);
+    public void sendUpdate(String sessionId, int cmdId, TelemetrySubscriptionUpdate update) {
+        doSendUpdate(sessionId, cmdId, update);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
```diff
@@ -545,7 +545,7 @@ public void onSuccess(List<AttributeKvEntry> data) {
                 TbAttributeSubscription sub = TbAttributeSubscription.builder()
                         .serviceId(serviceId)
                         .sessionId(sessionId)
-                        .subscriptionId(cmd.getCmdId())
+                        .subscriptionId(sessionRef.getSessionSubIdSeq().incrementAndGet())
```
ViacheslavKlimov, line 548: What about notification subscriptions? We are planning on adding notification widget, so I guess we should already do it for notification subs as well.

ashvayka: The bug was while adding the TbSubscription to the set in TbEntityLocalSubsInfo
```
private final Set<TbSubscription<?>> subs = ConcurrentHashMap.newKeySet();
```

TbSubscription hashCode/equals uses TbSubscriptionType so we are safe in such cases.

PR #10547, opened 2024-04-11
Create alarm node: fix `AlarmModificationRequest` validation, handle null UUID alarm originator, minor code style refactoring

dskarzh:
## Pull Request description

- Changed `validateAlarmRequest(AlarmModificationRequest request)` logic: previously it was possible to pass validation and end up with `startTs` greater than `endTs` - if `startTs` was not set (0) and `endTs` was set to some positive value, `startTs` would be set to current timestamp and it could be greater than already present `endTs`. Now if both are not set then they are set to current timestamp, if only `startTs` is missing then it is set to `endTs` value and vice versa
Refactored script evaluation logs: now they are one place and script failures are logged as well
- When alarm data from incoming message is used and alarm data parsing failed, catch correct exception and report failure via futures, instead of directly throwing it to `RuleNodeActorMessageProcessor`
- Now it is not possible to create alarm with null UUID originator, when alarm data from incoming message is used. Message originator will be used in such case.
- Various minor code style changes

## General checklist

## Back-End feature checklist

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbAbstractAlarmNode.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbAbstractAlarmNode.java
@@ -45,7 +45,7 @@ public abstract class TbAbstractAlarmNode<C extends TbAbstractAlarmNodeConfigura
 
     @Override
     public void init(TbContext ctx, TbNodeConfiguration configuration) throws TbNodeException {
-        this.config = loadAlarmNodeConfig(configuration);
+        config = loadAlarmNodeConfig(configuration);
         scriptEngine = ctx.createScriptEngine(config.getScriptLang(),
                 ScriptLanguage.TBEL.equals(config.getScriptLang()) ? config.getAlarmDetailsBuildTbel() : config.getAlarmDetailsBuildJs());
     }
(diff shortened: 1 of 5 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbAbstractAlarmNode.java
```diff
@@ -73,32 +73,33 @@ public void onMsg(TbContext ctx, TbMsg msg) {
...
         try {
             TbMsg dummyMsg = msg;
             if (previousDetails != null) {
                 TbMsgMetaData metaData = msg.getMetaData().copy();
                 metaData.putValue(PREV_ALARM_DETAILS, JacksonUtil.toString(previousDetails));
                 dummyMsg = TbMsg.transformMsgMetadata(msg, metaData);
             }
-            return scriptEngine.executeJsonAsync(dummyMsg);
+            ctx.logJsEvalRequest();
+            ListenableFuture<JsonNode> alarmDetailsFuture = scriptEngine.executeJsonAsync(dummyMsg);
```
ashvayka, line 85: Consider improvement similar to the LogNode:
```
        if (standard) {
            logStandard(ctx, msg);
            return;
        }
```

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbCreateAlarmNode.java
```diff
@@ -85,49 +79,45 @@ protected ListenableFuture<TbAlarmResult> processAlarm(TbContext ctx, TbMsg msg)
...
             }
         }
 
-        Alarm existingAlarm = ctx.getAlarmService().findLatestActiveByOriginatorAndType(ctx.getTenantId(), msg.getOriginator(), alarmType);
-        if (existingAlarm == null || existingAlarm.getStatus().isCleared()) {
-            return createNewAlarm(ctx, msg, msgAlarm);
-        } else {
-            return updateAlarm(ctx, msg, existingAlarm, msgAlarm);
-        }
+        ListenableFuture<Alarm> existingAlarmFuture = ctx.getDbCallbackExecutor().submit(
```
ashvayka, line 93: Don't use DbCallbackExecutor to submit tasks. It is only for callbacks. Create a separate Async method in dao that uses jpaExecutor

dskarzh: b46cc1e

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/action/TbAbstractAlarmNode.java
dskarzh: [c5938b1](https://github.com/thingsboard/thingsboard/pull/10547/commits/c5938b13393798c7c17c33ee50b7c5b21da71fea)

PR #11140, opened 2024-07-03
Delay node: improvements

irynamatveieva:
## Pull Request description

- added tests for the delay node
- changed configuration to allow set time unit using patterns
- provided upgrade for the existing nodes
- cover new functionality with tests

PE PR: https://github.com/thingsboard/thingsboard-pe/pull/2834

## General checklist

## Back-End feature checklist


dskarzh:
- Use concurrent hash map instead of normal one, because there is not guarantee that map will always be accessed by the same thread
- Make time unit an enum
- Add correct validation for all configuration fields (period not null, time unit not null and after parsing it is one of the supported values meaning it should not be possible to use microseconds for example, max pending messages is positive and not greater than 100, etc, please consult UI to check how it validates)

dskarzh (approved):

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/delay/TbMsgDelayNodeTest.java
```diff
@@ -0,0 +1,319 @@
...
+                                    }
+                                """,
+                        true,
+                        """
+                                    {
+                                        "period": "60",
+                                        "timeUnit": "SECONDS",
+                                        "maxPendingMsgs": 1000
+                                    }
+                                """
```
ShvaykaD, line 254: why period was changed from 13 seconds to 60 seconds?

ShvaykaD (changes requested):

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/delay/TbMsgDelayNodeTest.java
ShvaykaD: > **If there is no useMetadataPeriodInSecondsPatterns then set period to "60"**

This feature didn't exist at the time the rule node was created. So potentially there might be cases where this parameter doesn't exist in the rule node configuration. Non-existing boolean parameters are treated as false during the parse of json config to the configuration class..

ShvaykaD (approved):

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/delay/TbMsgDelayNode.java
```diff
@@ -50,13 +56,17 @@
 )
 public class TbMsgDelayNode implements TbNode {
 
+    private static final Set<TimeUnit> supportedTimeUnits = EnumSet.of(TimeUnit.SECONDS, TimeUnit.MINUTES, TimeUnit.HOURS);
+    private static final String supportedTimeUnitsStr = String.join(",", TimeUnit.SECONDS.name(), TimeUnit.MINUTES.name(), TimeUnit.HOURS.name());
```
ViacheslavKlimov, line 60: Join supportedTimeUnits.
Also, add a space to "," separator

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/delay/TbMsgDelayNode.java
```diff
@@ -89,25 +99,81 @@ public void onMsg(TbContext ctx, TbMsg msg) {
...
+    private void validateConfig() throws TbNodeException {
+        if (config.getMaxPendingMsgs() < 1 || config.getMaxPendingMsgs() > 100000) {
+            throw new TbNodeException("Maximum pending messages should be in a range from 1 to 100000.", true);
+        }
+        if (config.getPeriod() == null) {
+            throw new TbNodeException("Period should be specified.", true);
+        }
+        if (config.getTimeUnit() == null) {
+            throw new TbNodeException("Time unit should be specified.", true);
+        }
```
ViacheslavKlimov, line 128: Please use javax constraints.

ViacheslavKlimov (changes requested):

ViacheslavKlimov (approved):

PR #11265, opened 2024-07-23
HTTP API: validate payload size

dashevchenko:
## Pull Request description

Added spring filter to filter large rpc requests

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
+++ b/application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
@@ -15,6 +15,7 @@
  */
 package org.thingsboard.server.config;
 
+import com.amazonaws.HttpMethod;
 import jakarta.servlet.FilterChain;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServletRequest;
@@ -22,24 +23,23 @@ import jakarta.servlet.http.HttpServletResponse;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
+import org.springframework.security.web.util.matcher.NegatedRequestMatcher;
+import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.stereotype.Component;
-import org.springframework.util.AntPathMatcher;
 import org.springframework.web.filter.OncePerRequestFilter;
 import org.thingsboard.server.common.msg.tools.TbMaxPayloadSizeExceededException;
 import org.thingsboard.server.exception.ThingsboardErrorResponseHandler;
 
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Set;
 
 @Slf4j
 @Component
 @RequiredArgsConstructor
 public class DeviceRpcRequestSizeFilter extends OncePerRequestFilter {
 
-    private final Set<String> urls = new HashSet<>(Arrays.asList("/api/v1/*/rpc/*", "/api/v1/*/rpc"));
-    private final AntPathMatcher pathMatcher = new AntPathMatcher();
+    private final RequestMatcher uriMatcher = new AntPathRequestMatcher("/api/v1/*/rpc/**", HttpMethod.POST.name());
+    private final RequestMatcher matcher = new NegatedRequestMatcher(uriMatcher);
     private final ThingsboardErrorResponseHandler errorResponseHandler;
     
     @Value("${transport.http.rpc_max_payload_size:65536}")
@@ -48,6 +48,10 @@ public class DeviceRpcRequestSizeFilter extends OncePerRequestFilter {
     @Override
     public void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain chain) throws IOException, ServletException {
         if (request.getContentLength() > rpcMaxPayloadSize) {
+            if (log.isDebugEnabled()) {
+                log.debug("Too large payload size. Url: {}, client ip: {}, content length: {}", request.getRequestURL(),
+                        request.getRemoteAddr(), request.getContentLength());
+            }
             errorResponseHandler.handle(new TbMaxPayloadSizeExceededException(), response);
             return;
         }
@@ -56,7 +60,7 @@ public class DeviceRpcRequestSizeFilter extends OncePerRequestFilter {
 
     @Override
     protected boolean shouldNotFilter(HttpServletRequest request) {
-        return urls.stream().noneMatch(url -> pathMatcher.match(url, request.getRequestURI()));
+        return matcher.matches(request);
     }
 
     @Override
--- a/application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
+++ b/application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
@@ -15,7 +15,6 @@
  */
 package org.thingsboard.server.config;
 
-import com.amazonaws.HttpMethod;
 import jakarta.servlet.FilterChain;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServletRequest;
@@ -23,23 +22,24 @@ import jakarta.servlet.http.HttpServletResponse;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
-import org.springframework.security.web.util.matcher.NegatedRequestMatcher;
-import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.stereotype.Component;
+import org.springframework.util.AntPathMatcher;
 import org.springframework.web.filter.OncePerRequestFilter;
 import org.thingsboard.server.common.msg.tools.TbMaxPayloadSizeExceededException;
 import org.thingsboard.server.exception.ThingsboardErrorResponseHandler;
 
 import java.io.IOException;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
 @Slf4j
 @Component
 @RequiredArgsConstructor
 public class DeviceRpcRequestSizeFilter extends OncePerRequestFilter {
 
-    private final RequestMatcher uriMatcher = new AntPathRequestMatcher("/api/v1/*/rpc/**", HttpMethod.POST.name());
-    private final RequestMatcher matcher = new NegatedRequestMatcher(uriMatcher);
+    private final Set<String> urls = new HashSet<>(List.of("/api/v1/*/rpc/**", "/api/plugins/rpc/**", "/api/rpc/**"));
+    private final AntPathMatcher pathMatcher = new AntPathMatcher();
     private final ThingsboardErrorResponseHandler errorResponseHandler;
     
     @Value("${transport.http.rpc_max_payload_size:65536}")
@@ -60,7 +60,12 @@ public class DeviceRpcRequestSizeFilter extends OncePerRequestFilter {
 
     @Override
     protected boolean shouldNotFilter(HttpServletRequest request) {
-        return matcher.matches(request);
+        for (String url : urls) {
+            if (pathMatcher.match(url, request.getRequestURI())) {
+                return false;
+            }
+        }
+        return true;
     }
 
     @Override
--- a/application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
+++ b/application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
@@ -25,7 +25,7 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import org.springframework.util.AntPathMatcher;
 import org.springframework.web.filter.OncePerRequestFilter;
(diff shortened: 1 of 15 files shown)


common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportService.java
```diff
@@ -824,6 +827,21 @@ public void log(TransportProtos.SessionInfoProto sessionInfo, String msg) {
         process(sessionInfo, postTelemetryMsg, TransportServiceCallback.EMPTY);
     }
 
+    private boolean checkMsgSize(TransportProtos.SessionInfoProto sessionInfo, TransportServiceCallback<?> callback, CharSequence... payload) {
```
smatvienko-tb, line 830: The algorithm complexity should be O(1). Any higher complexity will not help prevent DDoS attacks (CPU overload).
You may sacrifice the precise calculation and use an estimate. e.g. charSequence length is enough. the real size may vary from 1 to 2 bytes per char. but it might be well enough protection.

But? please, read another comments before you made any changes.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportService.java
```diff
@@ -685,7 +688,7 @@ private void processTimeout(String requestId) {
 
     @Override
     public void process(TransportProtos.SessionInfoProto sessionInfo, TransportProtos.ToServerRpcRequestMsg msg, TransportServiceCallback<Void> callback) {
-        if (checkLimits(sessionInfo, msg, callback)) {
+        if (checkLimits(sessionInfo, msg, callback) && checkMsgSize(sessionInfo, callback, msg.getMethodName(), msg.getParams())) {
```
smatvienko-tb, line 691: ### Long story short
It is too late to verify the message size at this stage. In a DDoS case, you may already be OOM at this moment.
### TL;DR
You should check RAW payload size as early as possible.
The proto you check is already prepared be some code earlier.
The main goal is to not use any memory before the max_message_size  chack passed.

Transports uses some protocols and libraries. I believe some of them already allows to set a hard limit for payload in bytes on the lowest level possible. For example , max buffer size or so. 
Imagine you are receiving a 1GiG RPC request. The channel would rather not receive the whole 1GiG and then do some checks. You simply might not have the memory or network bandwidth to handle even a few parallel requests.

If the channel gives up immediately after exceeding the payload limit - it will be the best solution possible.

If the channel does not provide an option to check the msg limit - check the payload raw size as soon as possible before any transformations.

image

common/message/src/main/java/org/thingsboard/server/common/msg/tools/MaxMessageSizeExceededException.java
```diff
@@ -0,0 +1,30 @@
...
+import lombok.Getter;
+import org.thingsboard.server.common.data.id.DeviceId;
+
+public class MaxMessageSizeExceededException extends RuntimeException {
+
+    @Getter
+    private final DeviceId deviceId;
+
+    public MaxMessageSizeExceededException(DeviceId deviceId) {
+        super(String.format("Max message size exceeded for device [%s]", deviceId));
```
smatvienko-tb, line 27: I'm not a believer that we need `String.format` here. Just check how complex the implementation is, and you might want to use `string + string + string`

application/src/test/java/org/thingsboard/server/system/BaseHttpDeviceApiTest.java
```diff
@@ -74,6 +79,26 @@ public void testGetAttributes() throws Exception {
...
+    @Test
+    public void testReplyToCommandWithLargeResponse() throws Exception {
+        String errorResponse = doPostAsync("/api/v1/" + deviceCredentials.getCredentialsId() + "/rpc/5",
+                JacksonUtil.toString(createRpcResponsePayload(101)),
+                String.class,
+                status().is5xxServerError());
+        assertThat(errorResponse).isEqualTo("Max message size exceeded for device [" + device.getId() + "]");
+
+        doPostAsync("/api/v1/" + deviceCredentials.getCredentialsId() + "/rpc/5",
+                JacksonUtil.toString(createRpcResponsePayload(10)),
```
smatvienko-tb, line 91: I would like to see the corner case. like Maximum allowed payload ok. And failed `max + 1` or any expected overflow

application/src/main/resources/thingsboard.yml
```diff
@@ -951,6 +951,7 @@ transport:
     max_wrong_credentials_per_ip: "${TB_TRANSPORT_MAX_WRONG_CREDENTIALS_PER_IP:10}"
     # Timeout (in milliseconds) to expire block IP addresses
     ip_block_timeout: "${TB_TRANSPORT_IP_BLOCK_TIMEOUT:60000}"
+  max_message_size: "${TB_TRANSPORT_MAX_MESSAGE_SIZE:500000}" # max transport message size
```
smatvienko-tb, line 954: How did you come up with a 500k limit ?

smatvienko-tb (changes requested):
Good start. But please, read my notes carefully and try to avoid any overhead

Please, also consider that some protocols already have a protection like MQTT

```
max_payload_size: "${NETTY_MAX_PAYLOAD_SIZE:65536}"
```

common/message/src/main/java/org/thingsboard/server/common/msg/tools/TbMaxPayloadSizeExceededException.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.common.msg.tools;
+
+public class TbMaxPayloadSizeExceededException extends RuntimeException {
+
+    public TbMaxPayloadSizeExceededException() {
+        super("Payload size exceeds the configured maximum");
```
smatvienko-tb, line 21: Payload size exceeds the limit

application/src/test/java/org/thingsboard/server/system/BaseHttpDeviceApiTest.java
```diff
@@ -74,6 +79,54 @@ public void testGetAttributes() throws Exception {
...
+    }
+
+    private ObjectNode createRpcRequestPayload(int size) {
+        ObjectNode rpcResponseBody = JacksonUtil.newObjectNode();
+        char[] chars = new char[size-38];
+        Arrays.fill(chars, 'a');
+        rpcResponseBody.put("method", "get");
+        ObjectNode params = JacksonUtil.newObjectNode();
+        params.put("value", new String(chars));
+        rpcResponseBody.set("params", params);
```
smatvienko-tb, line 125: kind of overengineered.
Lets compose some simple string that are easy to read and understand like
```
{"method":"get","params":{"value":"aaaaaaaaaaaaaa"}}
```
the only "aaaaaaaa" can be generated. other just a static string . much clear as for me

application/src/main/java/org/thingsboard/server/exception/ThingsboardErrorResponseHandler.java
```diff
@@ -184,6 +187,13 @@ private void handleRateLimitException(HttpServletResponse response, TbRateLimits
                         ThingsboardErrorCode.TOO_MANY_REQUESTS, HttpStatus.TOO_MANY_REQUESTS));
     }
 
+    private void handleMaxPayloadSizeExceededException(HttpServletResponse response, TbMaxPayloadSizeExceededException exception) throws IOException {
+        response.setStatus(HttpStatus.BAD_REQUEST.value());
```
smatvienko-tb, line 191: Http has a special status for large payload
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413

smatvienko-tb (changes requested):
Now I believe it  works. 
Please, workout some minor stuff in my comment:

application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
```diff
@@ -15,31 +15,31 @@
  */
 package org.thingsboard.server.config;
 
+import com.amazonaws.HttpMethod;
```
smatvienko-tb, line 18: why Amazon is here?

application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
```diff
@@ -15,31 +15,31 @@
...
+import java.util.Set;
 
 @Slf4j
 @Component
 @RequiredArgsConstructor
 public class DeviceRpcRequestSizeFilter extends OncePerRequestFilter {
 
-    private final RequestMatcher uriMatcher = new AntPathRequestMatcher("/api/v1/*/rpc/**", HttpMethod.POST.name());
-    private final RequestMatcher matcher = new NegatedRequestMatcher(uriMatcher);
+    private final Set<String> urls = new HashSet<>(List.of("/api/v1/*/rpc/**", "/api/plugins/rpc/**", "/api/rpc/**"));
```
smatvienko-tb, line 41: immutable `List.of` is enough. Or use Set.of(). But you never search/modify by URL pattern. So the set makes no sense here

smatvienko-tb (approved):
look good to me

application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
```diff
@@ -0,0 +1,78 @@
...
+import org.thingsboard.server.common.msg.tools.TbMaxPayloadSizeExceededException;
+import org.thingsboard.server.exception.ThingsboardErrorResponseHandler;
+
+import java.io.IOException;
+import java.util.List;
+
+@Slf4j
+@Component
+@RequiredArgsConstructor
+public class DeviceRpcRequestSizeFilter extends OncePerRequestFilter {
```
ViacheslavKlimov, line 37: It's in application module, so this won't work in a microservice setup.

application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private final List<String> urls = List.of("/api/v1/*/rpc/**", "/api/plugins/rpc/**", "/api/rpc/**");
```
ViacheslavKlimov, line 39: Why only these endpoints? Device can as well supply a very large telemetry, attributes, etc.

application/src/main/java/org/thingsboard/server/config/DeviceRpcRequestSizeFilter.java
ViacheslavKlimov, line 37: After you cover more endpoints unrelated to RPC, this should be renamed to `RequestSizeFilter`.

application/src/main/resources/thingsboard.yml
```diff
@@ -959,6 +959,8 @@ transport:
     request_timeout: "${HTTP_REQUEST_TIMEOUT:60000}"
     # HTTP maximum request processing timeout in milliseconds
     max_request_timeout: "${HTTP_MAX_REQUEST_TIMEOUT:300000}"
+    # RPC maximum request size
+    rpc_max_payload_size: "${HTTP_RPC_MAX_PAYLOAD_SIZE:65536}" # max rpc payload size in bytes
```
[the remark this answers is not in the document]
ViacheslavKlimov: Rename to `max_payload_size`

common/message/src/main/java/org/thingsboard/server/common/msg/tools/TbMaxPayloadSizeExceededException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.common.msg.tools;
+
+public class TbMaxPayloadSizeExceededException extends RuntimeException {
```
ViacheslavKlimov, line 18: "Tb" prefix is unnecessary.

ViacheslavKlimov (changes requested):

common/transport/http/src/main/java/org/thingsboard/server/transport/http/DeviceApiController.java
```diff
@@ -265,14 +276,22 @@ public DeferredResult<ResponseEntity> subscribeToCommands(
...
             @Parameter(description = ACCESS_TOKEN_PARAM_DESCRIPTION, required = true , schema = @Schema(defaultValue = "YOUR_DEVICE_ACCESS_TOKEN"))
             @PathVariable("deviceToken") String deviceToken,
             @Parameter(description = "RPC request id from the incoming RPC request", required = true , schema = @Schema(defaultValue = "123"))
             @PathVariable("requestId") Integer requestId,
             @io.swagger.v3.oas.annotations.parameters.RequestBody(description = "Reply to the RPC request, JSON. For example: {\"status\":\"success\"}", required = true)
-            @RequestBody String json) {
+            @RequestBody String json, HttpServletRequest httpServletRequest) {
+        if (httpServletRequest.getContentLength() > rpcMaxPayloadSize) {
+            throw new MaxPayloadSizeExceededException();
+        }
```
ViacheslavKlimov, line 294: Extract to method `checkPayloadSize` or `checkRequestSize`

Issue #11410, opened 2024-08-13
Empty response when requesting from /api/entitiesQuery/find when using Cassandra


kitaniman:
**Describe the bug**
Empty response when requesting from /api/entitiesQuery/find with a specific request body and using Cassandra even though there is new data. 

**Your Server Environment**
* Own setup
* Deployment: monolith 
* Deployment type: deb
* PostgreSQL + Cassandra
* ThingsBoard Version 3.7.0
* Community Edition
* OS: Ubuntu 22.04

**Your Client Environment**
* OS: WIndows 11
* Browser: Chrome 127.0.6533.100 (Official Build) (64-bit)

**Your Device**

* Connectivity
  * MQTT
  * HTTP

**To Reproduce**
Send a POST request to the following path '/api/entitiesQuery/find' with the following request body:
```json
{
    "entityFilter": {
        "type": "singleEntity",
        "singleEntity": {
            "entityType": "DEVICE",
            "id": "<DEVICE_ID>"
        }
    },
    "pageLink": {
        "pageSize": 1
    },
    "latestValues": [
        {
            "type": "TIME_SERIES",
            "key": "temperature"
        }
    ]
}
```
The result will be:
```json
{
    "data": [
        {
            "entityId": {
                "entityType": "DEVICE",
                "id": "<DEVICE_ID>"
            },
            "latest": {
                "TIME_SERIES": {
                    "temperature": {
                        "ts": 0,
                        "value": ""
                    }
                }
            },
            "timeseries": {},
            "aggLatest": {}
        }
    ],
    "totalPages": 1,
    "totalElements": 1,
    "hasNext": false
}
```
More specifically:
```json
"temperature": {
    "ts": 0,
    "value": ""
}
```
**Expected behavior**
```json
"temperature": {
    "ts": 1723528586,
    "value": 23
}
```
**Additional Context**
* I have tried to do the same thing with a duplicate server but that uses PostgreSQL with no Cassandra and the issue did not appear.
* I have Upgraded to 3.7.0 recently.
* This issue results in an unusual behavior with widgets as they do not get the latest data from the WebSocket the first time but they get updated once the sensor sends new data. This happens each time the page is refreshed.

**Screenshots**
![image](https://github.com/user-attachments/assets/66b423d8-d5b8-4b8f-914a-152899f38dc2)

https://github.com/user-attachments/assets/4ca622f8-db1c-46e4-b8ce-59e457bdb330

kitaniman:
[...]

PR #11418, opened 2024-08-14
Fix Misleading DATABASE_TS_LATEST_TYPE Comment and Description

kitaniman:
## Pull Request description

This PR is related to [Issue 11410](https://github.com/thingsboard/thingsboard/issues/11410).

A comment related to `DATABASE_TS_LATEST_TYPE` in `application/src/main/resources/thingsboard.yml` is misleading, the same for the description in the [Core/Rule engine deployment parameters](https://thingsboard.io/docs/user-guide/install/config/#database-telemetry-parameters) page which uses the same comment; it suggests that `DATABASE_TS_LATEST_TYP`E needs to be set to `cassandra` when using PostgreSQL + Cassandra which is not true as answered in [Issue 11410](https://github.com/thingsboard/thingsboard/issues/11410). Hence, both the comment and the website need to be updated.

### DIFF
--- a/application/src/main/resources/thingsboard.yml
+++ b/application/src/main/resources/thingsboard.yml
@@ -203,7 +203,7 @@ database:
   ts:
     type: "${DATABASE_TS_TYPE:sql}" # cassandra, sql, or timescale (for hybrid mode, DATABASE_TS_TYPE value should be cassandra, or timescale)
   ts_latest:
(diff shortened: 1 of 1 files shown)


ashvayka:
This statement is incorrect. You may use all three types. Only timescale is deprecated. SQL is the recommended one, while Cassandra scales better but does not support the filtering API.

PR #11572, opened 2024-09-04
Fix last activity not being reported when activity reporting period ends

dskarzh:
When `onReportingPeriodEnd()` throws an exception it causes periodic task to stop all all further executions. Result of this was that activities were not reported at all.

[PE PR](https://github.com/thingsboard/thingsboard-pe/pull/2801)

### DIFF
--- a/common/queue/src/main/java/org/thingsboard/server/queue/scheduler/DefaultSchedulerComponent.java
+++ b/common/queue/src/main/java/org/thingsboard/server/queue/scheduler/DefaultSchedulerComponent.java
@@ -31,13 +31,11 @@ import java.util.concurrent.TimeUnit;
 @Component
 public class DefaultSchedulerComponent implements SchedulerComponent {
 
-    private static final String SCHEDULER_NAME = "queue-scheduler";
-
     private ScheduledExecutorService schedulerExecutor;
 
     @PostConstruct
     public void init() {
-        schedulerExecutor = Executors.newSingleThreadScheduledExecutor(ThingsBoardThreadFactory.forName(SCHEDULER_NAME));
+        schedulerExecutor = Executors.newSingleThreadScheduledExecutor(ThingsBoardThreadFactory.forName("queue-scheduler"));
     }
 
     @PreDestroy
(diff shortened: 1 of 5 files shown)


common/queue/src/main/java/org/thingsboard/server/queue/scheduler/DefaultSchedulerComponent.java
```diff
@@ -52,10 +56,19 @@ public <V> ScheduledFuture<V> schedule(Callable<V> callable, long delay, TimeUni
...
     public ScheduledFuture<?> scheduleWithFixedDelay(Runnable command, long initialDelay, long delay, TimeUnit unit) {
-        return schedulerExecutor.scheduleWithFixedDelay(command, initialDelay, delay, unit);
+        return schedulerExecutor.scheduleWithFixedDelay(() -> runSafely(command), initialDelay, delay, unit);
     }
+
+    private void runSafely(Runnable command) {
+        try {
+            command.run();
+        } catch (Exception e) {
+            log.error("[{}] Unexpected error occurred while executing task! {}", SCHEDULER_NAME, Thread.currentThread(), e);
```
ViacheslavKlimov, line 70: We don't need to duplicate scheduler name and thread name in the log

common/queue/src/main/java/org/thingsboard/server/queue/scheduler/DefaultSchedulerComponent.java
```diff
@@ -52,10 +56,19 @@ public <V> ScheduledFuture<V> schedule(Callable<V> callable, long delay, TimeUni
...
 
     public ScheduledFuture<?> scheduleWithFixedDelay(Runnable command, long initialDelay, long delay, TimeUnit unit) {
-        return schedulerExecutor.scheduleWithFixedDelay(command, initialDelay, delay, unit);
+        return schedulerExecutor.scheduleWithFixedDelay(() -> runSafely(command), initialDelay, delay, unit);
     }
+
+    private void runSafely(Runnable command) {
+        try {
+            command.run();
+        } catch (Exception e) {
```
ViacheslavKlimov, line 69: Let's catch `Throwable`, just in case

ViacheslavKlimov (approved):

PR #11689, opened 2024-09-18
Safe scheduled thread pool

dskarzh:
Customized scheduled thread pool executor that aims to log all exceptions from its tasks, also continues periodic tasks executions when task throws an exception (not error or throwable)

- replaced scheduled thread pools that have 1 core thread with single thread scheduler: DefaultMailService, DefaultUpdateService, TbLwM2mRedisRegistrationStore, TbInMemoryRegistrationStore

[PE PR](https://github.com/thingsboard/thingsboard-pe/pull/2864)


common/util/src/main/java/org/thingsboard/common/util/ThingsBoardScheduledThreadPoolExecutor.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    protected void afterExecute(Runnable r, Throwable t) {
+        super.afterExecute(r, t);
+        logExceptionsAfterExecute(r, t);
+    }
+
+    private static void logExceptionsAfterExecute(Runnable r, Throwable directThrowable) {
+        Throwable wrappedThrowable = extractThrowableWrappedInFuture(r);
```
smatvienko-tb, line 42: I see no point to extract all the time. We already log an error if the wrapped runnable has exception

Here we will get a Throwable if the Wrapper runnable `SafePeriodicRunnable` will get a Cancel or shutdown command. 

No go according performance as well.

common/util/src/main/java/org/thingsboard/common/util/ThingsBoardScheduledThreadPoolExecutor.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        return super.scheduleWithFixedDelay(new SafePeriodicRunnable(command), initialDelay, delay, unit);
+    }
+
+    private record SafePeriodicRunnable(Runnable runnable) implements Runnable {
+
+        public void run() {
+            try {
+                runnable.run();
+            } catch (Exception ex) {
```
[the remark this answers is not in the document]
smatvienko-tb: how does it process a standard lifecycle? if the runnable (e.g. future) is running and got a cancel command ?

common/util/src/test/java/org/thingsboard/common/util/ThingsBoardScheduledThreadPoolExecutorTest.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    @AfterEach
+    void cleanup() {
+        scheduler.shutdownNow();
+    }
+
+    @Test
+    @DisplayName("scheduleAtFixedRate() should continue periodic execution even if command throws exception")
+    void scheduleAtFixedRateShouldNotStopPeriodicExecutionWhenCommandThrowsException() {
```
smatvienko-tb, line 47: I would like to see an example of start and stop scheduled task  to keep a normal lifecycle.
To add more context: see the DeviceProfileRuleNode that can be stopped and started many times in the rule engine due to rebalance or update. Or The rule node deletion

smatvienko-tb (changes requested):
The mass replacement stuff is well done.
But the executor class itself looks over engineered. 
And I am afraid it might not respect the lifecycle of the scheduled task (e.g. stop the task intentionally at some point, thread pool shutdown, system shutdown)

common/util/src/main/java/org/thingsboard/common/util/ThingsBoardScheduledThreadPoolExecutor.java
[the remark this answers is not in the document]
smatvienko-tb: it proofed live. but test should be better

common/util/src/test/java/org/thingsboard/common/util/ThingsBoardScheduledThreadPoolExecutorTest.java
smatvienko-tb: image

smatvienko-tb: my suggestions here https://github.com/dskarzh/thingsboard/pull/1

PR #11861, opened 2024-10-15
Rule node debug strategies

ShvaykaD:
## Pull Request description

**Customer request:**

We want to be able to go to a rule node, and see any messages data, metadata and error when a failure happens.
We can use this to get more context when we see an exception in the API usage stats.
In addition to the debug checkbox, we add a debug failures checkbox to all rule nodes.
This way we have the option to turn on debug for failures only or all messages going through the rule node.

**Solution:**

We have decided to add debug strategies instead of adding an additional toggle for failure events only. Next strategies were added in the scope of this PR:

 - On failure(24/7) - with this strategy, we will persist to debug events storage only processing failures and only outgoing (OUT) messages. In the majority of cases, the outgoing message in case of failure is the same as the incoming (IN) message, so there is no need to add an incoming message to debug events storage.
 - All messages(for 15 min) - with this strategy, we will persist all messages (IN and OUT) to the debug events storage but only for x minutes since debug mode was enabled. The persistence of the debug events will be automatically stopped after passing the debug duration since last update of the rule node. The debug duration will be controlled in the tenant profile configuration and system settings.

image

**Addition ctx:**

We need to update the default values for tenant profiles for the cloud and demo environments to 15 and 60 minutes in accordance.

## General checklist

## Back-End feature checklist


ashvayka:
After end-to-end testing, we decided to follow a bit different approach:
We will remove the debug strategy and lastUpdateTs 
We will store two new columns in DB:
1. debugFailures boolean (corresponds to the ONLY_FAILURE_EVENTS)
2. debugAllUntil long (corresponds to the end of the ALL_EVENTS).

The data object will have three fields:
```
boolean debugFailures;
boolean debugAll;
long debugAllUntil;
```

We must implement the following logic:

A) When UI sends the `debugAll` equal to `true`, we reply with ```debugAllUntil = System.currentTimeMillis() + 15 minutes``` and we reset the `debugAll` to false. So, we never store the `debugAll` in DB.
B) UI will show `X minutes left` button. When the user opens the `Debug configuration` popup, we show 
`X minutes left` and two links/buttons: [Reset]() [Stop]().
C) If user press `Reset` - we will send the `debugAll` as `true` to server again. The server will behave as described in "A".
D) If user press `Stop` - we will send the `debugAll` as `false` and server will set `debugAllUntil = System.currentTimeMillis()`;

Motivation behind this is to avoid issues when someone exports/import the rule chain, or updates fields that are not related to the debug configuration.

ViacheslavKlimov (approved):

application/src/main/data/json/tenant/rule_chains/root_rule_chain.json
```diff
@@ -18,7 +18,9 @@
         },
         "type": "org.thingsboard.rule.engine.telemetry.TbMsgTimeseriesNode",
         "name": "Save Timeseries",
-        "debugMode": false,
+        "debugFailures": false,
```
ashvayka, line 21: Why in rule_chain_template we have 

```
        "debugFailures": false,
        "debugAll": false,
```

but here we have 

```
        "debugFailures": false,
        "debugAll": false,
        "debugAllUntil": 0,
```

? Any solution to skip all 3 parameters and use default values?

ashvayka: Next time we add something, we will also modify all resources?

application/src/main/data/upgrade/3.8.1/schema_update.sql
```diff
@@ -25,6 +25,23 @@ UPDATE user_credentials c SET failed_login_attempts = (SELECT (additional_info::
...
+-- UPDATE RULE NODE DEBUG MODE TO DEBUG STRATEGY START
+
+ALTER TABLE rule_node ADD COLUMN IF NOT EXISTS debug_failures boolean DEFAULT false;
+ALTER TABLE rule_node ADD COLUMN IF NOT EXISTS debug_all_until bigint NOT NULL DEFAULT 0;
+DO
+$$
+    BEGIN
+        IF EXISTS (SELECT 1 FROM information_schema.columns WHERE table_name = 'rule_node' AND column_name = 'debug_mode')
+            THEN
+                UPDATE rule_node SET debug_all_until = (extract(epoch from now()) + 3600) * 1000 WHERE debug_mode = true;
```
ashvayka, line 37: 1 hour of debug? Why not 15 minutes? yml references DEBUG_MODE_MAX_DURATION_MINUTES which is 15 minutes.

dao/src/main/java/org/thingsboard/server/dao/entity/AbstractEntityService.java
```diff
@@ -124,4 +135,15 @@ protected void checkAssignedEntityViewsToEdge(TenantId tenantId, EntityId entity
             }
         }
     }
+
+    protected void setDebugAllUntil(TenantId tenantId, HasDebugMode entity, long now) {
+        int debugDuration = tbTenantProfileCache.get(tenantId).getDefaultProfileConfiguration().getMaxDebugModeDurationMinutes(maxDebugModeDurationMinutes);
```
ashvayka, line 140: This code does not handle a case of missing tenant.

application/src/main/resources/thingsboard.yml
```diff
@@ -494,6 +494,14 @@ actors:
     # Actors statistic persistence frequency in milliseconds
     persist_frequency: "${ACTORS_STATISTICS_PERSIST_FREQUENCY:3600000}"
 
+debug_mode:
+  # Maximum duration (in minutes) for debug mode.
+  # Min value is 1 minute.
+  # If set in the tenant profile and valid, it will be used.
+  # Otherwise, the value from this setting is used.
+  # If value from this setting is invalid, the default value (15 minutes) will be used.
+  max_duration: "${DEBUG_MODE_MAX_DURATION_MINUTES:15}"
```
ashvayka, line 503: DEBUG_MODE_DEFAULT_DURATION_MINUTES (as i wrote to you in slack). This is a default value. Not maximum value.

application/src/main/resources/thingsboard.yml
```diff
@@ -494,6 +494,14 @@ actors:
     # Actors statistic persistence frequency in milliseconds
     persist_frequency: "${ACTORS_STATISTICS_PERSIST_FREQUENCY:3600000}"
 
+debug_mode:
+  # Maximum duration (in minutes) for debug mode.
```
ashvayka, line 498: Default duration (in minutes) for debug mode.

common/data/src/main/java/org/thingsboard/server/common/data/tenant/profile/DefaultTenantProfileConfiguration.java
```diff
@@ -197,4 +204,10 @@ public TenantProfileType getType() {
     public int getMaxRuleNodeExecsPerMessage() {
         return maxRuleNodeExecutionsPerMessage;
     }
+
+    @Override
+    public int getMaxDebugModeDurationMinutes(int systemMaxDebugModeDurationMinutes) {
+        return maxDebugModeDurationMinutes > 0 ? maxDebugModeDurationMinutes : systemMaxDebugModeDurationMinutes > 0 ?
```
ashvayka, line 210: I have received the brain damage while parsing this. Please simplify with if statements. Add note that profile value has more priority then system. Rename `systemMaxDebugModeDurationMinutes` to `systemDefaultDebugModeDurationMinutes`

ashvayka: We can discuss this, but it looks like spaghetti code to me. why do we pass `systemMaxDebugModeDurationMinutes` into the Configuration object to calculate the `maxDebugModeDurationMinutes` which is a field in the configuration?

common/util/src/main/java/org/thingsboard/common/util/DebugModeUtil.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.thingsboard.common.util;
+
+import org.thingsboard.server.common.data.HasDebugMode;
+import org.thingsboard.server.common.data.msg.TbNodeConnectionType;
+
+import java.util.Set;
+
+public final class DebugModeUtil {
```
ashvayka, line 23: Make this a service that has default value from yml and access to the tenantProfileCache and move all logic here.

application/src/main/data/json/tenant/rule_chains/root_rule_chain.json
ashvayka: According to the original design, we should have a 'debug' object instead of three fields. Absence of the object may mean that the debug is disabled. 

See https://github.com/thingsboard/thingsboard/pull/11861#issuecomment-2462606341

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -1010,4 +998,21 @@ private static String getFailureMessage(Throwable th) {
...
+        persistDebugOutput(msg, Set.of(relationType));
+    }
+
+    private void persistDebugOutput(TbMsg msg, Set<String> relationTypes) {
+        persistDebugOutput(msg, relationTypes, null, null);
+    }
+
+    private void persistDebugOutput(TbMsg msg, Set<String> relationTypes, Throwable error, String failureMessage) {
+        RuleNode ruleNode = nodeCtx.getSelf();
+        if (DebugModeUtil.isDebugAllAvailable(ruleNode)) {
```
ashvayka, line 1011: why not `ruleNode.isDebugEnabled`?

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -1010,4 +998,21 @@ private static String getFailureMessage(Throwable th) {
...
+            relationTypes.forEach(relationType -> mainCtx.persistDebugOutput(getTenantId(), ruleNode.getId(), msg, relationType, error, failureMessage));
+        } else if (DebugModeUtil.isDebugFailuresAvailable(ruleNode, relationTypes)) {
```
ashvayka, line 1013: Why not `ruleNode.isDebugEnabled(relationTypes)`?

dao/src/main/java/org/thingsboard/server/dao/entity/AbstractEntityService.java
YevhenBondarenko: I think its too mach to check "missing tenant" during entity save.

application/src/main/data/upgrade/3.8.1/schema_update.sql
YevhenBondarenko: I think its better to set bigger value for upgrade, because we will spend some time for restart all services and customers dont know about this new feature, so they will have more time for debug. We can afford it.

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
YevhenBondarenko: I think this better for understanding where we should use it, 
isDebugAllAvailable - for all debug events
isDebugFailuresAvailable - only for failures
If we will rename it to isDebugEnabled, everyone will use it without relation types.

PR #11913, opened 2024-10-18
used enqueue for input node to avoid latency issue

YevhenBondarenko:
## Pull Request description

https://thingsboard-portal.atlassian.net/browse/PROD-4919
Detected latency issue if we do the infinite loop using input node.  

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
@@ -110,7 +110,6 @@ import org.thingsboard.server.dao.widget.WidgetTypeService;
 import org.thingsboard.server.dao.widget.WidgetsBundleService;
 import org.thingsboard.server.gen.transport.TransportProtos;
 import org.thingsboard.server.queue.TbQueueCallback;
-import org.thingsboard.server.queue.TbQueueMsgMetadata;
 import org.thingsboard.server.queue.common.SimpleTbQueueCallback;
 import org.thingsboard.server.service.executors.PubSubRuleNodeExecutorProvider;
 import org.thingsboard.server.service.script.RuleNodeJsScriptEngine;
@@ -177,24 +176,10 @@ class DefaultTbContext implements TbContext {
         if (!msg.isValid()) {
             return;
         }
-        TbMsg inputMsg = msg.copyWithRuleChainId(ruleChainId);
-        inputMsg.pushToStack(nodeCtx.getSelf().getRuleChainId(), nodeCtx.getSelf().getId());
-        TransportProtos.ToRuleEngineMsg toReMsg = TransportProtos.ToRuleEngineMsg.newBuilder()
-                .setTenantIdMSB(getTenantId().getId().getMostSignificantBits())
-                .setTenantIdLSB(getTenantId().getId().getLeastSignificantBits())
-                .setTbMsg(TbMsg.toByteString(inputMsg)).build();
-        TopicPartitionInfo tpi = mainCtx.resolve(ServiceType.TB_RULE_ENGINE, getQueueName(), getTenantId(), inputMsg.getOriginator());
-        mainCtx.getClusterService().pushMsgToRuleEngine(tpi, inputMsg.getId(), toReMsg, new TbQueueCallback() {
-            @Override
-            public void onSuccess(TbQueueMsgMetadata metadata) {
-                ack(msg);
-            }
-
-            @Override
-            public void onFailure(Throwable error) {
-                tellFailure(msg, error);
-            }
-        });
+        TbMsg tbMsg = msg.copyWithRuleChainId(ruleChainId);
+        tbMsg.pushToStack(nodeCtx.getSelf().getRuleChainId(), nodeCtx.getSelf().getId());
+        TopicPartitionInfo tpi = mainCtx.resolve(ServiceType.TB_RULE_ENGINE, getQueueName(), getTenantId(), tbMsg.getOriginator());
+        doEnqueue(tpi, tbMsg, new SimpleTbQueueCallback(md -> ack(msg), t -> tellFailure(msg, t)));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -172,8 +172,18 @@ public void tellSelf(TbMsg msg, long delayMs) {
...
+        }
         msg.pushToStack(nodeCtx.getSelf().getRuleChainId(), nodeCtx.getSelf().getId());
-        nodeCtx.getChainActor().tell(new RuleChainInputMsg(ruleChainId, msg));
+        TbMsg inputMsg = msg.copyWithRuleChainId(ruleChainId);
+        TransportProtos.ToRuleEngineMsg toReMsg = TransportProtos.ToRuleEngineMsg.newBuilder()
+                .setTenantIdMSB(getTenantId().getId().getMostSignificantBits())
+                .setTenantIdLSB(getTenantId().getId().getLeastSignificantBits())
+                .setTbMsg(TbMsg.toByteString(inputMsg)).build();
+        TopicPartitionInfo tpi = mainCtx.resolve(ServiceType.TB_RULE_ENGINE, getQueueName(), getTenantId(), inputMsg.getOriginator());
+        mainCtx.getClusterService().pushMsgToRuleEngine(tpi, inputMsg.getId(), toReMsg, null);
```
ashvayka, line 186: ack/fail message in callback

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -172,8 +174,27 @@ public void tellSelf(TbMsg msg, long delayMs) {
...
     @Override
     public void input(TbMsg msg, RuleChainId ruleChainId) {
-        msg.pushToStack(nodeCtx.getSelf().getRuleChainId(), nodeCtx.getSelf().getId());
-        nodeCtx.getChainActor().tell(new RuleChainInputMsg(ruleChainId, msg));
+        if (!msg.isValid()) {
+            return;
+        }
+        TbMsg inputMsg = msg.copyWithRuleChainId(ruleChainId);
+        inputMsg.pushToStack(nodeCtx.getSelf().getRuleChainId(), nodeCtx.getSelf().getId());
+        TransportProtos.ToRuleEngineMsg toReMsg = TransportProtos.ToRuleEngineMsg.newBuilder()
```
smatvienko-tb, line 182: below part are very similar to enqueue. please, try to reuse an existing enqueue methods

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -172,8 +174,27 @@ public void tellSelf(TbMsg msg, long delayMs) {
 
     @Override
     public void input(TbMsg msg, RuleChainId ruleChainId) {
-        msg.pushToStack(nodeCtx.getSelf().getRuleChainId(), nodeCtx.getSelf().getId());
-        nodeCtx.getChainActor().tell(new RuleChainInputMsg(ruleChainId, msg));
+        if (!msg.isValid()) {
+            return;
```
smatvienko-tb, line 178: as we are in the void method, the result of the method is supposed to use some callbacks. I believe if we just drop the processing without calling anyone, the rule-engine batch processing stratety will face the timeout all the time instead of fail-fast behaviour.

application/src/test/java/org/thingsboard/server/rules/flow/AbstractRuleEngineFlowIntegrationTest.java
```diff
@@ -331,6 +333,15 @@ public void testTwoRuleChainsWithTwoRules() throws Exception {
         RuleChain finalRuleChain = rootRuleChain;
         RuleNode lastRuleNode = secondaryMetaData.getNodes().stream().filter(node -> !node.getId().equals(finalRuleChain.getFirstRuleNodeId())).findFirst().get();
 
+        Awaitility.await().atMost(3, TimeUnit.SECONDS)
```
smatvienko-tb, line 336: 3 sec is prooved to rise a flacky errors. We are generally using a constant like MAX_TIMEOUT with 30 sec. it is anough for anything and never affects a good tests

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -172,8 +174,27 @@ public void tellSelf(TbMsg msg, long delayMs) {
 
     @Override
     public void input(TbMsg msg, RuleChainId ruleChainId) {
```
smatvienko-tb, line 176: ok. input has changed, but why an output has no changes?

smatvienko-tb (changes requested):
The changes proposed might not suits for all users. Lets imagine that a customer use a lot of sub-rule chains to make some algorithmic calculation from many places.

For me it is a good choice to set an option in the yaml to enqueue or go straight as before.

Does it compatible with a singleton rule-nodes? Because we are gonna to enque first on rule chain input and the first rule node can be a singletone than we will do another enqueue to particular topic. I suggest to check and avoid this if possible

Other comments below:

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
YevhenBondarenko: The main idea its break the cycle for avoid latency using enqueue, I did it for input node.
We can do it for output too, but due to our realisation after send response from output node we should log out msg in input node but we cant do it after using enqueue.
image

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
YevhenBondarenko: Yep, enqueue is similar but not suitable for this case, but ok, I will extract duplicated code.

PR #11994, opened 2024-11-04
Added ability to close transport session on RPC delivery timeout

ShvaykaD:
## Pull Request description

**Customer request:** 

In case of the high rate of RPC requests to the device or in case of a request with the huge RPC request params the device might crash with no response and didn't close the MQTT session as expected. 
As a result, the device will be still considered alive for a while due to session timeout. In addition, RPC will be registered as failed, and the only way to re-process it again will be to create a new one with the same body. 

**Solution:**

Customer asked to add a configuration parameter that will allow closing the session immediately in case of RPC delivery timeout and return RPC to the QUEUED state, so they can decide on their own if the RPC needs to be removed from the map of pending RPCs based on RPC status updates history.

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
@@ -565,6 +565,10 @@ public class ActorSystemContext {
     @Getter
     private String rpcSubmitStrategy;
 
+    @Value("${actors.rpc.close_session_on_rpc_delivery_timeout:false}")
+    @Getter
+    private boolean closeTransportSessionOnRpcDeliveryTimeout;
+
     @Value("${actors.rpc.response_timeout_ms:30000}")
     @Getter
     private long rpcResponseTimeout;
--- a/application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
+++ b/application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
@@ -132,6 +132,7 @@ public class DeviceActorMessageProcessor extends AbstractContextAwareMsgProcesso
     private final boolean rpcSequential;
     private final RpcSubmitStrategy rpcSubmitStrategy;
     private final ScheduledExecutorService scheduler;
+    private final boolean closeTransportSessionOnRpcDeliveryTimeout;
 
     private int rpcSeq = 0;
     private String deviceName;
@@ -145,6 +146,7 @@ public class DeviceActorMessageProcessor extends AbstractContextAwareMsgProcesso
         this.tenantId = tenantId;
         this.deviceId = deviceId;
         this.rpcSubmitStrategy = RpcSubmitStrategy.parse(systemContext.getRpcSubmitStrategy());
+        this.closeTransportSessionOnRpcDeliveryTimeout = systemContext.isCloseTransportSessionOnRpcDeliveryTimeout();
         this.rpcSequential = !rpcSubmitStrategy.equals(RpcSubmitStrategy.BURST);
         this.attributeSubscriptions = new HashMap<>();
         this.rpcSubscriptions = new HashMap<>();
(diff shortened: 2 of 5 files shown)


application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
```diff
@@ -705,10 +706,19 @@ private void processRpcResponseStatus(SessionInfoProto sessionInfo, ToDeviceRpcR
...
-                    status = RpcStatus.FAILED;
-                    response = JacksonUtil.newObjectNode().put("error", "There was a Timeout and all retry " +
-                            "attempts have been exhausted. Retry attempts set: " + maxRpcRetries);
+                    if (closeTransportSessionOnRpcDeliveryTimeout) {
+                        md.setRetries(0);
+                        status = RpcStatus.QUEUED;
+                        sessions.forEach(this::notifyTransportAboutClosedSessionRpcDeliveryTimeout);
+                        attributeSubscriptions.clear();
+                        rpcSubscriptions.clear();
+                        dumpSessions();
```
ViacheslavKlimov, line 715: I see that this is duplicated, consider extracting to a method and refactor other usages

application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
```diff
@@ -705,10 +706,19 @@ private void processRpcResponseStatus(SessionInfoProto sessionInfo, ToDeviceRpcR
                 maxRpcRetries = maxRpcRetries == null ?
                         systemContext.getMaxRpcRetries() : Math.min(maxRpcRetries, systemContext.getMaxRpcRetries());
                 if (maxRpcRetries <= md.getRetries()) {
-                    toDeviceRpcPendingMap.remove(requestId);
-                    status = RpcStatus.FAILED;
-                    response = JacksonUtil.newObjectNode().put("error", "There was a Timeout and all retry " +
-                            "attempts have been exhausted. Retry attempts set: " + maxRpcRetries);
+                    if (closeTransportSessionOnRpcDeliveryTimeout) {
+                        md.setRetries(0);
+                        status = RpcStatus.QUEUED;
```
ViacheslavKlimov, line 711: Why do we use QUEUED status in this case?

ShvaykaD: Customer request.

> Customer asked to add a configuration parameter that will allow closing the session immediately in case of RPC delivery timeout and return RPC to the QUEUED state, so they can decide on their own if the RPC needs to be removed from the map of pending RPCs based on RPC status updates history.

ShvaykaD: Just to clarify about RPC status updates history: 

With a default TB functionality, we can't see a history of RPC updates. However, since we are sending rule engine notifications about RPC status changes, we can track the history of RPC updates based on those events. This is basically how we do this.

ViacheslavKlimov (approved):

application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
```diff
@@ -565,6 +565,10 @@ public void printStats() {
     @Getter
     private String rpcSubmitStrategy;
 
+    @Value("${actors.rpc.close_session_on_rpc_delivery_timeout:false}")
```
ashvayka, line 568: `close_session_on_delivery_timeout` - you have `rpc` in the value name already.

application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
```diff
@@ -145,6 +146,7 @@ public class DeviceActorMessageProcessor extends AbstractContextAwareMsgProcesso
         this.tenantId = tenantId;
         this.deviceId = deviceId;
         this.rpcSubmitStrategy = RpcSubmitStrategy.parse(systemContext.getRpcSubmitStrategy());
```
ashvayka, line 148: add refactoring task to move all rpc related setting to ab object. this will simplify code and avoid beating a dead horse

PR #12191, opened 2024-12-04
Reduce API usage traffic

YevhenBondarenko:
## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -402,11 +402,11 @@ protected void handleNotification(UUID id, TbProtoQueueMsg<ToCoreNotificationMsg
         }
     }
 
-    private void processUsageStatsMsg(List<TbProtoQueueMsg<ToUsageStatsServiceMsg>> msgs, TbQueueConsumer<TbProtoQueueMsg<ToUsageStatsServiceMsg>> consumer) throws Exception {
-        ConcurrentMap<UUID, TbProtoQueueMsg<ToUsageStatsServiceMsg>> pendingMap = msgs.stream().collect(
+    private void processUsageStatsMsg(List<TbProtoQueueMsg<ToUsageStatsServiceMsgPack>> msgs, TbQueueConsumer<TbProtoQueueMsg<ToUsageStatsServiceMsgPack>> consumer) throws Exception {
```
ViacheslavKlimov, line 405: Other services are still running and submit the old proto structure.

common/queue/src/main/java/org/thingsboard/server/queue/usagestats/DefaultTbApiUsageReportClient.java
```diff
@@ -110,31 +120,50 @@ private void reportStats() {
...
 
         if (!report.isEmpty()) {
             log.debug("Reporting API usage statistics for {} tenants and customers", report.size());
         }
     }
 
+    private List<ToUsageStatsServiceMsgPack> toMsgPack(List<ToUsageStatsServiceMsg> list) {
+        return IntStream.range(0, (list.size() + packSize - 1) / packSize)
+                .mapToObj(i -> {
+                    var packList = list.subList(i * packSize, Math.min((i + 1) * packSize, list.size()));
```
ViacheslavKlimov, line 160: `Lists.partition(...)`

common/queue/src/main/java/org/thingsboard/server/queue/usagestats/DefaultTbApiUsageReportClient.java
```diff
@@ -110,31 +120,50 @@ private void reportStats() {
                 });
```
smatvienko-tb, line 121: The good point done is to replace string keys with proto.

I would like to see the resulted message with Kafka UI. To be sure that the output message has no duplicated data.
Unfortunately, I see the duplicated data . Please, visualise the output to proof yourself that the optimisation job done.

here are two hints on the picture. (the service id, tenantId if the customer stats enabled) 

image

application/src/main/resources/thingsboard.yml
```diff
@@ -188,6 +188,8 @@ usage:
       enabled_per_customer: "${USAGE_STATS_REPORT_PER_CUSTOMER_ENABLED:false}"
       # Statistics reporting interval, set to send summarized data every 10 seconds by default
       interval: "${USAGE_STATS_REPORT_INTERVAL:10}"
+      # Amount of statistic messages in pack
```
smatvienko-tb, line 191: I believe we agreed on 60 seconds by default

smatvienko-tb (changes requested):
good progress. please do further optimizations

common/queue/src/main/java/org/thingsboard/server/queue/usagestats/DefaultTbApiUsageReportClient.java
YevhenBondarenko: I agree with serviceId, but with tenantId no, because if we create some group msg per one tenant, this message will include stats from all customers per tenant so this message can be huge and will processed in one tb-core.

ViacheslavKlimov (approved):

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
ashvayka: I don't like this at all. You made an important decision and did not even logged it here. How will the system behave when it can't read the data  (in old format) with the new code? Will the administrators see a ton of exceptions in the log? 

The best option is to maintain backward compatibility. You may ignore old messages, but you can't fail to process them. 

The worst option is to rename the topic and use new topic for the data with new format. This will at least avoid tons of error logs.

PR #12327, opened 2024-12-23
fix_bug_coaps_x509_docker_many_devices

nickAS21:
## Pull Request description

[Put your PR description here instead of this sentence.   ](https://thingsboard-portal.atlassian.net/browse/CP-12625)

### Before:
During the test, Docker reused the same ports for two different devices several times:
```
- 18:14 -> Device №9<->Device №7
- 18:15 -> Device №8<->Device №2
- 18:16 -> Device №5<->Device №3
- 18:17 -> Device №4<->Device №1
- 18:17 -> Device №1<->Device №8
```
зображення

### After:
This is the result of a single test session:
11:41 - 12:44
During this time, Docker reused the same ports for two different devices 85 times.
Everything is okay.
Pull request for PE.
зображення

### Problem:
Docker containers typically generate source ports for outgoing requests automatically. This can lead to port reuse, especially when multiple devices are sending requests from the same IP address (e.g., 127.0.0.1). This behavior might cause conflicts in scenarios where distinct devices are expected to use unique source ports.

Example:
```

    Device #1 sends a request to the server with InetSocketAddress addr = /127.0.0.1:54521.

    One minute later, Device #2 sends a request using InetSocketAddress addr = /127.0.0.1:54548.

    While the response for Device #1 is being processed, Device #3 sends a request with the same InetSocketAddress addr = /127.0.0.1:54521.

```
This can cause issues, especially in multi-threaded environments where device identification depends on source port uniqueness. Ideally, each device should always use a unique port, or at least each device should be assigned specific ports that won't be reused by others.
Possible Solutions:

    Use Static Source Ports:
    Configure each device or container to use predefined static source ports for outgoing connections. This ensures each device always communicates through a dedicated port, eliminating the possibility of reuse.

    Assign Unique IP Addresses to Containers:
    Bind each container to a unique IP address (e.g., using Docker’s custom network drivers). This ensures the combination of IP and port remains unique for each device.

    Use NAT with Static Rules:
    Implement Network Address Translation (NAT) with static port forwarding rules to map each device or container to a fixed source port.

    Maintain a "Device → Port" Mapping on the Server:
    Use the server-side logic to maintain a persistent mapping between devices and their assigned source ports. The server can reject requests using incorrect or reused ports.

    Use a Load Balancer:
    Place a load balancer in front of the server to distribute and manage incoming requests. It can rewrite source ports or handle device-specific routing, ensuring port reuse doesn’t cause conflicts.

### Server-side solutions:

We have addressed this issue on the server side to ensure that such scenarios are handled properly as follows:
#### Enhanced Request Validation:

```
- The server stores information about each incoming request using additional metadata (deviceId ) rather than relying solely on the source port. This reduces the dependency on unique source ports.
- Timeout-based resource cleanup:
Implement timeout mechanisms to clean up expired or inactive device sessions, freeing up resources and reducing the risk of port conflicts.
```

By combining client and server solutions, we aim to create a robust system that prevents port conflicts while maintaining reliable communication between devices and the server.

#### On the client side
The best solution depends on the specific use case and network setup. For smaller-scale deployments or tightly controlled environments, static source ports or device-to-port mapping can be effective. For large-scale, dynamic setups, unique IP addresses or load balancers offer more scalability and flexibility.

## General checklist

## Back-End feature checklist


common/coap-server/src/main/java/org/thingsboard/server/coapserver/TbCoapDtlsDeviceAddr.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.thingsboard.server.coapserver;
+
+import lombok.Data;
+import org.thingsboard.server.common.data.id.DeviceId;
+
+import java.net.InetSocketAddress;
+
+@Data
+public class TbCoapDtlsDeviceAddr {
```
ViacheslavKlimov, line 24: Let's rename to `TbCoapDtlsSessionKey`, as it is used for dtlsSessionsMap

common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportResource.java
```diff
@@ -458,5 +469,31 @@ public void removedObserveRelation(ObserveRelation relation) {
...
+        if (StringUtils.isNotBlank(certPemStr)) {
+            for (var entry : dtlsSessionsMap.entrySet()) {
+                TbCoapDtlsDeviceAddr tbCoapDtlsDeviceAddr = entry.getKey();
+                TbCoapDtlsSessionInfo sessionInfo = entry.getValue();
+                String credentials = sessionInfo.getMsg().getCredentials();
+                if (tbCoapDtlsDeviceAddr.getInetSocketAddress().equals(peerAddress) && certPemStr.equals(credentials)) {
+                    return tbCoapDtlsDeviceAddr;
+                }
+            }
+        }
```
ViacheslavKlimov, line 484: I don't like the idea of iterating all sessions on each request.
We could leave the key in the dtlsSessionsMap as InetSocketAddress, but the value would be a list of conflicting sessions:
`ConcurrentMap<InetSocketAddress, List<TbCoapDtlsSessionInfo>> dtlsSessionsMap;`
This way, we would need to only iterate a small list, not the whole map.

common/coap-server/src/main/java/org/thingsboard/server/coapserver/TbCoapDtlsDeviceAddr.java
nickAS21: Ok

common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/CoapTransportResource.java
nickAS21: TbCoapDtlsSessionKey tbCoapDtlsSessionKey = new TbCoapDtlsSessionKey(peerAddress, certPemStr); return dtlsSessionsMap.get(key);

common/coap-server/src/main/java/org/thingsboard/server/coapserver/TbCoapDtlsSessionKey.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.coapserver;
+
+import java.net.InetSocketAddress;
+import java.util.Objects;
+
+public record TbCoapDtlsSessionKey(InetSocketAddress peerAddress, String credentials) {
```
ViacheslavKlimov, line 21: No need to override "equals", it's autogenerated for records

Issue #12639, opened 2025-02-09
Basic authentication in REST API call node is broken


devaskim:
PR https://github.com/thingsboard/thingsboard/pull/12640

This node, actually underlying HTTP client, [try to decode username/password pair](https://github.com/thingsboard/thingsboard/blob/master/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java#L347C1-L352C10) instead of encoding it while creating request.
```
    private void prepareHeaders(HttpHeaders headers, TbMsg msg) {
        config.getHeaders().forEach((k, v) -> headers.add(TbNodeUtils.processPattern(k, msg), TbNodeUtils.processPattern(v, msg)));
        ClientCredentials credentials = config.getCredentials();
        if (CredentialsType.BASIC == credentials.getType()) {
            BasicCredentials basicCredentials = (BasicCredentials) credentials;
            String authString = basicCredentials.getUsername() + ":" + basicCredentials.getPassword();
            String encodedAuthString = new String(Base64.getDecoder().decode(authString.getBytes(StandardCharsets.UTF_8)));
            headers.add("Authorization", "Basic " + encodedAuthString);
        }
    }
```

**Workaround**:
Encode creadentials manually and add `Authorization` header with `Basic encodedCredentialsString` value.

ShvaykaD:
Hi @devaskim, thank you for identifying this bug. Great catch!

Hi @nickAS21, based on the Git annotations, it seems this issue was introduced during the [replacement](https://github.com/thingsboard/thingsboard/commit/03641bc486bf9bda067c4f1f1f567bab4cb3d3bc) of the deprecated Base64Utils class. Could you please address this by applying the following fix:

```java
String encodedAuthString = Base64.getEncoder().encodeToString(authString.getBytes(StandardCharsets.UTF_8));
```

Additionally, please ensure that the updated logic is covered with unit tests to prevent similar issues in the future. Let me know if you need assistance with this. Thank you!

ViacheslavKlimov:
Thanks @devaskim, I've merged the fix

PR #12640, opened 2025-02-09
Encode username/password pair instead of decoding it.

devaskim:
Issue https://github.com/thingsboard/thingsboard/issues/12639


ViacheslavKlimov:
Thanks a lot!

PR #13054, opened 2025-03-27
MQTT client: prevent potential memory leak in retransmission mechanism

dskarzh:
## Pull Request description

This PR addresses an issue in the MQTT Node related to uncontrolled retransmission of QoS 1 and 2 messages, which under certain conditions can lead to a memory leak and eventual OutOfMemoryError (OOM).

The MQTT Node uses a retransmission mechanism to ensure message delivery for QoS 1 and 2. However, this mechanism makes it possible for the following to happen:

- The broker or network is overloaded, so ACKs are delayed.
- The broker never sends an ACK due to misbehavior or misconfiguration.
- The system keeps retransmitting the same message indefinitely while waiting for an ACK that may never come.

This leads to a scenario for OOM:

- The MQTT broker becomes overloaded, or for whatever reason processes messages slower than we send them.
- As a result, the broker’s TCP receive buffer fills up.
- Our TCP send buffer fills up next, because we keep trying to send messages (most of them retransmission attepmts).
- Netty starts queuing messages into its internal unbounded ChannelOutboundBuffer.
- That buffer keeps growing and growing.
- Eventually, the JVM runs out of memory and throws an OutOfMemoryError.

This PR introduces the following changes to mitigate the issue:

- Exponential backoff with jitter is now used between retransmission attempts.
- A maximum number of retransmission attempts is enforced (default: 3).
- Once the maximum number of attempts is reached, the message is dropped completely.

Just limiting the number of retransmissions is not sufficient on its own. If the pending message is retained in memory indefinitely while waiting for an ACK that may never arrive, it can still lead to a memory leak over time. The message is dropped at the point where the attempt after the last one would have been made.

Note: If an ACK is received after the message has already been dropped, it will be ignored.

With the default configuration:
- The first retransmission occurs after approximately 5,000 ms ± jitter.
- Each subsequent delay doubles: 10,000 ms, then 20,000 ms ± jitter.
- The system will stop retransmitting and drop the message before the fourth attempt.

These parameters are configurable in thingsboard.yml:

```yaml
mqtt:
  client:
    retransmission:
      max_attempts: "${TB_MQTT_CLIENT_RETRANSMISSION_MAX_ATTEMPTS:3}"
      initial_delay_millis: "${TB_MQTT_CLIENT_RETRANSMISSION_INITIAL_DELAY_MILLIS:5000}"
      jitter_factor: "${TB_MQTT_CLIENT_RETRANSMISSION_JITTER_FACTOR:0.15}"
````

[Documentation PR](https://github.com/thingsboard/thingsboard.github.io/pull/1950)

### CODE REVIEW

File: netty-mqtt/src/main/java/org/thingsboard/mqtt/MqttClientImpl.java [hunk 1]
```diff
@@ -384,8 +385,33 @@ public Future<Void> publish(String topic, ByteBuf payload, MqttQoS qos, boolean
         MqttFixedHeader fixedHeader = new MqttFixedHeader(MqttMessageType.PUBLISH, false, qos, retain, 0);
         MqttPublishVariableHeader variableHeader = new MqttPublishVariableHeader(topic, getNewMessageId().messageId());
         MqttPublishMessage message = new MqttPublishMessage(fixedHeader, variableHeader, payload);
-        MqttPendingPublish pendingPublish = new MqttPendingPublish(variableHeader.packetId(), future,
-                payload.retain(), message, qos, () -> !pendingPublishes.containsKey(variableHeader.packetId()));
+
+        final var pendingPublish = MqttPendingPublish.builder()
+                .messageId(variableHeader.packetId())
+                .future(future)
+                .payload(payload.retain())
```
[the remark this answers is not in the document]
dskarzh: This was checked and no memory leaks were found

application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
```diff
@@ -639,6 +639,18 @@ public void init() {
     @Getter
     private long cfCalculationResultTimeout;
 
+    @Value("${mqtt.client.retransmission.max_attempts:3}")
+    @Getter
+    private int mqttClientRetransmissionMaxAttempts;
```
ashvayka, line 644: Move this 3 parameters to a separate @Component and use them as one object here

netty-mqtt/src/main/java/org/thingsboard/mqtt/MqttChannelHandler.java
```diff
@@ -282,17 +284,16 @@ private void handleUnsuback(MqttUnsubAckMessage message) {
...
-            return;
-        }
-        pendingPublish.getFuture().setSuccess(null);
-        pendingPublish.onPubackReceived();
-        this.client.getPendingPublishes().remove(message.variableHeader().messageId());
-        pendingPublish.getPayload().release();
-        if (this.client.getCallback() != null) {
-            this.client.getCallback().onPubAck(message);
-        }
+        log.debug("{} Handling PUBACK", client.getClientConfig().getOwnerId());
```
ashvayka, line 287: CONNECT/CONACK - debug level, everything else - trace.

application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
dskarzh: I agree that it would be better to use a single autowired object here, but I suggest keeping it as is because:

- The `@Component` approach was already used in the PE for configuring the MQTT integration client.  
- It makes sense to reuse this component in two places - MQTT integration and the MQTT rule node - to avoid having two identical classes in two different places
- I did not find a suitable module in which to place this component so that it’s a `@Component` available to both integration and rule engine (except `common/queue`, but placing MQTT client settings in a queue-dedicated module doesn’t feel appropriate).

So we either duplicate essentially the same class twice, or place the component into a strange module, or leave it as is.

netty-mqtt/src/main/java/org/thingsboard/mqtt/MqttChannelHandler.java
dskarzh: I fully agree that logging PUBACKs at DEBUG level would flood the logs with unnecessary information, but I left DISCONNECT at TRACE because if you want to see CONNECT events, you’d probably want to see DISCONNECT events as well.

application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
ashvayka: Sometimes It is better to duplicate then to leave as is. CE code quality don't need to suffer because we have identical lines in PE. Please change the code as i have requested or it will not get merged

PR #13107, opened 2025-04-04
Fixed time series redis cache

irynamatveieva:
## Pull Request description

If there is no telemetry in the database (or if the version is null), we do not put a null value in the cache (we do not overwrite the previous value) and it remains unchanged, so the cache will return the old value that was deleted.

## General checklist

## Back-End feature checklist

### DIFF
--- a/dao/src/main/java/org/thingsboard/server/dao/sqlts/CachedRedisSqlTimeseriesLatestDao.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/sqlts/CachedRedisSqlTimeseriesLatestDao.java
@@ -33,18 +33,14 @@ import org.thingsboard.server.common.data.kv.BasicTsKvEntry;
 import org.thingsboard.server.common.data.kv.DeleteTsKvQuery;
 import org.thingsboard.server.common.data.kv.TsKvEntry;
 import org.thingsboard.server.common.data.kv.TsKvLatestRemovingResult;
-import org.thingsboard.server.common.data.page.PageData;
-import org.thingsboard.server.common.data.page.PageLink;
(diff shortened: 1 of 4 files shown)


smatvienko-tb (changes requested):
I believe you know the problem well and providing a good fix, but:
1. No cache related classes changed
2. No TEST that uncover the problem before the fix
3. To be on the safe side, we need to have some eviction strategy in the cache
4. blocking call  transactionTemplate.execute() in the Async method - not your fault, but it is on you since this fix
5. Manipulating with versions should be tested with EDQS events pipeline
6. Manipulating with versions probably might affect CalculatedFields

OK. that is my quick dive into problem. Please, walk through. May be no drama at all, but lets be assured before such a core feature change

ashvayka:
> I believe you know the problem well and providing a good fix, but:
> 
> 1. No cache related classes changed
> 2. No TEST that uncover the problem before the fix
> 3. To be on the safe side, we need to have some eviction strategy in the cache
> 4. blocking call  transactionTemplate.execute() in the Async method - not your fault, but it is on you since this fix
> 5. Manipulating with versions should be tested with EDQS events pipeline
> 6. Manipulating with versions probably might affect CalculatedFields
> 
> OK. that is my quick dive into problem. Please, walk through. May be no drama at all, but lets be assured before such a core feature change
> 
> As a bonus, you can store in the cache "not existed status".

I agree that we need to add tests but either don't understand or disagree with any feature requests or change of responsibility here. This was discovered during CF testing so CFs are definitely affected in positive way. 

Let's do item (2) only and review the code again.

smatvienko-tb (approved):
Since p2 (tests) added I have approved this PR

PR #13221, opened 2025-04-16
invalidate async js scripts

smatvienko-tb:
## Pull Request description

invalidate async js scripts

validating in two steps
1. regex (quick check)
2. JavaScript syntax wise (//Recheck if quick check failed. Ignoring comments and strings)

downside:

isExecEnabled and scriptBodySizeExceeded performed twice per eval. But it relatively cheap.

key messages to a user:
```
"Script must not contain 'async' keyword."
"Script must not contain 'await' keyword."
 "Script must not use 'Promise'."
 "Script must not use 'setTimeout' method."
```

## General checklist

## Back-End feature checklist

### DIFF
--- a/common/script/script-api/src/main/java/org/thingsboard/script/api/js/AbstractJsInvokeService.java
+++ b/common/script/script-api/src/main/java/org/thingsboard/script/api/js/AbstractJsInvokeService.java
@@ -96,18 +96,12 @@ public abstract class AbstractJsInvokeService extends AbstractScriptInvokeServic
     }
 
     @Override
-    public ListenableFuture<UUID> eval(TenantId tenantId, ScriptType scriptType, String scriptBody, String... argNames) {
-        if (!isExecEnabled(tenantId)) {
-            return error("Script Execution is disabled due to API limits!");
+    public String validate(TenantId tenantId, String scriptBody) {
+        String errorMessage = super.validate(tenantId, scriptBody);
+        if (errorMessage == null) {
+            return JsValidator.validate(scriptBody);
         }
-        if (scriptBodySizeExceeded(scriptBody)) {
-            return error(format("Script body exceeds maximum allowed size of %s symbols", getMaxScriptBodySize()));
-        }
-        final String validationIssue = JsValidator.validate(scriptBody);
-        if (validationIssue != null ) {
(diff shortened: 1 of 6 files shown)


common/script/script-api/src/main/java/org/thingsboard/script/api/js/AbstractJsInvokeService.java
```diff
@@ -93,6 +95,21 @@ protected void doRelease(UUID scriptId) throws Exception {
...
     }
 
+    @Override
+    public ListenableFuture<UUID> eval(TenantId tenantId, ScriptType scriptType, String scriptBody, String... argNames) {
+        if (!isExecEnabled(tenantId)) {
+            return error("Script Execution is disabled due to API limits!");
+        }
+        if (scriptBodySizeExceeded(scriptBody)) {
+            return error(format("Script body exceeds maximum allowed size of %s symbols", getMaxScriptBodySize()));
+        }
```
ViacheslavKlimov, line 105: I don't like the copy-paste. Introduce method `validate(String scriptBody)` in the `AbstractScriptInvokeService` with no implementation by default and extend it in the JS service. Then call in the `AbstractScriptInvokeService#eval`

giltwist:
Just want to put out there that a potentially legitimate use case for async js would include utilizing entityRelationService.findByQuery in, for example, an entities hierarchy widget to append the number of child nodes meeting some condition (such as enabled == true) to a parent node's text description.  I was actually looking at the possibility of incorporating the deasync package to my deployment so that I could do this.

smatvienko-tb:
> Just want to put out there that a potentially legitimate use case for async js would include utilizing entityRelationService.findByQuery in, for example, an entities hierarchy widget to append the number of child nodes meeting some condition (such as enabled == true) to a parent node's text description. I was actually looking at the possibility of incorporating the deasync package to my deployment so that I could do this.

Hi, @giltwist 

The JS or TBEL script rule nodes are designed to exec scripts in a short time window. max to batch processing timeout. scripts are also should be non-blocking to not affect other script nodes. So any wait or speculative delays are the anti-pattern. 

If you  really need some async logic, consider to develop a custom rule node. See https://github.com/thingsboard/rule-node-examples

PR #13312, opened 2025-05-06
Fix/get attribute requests

imbeacon:
## Pull Request description

This PR contains the following changes for attribute requests logic:

1. Fix for response format on getting all attributes from any device scope when there is only one attribute.

  |Old response|New response|
  |-|-|
  |{"id":1,"value":20}|{"id":1,"shared":{"sharedAttr":20}}|

2. Fix for returning all attributes and their values by scope (earlier you was unable to request only client attributes, you receive shared and client attributes simultaneously)

3. Fix for inability to retrieve all attributes for regular clients - when sharedKeys and/or clientKeys present in the request message, but empty.

4. Added additional topics for gateway API - `v2/gateway/attributes/request` and `v2/gateway/attributes/response` to have ability to get responses for attributes in the same format as it is in regular device session, it allows easily to determine scope of attribute, useful in case of requesting all attributes in scope:

Requests example:

|Regular device attribute request|V1 gateway attribute request|V2 gateway attribute request|
|-|-|-|
|{"clientKeys":"attribute1,attribute2", "sharedKeys":"shared1,shared2"}|{"keys": ["shared1","shared2"], "device": "DeviceName", "client": False, "id": 2}|{"sharedKeys": ["shared1","shared2"], "clientKeys":["attribute1","attribute2"], "device": "DeviceName", "id": 1}|

Responses example:

|Regular device attribute response|V1 gateway attribute response|V2 gateway attribute response|
|-|-|-|
|{"client":{"attribute1":"value1","attribute2":true},"shared":{"shared1":"value1","shared2":"value2"}}|{"id": 1, "device": "DeviceName", "values": {"shared1": "value1", "shared2": "value2"}}|{"shared": {"shared1":"value1","shared2":"value2"}, "client":{"attribute1":"value1","attribute2":"value2"}, "device": "DeviceName", "id": 1}|

Most useful case - requesting all client and shared attributes, for example for device state synchronisation on device connection:

||**Request**|**Response**|**Comment**|
|-|-|-|-|
|**V1 format**|{"keys": [], "device": "DeviceName", "client": True, "id": 2}|{"id": 1, "device": "DeviceName", "values": {"attribute1":"value1","attribute2":"value2","shared1": "value1", "shared2": "value2"}}| *It is impossible to figure out scope of received attributes*|
|**V2 format**|{"sharedKeys": [], "clientKeys":[], "device": "DeviceName", "id": 1}|{"shared": {"shared1":"value1","shared2":"value2"}, "client":{"attribute1":"value1","attribute2":"value2"}, "device": "DeviceName", "id": 1}| *It is easy to determine the scope of retrieved attributes*|

## General checklist

## Back-End feature checklist


ashvayka:
@imbeacon please confirm that no API was changed, only new API calls were added ( I am concerned about your fix you mention in (1). We can't change behavior of existing APIs).

Also, please add the corresponding PR for documentation before we can proceed to review. Doc PR should target develop/4.1 branch

imbeacon:
I should say that there is 1 change present - I returned ability to request all attributes for devices using MQTT transport. These ability was in ThingsBoard MQTT till 2021, but was missed during refactoring and adding protobuf payload. At the same time it was present in gateway SDK, so I think it was missed, and not removed.

I have tried to run black-box tests  from release 4.0 with new code and they all passed. According to the point 1, that you mentioned - it is impossible to request all attributes using regular device, but when the gateway requests all attributes, for any scope - if device has only one attribute present response was in wrong format and it was impossible to determine key of the attribute, you can see this in the following screenshot (It is screenshot of newly added test with TB 4.0):
Format of the message is like it is in description of PR.
image (7)

The issue of this point - was in determination of type of request, is it multiply request or request for a single attribute. Code analysed the response from core, not the initial request. Such as MQTT devices was unable to retrieve all attribute, sending empty list of attributes - they were unaffected and this bug was not found earlier.


Additionaly you can find results of new tests with old TB

From the following test we can see that for CoAP transport - ability to get all attributes. And this transport doesn't have this issue.
image (8)
In the following screenshot - we can see that we cannot request all shared or all client attributes in MQTT transport.
image (9)


I will add documentation and update the description of PR, should I add label WIP while I will do this?

imbeacon:
I have updated the PR. I decided to move v2 topic for the gateway into separated PR, this contains only fix for retrieving one attribute format and new API endpoints for HTTP and CoAP.

common/proto/src/main/proto/queue.proto
```diff
@@ -414,6 +414,8 @@ message GetAttributeRequestMsg {
   repeated string clientAttributeNames = 2;
   repeated string sharedAttributeNames = 3;
   bool onlyShared = 4;
+  bool addClient = 5;
+  bool addShared = 6;
```
ViacheslavKlimov: Add `optional` for backward compatibility with yet unprocessed requests. Don't forget to add corresponding "has" checks

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/adaptors/JsonMqttAdaptor.java
```diff
@@ -185,9 +185,15 @@ private TransportProtos.GetAttributeRequestMsg processGetAttributeRequestMsg(Mqt
...
                 result.addAllClientAttributeNames(clientKeys);
+                result.setAddClient(true);
             }
             if (sharedKeys != null) {
                 result.addAllSharedAttributeNames(sharedKeys);
+                result.setAddShared(true);
+            }
+            if (clientKeys == null && sharedKeys == null) {
+                result.setAddClient(true);
+                result.setAddShared(true);
```
ViacheslavKlimov: Explicitly set to false since these fields are optional. This is not only about MQTT

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/adaptors/JsonMqttAdaptor.java
```diff
@@ -178,16 +178,24 @@ private TransportProtos.GetAttributeRequestMsg processGetAttributeRequestMsg(Mqt
...
             if (clientKeys != null) {
                 result.addAllClientAttributeNames(clientKeys);
+                result.setAddClient(true);
             }
             if (sharedKeys != null) {
                 result.addAllSharedAttributeNames(sharedKeys);
+                result.setAddShared(true);
+            }
+            if (clientKeys == null && sharedKeys == null) {
+                result.setAddClient(true);
```
ViacheslavKlimov: Just extract addClient and addShared to variables, false by default. Then set them at the end into result

common/transport/http/src/main/java/org/thingsboard/server/transport/http/DeviceApiController.java
```diff
@@ -620,6 +665,39 @@ public void onDeviceDeleted(DeviceId deviceId) {
 
     }
 
+    private DeferredResult<ResponseEntity> handleAttributeRequest(String deviceToken, AttributeScope scope, String keys) {
+        DeferredResult<ResponseEntity> responseWriter = new DeferredResult<>();
+        List<String> keySet = !StringUtils.isEmpty(keys) ? Arrays.asList(keys.split(",")) : List.of();
+        GetAttributeRequestMsg.Builder builder = GetAttributeRequestMsg.newBuilder()
+                .setRequestId(0);
+        switch (scope) {
+            case CLIENT_SCOPE:
```
ViacheslavKlimov, line 674: This switch only sets the *requested* scope's flag to `true` but never explicitly sets the other scope to `false`. Since `addClient`/`addShared` are `optional bool` in proto, when they're never set on the builder, `hasAddClient()`/`hasAddShared()` returns `false`. Then in `DeviceActorMessageProcessor.getAttributesKvEntries`, the fallback logic:

```java
addShared = request.hasAddShared() ? request.getAddShared() : true;
```

defaults the unset field to `true`. This means a request to `/attributes/client` will also fetch ALL shared attributes (and vice versa), defeating the purpose of these scoped endpoints.

Every other transport adaptor in this PR initializes both flags explicitly at the top:
```java
result.setAddClient(false);
result.setAddShared(false);
```

This method should do the same — add `builder.setAddClient(false); builder.setAddShared(false);` before the switch, then selectively set the correct one to `true` inside the case.

application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
```diff
@@ -514,74 +513,70 @@ private void reportSessionClose() {
...
-                            .setIsMultipleAttributesRequest(
-                                    request.getSharedAttributeNamesCount() + request.getClientAttributeNamesCount() > 1)
-                            .build();
-                    sendToTransport(responseMsg, sessionInfo);
-                }
+        log.trace("[{}][{}] Processing get attributes request: {}", deviceId, requestId, request);
+        Futures.addCallback(getAttributesKvEntries(request), new FutureCallback<>() {
+            @Override
+            public void onSuccess(@Nullable List<List<AttributeKvEntry>> result) {
+                boolean isMultipleAttributesRequest = ((request.getSharedAttributeNamesCount() + request.getClientAttributeNamesCount()) != 1) ||
```
ViacheslavKlimov, line 522: The mix of `&&` and `||` without explicit parentheses makes this hard to reason about. Java's `&&` does bind tighter than `||`, so the current logic is technically correct, but a reader has to pause and think about precedence. Consider adding parentheses for clarity:

```java
boolean isMultipleAttributesRequest = ((sharedCount + clientCount) != 1)
        || (sharedCount == 0 && request.getAddShared())
        || (clientCount == 0 && request.getAddClient());
```

application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
```diff
@@ -514,74 +513,70 @@ private void reportSessionClose() {
...
+        boolean addClient;
+        boolean addShared;
+
+        if (onlyShared) {
+            addClient = false;
+            addShared = true;
         } else {
-            sharedAttributesFuture = Futures.immediateFuture(Collections.emptyList());
-            clientAttributesFuture = findAttributesByScope(toSet(request.getClientAttributeNamesList()), AttributeScope.CLIENT_SCOPE);
+            addClient = request.hasAddClient() ? request.getAddClient() : true;
```
ViacheslavKlimov, line 560: This backward-compat fallback (default to `true` when the field isn't present) is the right approach for old transports that don't set the new flags. However, it changes behavior for existing request patterns during rolling upgrades.

Consider: an old MQTT transport sends a request with only `sharedAttributeNames=["temp"]` and no `addClient`/`addShared` set. Old behavior: client scope returns empty list. New behavior: `addClient` defaults to `true`, client names are empty → `findAllAttributesByScope(CLIENT_SCOPE)` fetches ALL client attributes.

A safer fallback might be to default to `true` only when *both* key lists are empty (the "fetch everything" case), and default to `false` when the other scope's key list is empty:

```java
addClient = request.hasAddClient() ? request.getAddClient()
        : request.getClientAttributeNamesList().isEmpty() && request.getSharedAttributeNamesList().isEmpty();
```

Or simply default to `!names.isEmpty() || bothEmpty` to match the old four-branch logic exactly. This would avoid returning unexpected extra data during mixed-version deployments.

msa/black-box-tests/src/test/java/org/thingsboard/server/msa/connectivity/CoapClientTest.java
```diff
@@ -17,35 +17,113 @@
...
+            coapClient.shutdown();
+        }
         disconnect();
     }
 
+    private void initializeScopedCoapClient(String token, AttributeScope scope, String keys) {
+        if (scope == null) {
+            throw new IllegalArgumentException("Scope must not be null for scoped requests.");
+        }
+        String uri = constructUri(token, "/attributes" + getScopePath(scope), keys, null);
```
ViacheslavKlimov, line 75: The `keys` parameter is passed as `clientKeys` to `constructUri`, which produces a query string like `?clientKeys=booleanKey,stringKey`. But the server-side `CoapAdaptorUtils.processScopedRequest` calls `addKeysFromQuery(queryParameters, ...)` which looks for a query parameter named `keys`, not `clientKeys`.

So the server never sees the requested keys, returns all attributes for the scope, and the test passes coincidentally because the requested keys are a subset of all attributes.

To fix, `initializeScopedCoapClient` should build the URI with `?keys=...` instead. One approach: add a separate `constructScopedUri` method, or add a parameter to `constructUri` that controls the key parameter name.

PR #13424, opened 2025-05-23
Docker file improvements

ashvayka:
Docker file improvements to support installation **without mounting the configuration folder**. The idea is to simplify the installation instruction and avoid using the 'tb-posgres':

Contains correct built-in logback.xml to avoid spamming to file system.

Example of usage:

```
services:
  postgres:
    restart: always
    image: "postgres:16"
    ports:
      - "5432"
    environment:
      POSTGRES_DB: thingsboard
      POSTGRES_PASSWORD: postgres
    volumes:
      - postgres-data:/var/lib/postgresql/data
  thingsboard-ce:
    restart: always
    image: "thingsboard/tb-node:4.0.1"
    ports:
      - "8080:8080"
      - "7070"
      - "1883"
      - "5683-5688:5683-5688/udp"
    logging:
      driver: "json-file"
      options:
        max-size: "100m"
        max-file: "10"
    environment:
      TB_SERVICE_ID: tb-ce-node
      EDGES_ENABLED: "true"
      SPRING_DATASOURCE_URL: jdbc:postgresql://postgres:5432/thingsboard
    depends_on:
      - postgres

volumes:
  postgres-data:
    name: tb-ce-postgres-data
```

To install the DB, first launch the command:

```
docker compose run --rm -e INSTALL_TB=true -e LOAD_DEMO=true thingsboard-ce
```

Then, simply launch the app:

```
docker compose up -d && docker compose logs -f
```

### DIFF
--- a/msa/tb-node/docker/Dockerfile
+++ b/msa/tb-node/docker/Dockerfile
@@ -16,12 +16,14 @@
 
 FROM thingsboard/openjdk17:bookworm-slim
 
-COPY start-tb-node.sh ${pkg.name}.deb /tmp/
+COPY logback.xml start-tb-node.sh ${pkg.name}.deb /tmp/
 
 RUN chmod a+x /tmp/*.sh \
     && mv /tmp/start-tb-node.sh /usr/bin && \
     (yes | dpkg -i /tmp/${pkg.name}.deb) && \
     rm /tmp/${pkg.name}.deb && \
+    mv /tmp/logback.xml ${pkg.installFolder}/conf && \
+    chown -R ${pkg.user}:${pkg.user} ${pkg.installFolder}/conf/logback.xml && \
     (systemctl --no-reload disable --now ${pkg.name}.service > /dev/null 2>&1 || :) && \
     chown -R ${pkg.user}:${pkg.user} /tmp && \
     chmod 555 ${pkg.installFolder}/bin/${pkg.name}.jar
--- a/msa/tb-node/docker/logback.xml
+++ b/msa/tb-node/docker/logback.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+
+    Copyright © 2016-2025 The Thingsboard Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<!DOCTYPE configuration>
+<configuration scan="true" scanPeriod="10 seconds">
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{ISO8601} [%thread] %-5level %logger{36} - %msg%n</pattern>
+        </encoder>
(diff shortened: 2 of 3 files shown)


msa/tb-node/docker/start-tb-node.sh
```diff
@@ -15,14 +15,21 @@
...
-source "${CONF_FOLDER}/${configfile}"
+CONF_FOLDER="/config"
+if [ -d "${CONF_FOLDER}" ]; then
+  LOGGING_CONFIG="${CONF_FOLDER}/logback.xml"
+  source "${CONF_FOLDER}/${configfile}"
+  export LOADER_PATH=${CONF_FOLDER},${LOADER_PATH}
+else
+  CONF_FOLDER="/usr/share/${pkg.name}/conf"
+  LOGGING_CONFIG="/usr/share/${pkg.name}/conf/logback.xml"
+  source "${CONF_FOLDER}/${configfile}"
```
ViacheslavKlimov, line 30: Is it ok we don't add conf folder to LOADER_PATH?

msa/tb-node/docker/logback.xml
```diff
@@ -0,0 +1,38 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<!DOCTYPE configuration>
+<configuration scan="true" scanPeriod="10 seconds">
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
```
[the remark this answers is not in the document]
ashvayka: STDOUT is a correct way for docker. that is how it is available in "docker compose logs -f"

msa/tb-node/docker/start-tb-node.sh
ashvayka: It is already added. See https://github.com/thingsboard/thingsboard/blob/master/application/src/main/conf/thingsboard.conf#L23

msa/tb-node/docker/Dockerfile
```diff
@@ -16,12 +16,14 @@
...
+COPY logback.xml start-tb-node.sh ${pkg.name}.deb /tmp/
 
 RUN chmod a+x /tmp/*.sh \
     && mv /tmp/start-tb-node.sh /usr/bin && \
     (yes | dpkg -i /tmp/${pkg.name}.deb) && \
     rm /tmp/${pkg.name}.deb && \
+    mv /tmp/logback.xml ${pkg.installFolder}/conf && \
+    chown -R ${pkg.user}:${pkg.user} ${pkg.installFolder}/conf/logback.xml && \
     (systemctl --no-reload disable --now ${pkg.name}.service > /dev/null 2>&1 || :) && \
     chown -R ${pkg.user}:${pkg.user} /tmp && \
```
smatvienko-tb, line 28: move to first step
```
chown -R ${pkg.user}:${pkg.user} /tmp
```

msa/tb-node/docker/logback.xml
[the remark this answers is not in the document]
smatvienko-tb: STDOUT is the right Choice for docker.

smatvienko-tb (approved):
it looks good to me

msa/tb-node/docker/Dockerfile
ashvayka: Actually I was going to remove this chown line because I don't understand why do we need it. Do you have any idea?

smatvienko-tb: > Actually I was going to remove this chown line because I don't understand why do we need it. Do you have any idea?

it is obsolete I guess. the approach is to copy everything to tmp. set the owner for everything and then move to the right place

I believe the /tmp folder should be writeable for anyone by default from in the source image

PR #13541, opened 2025-06-06
Update cached device activity status only after a successful database save

dskarzh:
## Pull Request Description

In the event of a transient database outage, the new value for the `active` key may fail to persist while the in-memory cache is still updated, leading to an inconsistency where the UI shows a device as inactive (or vice versa). For example:

* A device is initially inactive.
* The device sends a message that should mark it active.
* The service processes the message and attempts to update `active` to `true` in both the database and the cache.
* A brief database outage causes the database update to fail, but the cache is still updated `active = true`.
* The device continues sending data, and the UI shows it as inactive.
* Subsequent activity messages update the last activity timestamp in both the database and cache, but since the service believes the device is already active, it skips re-saving the `active` flag.
* As a result, the database remains `active = false`, and the UI is showing the device as inactive, when really it *is* active.

**Fix**

* Only update the in-memory cache after a successful database save.
* If the database save fails, next activity message or a periodic check will update the value of `active` in DB to reflect true state of the device.

**Changes**

* Return a `ListenableFuture<List<Long>>` instead of `void` in `InternalTelemetryService::saveAttributesInternal`.
* Use callbacks to update the in-memory cache in `DefaultDeviceStateService`
* Refactored some existing tests for `DefaultDeviceStateService`.
* Added new tests covering the scenario described above.
* Added the missing `states.initFetchPackSize` property to `thingsboard.yml`.

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/data/json/demo/dashboards/rule_engine_statistics.json
+++ b/application/src/main/data/json/demo/dashboards/rule_engine_statistics.json
@@ -564,6 +564,7 @@
             },
             "tooltipDateColor": "rgba(0, 0, 0, 0.76)",
             "tooltipDateInterval": true,
+            "tooltipHideZeroFalse": true,
             "tooltipBackgroundColor": "rgba(255, 255, 255, 0.76)",
             "tooltipBackgroundBlur": 4,
             "animation": {
@@ -977,6 +978,7 @@
             },
             "tooltipDateColor": "rgba(0, 0, 0, 0.76)",
             "tooltipDateInterval": true,
+            "tooltipHideZeroFalse": true,
             "tooltipBackgroundColor": "rgba(255, 255, 255, 0.76)",
             "tooltipBackgroundBlur": 4,
             "animation": {
--- a/application/src/main/java/org/thingsboard/server/controller/BaseController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/BaseController.java
@@ -199,6 +199,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.function.BiConsumer;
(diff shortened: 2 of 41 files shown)


application/src/main/resources/thingsboard.yml
```diff
@@ -898,6 +898,8 @@ state:
   # Used only when state.persistToTelemetry is set to 'true' and Cassandra is used for timeseries data.
   # 0 means time-to-live mechanism is disabled.
   telemetryTtl: "${STATE_TELEMETRY_TTL:0}"
+  # Number of device records to fetch per batch when initializing device activity states
+  initFetchPackSize: "${TB_DEVICE_STATE_INIT_FETCH_PACK_SIZE:50000}"
```
ViacheslavKlimov, line 902: 👍

application/src/main/java/org/thingsboard/server/service/telemetry/InternalTelemetryService.java
```diff
@@ -23,14 +23,16 @@
...
+
 /**
  * Created by ashvayka on 27.03.18.
  */
 public interface InternalTelemetryService extends RuleEngineTelemetryService {
 
     ListenableFuture<TimeseriesSaveResult> saveTimeseriesInternal(TimeseriesSaveRequest request);
 
-    void saveAttributesInternal(AttributesSaveRequest request);
+    ListenableFuture<List<Long>> saveAttributesInternal(AttributesSaveRequest request);
```
ViacheslavKlimov: It would be good to wrap the result into AttributesSaveResult, similar to TimeseriesSaveResult. I understand this is slightly out of scope of this PR, so not required

dskarzh: This is a nice improvement, added it to this PR

PR #13630, opened 2025-06-24
Added missing fields for RPC message in cluster mode

ShvaykaD:
## Pull Request description

In cluster mode, the internal RPC notification message sent between core nodes was missing two required fields: 'additionalInfo' and 'retries'.

The 'additionalInfo' field is necessary for post-processing logic during the handling of RPC status messages (e.g., RPC_QUEUED, RPC_SENT, RPC_SUCCESSFUL, RPC_EXPIRED), and its absence resulted in null values in those messages, breaking functionality in use cases that depend on this data.

The 'retries' field is used for internal retry logic. Its omission led to a default value of 0 being applied during inter-node communication, which contradicted user expectations regarding the configured number of retry attempts for the RPC.

This fix ensures that both 'additionalInfo' and 'retries' are correctly populated when routing RPC requests between core nodes in clustered deployments.

## Explanation of the fix:

In cluster mode, the `retries` field was missing from the internode RPC notification message. As a result, it was explicitly set to 0 on the receiving node.

This caused the following logic to always treat it as 0:

```java
Integer maxRpcRetries = toDeviceRpcRequest.getRetries();
maxRpcRetries = maxRpcRetries == null ?
        systemContext.getMaxRpcRetries() : Math.min(maxRpcRetries, systemContext.getMaxRpcRetries());
```

Because retries was never null, the system default was never applied — leading to no retries being performed, even if they were expected in the original request.

In cluster mode, the `additionalInfo` field was defaulted to an empty string. However, during RPC persistence, the following code converts empty strings to null:

```java
rpc.setAdditionalInfo(JacksonUtil.toJsonNode(request.getAdditionalInfo()));
```

```java
public static JsonNode toJsonNode(String value, ObjectMapper mapper) {
    if (value == null || value.isEmpty()) {
        return null;
    }
    try {
        return mapper.readTree(value);
    } catch (IOException e) {
        throw new IllegalArgumentException(e);
    }
}
```

Which effectively meant that additionalInfo was always null in RPC status updates, even if it had a value in the original request — breaking use cases that rely on this field.

## PE

This PR doesn't have conflicts with PE version.

## General checklist

## Back-End feature checklist

### DIFF
--- a/common/proto/src/main/java/org/thingsboard/server/common/util/ProtoUtils.java
+++ b/common/proto/src/main/java/org/thingsboard/server/common/util/ProtoUtils.java
@@ -522,7 +522,7 @@ public class ProtoUtils {
     }
 
     private static TransportProtos.ToDeviceRpcRequestActorMsgProto toProto(ToDeviceRpcRequestActorMsg msg) {
-        TransportProtos.ToDeviceRpcRequestMsg proto = TransportProtos.ToDeviceRpcRequestMsg.newBuilder()
+        TransportProtos.ToDeviceRpcRequestMsg.Builder builder = TransportProtos.ToDeviceRpcRequestMsg.newBuilder()
                 .setMethodName(msg.getMsg().getBody().getMethod())
                 .setParams(msg.getMsg().getBody().getParams())
                 .setExpirationTime(msg.getMsg().getExpirationTime())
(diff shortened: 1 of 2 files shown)


ShvaykaD:
Hi @AndriiLandiak, I would like you to review this PR as well since the change that I’ve made breaks backward compatibility of the additionalInfo field handling in edge queue messages.

Previously, in cluster mode, additionalInfo was missing and defaulted to an empty string (""). As a result, when forwarding to the edge queue, we had:

```java
body.put("additionalInfo", msg.getAdditionalInfo()); // always ""
```

After this fix, the field now preserves its actual value from the original request — which may be a valid JSON string or null. This means that edge messages may now contain either the correct additionalInfo or null, instead of the legacy empty string. Please take a look and let me know if this change in behavior is acceptable, or if we should explicitly preserve the previous format for edge compatibility.

ViacheslavKlimov:
Closing this in favor of https://github.com/thingsboard/thingsboard/pull/13641

PR #13806, opened 2025-07-31
Prohibited blank relation types

dskarzh:
Changes:
- The `checkParameter(String name, String param)` method in `BaseController` now checks that the parameter value is not blank, rather than just not empty. This change synchronizes the behavior between CE and PE - PE has been checking for blanks for some time now.
- Refactored `BaseRelationService` and `EntityRelation` to use validation annotations.
- It is now not possible to save a relation with a blank type. However, deleting a relation with a blank type is still allowed (for example, to remove legacy relations with blank types).
- Minor code cleanup and refactoring.


application/src/main/java/org/thingsboard/server/controller/EntityRelationController.java
```diff
@@ -88,29 +87,26 @@ public void saveRelation(@Parameter(description = "A JSON value representing the
...
 
     private EntityRelation doSave(EntityRelation relation) throws ThingsboardException {
-        checkNotNull(relation);
-        checkCanCreateRelation(relation.getFrom());
-        checkCanCreateRelation(relation.getTo());
         if (relation.getTypeGroup() == null) {
             relation.setTypeGroup(RelationTypeGroup.COMMON);
         }
-
+        ConstraintValidator.validateFields(relation);
```
ViacheslavKlimov, line 100: Already happens in the BaseRelationService, no?

dskarzh: This is for nicer error messages and to avoid recording bad requests in audit logs (idk, maybe we should). Without this, when to or from are not specified, error returned is this - `Parameter entityId can't be empty!` - which is not very informative. With this changes, error message is - `Validation error: to must not be null`

PR #13843, opened 2025-08-10
Fix: Robust Edge session cleanup to prevent resource leaks and message lag under unstable network conditions

bcblr1993:
### The Problem

Under unstable network conditions, ThingsBoard Edge instances can experience frequent disconnections and reconnections. This can lead to a state where the Edge stops receiving any downstream messages from the Cloud, even though it appears to be connected.The primary symptom is a continuously growing consumer lag on the Edge-specific Kafka topic (tb_edge_event.notifications.*). Restarting the Edge client does not resolve the issue; only a full restart of the ThingsBoard server clears the message backlog and restores normal data flow. This indicates a persistent state issue on the server side.
[iotcloud.2025-07-21.log](https://github.com/user-attachments/files/21703269/iotcloud.2025-07-21.log)

### Root Cause Analysis
The root cause is a race condition combined with incomplete resource cleanup, leading to both application-level resource leaks and Kafka-level "zombie consumers".
1. **Initial Trigger**: An Edge client on an unstable network triggers rapid onEdgeConnect and onEdgeDisconnect events in EdgeGrpcService.
2. **Cleanup Failure**: Under load or due to timeouts, the KafkaEdgeGrpcSession.destroy() method can fail to gracefully stop its underlying Kafka consumer. This was observed in logs with warnings like Failed to stop edge event consumer.
3. **Creation of a "Zombie Consumer"**: When destroy() fails, the KafkaConsumer thread remains active. This "zombie" consumer is still a member of the consumer group from the Kafka broker's perspective, but its corresponding gRPC stream to the Edge is gone.
4. **Application-Level Leak**: The onEdgeDisconnect method removes the EdgeGrpcSession object from the main sessions map before attempting the destroy() operation. If destroy() fails, the session object and its associated ExecutorService are orphaned in memory. They are no longer reachable by the sessions map, so the built-in periodic cleanup task (destroyKafkaSessionIfDisconnectedAndConsumerActive) cannot find and terminate them.
5. **Message Hijacking & Lag**: When the Edge reconnects, a new, valid session and consumer are created. However, the zombie consumer may still be assigned the topic partition by the Kafka group coordinator. When this zombie polls messages, it has no gRPC stream to forward them to. The messages are effectively lost, the offset is never committed, and the consumer lag grows indefinitely.

**The Solution**
This PR implements a “periodic defense” strategy to make the Edge session lifecycle management resilient to these failures.
1. **Long-Term Resource Management: Zombie Session Cleanup Queue** 
This addresses the application-level memory and thread leaks.
- A new ConcurrentLinkedQueue<EdgeGrpcSession> zombieSessions is introduced in EdgeGrpcService to track session objects that failed to be destroyed.
- The onEdgeDisconnect logic is modified: if destroySession() fails, the failed session object is added to this zombieSessions queue.
- A new periodic background task, cleanupZombieSessions(), iterates through this queue and repeatedly attempts to destroy the zombie sessions until it succeeds, ensuring the eventual reclamation of all leaked resources.

**How to Reproduce the Issue**
1. **Simulate Cleanup Failure**: Modify the KafkaEdgeGrpcSession.destroy() method to always return false; and bypass the actual cleanup logic to reliably create an orphaned session.
2. **Simulate Network Churn**: Create a simple gRPC test client that connects to the ThingsBoard server as an Edge instance. The client should run in a loop where it:
a. Connects and sends the ConnectRequestMsg.
b. Waits for 1-2 seconds.
c. Abruptly shuts down its channel (e.g., channel.shutdownNow()) to simulate a network drop.
d. Waits for 1 second before reconnecting.
3. **Generate Load**: Use a separate Kafka producer to continuously send messages to the target Edge's event topic (tb_edge_event.notifications.<tenantId>.<edgeId>).
4. **Observe**: Without the fix, the Kafka consumer lag for the Edge's consumer group will increase steadily. With the fix applied, the lag is correctly processed after each reconnection, and the system remains stable without resource leaks.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
+++ b/application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
@@ -219,6 +219,9 @@ public class EdgeGrpcService extends EdgeRpcServiceGrpc.EdgeRpcServiceImplBase i
         if (executorService != null) {
             executorService.shutdownNow();
         }
+        if(zombieSessionsExecutorService != null){
+            zombieSessionsExecutorService.shutdownNow();
+        }
     }
 
     @Override
--- a/application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
+++ b/application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
@@ -69,17 +69,8 @@ import org.thingsboard.server.service.telemetry.TelemetrySubscriptionService;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
+import java.util.*;
+import java.util.concurrent.*;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


bcblr1993:
Hi, I’ve submitted this PR: “Fix: Improve Edge session cleanup to prevent resource leaks and message backlog due to unstable network conditions and Kafka busy timeout.”
Please review it when you have a chance. Thank you! 🙏

bcblr1993:
Hi, apologies for the earlier conflict in this PR. I’ve resolved it now, so it’s ready for review again. Thanks for taking another look!

volodymyr-babak:
@bcblr1993

Hello from the Edge team! Thanks a lot for submitting this PR to fix the bug, and apologies for the delayed response — it had slipped out of my view.

We’ll review it shortly and proceed with merging, as this issue has already been observed in other setups as well. In the meantime, please avoid merging the latest master or rc branches — if needed, we’ll handle that as part of the final review process.

Thanks again for your contribution!

bcblr1993:
@volodymyr-babak 
Got it, thanks for the heads-up!  
I’ll avoid merging master/rc for now and will wait for your review.  
Happy to know this fix is useful for other setups too!

volodymyr-babak:
@bcblr1993

I’ve reviewed the code and simplified it a bit.
Please take a look from your side, and if you have no objections, we’ll proceed with merging it into master.

Thanks again for your contribution - it’s really important and much appreciated from our side.

bcblr1993:
@volodymyr-babak 
I’ve reviewed your simplified code, and it looks great. I have no objections — please go ahead with merging. Thanks!

PR #13882, opened 2025-08-16
Fix the extra ']' in the help

371518473:
Fix the extra ']' in the help
image
image


vvlladd28:
Wow, super fix! Thanks, @371518473, for noticing and fixing that extra bracket ']' in the ThingsBoard help section. It seems like a small thing, but such details make the UI cleaner and more pleasant for users. You're great for contributing to open source – the project gets even better thanks to such pull requests. Keep it up! 🚀

PR #13910, opened 2025-08-22
AI Request Node: added ability to attach files

dashevchenko:
## Pull Request description

Added the ability to attach a file for an AI request. TB_RESOURCE entity type was used to save files in TB.

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
```diff
@@ -511,6 +512,11 @@ public ConcurrentMap<TenantId, DebugTbRateLimits> getDebugPerTenantLimits() {
     @Getter
     private ResourceService resourceService;
 
+    @Lazy
+    @Autowired(required = false)
+    @Getter
+    private ResourceDataCache resourceDataCache;
```
ViacheslavKlimov: Do you need `Lazy`? Why `required = false`?

application/src/main/java/org/thingsboard/server/controller/TbResourceController.java
```diff
@@ -272,6 +272,8 @@ public PageData<TbResourceInfo> getTenantResources(@Parameter(description = PAGE
                                                        @RequestParam int pageSize,
                                                        @Parameter(description = PAGE_NUMBER_DESCRIPTION, required = true)
                                                        @RequestParam int page,
+                                                       @Parameter(description = RESOURCE_TYPE, schema = @Schema(allowableValues = {"LWM2M_MODEL", "JKS", "PKCS_12", "JS_MODULE"}))
+                                                       @RequestParam(required = false) String resourceType,
```
ViacheslavKlimov: What for? There is API right above with type filter

common/data/src/main/java/org/thingsboard/server/common/data/ResourceType.java
```diff
@@ -25,7 +25,8 @@ public enum ResourceType {
     PKCS_12("application/x-pkcs12", false, false),
     JS_MODULE("application/javascript", true, true),
     IMAGE(null, true, true),
-    DASHBOARD("application/json", true, true);
+    DASHBOARD("application/json", true, true),
+    AI_PROMT(null, true, true);
```
ViacheslavKlimov: PROM**P**T

ViacheslavKlimov: customerAccess = true?

dao/src/main/java/org/thingsboard/server/dao/resource/DefaultResourceDataCache.java
```diff
@@ -0,0 +1,79 @@
...
+                    }
+                }, executor));
+    }
+
+    public CompletableFuture<byte[]> getResourceData(TenantId tenantId, TbResourceId resourceId) {
+        log.trace("Retrieving resource data by id [{}], tenant id [{}] from cache", resourceId, tenantId);
+        return cache.get(new ResourceDataKey(tenantId, resourceId));
+    }
+
+    @TransactionalEventListener(classes = ResourceInfoEvictEvent.class)
```
ViacheslavKlimov: Cluster...

application/src/main/resources/thingsboard.yml
```diff
@@ -676,6 +676,9 @@ cache:
       maxSize: "${CACHE_SPECS_IMAGE_ETAGS_MAX_SIZE:10000}" # 0 means the cache is disabled
     systemImagesBrowserTtlInMinutes: "${CACHE_SPECS_IMAGE_SYSTEM_BROWSER_TTL:0}" # Browser cache TTL for system images in minutes. 0 means the cache is disabled
     tenantImagesBrowserTtlInMinutes: "${CACHE_SPECS_IMAGE_TENANT_BROWSER_TTL:0}" # Browser cache TTL for tenant images in minutes. 0 means the cache is disabled
+    tbResourceData:
+      timeToLiveInMinutes: "${CACHE_SPECS_CLAIM_DEVICES_TTL:44640}" # TB resource data cache TTL
+      maxSize: "${CACHE_SPECS_CLAIM_DEVICES_MAX_SIZE:100000}" # 0 means the cache is disabled
```
ViacheslavKlimov: Why under image config?

common/dao-api/src/main/java/org/thingsboard/server/dao/resource/ResourceDataCache.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package org.thingsboard.server.dao.resource;
+
+import org.thingsboard.server.common.data.id.TbResourceId;
+import org.thingsboard.server.common.data.id.TenantId;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface ResourceDataCache {
```
ViacheslavKlimov: Tests

application/src/main/java/org/thingsboard/server/controller/TbResourceController.java
dashevchenko: It is needed for UI

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -396,6 +401,8 @@ protected void handleNotification(UUID id, TbProtoQueueMsg<ToCoreNotificationMsg
             callback.onSuccess();
         } else if (toCoreNotification.hasResourceCacheInvalidateMsg()) {
             forwardToResourceService(toCoreNotification.getResourceCacheInvalidateMsg(), callback);
+        } else if (toCoreNotification.hasResourceDataCacheInvalidateMsg()) {
```
ViacheslavKlimov: Can't you **reuse** ResourceCacheInvalidateMsg?

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -157,6 +184,23 @@ public void onFailure(@NonNull Throwable t) {
...
+                    .thenApply(bytes -> {
+                        if (bytes == null || bytes.length == 0) {
+                            return null;
+                        }
+                        return new TextContent(new String(bytes, StandardCharsets.UTF_8));
+                    });
+            contentFutures.add(f);
+        }
+        return CompletableFuture.allOf(contentFutures.toArray(CompletableFuture[]::new))
+                .thenApply(v -> contentFutures.stream().map(CompletableFuture::join).filter(Objects::nonNull).toList());
```
ViacheslavKlimov: Futures.allAsList

application/src/main/resources/thingsboard.yml
```diff
@@ -676,6 +676,9 @@ cache:
       maxSize: "${CACHE_SPECS_IMAGE_ETAGS_MAX_SIZE:10000}" # 0 means the cache is disabled
     systemImagesBrowserTtlInMinutes: "${CACHE_SPECS_IMAGE_SYSTEM_BROWSER_TTL:0}" # Browser cache TTL for system images in minutes. 0 means the cache is disabled
     tenantImagesBrowserTtlInMinutes: "${CACHE_SPECS_IMAGE_TENANT_BROWSER_TTL:0}" # Browser cache TTL for tenant images in minutes. 0 means the cache is disabled
+  tbResourceData:
+    timeToLiveInMinutes: "${CACHE_SPECS_CLAIM_DEVICES_TTL:44640}" # TB resource data cache TTL
+    maxSize: "${CACHE_SPECS_CLAIM_DEVICES_MAX_SIZE:100000}" # 0 means the cache is disabled
```
ViacheslavKlimov: CLAIM_DEVICES ...
Please be more attentive

application/src/main/java/org/thingsboard/server/controller/TbResourceController.java
```diff
@@ -281,7 +302,7 @@ public PageData<TbResourceInfo> getTenantResources(@Parameter(description = PAGE
         PageLink pageLink = createPageLink(pageSize, page, textSearch, sortProperty, sortOrder);
         TbResourceInfoFilter filter = TbResourceInfoFilter.builder()
                 .tenantId(getTenantId())
-                .resourceTypes(EnumSet.allOf(ResourceType.class))
+                .resourceTypes(StringUtils.isNotEmpty(resourceType) ? Set.of(ResourceType.valueOf(resourceType)) : EnumSet.allOf(ResourceType.class))
```
ViacheslavKlimov: allOf? You are returning images as well

dao/src/main/java/org/thingsboard/server/dao/resource/DefaultTbResourceDataCache.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public CompletableFuture<TbResourceDataInfo> getResourceDataInfo(TenantId tenantId, TbResourceId resourceId) {
+        log.trace("Retrieving resource data info by id [{}], tenant id [{}] from cache", resourceId, tenantId);
+        return cache.get(new ResourceDataKey(tenantId, resourceId));
+    }
+
+    @Override
+    public void evictResourceData(TenantId tenantId, TbResourceId resourceId) {
+        cache.asMap().remove(new ResourceDataKey(tenantId, resourceId));
```
ViacheslavKlimov, line 66: Why call asMap? Do invalidate.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -192,11 +242,56 @@ private static String wrapInJsonObject(String response) {
...
+            return Optional.of(new TextContent(new String(data, StandardCharsets.UTF_8)));
+        }
+        if (mediaType.startsWith("image/")) {
+            return Optional.of(new ImageContent(Base64.getEncoder().encodeToString(data), mediaType));
+        }
+        if (mediaType.equals("application/pdf")) {
+            return Optional.of(new PdfFileContent(Base64.getEncoder().encodeToString(data), mediaType));
+        }
+
+        log.warn("Unsupported media type: {} for resource", resource.getDescriptor());
```
ViacheslavKlimov: Log on debug level and fallback to TextContent

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -192,11 +242,56 @@ private static String wrapInJsonObject(String response) {
...
+        List<Content> contents = new ArrayList<>();
+        contents.add(new TextContent(userPrompt));
+
+        for (TbResourceDataInfo resource : resources) {
+            toContent(resource).ifPresent(contents::add);
+        }
+        return contents;
+    }
+
+    private Optional<Content> toContent(TbResourceDataInfo resource) {
```
ViacheslavKlimov: Don't swallow errors and print the logs the user will never see and know what's wrong. 
Throw the error and fail message processing.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -77,6 +92,7 @@ public final class TbAiNode extends TbAbstractExternalNode implements TbNode {
 
     private String systemPrompt;
     private String userPrompt;
+    private List<TbResourceId> resourceIds;
```
ViacheslavKlimov: Let's rename to `resources`?

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -192,11 +242,56 @@ private static String wrapInJsonObject(String response) {
...
+            return Optional.of(new ImageContent(Base64.getEncoder().encodeToString(data), mediaType));
+        }
+        if (mediaType.equals("application/pdf")) {
+            return Optional.of(new PdfFileContent(Base64.getEncoder().encodeToString(data), mediaType));
+        }
+
+        log.warn("Unsupported media type: {} for resource", resource.getDescriptor());
+        return Optional.empty();
+    }
+
```
ViacheslavKlimov, line 310: Add upgrade for the rule node (`"resources": []`)

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNodeConfiguration.java
```diff
@@ -41,6 +43,8 @@ public class TbAiNodeConfiguration implements NodeConfiguration<TbAiNodeConfigur
     @Length(min = 1, max = 500_000)
     private String userPrompt;
 
+    private List<UUID> resourceIds;
```
dskarzh:
```suggestion
    private Set<@NotNull(message = "references to resources cannot be null") TbResourceId> resources;
```

This is a minor suggestion, but I think it could be worth considering for a few reasons:

1. Ensure uniqueness by using `Set`
It doesn't really make sense to have duplicate resource IDs in the rule node configuration, so changing from `List` to `Set` would better express this constraint. This might also help in edge cases where users import corrupted rule chains.
2. Ensure that elements in the set cannot be null
Similar reasoning as above - null elements don't make sense in this context, so it would be good to disallow them explicitly.
3. Consistency
The `modelId` field is already present and uses the full ID object (with entity type). It would be nice to stay consistent and use the same format for resource IDs as well.

As a bonus, after this change you could remove the UUID-to-ID object mapping in the init method.

What do you think?

dao/src/main/java/org/thingsboard/server/dao/resource/DefaultTbResourceDataCache.java
```diff
@@ -0,0 +1,70 @@
...
+    private void init() {
+        cache = Caffeine.newBuilder()
+                .maximumSize(cacheMaxSize)
+                .expireAfterAccess(cacheValueTtl, TimeUnit.MINUTES)
+                .executor(executorService)
+                .buildAsync((key, executor) -> CompletableFuture.supplyAsync(() -> resourceService.getResourceDataInfo(key.tenantId(), key.resourceId()), executor));
+    }
+
+    @Override
+    public CompletableFuture<TbResourceDataInfo> getResourceDataInfo(TenantId tenantId, TbResourceId resourceId) {
```
dskarzh:
```suggestion
    public FluentFuture<TbResourceDataInfo> getResourceDataInfoAsync(TenantId tenantId, TbResourceId resourceId) {
```

This is an important consideration. I'd suggest avoiding introducing another async API to the system if possible. We already use Guava async APIs like `ListenableFuture` and `FluentFuture` throughout the codebase. Having mixed async APIs can potentially create conflicts where you might have two async methods that you can't easily compose because they use different async APIs, leading to verbose conversion code.
I understand that Caffeine forces you to use `CompletableFuture`, but we could encapsulate that usage within this class and convert it to `FluentFuture`. Here's how you could do it:
```java
@Override
public FluentFuture<TbResourceDataInfo> getResourceDataInfo(TenantId tenantId, TbResourceId resourceId) {
    log.trace("Retrieving resource data info by id [{}], tenant id [{}] from cache", resourceId, tenantId);
    SettableFuture<TbResourceDataInfo> future = SettableFuture.create();
    cache.get(new ResourceDataKey(tenantId, resourceId)).whenComplete((result, exception) -> {
        if (exception != null) {
            future.setException(exception);
        } else {
            future.set(result);
        }
    });
    return FluentFuture.from(future);
}
```
You could even create a reusable utility method and place it in `DonAsynchron` for future use:
```java
public static <T> FluentFuture<T> toFluentFuture(CompletableFuture<T> completable) {
    SettableFuture<T> future = SettableFuture.create();
    completable.whenComplete((result, exception) -> {
        if (exception != null) {
            future.setException(exception);
        } else {
            future.set(result);
        }
    });
    return FluentFuture.from(future);
}
```

What do you think?

dao/src/main/java/org/thingsboard/server/dao/resource/DefaultTbResourceDataCache.java
dskarzh: I think this approach is fine. The Caffeine author suggests it as one of the recommended ways to invalidate entries in async caches: https://github.com/ben-manes/caffeine/issues/422

application/src/main/java/org/thingsboard/server/controller/TbResourceController.java
```diff
@@ -263,6 +267,21 @@ public PageData<TbResourceInfo> getResources(@Parameter(description = PAGE_SIZE_
         }
     }
 
+    @ApiOperation(value = "Get Tenant Resource Infos by ids (getTenantResourcesByIds)")
+    @PreAuthorize("hasAnyAuthority('TENANT_ADMIN')")
+    @GetMapping(value = "/resource/tenant", params = {"resourceIds"})
+    public List<TbResourceInfo> getTenantResourcesByIds(
+            @Parameter(description = "A list of resource ids, separated by comma ','", array = @ArraySchema(schema = @Schema(type = "string")))
+            @RequestParam("resourceIds") String[] strResourceIds) throws ThingsboardException {
```
dskarzh:
```suggestion
            @RequestParam("resourceIds") Set<UUID> resourceIds) throws ThingsboardException {
```

I'd suggest moving away from using a raw string array to accept a set of IDs instead. This provides better type safety and ensures uniqueness of the resource IDs.

application/src/main/java/org/thingsboard/server/service/resource/DefaultTbResourceService.java
```diff
@@ -102,7 +108,7 @@ public TbResourceDeleteResult delete(TbResourceInfo tbResource, boolean force, U
             if (result.isSuccess()) {
                 logEntityActionService.logEntityAction(tenantId, resourceId, tbResource, actionType, user, resourceId.toString());
             }
-
+            evictResourceDataCache(tenantId, resourceId);
```
dskarzh: I think we should evict from cache only if deletion was successful.

application/src/main/java/org/thingsboard/server/service/resource/DefaultTbResourceService.java
```diff
@@ -191,4 +197,15 @@ private Comparator<? super LwM2mObject> getComparator(String sortProperty, Strin
...
 
+    private void evictResourceDataCache(TenantId tenantId, TbResourceId resourceId) {
+        tbResourceDataCache.evictResourceData(tenantId, resourceId);
+        clusterService.broadcastToCore(TransportProtos.ToCoreNotificationMsg.newBuilder()
+                .setResourceCacheInvalidateMsg(TransportProtos.ResourceCacheInvalidateMsg.newBuilder()
+                        .setTenantIdMSB(tenantId.getId().getMostSignificantBits())
+                        .setTenantIdLSB(tenantId.getId().getLeastSignificantBits())
+                        .setResourceIdMSB(resourceId.getId().getMostSignificantBits())
+                        .setResourceIdLSB(resourceId.getId().getLeastSignificantBits()).build())
+                .build());
```
dskarzh: As I see it, the resource data cache is deployed in every microservice we have. Let's say I'm in an AI request node and I'm using a resource - it gets cached in memory in the rule engine microservice I'm running in. Later, this resource gets updated and only the core microservices receive the invalidation message. Wouldn't the rule engines keep returning stale resource data from their local cache?

application/src/main/resources/thingsboard.yml
```diff
@@ -676,6 +676,9 @@ cache:
       maxSize: "${CACHE_SPECS_IMAGE_ETAGS_MAX_SIZE:10000}" # 0 means the cache is disabled
     systemImagesBrowserTtlInMinutes: "${CACHE_SPECS_IMAGE_SYSTEM_BROWSER_TTL:0}" # Browser cache TTL for system images in minutes. 0 means the cache is disabled
     tenantImagesBrowserTtlInMinutes: "${CACHE_SPECS_IMAGE_TENANT_BROWSER_TTL:0}" # Browser cache TTL for tenant images in minutes. 0 means the cache is disabled
+  tbResourceData:
+    timeToLiveInMinutes: "${CACHE_SPECS_RESOURCE_DATA_TTL:44640}" # TB resource data cache TTL
```
dskarzh:
```suggestion
    timeToLiveInMinutes: "${CACHE_SPECS_RESOURCE_DATA_TTL:10080}" # TB resource data cache TTL
```
I think one month is too much. Let's go down to one week maybe?

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNodeConfiguration.java
dashevchenko: 1. Ok
2. I don`t think we do it across the project.
3. The UI team requested it. They usually send a list of UUIDs.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -111,6 +127,12 @@ public void init(TbContext ctx, TbNodeConfiguration configuration) throws TbNode
             // LangChain4j AnthropicChatModel rejects requests with non-null ResponseFormat even if ResponseFormatType is TEXT
             responseFormat = config.getResponseFormat().toLangChainResponseFormat();
         }
+        if (config.getResourceIds() != null && !config.getResourceIds().isEmpty()) {
```
dskarzh, line 137: Ensure that such resources actually exist and belong to the current tenant. You can fetch them using your resource data cache during initialization, perform the checks, and have the resources already cached since they will likely be requested when the AI node is used.

dashevchenko: Do you mean store resources inside the rule node? What if resource will be deleted?

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -77,9 +97,11 @@ public final class TbAiNode extends TbAbstractExternalNode implements TbNode {
 
     private String systemPrompt;
     private String userPrompt;
+    private Set<TbResourceId> resourceIds;
     private ResponseFormat responseFormat;
     private int timeoutSeconds;
     private AiModelId modelId;
+    private ExecutorService aiMessageExecutor;
```
dskarzh: Please do not do this, as creating a thread pool per rule node could lead to some significant issues:

-  Unbounded thread pool: This will create a new thread for each task when no threads are currently available, which risks uncontrolled resource usage
- Resource overhead: Since this executor performs a single lightweight callback, creating a whole thread for this is overkill

To illustrate the potential impact: imagine we have, say, 40 AI nodes using resources that receive messages around the same time. This executor would create 40 threads just to perform a method that does no I/O and only manipulates Java objects. Those threads would not be removed and would stay in memory indefinitely, consuming resources.

Use `ctx.getDbCallbackExecutor()` for callbacks after I/O operations in rule nodes. This is a shared thread pool that exists exactly for this purpose and will handle the workload much more efficiently.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -192,11 +253,67 @@ private static String wrapInJsonObject(String response) {
         return JacksonUtil.newObjectNode().put("response", response).toString();
     }
 
+    private void validateResource(TbContext ctx, TbResourceId tbResourceId) throws TbNodeException {
+        TbResourceInfo resource = ctx.getResourceService().findResourceInfoById(ctx.getTenantId(), tbResourceId);
+        if (resource == null) {
+            throw new TbNodeException("[" + ctx.getTenantId() + "] Resource with ID: [" + tbResourceId + "] was not found");
```
dskarzh:
```suggestion
            throw new TbNodeException("[" + ctx.getTenantId() + "] Resource with ID: [" + tbResourceId + "] was not found", true);
```

This is unrecoverable error that requires config change to remove resource that does not exist.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/TbAiNode.java
```diff
@@ -192,11 +253,67 @@ private static String wrapInJsonObject(String response) {
...
+        }
+        if (!ResourceType.GENERAL.equals(resource.getResourceType())) {
+            throw new TbNodeException("[" + ctx.getTenantId() + "] Resource with ID: [" + tbResourceId + "] has unsupported resource type: " + resource.getResourceType());
```
dskarzh:
```suggestion
            throw new TbNodeException("[" + ctx.getTenantId() + "] Resource with ID: [" + tbResourceId + "] has unsupported resource type: " + resource.getResourceType(), true);
```

This is unrecoverable error that requires config change to remove resource that has wrong type.

rule-engine/rule-engine-api/src/main/java/org/thingsboard/rule/engine/api/TbContext.java
```diff
@@ -431,4 +436,5 @@ public interface TbContext {
 
     MqttClientSettings getMqttClientSettings();
 
+    <E extends HasId<I> & HasTenantId, I extends EntityId> void checkTenantEntity(E entity) throws TbNodeException;
```
dskarzh: Let's move this method to be right below existing 
`void checkTenantEntity(EntityId entityId) throws TbNodeException;`

They are doing basically the same thing but are placed in two different locations

dao/src/test/java/org/thingsboard/server/dao/sql/rule/JpaRuleNodeDaoTest.java
```diff
@@ -98,6 +99,12 @@ public void testFindRuleNodesByTenantIdAndType() {
         assertEquals(20, ruleNodes2.size());
     }
 
+    @Test
+    public void testFindRuleNodesByResourceId() {
```
dskarzh: If this test is designed to verify that nothing should be found when no rule nodes use resources, then it needs a better name.
We also need a positive test that verifies rule nodes using specific resources are correctly found by their resource ID.

application/src/main/java/org/thingsboard/server/controller/TbResourceController.java
```diff
@@ -263,6 +266,20 @@ public PageData<TbResourceInfo> getResources(@Parameter(description = PAGE_SIZE_
...
+    @ApiOperation(value = "Get Resource Infos by ids (getSystemOrTenantResourcesByIds)")
+    @PreAuthorize("hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN')")
+    @GetMapping(value = "/resource", params = {"resourceIds"})
+    public List<TbResourceInfo> getSystemOrTenantResourcesByIds(
+            @Parameter(description = "A list of resource ids, separated by comma ','", array = @ArraySchema(schema = @Schema(type = "string")))
+            @RequestParam("resourceIds") Set<UUID> resourceUuids) throws ThingsboardException {
+        SecurityUser user = getCurrentUser();
+        List<TbResourceId> resourceIds = new ArrayList<>();
+        for (UUID resourceId : resourceUuids) {
+            resourceIds.add(new TbResourceId(resourceId));
```
ViacheslavKlimov, line 278: Where are permission checks

dao/src/main/java/org/thingsboard/server/dao/resource/BaseResourceService.java
```diff
@@ -344,32 +355,48 @@ public TbResourceDeleteResult deleteResource(TenantId tenantId, TbResourceId res
...
-                    if (!entities.isEmpty()) {
-                        affectedEntities.put(entityType.name(), entities);
-                    }
-                });
-
-                if (!affectedEntities.isEmpty()) {
-                    success = false;
-                    result.references(affectedEntities);
-                }
+            Map<String, List<? extends HasId<?>>> references = findResourceReferences(tenantId, resource);
```
ViacheslavKlimov: You don't need `? extends HasId<?>` anymore, EntityInfo

PR #14082, opened 2025-09-30
Housekeeper - run LatestTsDeletion and TsHistoryDeletion in parallel

AndriiLandiak:
## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/service/housekeeper/HousekeeperService.java
```diff
@@ -90,33 +110,112 @@ public void afterStartUp() {
...
+
+                    processFuture.addListener(() -> {
+                        timeoutFuture.cancel(false);
+                        try {
+                            processFuture.get();
+                            long timing = System.currentTimeMillis() - startTs;
+                            if (log.isDebugEnabled()) {
+                                log.debug("[{}] Processed {} in {} ms (attempt {})", task.getTenantId(), task.getDescription(), timing, msg.getTask().getAttempt());
+                            }
+                            statsService.ifPresent(s -> s.reportProcessed(taskType, msg, timing));
```
ViacheslavKlimov: The success branch calls `reportProcessed` (and logs the timing) unconditionally, before/regardless of whether `completionFuture.set(null)` actually wins — unlike the two failure branches, which guard their side effects on `completionFuture.setException(...)` returning `true`. If the task completes successfully at the same instant the timeout fires, `cancel(true)` is a no-op, `processFuture.get()` returns normally, and you get both a `reportProcessed` here AND a `reportFailure` + `submitForReprocessing` from the timeout path for the same task (plus a redundant reprocessing of an already-completed deletion). Could the success side mirror the failure side — only report/log when `completionFuture.set(null)` returns `true`?

application/src/main/java/org/thingsboard/server/service/housekeeper/HousekeeperService.java
```diff
@@ -90,33 +110,112 @@ public void afterStartUp() {
...
+                allFutures.get();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                return;
+            } catch (ExecutionException e) {
+                // successfulAsList doesn't throw ExecutionException, but it's a checked exception from Future.get()
+            }
+        }
+
         consumer.commit();
```
ViacheslavKlimov, line 144: A subtle ordering shift versus the sync path worth a sanity check: `completionFuture` is completed (releasing the consumer thread from `allFutures.get()` to reach `consumer.commit()`) *before* `handleFailure` runs, since `setException(...)` is called first and the listener executes on `directExecutor`. So the offset for a failed task can be committed before `submitForReprocessing` has even called `producer.send`. In the old sync path the resubmit was initiated on the consumer thread before commit. The producer send was always async/best-effort, so this may be acceptable, but a crash in that window now loses the task entirely (offset committed, no reprocessing message sent). Was the weaker ordering intentional?

application/src/main/java/org/thingsboard/server/service/housekeeper/HousekeeperService.java
```diff
@@ -90,33 +110,112 @@ public void afterStartUp() {
...
     }
 
     @SuppressWarnings("unchecked")
-    protected <T extends HousekeeperTask> void processTask(ToHousekeeperServiceMsg msg) throws Exception {
+    protected <T extends HousekeeperTask> ListenableFuture<Void> processTask(ToHousekeeperServiceMsg msg, boolean isReprocessing) throws Exception {
```
ViacheslavKlimov, line 148: This method now signals its mode two ways at once: a `boolean isReprocessing` in, and a nullable `ListenableFuture<Void>` out where `null` implicitly means "handled synchronously, nothing to await." Callers have to know that contract and null-check it (the `if (asyncFuture != null)` guard in `processMsgs`), and a bare boolean at the call site is easy to misread. Would it read more clearly to either return a non-null future in every case (`Futures.immediateFuture(null)` on the sync path, dropping the null-check), or to have the two consumers call distinct `processTaskSync` / `processTaskAsync` entry points directly?

application/src/main/java/org/thingsboard/server/service/housekeeper/HousekeeperService.java
```diff
@@ -90,33 +110,112 @@ public void afterStartUp() {
...
         }
+
+        if (!asyncFutures.isEmpty()) {
+            ListenableFuture<List<Void>> allFutures = Futures.successfulAsList(asyncFutures);
+            try {
+                allFutures.get();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                return;
+            } catch (ExecutionException e) {
```
ViacheslavKlimov: This empty `catch` whose only body is a comment explaining it can't actually fire reads as dead code — a future reader can't tell if a real handler was dropped by accident. Since `Futures.successfulAsList` never completes exceptionally, maybe drop the `ExecutionException` catch and move the explanation next to the `successfulAsList` call, or at least log if it somehow does occur.

application/src/main/java/org/thingsboard/server/service/housekeeper/HousekeeperService.java
```diff
@@ -90,33 +110,112 @@ public void afterStartUp() {
...
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T extends HousekeeperTask> ListenableFuture<Void> processTaskAsync(ToHousekeeperServiceMsg msg, HousekeeperTask task, HousekeeperTaskType taskType, HousekeeperTaskProcessor<T> taskProcessor) {
+        long startTs = System.currentTimeMillis();
+        SettableFuture<Void> completionFuture = SettableFuture.create();
+
+        try {
+            asyncTaskExecutor.submit(() -> {
```
ViacheslavKlimov, line 180: Since `processAsync` returns a future and the real deletion runs on the timeseries/Cassandra/JDBC executors (with all completion handling on `directExecutor`), this fixed pool only absorbs the synchronous prologue of `processAsync` — query building, the `removeLatest`/`remove` call that returns immediately. That's a fairly subtle role for a whole configurable, auto-sized pool: `async-processing-threads` bounds concurrent *submissions*, not concurrent deletions, which makes the knob hard to reason about. A short comment on what work is expected to run on these threads would help the next maintainer size it.

application/src/main/java/org/thingsboard/server/service/housekeeper/HousekeeperService.java
```diff
@@ -90,33 +110,112 @@ public void afterStartUp() {
...
+                try {
+                    ListenableFuture<Void> processFuture = taskProcessor.processAsync((T) task);
+
+                    ScheduledFuture<?> timeoutFuture = timeoutScheduler.schedule(() -> {
```
ViacheslavKlimov, line 186: There are now two independent timeout implementations for the same concept: the sync path relies on `taskExecutor.submit(...).get(timeout)`, while the async path uses this dedicated single-thread `ScheduledExecutorService` plus manual `cancel`/`setException` juggling. Both read `config.getTaskProcessingTimeout()` but enforce it differently, so any future change to timeout semantics has to be made and verified in two places. Could the "apply a timeout to a ListenableFuture" concern be factored into one helper shared by both paths (or at least cross-referenced in a comment)?

application/src/main/java/org/thingsboard/server/service/housekeeper/HousekeeperService.java
```diff
@@ -134,38 +233,48 @@ protected <T extends HousekeeperTask> void processTask(ToHousekeeperServiceMsg m
...
             }
         }
     }
 
+    private void handleFailure(HousekeeperTaskType taskType, ToHousekeeperServiceMsg msg, HousekeeperTask task, Throwable error) {
+        if (error instanceof ExecutionException && error.getCause() != null) {
+            error = error.getCause();
+        }
+        if (error instanceof TimeoutException) {
+            error = new TimeoutException("Timeout after " + config.getTaskProcessingTimeout() + " ms");
```
ViacheslavKlimov, line 263: For the async caller this rebuilds a brand-new `TimeoutException` with the exact same `"Timeout after ... ms"` message that was already constructed at line 180 before calling `handleFailure`, discarding the original. The literal now lives in two spots. Minor, but collapsing to a single construction site (e.g. always build it here) would avoid the duplication and the throwaway exception.

application/src/test/java/org/thingsboard/server/service/housekeeper/HousekeeperServiceAsyncTest.java
```diff
@@ -0,0 +1,190 @@
...
+        timeseriesService.save(tenantId, device.getId(),
+                new BasicTsKvEntry(System.currentTimeMillis(), new StringDataEntry(TELEMETRY_KEY, KV_VALUE))).get();
+
+        doDelete("/api/device/" + device.getId()).andExpect(status().isOk());
+
+        await().atMost(30, TimeUnit.SECONDS).untilAsserted(() -> {
+            assertThat(getTimeseriesHistory(device.getId())).isEmpty();
+        });
+
+        verify(housekeeperReprocessingService, times(2)).submitForReprocessing(any(), any());
```
ViacheslavKlimov, line 142: Nice coverage of the failure→reprocess and timeout→reprocess paths. One gap: the terminal branch in `handleFailure` — attempts exceeding `maxReprocessingAttempts`, which fires `notificationRuleProcessor.process(...)` — is only exercised on the sync path (`HousekeeperServiceTest` with async disabled). Since `handleFailure` is now shared and reachable from the async path too, a test that drives an async task past max attempts and asserts the notification trigger would close the gap on the new code.

ViacheslavKlimov (commented):
## Review summary

Reviewed 11 changed files in [Housekeeper - run LatestTsDeletion and TsHistoryDeletion in parallel](https://github.com/thingsboard/thingsboard/pull/14082). The async-processing design is sound overall — futures are aggregated and waited on before commit, failures unwind through a shared `handleFailure`, and reprocessed tasks are deliberately kept on the sync path. Left 8 comment(s) inline: two on correctness (a success/timeout reporting race and a commit-ordering question), the rest on design, readability, and a test-coverage gap.

---

application/src/main/java/org/thingsboard/server/service/housekeeper/HousekeeperService.java
AndriiLandiak: I left the two implementations as-is but added cross-referencing comments on both. The shapes are different enough (blocking future.get(timeout) vs a ListenableFuture + scheduled cancellation) that a shared helper felt like it'd obscure more than it'd unify. Open to extracting one if you feel strongly - the comments at least make the "change both" requirement explicit for now.

PR #14195, opened 2025-10-21
Fixed incorrect CF calculation when same key is used across multiple arguments

irynamatveieva:
## Pull Request description

When the same key from an entity is used in different arguments within a calculated field, the system does not handle it properly.
Only one argument value gets updated, while the others remain stale.

PE PR: https://github.com/thingsboard/thingsboard-pe/pull/3960

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/cf/ctx/state/CalculatedFieldCtx.java
+++ b/application/src/main/java/org/thingsboard/server/service/cf/ctx/state/CalculatedFieldCtx.java
@@ -35,6 +35,7 @@ import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.common.data.kv.AttributeKvEntry;
 import org.thingsboard.server.common.data.kv.TsKvEntry;
 import org.thingsboard.server.common.data.tenant.profile.DefaultTenantProfileConfiguration;
+import org.thingsboard.server.common.data.util.CollectionsUtil;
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/thingsboard/server/service/cf/ctx/state/CalculatedFieldCtx.java
```diff
@@ -88,9 +90,10 @@ public CalculatedFieldCtx(CalculatedField calculatedField, TbelInvokeService tbe
             var refId = entry.getValue().getRefEntityId();
             var refKey = entry.getValue().getRefEntityKey();
             if (refId == null || refId.equals(calculatedField.getEntityId())) {
-                mainEntityArguments.put(refKey, entry.getKey());
+                mainEntityArguments.computeIfAbsent(refKey, key -> new HashSet<>()).add(entry.getKey());
             } else {
-                linkedEntityArguments.computeIfAbsent(refId, key -> new HashMap<>()).put(refKey, entry.getKey());
+                linkedEntityArguments.computeIfAbsent(refId, key -> new HashMap<>())
+                        .computeIfAbsent(refKey, key -> new HashSet<>()).add(entry.getKey());
```
ViacheslavKlimov: I don't like that a HashSet<>(16) is created for each refKey. In most cases, there will only be one arg for a ref key.
Please do entityArguments.replace(..) with `Set.of(x)`

PR #14237, opened 2025-10-28
Add tests for rule engine consumer loop

dskarzh:
Changes:
- Added test for rule engine consumer loop logic
- Added `TbMsgPackProcessingContextFactory` to be able to verify `TbMsgPackProcessingContext::await` calls
- Added tests for `TbMsgPackProcessingContext::await`
- Added missing `queue.rule-engine.prometheus-stats.enabled` to `thingsboard.yml`
- Refactored `IdMsgPair` from regular class to a record


application/src/main/resources/thingsboard.yml
```diff
@@ -1911,6 +1911,9 @@ queue:
       print-interval-ms: "${TB_QUEUE_RULE_ENGINE_STATS_PRINT_INTERVAL_MS:60000}"
       # Max length of the error message that is printed by statistics
       max-error-message-length: "${TB_QUEUE_RULE_ENGINE_MAX_ERROR_MESSAGE_LENGTH:4096}"
+    prometheus-stats:
+      # Enable/disable Prometheus statistics for individual Rule Engine message processing (records time in ms for success/failure).
+      enabled: "${TB_QUEUE_RULE_ENGINE_PROMETHEUS_STATS_ENABLED:true}"
```
ViacheslavKlimov: Make false by default, as it was before

dskarzh: nice catch

PR #14417, opened 2025-11-25
Added automatic SSL/TLS certificate reload for transports without service restart

AndriiLandiak:
## Pull Request description

  Introduces automatic SSL/TLS certificate reload functionality for all ThingsBoard transport protocols (MQTT, CoAP/DTLS, LwM2M, HTTP), eliminating the need for service restarts when certificates are updated or rotated.

## General checklist

## Back-End feature checklist


common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
```diff
@@ -67,11 +68,24 @@ public SslCredentialsConfig mqttSslCredentials() {
...
 
     public SslHandler getSslHandler() {
         if (sslContext == null) {
-            sslContext = createSslContext();
+            synchronized (this) {
+                if (sslContext == null) {
+                    sslContext = createSslContext();
+                }
+            }
         }
```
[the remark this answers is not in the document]
AndriiLandiak: If createSslContext() throws a RuntimeException, the exception propagates to the caller, and sslContext remains null. This is fine! The next connection attempt will retry the creation - it's self-healing.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/DefaultLwM2mTransportService.java
```diff
@@ -214,6 +230,28 @@ private void setServerWithCredentials(LeshanServerBuilder builder) {
...
+
+            server.destroy();
+            log.info("Old LwM2M server stopped.");
+        }
+
+        log.info("Creating new LwM2M server with updated certificates...");
+        this.server = getLhServer();
+        this.context.setServer(server);
+        this.startLhServer();
+        log.info("New LwM2M server started successfully.");
```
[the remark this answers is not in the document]
AndriiLandiak: I am not sure I can build two servers on the same port, so. Suggest a solution, copilot....

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
```diff
@@ -66,11 +67,24 @@ public SslCredentialsConfig mqttSslCredentials() {
...
+            log.info("MQTT SSL certificates reloaded. Invalidating SSL context...");
+            sslContext = null;
+            log.info("MQTT SSL context invalidated. Will be recreated on next connection.");
+        });
+    }
 
     public SslHandler getSslHandler() {
         if (sslContext == null) {
-            sslContext = createSslContext();
+            synchronized (this) {
```
ViacheslavKlimov, line 84: Race condition: between the `if (sslContext == null)` check (line 82) returning `false` and `sslContext.createSSLEngine()` (line 89), the reload callback on another thread can set `sslContext = null`, causing a `NullPointerException`.

The double-checked locking pattern is correct for lazy initialization, but the field is read again *after* the synchronized block without protection.

Fix: use a local variable throughout the method:
```java
public SslHandler getSslHandler() {
    SSLContext ctx = sslContext;
    if (ctx == null) {
        synchronized (this) {
            ctx = sslContext;
            if (ctx == null) {
                ctx = createSslContext();
                sslContext = ctx;
            }
        }
    }
    SSLEngine sslEngine = ctx.createSSLEngine();
    ...
}
```

common/coap-server/src/main/java/org/thingsboard/server/coapserver/DefaultCoapServerService.java
```diff
@@ -137,4 +140,54 @@ private long getDtlsSessionReportTimeout() {
...
+    }
+
+    private synchronized void recreateDtlsEndpoint() throws IOException {
+        CoapEndpoint oldDtlsEndpoint = dtlsCoapEndpoint;
+        DTLSConnector oldDtlsConnector = dtlsConnector;
+
+        Configuration networkConfig = createNetworkConfiguration();
+
+        log.info("Creating new DTLS endpoint with updated certificates...");
+        createDtlsEndpoint(networkConfig);
```
ViacheslavKlimov: `createDtlsEndpoint(networkConfig)` mutates instance fields (`dtlsConnector`, `dtlsCoapEndpoint`, `tbDtlsCertificateVerifier`) and calls `server.addEndpoint(dtlsCoapEndpoint)` *before* `dtlsCoapEndpoint.start()` on line 178. If `start()` throws:

1. The instance fields already point to the new (un-started) objects — the old references are lost.
2. The new endpoint was added to the server but never started.
3. The old endpoint is never cleaned up (the `if (oldDtlsEndpoint != null)` block is skipped because the exception propagates).

Compare with the LwM2M equivalent (`recreateLwM2mServer`) which correctly uses a create-then-swap pattern: it assigns `this.server = newServer` only *after* `newServer.start()` succeeds.

Consider refactoring `createDtlsEndpoint` to return the created objects rather than mutating fields, and only assign them to instance fields after `start()` succeeds.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/CertificateReloadManager.java
```diff
@@ -0,0 +1,229 @@
...
+
+                    List<Path> filePaths = credentials.getCertificateFilePaths();
+                    if (filePaths == null || filePaths.isEmpty()) {
+                        log.debug("No certificate files to watch for: {} ({})", config.getName(), beanName);
+                        continue;
+                    }
+
+                    for (Path filePath : filePaths) {
+                        if (filePath != null && Files.exists(filePath)) {
+                            String watcherKey = config.getName() + " - " + filePath.getFileName();
```
ViacheslavKlimov: Watcher key collision risk: `config.getName() + " - " + filePath.getFileName()` uses only the filename, not the full path. If two different `SslCredentialsConfig` beans have the same name and watch files with the same filename but different directories (e.g., `/certs/a/server.pem` and `/certs/b/server.pem`), the second `watchers.put()` on line 62 will silently overwrite the first watcher.

Consider using the full absolute path in the key, or at minimum `filePath.toString()` instead of `filePath.getFileName()`.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/CertificateReloadManager.java
```diff
@@ -0,0 +1,229 @@
...
+                            registerWatcher(watcherKey, filePath, config::onCertificateFileChanged);
```
ViacheslavKlimov: **Low-confidence observation:** When `PemSslCredentials` has both `certFile` and `keyFile`, `getCertificateFilePaths()` returns two paths, and two separate watchers are registered here — both pointing to the same `config::onCertificateFileChanged` callback. During certificate rotation (both files updated), this causes the reload to fire twice:

1. First reload may use the new cert with the old key (or vice versa) if the files are not updated atomically, potentially causing a transient failure.
2. Even if both succeed, it unnecessarily recreates DTLS endpoints / LwM2M servers twice.

Consider debouncing or deduplicating reloads per `SslCredentialsConfig` — for example, by registering a single watcher per config that checks all its files, rather than one watcher per file.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttSslHandlerProvider.java
```diff
@@ -98,16 +115,16 @@ private SSLContext createSslContext() {
             sslContext.init(km, tm, null);
             return sslContext;
         } catch (Exception e) {
-            log.error("Unable to set up SSL context. Reason: " + e.getMessage(), e);
+            log.error("Unable to set up SSL context. Reason: {}", e.getMessage(), e);
```
smatvienko-tb, line 118: log.error("Unable to set up SSL context", e)

PR #14458, opened 2025-12-01
Fixed REST client for `/entitiesQuery/find/keys` API

dskarzh:
Changes:
- Fixed rest client method for  `/entitiesQuery/find/keys` API
- Refactoring and improvements to the same API:
    - Return response type instead of raw `JsonNode`
    - Improve API docs
    - Do use `DbCallbackExecutor` for fetching time series keys, use `JpaExecutorService` instead
    - Minor code style refactorings


common/data/src/main/java/org/thingsboard/server/common/data/query/AvailableEntityKeys.java
```diff
@@ -0,0 +1,67 @@
...
+        )
+        List<String> timeseries,
+
+        @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
+        @ArraySchema(
+                arraySchema = @Schema(description = "List of unique attribute key names available on the matched entities."),
+                schema = @Schema(implementation = String.class, example = "serialNumber"),
+                uniqueItems = true
+        )
+        List<String> attribute
```
ViacheslavKlimov, line 54: attribute**s**

dskarzh: image

This is a breaking change, API was returning `attribute`

PR #14509, opened 2025-12-04
Added support of relations lifecycle updates for propagation CF

ShvaykaD:
## Pull Request description

Added support of relations lifecycle updates for propagation CF

PE PR: https://github.com/thingsboard/thingsboard-pe/pull/4127

## General checklist

- [x] You have reviewed the guidelines [document](https://docs.google.com/document/d/1wqcOafLx5hth8SAg4dqV_LV3un3m5WYR8RdTJ4MbbUM/edit?usp=sharing).
- [x] [Labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#about-labels) that classify your pull request have been added.
- [x] The [milestone](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) is specified and corresponds to fix version.  
- [x] Description references specific [issue](https://github.com/thingsboard/thingsboard/issues).
- [x] Description contains human-readable scope of changes.
- [x] Description contains brief notes about what needs to be added to the documentation.
- [x] No merge conflicts, commented blocks of code, code formatting issues.
- [x] Changes are backward compatible or upgrade script is provided.
- [x] Similar PR is opened for PE version to simplify merge. Crosslinks between PRs added. Required for internal contributors only.
  
## Front-End feature checklist

- [x] Screenshots with affected component(s) are added. The best option is to provide 2 screens: before and after changes;
- [x] If you change the widget or other API, ensure it is backward-compatible or upgrade script is present.
- [x] Ensure new API is documented [here](https://github.com/thingsboard/thingsboard-ui-help)

## Back-End feature checklist

- [x] Added corresponding unit and/or integration test(s). Provide written explanation in the PR description if you have failed to add tests.
- [x] If new dependency was added: the dependency tree is checked for conflicts.
- [x] If new service was added: the service is marked with corresponding @TbCoreComponent, @TbRuleEngineComponent, @TbTransportComponent, etc.
- [x] If new REST API was added: the RestClient.java was updated, issue for [Python REST client](https://github.com/thingsboard/thingsboard-python-rest-client) is created.
- [x] If new yml property was added: make sure a description is added (above or near the property).

PR #14536, opened 2025-12-08
Fixed key dictionary race condition causing Hibernate to cache zero keyId in cluster mode

ShvaykaD:
## Pull Request description

This PR fixes a concurrency issue in `key_dictionary` where, under clustered execution, one of the rule-engine nodes can resolve a dictionary key with keyId = 0. This incorrect value propagates into `keyDictionaryMap` and later breaks latest-telemetry writes and search queries.

### Root Cause

Here is a log screenshot from local MSA black-box testing for the branch from this [PR](https://github.com/thingsboard/thingsboard/pull/14509).  

image

It shows two TB rule engine nodes processing the same telemetry key (`temperatureComputed`) at the same time. Node 1 resolves and uses `keyId = 37` and successfully updates `ts_kv_latest`, while node 2 logs `HHH000502` and then reports `keyId = 0` for the same dictionary key.

When two TB rule-engine nodes process the same telemetry key simultaneously, the first node successfully inserts the dictionary entry and the database generates the correct `key_id` (e.g., `37`).
The second node creates a new `KeyDictionaryEntry` where `keyId` starts as the default primitive value `0`.

Because the field was mapped as:

```java
@Generated
private int keyId;
```

Hibernate treats keyId as a database-generated and immutable property.

In this concurrent scenario Hibernate:

 - Detects that the property is “immutable” (`HHH000502` warning),
 - Does not populate `key_id` from the database, and
 - Returns the in-memory entity with the default primitive `keyId = 0`.

No fallback or reload happens — Hibernate simply reuses the stale managed entity. As a result, 0 is cached in keyDictionaryMap and used for saving latest telemetry, which causes missing ts_kv_latest entries and test flakiness in cluster mode.

### Fix

The fix replaces the JPA `save()` + fallback logic with a single atomic Postgres UPSERT:

`INSERT ... ON CONFLICT (key) DO UPDATE ... RETURNING key_id`, and treats `keyId` as a read-only, DB-generated `Integer`. This guarantees that `keyId` is always resolved directly from the database (the same value on all nodes) and never from a failed cached entity, eliminating the `keyId = 0` scenario and stabilizing latest telemetry writes in cluster mode.

## PE PR

No conflicts, so I didn't create separate PR for PE. 

## General checklist

## Back-End feature checklist

### DIFF
--- a/dao/src/main/java/org/thingsboard/server/dao/sqlts/dictionary/JpaKeyDictionaryDao.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/sqlts/dictionary/JpaKeyDictionaryDao.java
@@ -51,9 +51,7 @@ public class JpaKeyDictionaryDao extends JpaAbstractDaoListeningExecutorService 
             return cached;
         }
         var compositeKey = new KeyDictionaryCompositeKey(strKey);
-        Optional<Integer> existingId = keyDictionaryRepository.findById(compositeKey)
-                .map(KeyDictionaryEntry::getKeyId)
-                .filter(id -> id != 0);
+        Optional<Integer> existingId = keyDictionaryRepository.findById(compositeKey).map(KeyDictionaryEntry::getKeyId);
         if (existingId.isPresent()) {
             return cacheAndReturn(strKey, existingId.get());
         }
@@ -64,13 +62,12 @@ public class JpaKeyDictionaryDao extends JpaAbstractDaoListeningExecutorService 
                 return fromCache;
             }
             Integer keyId = keyDictionaryRepository.upsertAndGetKeyId(strKey);
-            if (keyId != null && keyId != 0) {
+            if (keyId != null) {
                 return cacheAndReturn(strKey, keyId);
             }
             log.warn("upsertAndGetKeyId returned: [{}] for key: [{}], falling back to findById", keyId, strKey);
             keyId = keyDictionaryRepository.findById(compositeKey)
                     .map(KeyDictionaryEntry::getKeyId)
-                    .filter(id -> id != 0)
                     .orElseThrow(() -> new IllegalStateException(
                             "Failed to resolve keyId for string key: " + strKey + " after fallback."));
             return cacheAndReturn(strKey, keyId);


dao/src/main/java/org/thingsboard/server/dao/sqlts/dictionary/KeyDictionaryRepository.java
```diff
@@ -31,4 +32,7 @@ public interface KeyDictionaryRepository extends JpaRepository<KeyDictionaryEntr
     @Query("SELECT e FROM KeyDictionaryEntry e ORDER BY e.keyId ASC")
     Page<KeyDictionaryEntry> findAll(Pageable pageable);
 
-}
\ No newline at end of file
+    @Query(value = "INSERT INTO key_dictionary (key) VALUES (:key) ON CONFLICT (key) DO UPDATE SET key = EXCLUDED.key RETURNING key_id", nativeQuery = true)
```
ShvaykaD, line 35: @ViacheslavKlimov we need to use `EXCLUDED` here, because `ON CONFLICT DO NOTHING` returns no `key_id` in the conflict case and would require an extra SELECT, while the `EXCLUDED` no-op update guarantees we always get the `key_id` back in a single atomic statement.

dao/src/main/java/org/thingsboard/server/dao/sqlts/dictionary/JpaKeyDictionaryDao.java
```diff
@@ -48,43 +46,32 @@ public class JpaKeyDictionaryDao extends JpaAbstractDaoListeningExecutorService
...
-                    }
-                } finally {
-                    creationLock.unlock();
+        Integer cached = keyDictionaryMap.get(strKey);
+        if (cached != null) {
+            return cached;
+        }
+        creationLock.lock();
+        try {
+            Integer keyId = keyDictionaryMap.get(strKey);
```
ViacheslavKlimov: We should findById first, then if null - lock and try to insert.

dao/src/main/java/org/thingsboard/server/dao/sqlts/dictionary/JpaKeyDictionaryDao.java
```diff
@@ -48,43 +46,40 @@ public class JpaKeyDictionaryDao extends JpaAbstractDaoListeningExecutorService
...
+            if (keyId == null || keyId == 0) {
+                log.warn("upsertAndGetKeyId returned: [{}] for key: [{}], falling back to findById", keyId, strKey);
+                entryOpt = keyDictionaryRepository.findById(compositeKey);
+                if (entryOpt.isEmpty() ||
+                    entryOpt.get().getKeyId() == null ||
+                    entryOpt.get().getKeyId() == 0) {
+                    throw new IllegalStateException("Failed to resolve keyId for string key: " + strKey + " after fallback. keyId: " + keyId);
+                }
             }
+            keyDictionaryMap.put(strKey, keyId);
```
ViacheslavKlimov: You forgot to update `keyId` after falling back to `findById`. `keyId` at this point may be null or zero

PR #14553, opened 2025-12-09
[Transport] Closing device sessions after tenant deletion

dashevchenko:
## Pull Request description

Closing device sessions after tenant deletion

## General checklist

## Back-End feature checklist


common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportService.java
```diff
@@ -966,6 +968,18 @@ protected void processToTransportMsg(ToTransportMsg toSessionMsg) {
         }
     }
 
+    private void onTenantDeleted(TenantId tenantId) {
+        sessions.forEach((id, md) -> {
+            TenantId sessionTenantId = getTenantId(md.getSessionInfo());
+            if (sessionTenantId.equals(tenantId)) {
+                DeviceId sessionDeviceId = getDeviceId(md.getSessionInfo());
+                transportCallbackExecutor.submit(() -> {
+                    md.getListener().onDeviceDeleted(sessionDeviceId);
```
ViacheslavKlimov: Some handlers for onDeviceDeleted call `checkLimits(...)` which adds rate limits to rateLimitService cache.
Which was supposed to be cleaned by rateLimitService.remove(tenantId);

ViacheslavKlimov: Don't you see it passes onDeviceDeleted to executor? What is the point of moving partly async onDevicesDeleted before the `rateLimitService.remove(..)`?

ViacheslavKlimov:
Please make sure that **NO** unnecessary calls to partition service, transport service, rate limit service, etc. are made when handling the tenant deletion and closing device sessions.

I would also like to see tests that verify that.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportService.java
ViacheslavKlimov: Also, moving the onDevicesDeleted **before** the cleanup of partition service - will cause some messages to be pushed to cores and rule engines

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -1598,6 +1604,12 @@ public void onDeviceDeleted(DeviceId deviceId) {
         }
     }
 
+    @Override
+    public void onTenantDeleted(DeviceId deviceId) {
+        deviceSessionCtx.setSessionCloseReason(SessionCloseReason.TENANT_DELETED);
+        onDeviceDeleted(deviceId);
```
ViacheslavKlimov, line 1633: `onTenantDeleted` delegates to `onDeviceDeleted`, whose first line is `context.onAuthFailure(address)` → `rateLimitService.onAuthFailure(address)`. So deleting a tenant increments the per-IP wrong-credentials counter for every connected device/gateway of that tenant, and once it crosses `max_wrong_credentials_per_ip` that IP gets blocked. With many devices behind one NAT/public IP this can block connectivity for *other* tenants' devices sharing the IP. Tenant deletion is an admin action, not a credential failure, so reusing `onDeviceDeleted` here looks unintended — and it contradicts the PR's stated goal of not touching the rate-limit service during tenant deletion. Consider gating the `onAuthFailure` call (e.g. skip it when `!deviceSessionCtx.shouldNotifyCore()`) or extracting the channel-close logic from `onDeviceDeleted`. It's currently harmless only because `transport.rate_limits.ip_limits_enabled` defaults to false.

common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
```diff
@@ -252,4 +259,11 @@ public void givenMqttMessage_whenDeviceProfileMqttTransport_thenTopicAddedToMeta
         verify(transportService, times(1)).process(any(), (TransportProtos.PostTelemetryMsg) any(), eq(expectedMd), any());
     }
 
+    @Test
+    public void givenOnTenantDeleted_thenNoMessagesToCore() {
+        handler.deviceSessionCtx.setChannel(ctx);
+        handler.onTenantDeleted(new DeviceId(UUID.randomUUID()));
+        verifyNoInteractions(transportService, transportRateLimitService);
```
ViacheslavKlimov, line 269: `verifyNoInteractions(transportRateLimitService)` passes here only because `context` is a `@Mock`: `onDeviceDeleted` calls `context.onAuthFailure(address)`, and the mock swallows it rather than forwarding to the real `transportRateLimitService`. In production `MqttTransportContext.onAuthFailure` *does* call `rateLimitService.onAuthFailure(...)`, so this assertion gives false confidence that tenant deletion produces no rate-limit interaction. To actually pin the intended behavior, either stub `context.onAuthFailure` to delegate to the spy, or assert on the real context's rate-limit service.

ViacheslavKlimov (commented):
## Review summary

Reviewed 23 changed files in [[Transport] Closing device sessions after tenant deletion](https://github.com/thingsboard/thingsboard/pull/14553). The core approach is sound: a new `SessionCloseReason.TENANT_DELETED` flag flows into each transport's `shouldNotifyCore()` check so the `SESSION_EVENT_MSG_CLOSED` / lifecycle / sparkplug messages are skipped, and the new `TenantNotFoundException` catch in `sendToCore` acts as a safety net for any in-flight message after `partitionService.removeTenant`. Local `deregisterSession` (just removes from the `sessions` map) and the `DeviceDeletedEvent` (only does `clients.remove` in the CoAP context) are correctly still invoked. Left 10 comments inline.

On the specific ask — *verify NO unnecessary calls to partition / transport / rate-limit service on tenant deletion* — the gating is correct for the `SESSION_EVENT_MSG_CLOSED`, SNMP `lifecycleEvent`, CoAP subscription-cancel and sparkplug-OFFLINE paths, but I found **two paths that still reach those services**:

- **MQTT**: `onTenantDeleted` → `onDeviceDeleted` → `context.onAuthFailure(address)` → `rateLimitService.onAuthFailure(...)`. This increments the per-IP wrong-credentials counter on tenant deletion (and can block a shared IP). The unit test misses it because `context` is mocked; the integration test misses it because it only asserts on `checkLimits`.
- **LwM2M**: `doUnReg` still calls `logService.log(...)` which sends a telemetry `PostTelemetryMsg` to core via `transportService.process` → `sendToCore` (now swallowed by the new `TenantNotFoundException` catch, but still an unnecessary transport/partition call + a WARN per device).

The rest are architecture/cleanliness items (duplicated `shouldNotifyCore`, unused enum constant, non-volatile shared field, a stray whitespace change) and a possible NPE in the LwM2M path.

---

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/uplink/DefaultLwM2mUplinkMsgHandler.java
```diff
@@ -447,6 +448,13 @@ public void onDeviceDelete(DeviceId deviceId) {
         clearAndUnregister(clientContext.getClientByDeviceId(deviceId.getId()));
     }
 
+    @Override
+    public void onTenantDelete(DeviceId deviceId) {
+        LwM2mClient client = clientContext.getClientByDeviceId(deviceId.getId());
+        client.setSessionCloseReason(SessionCloseReason.TENANT_DELETED);
```
ViacheslavKlimov: `getClientByDeviceId` returns `null` (`...orElse(null)`), so `client.setSessionCloseReason(...)` can throw an NPE. This is reachable: `onTenantDeleted` is dispatched asynchronously per session on `transportCallbackExecutor`, and the client may have deregistered between the `sessions` snapshot and this lookup. `onDeviceDelete` above passes the possibly-null result straight into `clearAndUnregister` (which would also NPE, but only inside the lock), whereas here you dereference it one line earlier. Add a null guard before setting the close reason.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/uplink/DefaultLwM2mUplinkMsgHandler.java
```diff
@@ -293,7 +294,7 @@ private void doUnReg(Registration registration, LwM2mClient client) {
             clientContext.unregister(client, registration);
             SessionInfoProto sessionInfo = client.getSession();
             if (sessionInfo != null) {
-                sessionManager.deregister(sessionInfo);
+                sessionManager.deregister(sessionInfo, client.shouldNotifyCore());
```
ViacheslavKlimov, line 299: Gating `deregister` on `shouldNotifyCore()` is correct, but a few lines up (start of `doUnReg`) the method still calls `logService.log(client, ...)` with the client-unregistration message, which sends a `PostTelemetryMsg` to core via `transportService.process` → `sendToCore`. On tenant deletion that `partitionService.resolve` now throws `TenantNotFoundException` and is swallowed — but it's still an unnecessary transport + partition-service call and produces a WARN log line per LwM2M device. To fully meet the PR's goal of not messaging core during tenant deletion, skip this log too when `!client.shouldNotifyCore()`.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/session/DeviceAwareSessionContext.java
```diff
@@ -87,6 +90,10 @@ public void setDisconnected() {
         this.connected = false;
     }
 
+    public boolean shouldNotifyCore() {
+        return sessionCloseReason == null || sessionCloseReason != SessionCloseReason.TENANT_DELETED;
```
ViacheslavKlimov: `sessionCloseReason == null || sessionCloseReason != SessionCloseReason.TENANT_DELETED` simplifies to just `sessionCloseReason != SessionCloseReason.TENANT_DELETED` — when the field is null the `!=` is already true, so the null check is dead code. This exact method is also duplicated verbatim in `TbCoapClientState` and `LwM2mClient`. Consider moving the predicate onto the `SessionCloseReason` enum (e.g. a `notifiesCore()` method) so the three copies can't drift apart.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/session/DeviceAwareSessionContext.java
```diff
@@ -50,6 +50,9 @@ public abstract class DeviceAwareSessionContext implements SessionContext {
     @Getter
     @Setter
     protected volatile TransportProtos.SessionInfoProto sessionInfo;
+    @Getter
+    @Setter
+    protected SessionCloseReason sessionCloseReason;
```
ViacheslavKlimov: The sibling fields right above (`sessionInfo`, `connected`) are declared `volatile`, but `sessionCloseReason` is a plain field. For MQTT it's written on the `transportCallbackExecutor` thread (via `onTenantDeleted`) and later read on the Netty event-loop thread inside `doDisconnect`/`shouldNotifyCore`. The Netty channel-close hand-off most likely provides the happens-before edge, but since the field is shared across threads and its neighbors are volatile, marking it `volatile` would make the intent explicit and remove the ambiguity.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/session/SessionCloseReason.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.common.transport.session;
+
+public enum SessionCloseReason {
+
+    DEFAULT, TENANT_DELETED
```
ViacheslavKlimov: `DEFAULT` isn't referenced anywhere — the code only ever sets/checks `TENANT_DELETED` and treats `null` as the default. Either drop `DEFAULT`, or actually use it (e.g. initialize `sessionCloseReason` to `DEFAULT` instead of leaving it null) so the enum and the `shouldNotifyCore` checks are consistent.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
```diff
@@ -227,7 +227,7 @@ public void onDeviceDeleted(String deviceName) {
         deregisterSession(deviceName);
     }
 
-    public String getNodeId() {
+   public String getNodeId() {
```
ViacheslavKlimov: Accidental whitespace change — this line lost one leading space (3 instead of 4) and is unrelated to the PR. Worth reverting to keep the diff clean and avoid a spurious indentation.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/LwM2mSessionMsgListener.java
```diff
@@ -114,4 +114,10 @@ public void onDeviceDeleted(DeviceId deviceId) {
         log.trace("[{}] Device on delete", deviceId);
         this.handler.onDeviceDelete(deviceId);
     }
+
+    @Override
+    public void onTenantDeleted(DeviceId deviceId) {
+        log.trace("[{}] Device on delete", deviceId);
```
ViacheslavKlimov: The log message is copy-pasted from `onDeviceDeleted` — it reads `Device on delete` but this is the tenant-deleted path. Minor, but it'll be confusing when tracing logs during a tenant deletion; consider `Tenant on delete` or similar.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportService.java
```diff
@@ -966,6 +968,18 @@ protected void processToTransportMsg(ToTransportMsg toSessionMsg) {
         }
     }
 
+    public void onTenantDeleted(TenantId tenantId) {
```
ViacheslavKlimov: Two small things on this new method. (1) It's `public` while its sibling `onDeviceDeleted` is `private` and it's only called internally (line 935) — `private` seems more appropriate unless something outside needs it. (2) Unlike `onDeviceDeleted`, which publishes `DeviceDeletedEvent` unconditionally after the loop, here the event is only published for devices that currently have an entry in `sessions`. The CoAP context keeps `TbCoapClientState` in its `clients` map even without an active session, and that map is cleaned only via this `DeviceDeletedEvent`; on tenant deletion no per-device delete message is sent, so such states would never be reached and could linger. Worth confirming that's acceptable, or iterate clients independently for CoAP.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
dashevchenko: Fixed. Calling onAuthFailure only when needed.

common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
dashevchenko: Fixed. Verifying that no onAuthFailure was invoked.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/uplink/DefaultLwM2mUplinkMsgHandler.java
dashevchenko: Fixed. Added null check

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/uplink/DefaultLwM2mUplinkMsgHandler.java
dashevchenko: Fixed. Calling logService.log only when needed.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/session/DeviceAwareSessionContext.java
dashevchenko: Fixed. Simplified expression.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/session/DeviceAwareSessionContext.java
dashevchenko: Fixed. Added volatile.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/session/SessionCloseReason.java
dashevchenko: Fixed. Deleted the DEFAULT enum option.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
dashevchenko: Fixed.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/LwM2mSessionMsgListener.java
dashevchenko: Fixed.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportService.java
dashevchenko: Fixed. Made the method private. Also added publishing of TenantDeletedEvent to do right clear in CoAP. DeviceDeletedEvent became redundant.

dashevchenko (commented):
## Re-review summary

Re-reviewed [[Transport] Closing device sessions after tenant deletion](https://github.com/thingsboard/thingsboard/pull/14553) — verified 10 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/14553#pullrequestreview-4427829436).

| Status | Count |
|--------|-------|
| ✅ Resolved | 10 |
| 💬 Acknowledged | 0 |

All previous findings have been addressed. The two functional gaps (MQTT `onAuthFailure` → rate-limit counter, and LwM2M `logService.log` → core) are now correctly gated on `shouldNotifyCore()`, and the lingering-CoAP-state concern is solved cleanly by publishing a `TenantDeletedEvent` that `DefaultCoapClientContext` consumes to evict client states by tenant id. The `shouldNotifyCore` predicate was consolidated onto the `SessionCloseReason` enum, so the three former copies no longer duplicate it.

### Finding details

- ✅ **MqttTransportHandler.java** — `onAuthFailure` was called on tenant deletion, incrementing the per-IP wrong-credentials counter — *Fixed: gated behind `shouldNotifyCore()`; `onTenantDeleted` sets `TENANT_DELETED` before delegating.*
- ✅ **MqttTransportHandlerTest.java** — assertion passed only because the mocked `context` swallowed `onAuthFailure` — *Fixed: new `givenOnTenantDeleted_thenNoMessagesToCore` asserts `onAuthFailure` is never called and `verifyNoInteractions(transportService, transportRateLimitService)`, plus a positive `givenOnDeviceDeleted_thenOnAuthFailureReported`.*
- ✅ **DefaultLwM2mUplinkMsgHandler.java** — possible NPE dereferencing a null client in the tenant-deleted path — *Fixed: `onTenantDelete` now null-checks the client before use.*
- ✅ **DefaultLwM2mUplinkMsgHandler.java** — `doUnReg` still sent a telemetry log to core on tenant deletion — *Fixed: `logService.log` and `sessionManager.deregister(..., notifyCore)` now gated on `client.shouldNotifyCore()`.*
- ✅ **DeviceAwareSessionContext.java** — dead null check in the predicate and the predicate duplicated across three classes — *Fixed: simplified to `reason != TENANT_DELETED` and extracted to `SessionCloseReason.shouldNotifyCore(...)`; `DeviceAwareSessionContext`, `TbCoapClientState` and `LwM2mClient` all delegate to it.*
- ✅ **DeviceAwareSessionContext.java** — `sessionCloseReason` shared across threads but not volatile — *Fixed: field marked `volatile`.*
- ✅ **SessionCloseReason.java** — unused `DEFAULT` constant — *Fixed: removed; only `TENANT_DELETED` remains.*
- ✅ **AbstractGatewaySessionHandler.java** — stray whitespace/indentation change — *Fixed: indentation reverted to 4 spaces.*
- ✅ **LwM2mSessionMsgListener.java** — copy-pasted `Device on delete` log on the tenant path — *Fixed: now logs `Tenant on delete`.*
- ✅ **DefaultTransportService.java** — `onTenantDeleted` was `public`, and CoAP per-device states with no active session could linger — *Fixed: made `private`; now publishes `TenantDeletedEvent`, and `DefaultCoapClientContext` evicts states via `clients.values().removeIf(state -> event.getTenantId().equals(state.getTenantId()))`.*

### Fix-commit notes

Reviewed the new code (the `TenantDeletedEvent` flow, the CoAP event listener, the consolidated enum predicate) for regressions — no functional issues found. `TbCoapClientState.getTenantId()` can return null, but the listener calls `event.getTenantId().equals(state.getTenantId())` with the non-null event tenant on the left, so there's no NPE. One trivial nit: `MqttTransportHandlerTest.java` no longer ends with a trailing newline (`\ No newline at end of file`) — worth re-adding to keep the file clean.

---

ViacheslavKlimov (commented):
## Review summary

Reviewed 24 changed files in [[Transport] Closing device sessions after tenant deletion](https://github.com/thingsboard/thingsboard/pull/14553). Left 8 comment(s) inline.

The approach — a `SessionCloseReason.TENANT_DELETED` flag gated by `shouldNotifyCore()`, a `TenantDeletedEvent` to purge per-device state, and a `TenantNotFoundException` guard in `sendToCore` — is sound and applied consistently across all transports (MQTT, CoAP, LwM2M, SNMP, HTTP, gateway). I didn't find a hard functional bug. The comments are mostly about the design of the suppression mechanism (single-value enum, implicit close-reason side effects, boolean flag), test-coverage gaps (SNMP/LwM2M/HTTP untested; the CoAP unit test doesn't actually validate the linkage), and one potential race on the gateway sub-device cleanup path.

---

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
```diff
@@ -802,12 +802,14 @@ protected void ackOrClose(int msgId, AtomicBoolean ackSent) {
...
 
     private void deregisterSession(String deviceName, MqttDeviceAwareSessionContext deviceSessionCtx) {
-        if (this.deviceSessionCtx.isSparkplug()) {
+        if (this.deviceSessionCtx.isSparkplug() && deviceSessionCtx.shouldNotifyCore()) {
             sendSparkplugStateOnTelemetry(deviceSessionCtx.getSessionInfo(),
                     deviceSessionCtx.getDeviceInfo().getDeviceName(), OFFLINE, new Date().getTime());
         }
         transportService.deregisterSession(deviceSessionCtx.getSessionInfo());
-        transportService.process(deviceSessionCtx.getSessionInfo(), SESSION_EVENT_MSG_CLOSED, null);
+        if (deviceSessionCtx.shouldNotifyCore()) {
```
ViacheslavKlimov, line 810: For gateway sub-devices, two paths can deregister the session on tenant deletion: the sub-device's own `onTenantDeleted` callback (which sets `TENANT_DELETED` on its own context first), and the gateway's `doDisconnect → onDevicesDisconnect` (fired when the gateway channel closes), which deregisters sub-devices here without touching their close reason. These run on different threads (the transport callback executor vs. the Netty event loop), so if `onDevicesDisconnect` wins the `devices.remove()` race before the sub-device callback has set the reason, `shouldNotifyCore()` is still `true` and a `SESSION_EVENT_MSG_CLOSED` goes out for the just-deleted tenant — which reaches `checkLimits()` (the very thing this PR is trying to avoid) and could make `testGatewayShouldBeDisconnectedAfterTenantDeletion`'s `verify(rateLimitService, never())` flaky. The new `TenantNotFoundException` catch swallows the message so there's no functional fallout, but would it be more robust to set the close reason on the sub-device contexts up front (e.g. inside `onDevicesDisconnect`/`onGatewayDelete`) so the suppression is deterministic rather than timing-dependent?

common/transport/coap/src/test/java/org/thingsboard/server/transport/coap/CoapSessionListenerTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @BeforeEach
+    void setUp() {
+        coapClientContext = new DefaultCoapClientContext(config, transportContext, transportService, profileCache, partitionService);
+    }
+
+    @Test
+    void onTenantDeleted_shouldNotInteractWithCore() {
+        // GIVEN
+        TbCoapClientState state = mock(TbCoapClientState.class);
```
ViacheslavKlimov, line 61: Because `state` is a Mockito mock, `state.shouldNotifyCore()` returns the default `false` regardless of the `setSessionCloseReason(TENANT_DELETED)` call inside `onTenantDeleted` — the real method body never runs. So this test would still pass even if that `setSessionCloseReason(...)` line were deleted from `CoapSessionListener.onTenantDeleted`; it doesn't actually verify the reason→suppression linkage. For the same reason, the stubbed `getRpc()`/`getAttrs()` keep returning non-null after `setRpc(null)`/`setAttrs(null)` (no-ops on a mock), so `closeAndCleanup` — and therefore `deregisterSession` — is never reached either. A spy over a real `TbCoapClientState` (or relying on the CoAP integration test, which does exercise the real path) would give genuine coverage.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/session/SessionCloseReason.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.common.transport.session;
+
+public enum SessionCloseReason {
+
+    TENANT_DELETED;
+
+    public static boolean shouldNotifyCore(SessionCloseReason reason) {
```
ViacheslavKlimov, line 22: An enum with a single value plus a static helper that means "is this reason NOT the one value we have" is an awkward shape that won't age well. As soon as a second reason is added (say `RATE_LIMIT_EXCEEDED` that also shouldn't notify core), `reason != TENANT_DELETED` silently does the wrong thing and every call site keeps compiling. Would it be cleaner to make the policy intrinsic to the enum — e.g. a `boolean notifyCore` field per constant (or an abstract method) and an instance `shouldNotifyCore()`, with a static helper only to handle the null case? That keeps the per-reason decision next to the reason itself rather than in a comparison future authors must remember to update.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -1611,14 +1617,22 @@ public void onDeviceUpdate(TransportProtos.SessionInfoProto sessionInfo, Device
...
         ChannelHandlerContext ctx = deviceSessionCtx.getChannel();
         closeCtx(ctx, MqttReasonCodes.Disconnect.ADMINISTRATIVE_ACTION);
         if (gatewaySessionHandler != null) {
             gatewaySessionHandler.onGatewayDelete(deviceId);
         }
     }
 
+    @Override
+    public void onTenantDeleted(DeviceId deviceId) {
+        deviceSessionCtx.setSessionCloseReason(SessionCloseReason.TENANT_DELETED);
```
ViacheslavKlimov, line 1632: This "set a close-reason flag on the session context, then call `onDeviceDeleted` which reads it back via `shouldNotifyCore()`" pattern is repeated near-verbatim across five places (here, `AbstractGatewayDeviceSessionContext`, `DefaultCoapClientContext.CoapSessionListener`, SNMP's `DeviceSessionContext`, and the LwM2M path). The control flow is implicit — whether `onDeviceDeleted` suppresses the core notification now depends on hidden state mutated by a different method, so someone reading `onDeviceDeleted` in isolation can't see why `context.onAuthFailure`/`SESSION_EVENT_MSG_CLOSED` is sometimes skipped. Have you considered passing the reason explicitly (e.g. a shared `closeSession(deviceId, SessionCloseReason)` or an extra parameter) so the suppression is visible at the call site instead of via a side-effecting field?

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/session/LwM2MSessionManager.java
```diff
@@ -23,5 +23,6 @@ public interface LwM2MSessionManager {
 
     void deregister(TransportProtos.SessionInfoProto sessionInfo);
 
+    void deregister(TransportProtos.SessionInfoProto sessionInfo, boolean notifyCore);
```
ViacheslavKlimov, line 26: The new `boolean notifyCore` overload is a classic boolean-flag-at-the-call-site: `sessionManager.deregister(sessionInfo, client.shouldNotifyCore())` reads fine, but a bare `deregister(sessionInfo, false)` elsewhere would be opaque. The rest of this PR models the same concept as a `SessionCloseReason` enum — passing that reason (or null) here instead of a raw boolean would be self-documenting and consistent with the abstraction the PR introduces everywhere else. Minor, since there's only one caller today, but worth aligning the API with the enum the PR otherwise leans on.

application/src/test/java/org/thingsboard/server/transport/mqtt/mqttv3/client/MqttClientConnectionTest.java
```diff
@@ -45,4 +76,95 @@ public void testClientWithWrongAccessToken() throws Exception {
...
+        tenant.setTitle("Mqtt Client Connection Test Tenant");
+        Tenant savedTenant = saveTenant(tenant);
+
+        User tenantAdmin = new User();
+        tenantAdmin.setAuthority(Authority.TENANT_ADMIN);
+        tenantAdmin.setTenantId(savedTenant.getTenantId());
+        tenantAdmin.setEmail("mqttTestClient@thingsboard.org");
+        createUserAndLogin(tenantAdmin, TENANT_ADMIN_PASSWORD);
+
+        savedDevice = createDevice(RandomStringUtils.randomAlphabetic(10), "default", false);
```
ViacheslavKlimov, line 93: The two new tenant-deletion tests here share ~20 lines of identical setup (login sysadmin, build tenant, `saveTenant`, build `TENANT_ADMIN` user, `createUserAndLogin`, `createDevice`, fetch credentials, set `accessToken`), and the same create-tenant-plus-admin block is duplicated a third time in `CoapClientIntegrationTest`. The trailing assertion block (`clearInvocations`, `deleteTenant`, await sessions empty, await client disconnected, verify `checkLimits` never called) is also copy-pasted. Extracting a small helper (e.g. `createTenantWithAdminAndDevice(...)` and an `awaitSessionsClearedWithoutCoreCalls(...)`) would cut the boilerplate and let the actual behavior under test stand out. Also note the literal tenant title `"Mqtt Client Connection Test Tenant"` was copied verbatim into the CoAP test, where it's misleading.

common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/SnmpTransportContext.java
```diff
@@ -207,7 +207,9 @@ private void destroyDeviceSession(DeviceSessionContext sessionContext) {
         transportService.deregisterSession(sessionContext.getSessionInfo());
         snmpTransportService.cancelQueryingTasks(sessionContext);
         sessions.remove(sessionContext.getDeviceId());
-        transportService.lifecycleEvent(sessionContext.getTenantId(), sessionContext.getDeviceId(), ComponentLifecycleEvent.STOPPED, true, null);
+        if (sessionContext.shouldNotifyCore()) {
```
ViacheslavKlimov, line 210: The new `shouldNotifyCore()` guards in the SNMP (`lifecycleEvent` suppression here) and LwM2M (`doUnReg` / `deregister(..., notifyCore)`) paths, plus the HTTP `onTenantDeleted`, have no test coverage — only MQTT and CoAP got tests (integration + unit). There's established integration-test infrastructure for both transports (`SnmpTestV2`/`V3`, `AbstractLwM2MIntegrationTest`), so exercising tenant deletion through them looks achievable rather than needing a unit fallback. Since the PR adds parallel suppression logic to each transport, asserting at least one path per transport (SNMP `STOPPED` lifecycle not sent, LwM2M deregister not notifying core) would keep a future change to one transport's guard from silently regressing.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/TenantDeletedEvent.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.thingsboard.server.common.transport;
+
+import lombok.Getter;
+import org.thingsboard.server.common.data.id.TenantId;
+import org.thingsboard.server.queue.discovery.event.TbApplicationEvent;
+
+public final class TenantDeletedEvent extends TbApplicationEvent {
+
+    private static final long serialVersionUID = -3801673609913423334L;
```
ViacheslavKlimov, line 24: Small consistency nit: the sibling `DeviceDeletedEvent` (which this is otherwise a faithful copy of) annotates `serialVersionUID` with `@Serial` and puts a blank line before the `@Getter` field; this one drops the `@Serial` annotation and the spacing. Worth matching the sibling since they're clearly meant to be parallel.

PR #14618, opened 2025-12-16
Refactor Edge related classes: session, service, server

MazurenkoNick:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/PostgresBasedEdgeGrpcSessionManager.java
+++ b/application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/PostgresBasedEdgeGrpcSessionManager.java
@@ -46,7 +46,6 @@ public class PostgresBasedEdgeGrpcSessionManager extends AbstractEdgeGrpcSession
 
     private final EdgeSessionsHolder edgeSessions;
 
-    // todo: verify handled properly
     private final Lock newEventsLock = new ReentrantLock();
     private final AtomicReference<ScheduledFuture<?>> edgeEventCheckFutureRef = new AtomicReference<>();
     private PostgresGeneralEdgeEventsDispatcher generalEdgeEventsDispatcher;


application/src/main/java/org/thingsboard/server/service/edge/rpc/session/EdgeSessionsHolder.java
```diff
@@ -0,0 +1,79 @@
...
+@ConditionalOnProperty(prefix = "edges", value = "enabled", havingValue = "true")
+@TbCoreComponent
+public class EdgeSessionsHolder {
+
+    private final ConcurrentMap<EdgeId, EdgeGrpcSessionManager> sessions = new ConcurrentHashMap<>();
+    private final ConcurrentMap<UUID, EdgeGrpcSessionManager> sessionsById = new ConcurrentHashMap<>();
+
+    public void forEach(Consumer<EdgeGrpcSessionManager> consumer) {
+        sessions.values().forEach(consumer);
+        sessionsById.values().forEach(consumer);
```
ViacheslavKlimov: `forEach` iterates over both `sessions.values()` and `sessionsById.values()`. Since `put()` stores the same `EdgeGrpcSessionManager` instance in both maps, the consumer is invoked **twice** per session.

This is called from `EdgeGrpcService.preDestroy()` as `sessions.forEach(EdgeGrpcSessionManager::onEdgeDisconnect)`, so every session gets `onEdgeDisconnect()` called twice on shutdown. For `PostgresBasedEdgeGrpcSessionManager` this means double `cancelScheduleEdgeEventsCheck()` calls; for Kafka it's currently a no-op, but the pattern is fragile.

Should this iterate only one of the maps (e.g., `sessions.values()`)?

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/EdgeSessionsHolder.java
```diff
@@ -0,0 +1,79 @@
...
+        return sessionsById.remove(sessionId);
+    }
+
+    public void remove(EdgeGrpcSessionManager session) {
+        if (session == null) {
+            log.warn("Can't remove session from holder because it's null");
+            return;
+        }
+        EdgeSessionState sessionState = session.getState();
+        removeByEdgeId(sessionState.getEdge().getId()); // todo: react to warnings
```
ViacheslavKlimov: `sessionState.getEdge().getId()` will throw NPE if `getEdge()` returns null (e.g., session removed before `processConnect` completes). `EdgeSessionState.getEdgeId()` already handles this with a null check — consider using it here instead.

Also, there's a `// todo: react to warnings` comment left in production code.

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/AbstractEdgeGrpcSessionManager.java
```diff
@@ -0,0 +1,140 @@
...
+            return;
+        }
+        EdgeUpdateMsg edgeConfig = EdgeUpdateMsg.newBuilder()
+                .setConfiguration(EdgeMsgConstructorUtils.constructEdgeConfiguration(edge)).build();
+
+        ResponseMsg edgeConfigMsg = ResponseMsg.newBuilder()
+                .setEdgeUpdateMsg(edgeConfig)
+                .build();
+
+        session.sendDownlinkMsg(edgeConfigMsg);
```
ViacheslavKlimov, line 122: If `session` is null, `getState()` returns null on line 103, the `stateCustomerId` stays null, the guard on line 109 is skipped, and execution reaches `session.sendDownlinkMsg(edgeConfigMsg)` here — which throws NPE.

Should there be an early return when `state` is null (or `session` is null)?

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/AbstractEdgeGrpcSessionManager.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public void destroyAndMarkAsZombieIfFailed() {
+        EdgeSessionState state = getState();
+        final var finalSession = session;
+
+        try (finalSession) {
+            if (!destroy()) {
+                log.warn("[{}][{}] Session destroy failed for edge [{}] with session id [{}]. Adding to zombie queue for later cleanup.",
```
ViacheslavKlimov: If `getState()` returns null (session not yet initialized when destroy is triggered), `state.getTenantId()`, `state.getEdgeId()`, and `state.getEdge().getName()` will all NPE. The same applies in the catch block at line 136-137. Consider guarding against null `state` here.

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/EdgeGrpcSession.java
```diff
@@ -0,0 +1,621 @@
...
+        try {
+            if (msg.getSuccess()) {
+                state.getPendingMsgsMap().remove(msg.getDownlinkMsgId());
+                ctx.getStatsCounterService().ifPresent(statsCounterService -> statsCounterService.recordEvent(EdgeStatsKey.DOWNLINK_MSGS_PUSHED, getTenantId(), getEdgeId(), 1));
+                log.debug("[{}][{}][{}] Msg has been processed successfully! Msg Id: [{}], Msg: {}", getTenantId(), getEdgeId(), getSessionId(), msg.getDownlinkMsgId(), msg);
+            } else {
+                log.debug("[{}][{}][{}] Msg processing failed! Msg Id: [{}], Error msg: {}", getTenantId(), getEdgeId(), getSessionId(), msg.getDownlinkMsgId(), msg.getErrorMsg());
+                DownlinkMsg downlinkMsg = state.getPendingMsgsMap().get(msg.getDownlinkMsgId());
+                // if NOT timeseries or attributes failures - ack failed downlink
+                if (downlinkMsg.getEntityDataCount() == 0) {
```
ViacheslavKlimov: `state.getPendingMsgsMap().get(msg.getDownlinkMsgId())` can return `null` if the message was already removed (e.g., by a concurrent `interruptPreviousSendDownlinkMsgsTask` clearing the map). The immediate `downlinkMsg.getEntityDataCount()` on the next line would then NPE. A null check is needed.

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/KafkaBasedEdgeGrpcSessionManager.java
```diff
@@ -0,0 +1,300 @@
...
+@Slf4j
+@TbKafkaComponent
+@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+@RequiredArgsConstructor
+public class KafkaBasedEdgeGrpcSessionManager extends AbstractEdgeGrpcSessionManager {
+
+    private static final int NO_INITIAL_DELAY_VALUE = 0;
+
+    private final TbCoreQueueFactory tbCoreQueueFactory;
+    private final DownlinkMessageMapper downlinkMessageMapper;
```
ViacheslavKlimov: This `downlinkMessageMapper` field shadows the `@Autowired protected DownlinkMessageMapper downlinkMessageMapper` in parent `AbstractEdgeGrpcSessionManager` (line 63). The child uses its own constructor-injected field (e.g., line 239), while the parent's `initInputStream` at line 84 uses the parent's autowired field. Since Spring injects the same singleton into both, this works accidentally, but it's confusing and fragile — if one injection path changes, they could diverge silently.

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/KafkaBasedEdgeGrpcSessionManager.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+
+    private void cancelMigrationAndProcessingInit() {
+        log.trace("[{}] cancelling edge migration & processing init for edge", getState().getEdgeId());
+        cancelFuture(initMigrationAndProcessingFutureRef);
+    }
+
+    private void cancelHighPriorityProcessing() {
+        log.trace("[{}] cancelling high priority processing task for edge", getState().getEdgeId());
```
ViacheslavKlimov: `getState().getEdgeId()` will NPE if `session` is null (making `getState()` return null). `destroy()` at line 101 calls `cancelHighPriorityProcessing()`, and `destroy()` can be called before `initInputStream()` has set up the session. The same issue exists in `cancelMigrationAndProcessingInit()` at line 255.

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/KafkaBasedEdgeGrpcSessionManager.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+
+    private void awaitConsumerTermination() {
+        EdgeSessionState state = getState();
+        try {
+            if (!consumerExecutor.awaitTermination(10, java.util.concurrent.TimeUnit.SECONDS)) {
+                consumerExecutor.shutdownNow(); // todo: verify it is acceptable
+            }
+        } catch (InterruptedException ie) {
```
ViacheslavKlimov, line 297: When `InterruptedException` is caught, the thread's interrupt status should be restored via `Thread.currentThread().interrupt()`. Swallowing the interrupt violates Java concurrency best practices and can cause callers higher up the stack to miss the interruption signal.

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/PostgresBasedEdgeGrpcSessionManager.java
```diff
@@ -0,0 +1,180 @@
...
+
+    public ListenableFuture<Boolean> processEdgeEvents() throws Exception {
+        if (generalEdgeEventsDispatcher == null) {
+            generalEdgeEventsDispatcher = new PostgresGeneralEdgeEventsDispatcher(session, ctx);
+        }
+        return generalEdgeEventsDispatcher.processNewEvents();
+    }
+
+    private void cancelScheduleEdgeEventsCheck() {
+        EdgeId edgeId = getState().getEdgeId();
```
ViacheslavKlimov: Same NPE risk as in the Kafka manager: `getState().getEdgeId()` will throw if `getState()` returns null. This is called from `destroy()` → `cancelScheduleEdgeEventsCheck()`, and `destroy()` can be triggered before session initialization.

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/DefaultZombieSessionCleanupService.java
```diff
@@ -0,0 +1,126 @@
...
+    @Autowired
+    private EdgeSessionsHolder edgeSessionsHolder;
+
+    private final Queue<EdgeGrpcSessionManager> zombieSessions;
+    private final ScheduledExecutorService zombieSessionsExecutorService;
+
+    public DefaultZombieSessionCleanupService() {
+        this.zombieSessions = new ConcurrentLinkedQueue<>();
+        this.zombieSessionsExecutorService = ThingsBoardExecutors.newSingleThreadScheduledExecutor("zombie-sessions");
+        this.zombieSessionsExecutorService.scheduleAtFixedRate(this::cleanupZombieSessions, 30, 60, TimeUnit.SECONDS);
```
ViacheslavKlimov, line 58: The `scheduleAtFixedRate` call is in the constructor, but `edgeSessionsHolder` is injected via `@Autowired` field injection, which happens *after* the constructor. If the 30-second initial delay expires before Spring finishes wiring (unlikely but not guaranteed), `edgeSessionsHolder` will be null in `cleanupZombieSessions()`. Moving the schedule to a `@PostConstruct` method would be safer.

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/DefaultZombieSessionCleanupService.java
```diff
@@ -0,0 +1,126 @@
...
+                            state.getTenantId(), state.getEdgeId(), state.getSessionId(), state.getEdge().getName());
+                    return true;
+                } else {
+                    log.warn("[{}][{}] Failed to remove zombie session [{}] for edge [{}].",
+                            state.getTenantId(), state.getEdgeId(), state.getSessionId(), state.getEdge().getName());
+                    return false;
+                }
+            });
+        } catch (Exception e) {
+            log.warn("Failed to cleanup kafka sessions", e);
```
ViacheslavKlimov: The log message says `"Failed to cleanup kafka sessions"` but this service handles both Kafka and Postgres session managers. The message is misleading in non-Kafka deployments.

application/src/main/java/org/thingsboard/server/service/edge/rpc/GrpcServer.java
```diff
@@ -0,0 +1,131 @@
...
+        if (sslEnabled) {
+            setupSsl(serverBuilder);
+        }
+        server = serverBuilder.build();
+        log.info("Going to start Edge RPC server using port: {}", rpcPort);
+        try {
+            server.start();
+        } catch (IOException e) {
+            log.error("Failed to start Edge RPC server!", e);
+            throw new RuntimeException("Failed to start Edge RPC server!");
```
ViacheslavKlimov: The `RuntimeException` doesn't include the original `IOException e` as the cause. This loses the root cause in the exception chain, making it harder to debug startup failures. Should be:
```java
throw new RuntimeException("Failed to start Edge RPC server!", e);
```

application/src/main/java/org/thingsboard/server/service/edge/rpc/service/EdgeGrpcService.java
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    private void onEdgeDisconnect(Edge edge, UUID sessionId) {
+        TenantId tenantId = edge.getTenantId();
+        EdgeId edgeId = edge.getId();
+        log.info("[{}][{}] edge disconnected!", edgeId, sessionId);
+        EdgeGrpcSessionManager current = sessions.getByEdgeId(edgeId);
+        if (current != null && current.getState().getSessionId().equals(sessionId)) {
+            EdgeGrpcSessionManager toRemove = sessions.removeByEdgeId(edgeId);
+            toRemove.onEdgeDisconnect();
```
ViacheslavKlimov: Between the check on line 222 (`current != null && ...equals(sessionId)`) and the `removeByEdgeId` call here, another thread could concurrently remove the same `edgeId` (e.g., via `deleteEdge`). If `removeByEdgeId` returns null, `toRemove.onEdgeDisconnect()` will NPE. A null check on `toRemove` would be safer.

application/src/main/java/org/thingsboard/server/service/edge/rpc/service/EdgeGrpcService.java
```diff
@@ -0,0 +1,390 @@
...
+                    .scope(AttributeScope.SERVER_SCOPE)
+                    .entry(new BooleanDataEntry(key, value))
+                    .callback(new EdgeAttributeSaveCallback(tenantId, edgeId, key, value))
+                    .build());
+        }
+    }
+
+    private void shutdownExecutorSafely(ExecutorService sendDownlinkExecutorService) {
+        if (sendDownlinkExecutorService != null && !sendDownlinkExecutorService.isShutdown()) {
+            sendDownlinkExecutorService.shutdown();
```
ViacheslavKlimov: The old `EdgeGrpcService.destroy()` used `shutdownNow()` for all executor services. The new `shutdownExecutorSafely` uses `shutdown()` (graceful), which means the application will wait for all pending scheduled tasks (sync timeouts, downlink sends) to finish before the JVM exits. Was this change intentional? If not, consider `shutdownNow()` to match the previous behavior.

Also, the parameter name `sendDownlinkExecutorService` shadows the class field of the same name, which is a bit confusing.

MazurenkoNick: `shutdown()` is fine here. The tasks are short-lived (send a message, schedule a timeout), and `preDestroy` already calls `onEdgeDisconnect` on all sessions which sets `connected=false`, so any running send tasks will bail out early on the `if (!state.isConnected())` check. There's no real stall risk. `shutdownNow()` would also work but gains little and adds interrupt noise in the logs.

MazurenkoNick:
General comment regarding all warning about NPE on `EdgeGrpcSessionManager#getState()`:

The safeguard checks for NPE were added even though it is practically impossible for getState() to throw an NPE, except in the case where the method is executed before sessionManager.initInputStream(...)

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/DefaultZombieSessionCleanupService.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public void add(EdgeGrpcSessionManager session) {
+        zombieSessions.add(session);
+    }
+
+    private void cleanupZombieSessions() {
+        try {
+            tryToDestroyZombieSessions(getZombieSessions(edgeSessionsHolder.getSessions().values()),
+                    s -> edgeSessionsHolder.removeByEdgeId(s.getState().getEdge().getId()));
```
ViacheslavKlimov, line 76: This uses `s.getState().getEdge().getId()` — the same pattern that was just fixed in `EdgeSessionsHolder.remove()` (switched to `getEdgeId()` for null safety). The same applies at line 85, 89, and 125 (`toRemove.getState().getEdge().getId()`). Would be more consistent to use `getEdgeId()` here too:
```java
s -> edgeSessionsHolder.removeByEdgeId(s.getState().getEdgeId())
```

application/src/main/java/org/thingsboard/server/service/edge/rpc/session/manager/PostgresBasedEdgeGrpcSessionManager.java
```diff
@@ -0,0 +1,184 @@
...
+@Service
+@Slf4j
+@ConditionalOnExpression("'${queue.type:null}'!='kafka'")
+@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+@RequiredArgsConstructor
+public class PostgresBasedEdgeGrpcSessionManager extends AbstractEdgeGrpcSessionManager {
+
+    private final EdgeSessionsHolder edgeSessions;
+
+    // todo: verify handled properly
```
ViacheslavKlimov: Leftover TODO: `// todo: verify handled properly`. Should this be resolved or removed before merging?

PR #14762, opened 2026-01-07
Fixed notification requests and RPC cleanup timeout on large datasets

AndriiLandiak:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/test/java/org/thingsboard/server/service/ttl/NotificationsCleanUpServiceTest.java
+++ b/application/src/test/java/org/thingsboard/server/service/ttl/NotificationsCleanUpServiceTest.java
@@ -125,15 +125,20 @@ public class NotificationsCleanUpServiceTest {
     }
 
     @Test
-    public void testNoPartitionsDropped_skipsRequestCleanup() {
+    public void testNoPartitionsDropped_stillCleansUpRequests() {
         TopicPartitionInfo myPartition = TopicPartitionInfo.builder().topic("tb_core").myPartition(true).build();
         when(partitionService.resolve(any(), any(), any())).thenReturn(myPartition);
         when(partitioningRepository.dropPartitionsBefore(anyString(), anyLong(), anyLong()))
                 .thenReturn(0L);
 
+        when(notificationRequestDao.removeByTenantIdAndCreatedTimeBeforeBatch(eq(TenantId.SYS_TENANT_ID), anyLong(), eq(BATCH_SIZE)))
+                .thenReturn(0);
+        when(tenantService.findTenantsIds(any()))
+                .thenReturn(new PageData<>(List.of(), 0, 0, false));
+
         cleanUpService.cleanUp();
 
-        verify(notificationRequestDao, never()).removeByTenantIdAndCreatedTimeBeforeBatch(any(), anyLong(), anyInt());
+        verify(notificationRequestDao).removeByTenantIdAndCreatedTimeBeforeBatch(eq(TenantId.SYS_TENANT_ID), anyLong(), eq(BATCH_SIZE));
     }
 
 }


application/src/main/java/org/thingsboard/server/service/ttl/NotificationsCleanUpService.java
```diff
@@ -20,33 +20,43 @@
...
+import org.thingsboard.server.queue.util.TbCoreComponent;
 
+import java.util.Date;
 import java.util.concurrent.TimeUnit;
 
 import static org.thingsboard.server.dao.model.ModelConstants.NOTIFICATION_TABLE_NAME;
 
+@Slf4j
 @Service
+@TbCoreComponent
```
ViacheslavKlimov: Rule engine doesn't persist notifications?

application/src/main/resources/thingsboard.yml
```diff
@@ -419,14 +419,16 @@ sql:
...
+      removal_batch_size: "${SQL_RPC_TTL_REMOVAL_BATCH_SIZE:10000}" # To delete outdated rpc not all at once but in batches
     audit_logs:
       enabled: "${SQL_TTL_AUDIT_LOGS_ENABLED:true}" # Enable/disable TTL (Time To Live) for audit log records
       ttl: "${SQL_TTL_AUDIT_LOGS_SECS:0}" # Disabled by default. The accuracy of the cleanup depends on the sql.audit_logs.partition_size
       checking_interval_ms: "${SQL_TTL_AUDIT_LOGS_CHECKING_INTERVAL_MS:86400000}" # Default value - 1 day
     notifications:
       enabled: "${SQL_TTL_NOTIFICATIONS_ENABLED:true}" # Enable/disable TTL (Time To Live) for notification center records
       ttl: "${SQL_TTL_NOTIFICATIONS_SECS:2592000}" # Default value - 30 days
-      checking_interval_ms: "${SQL_TTL_NOTIFICATIONS_CHECKING_INTERVAL_MS:86400000}" # Default value - 1 day
+      checking_interval_ms: "${SQL_TTL_NOTIFICATIONS_CHECKING_INTERVAL_MS:86400}" # Default value - 1 day
```
ViacheslavKlimov: Doesn't look like one day to me

application/src/main/resources/thingsboard.yml
```diff
@@ -419,6 +419,7 @@ sql:
     rpc:
       enabled: "${SQL_TTL_RPC_ENABLED:true}" # Enable/disable TTL (Time To Live) for rpc call records
       checking_interval: "${SQL_RPC_TTL_CHECKING_INTERVAL:7200000}" # Number of milliseconds. The current value corresponds to two hours
+      removal_batch_size: "${SQL_RPC_TTL_REMOVAL_BATCH_SIZE:10000}" # To delete outdated rpc not all at once but in batches
```
ViacheslavKlimov: Fix both comments to "Batch size for X records removal" or smth like that

application/src/main/java/org/thingsboard/server/service/ttl/NotificationsCleanUpService.java
```diff
@@ -63,9 +71,56 @@ public void cleanUp() {
...
+        try {
+            totalRemoved += cleanUpByTenant(TenantId.SYS_TENANT_ID, expirationTime);
+        } catch (Exception e) {
+            log.warn("Failed to clean up notification requests for sysadmin {}", TenantId.SYS_TENANT_ID, e);
+        }
+        // Clean up notification requests for tenants
+        PageDataIterable<TenantId> tenants = new PageDataIterable<>(tenantService::findTenantsIds, 10_000);
+        for (TenantId tenantId : tenants) {
+            try {
+                if (!isTenantPartitionMine(tenantId)) {
```
ViacheslavKlimov, line 93: Bug: In a multi-node cluster, the `cleanUpNotificationRequests` method is only reachable from the node that owns `SYS_TENANT_ID` (guarded by `isSystemTenantPartitionMine()` at line 65). This `isTenantPartitionMine(tenantId)` check will skip tenants whose partitions belong to other nodes — but no other node will ever run this cleanup code. Result: notification requests for those tenants are never cleaned up.

The old code did a single global `removeAllByCreatedTimeBefore(requestExpTime)` that cleaned ALL tenants' notification requests on the SYS node. The per-tenant partitioning check was likely added by analogy with `RpcCleanUpService`, but `RpcCleanUpService.cleanUp()` has no outer `isSystemTenantPartitionMine()` guard — every node independently cleans up its own tenants' RPCs.

Suggested fix: remove the `isTenantPartitionMine` check here, since only one node runs this code anyway. This preserves the original semantics (one node cleans everything) while gaining the batching benefit:

```java
for (TenantId tenantId : tenants) {
    try {
        int tenantRemoved = cleanUpByTenant(tenantId, expirationTime);
        // ...
    }
}
```

Alternatively, restructure so that notification request cleanup runs independently of partition dropping, allowing each node to clean its own tenants.

dao/src/main/java/org/thingsboard/server/dao/sql/rpc/RpcRepository.java
```diff
@@ -27,14 +27,19 @@
...
 
     Page<RpcEntity> findAllByTenantId(UUID tenantId, Pageable pageable);
 
     @Modifying
-    @Query(value = "DELETE FROM rpc WHERE tenant_id = :tenantId AND created_time < :expirationTime",
+    @Query(value = "DELETE FROM rpc WHERE id IN " +
+            "(SELECT id FROM rpc WHERE tenant_id = :tenantId AND created_time < :expirationTime LIMIT :batchSize)",
             nativeQuery = true)
-    int deleteOutdatedRpcByTenantId(@Param("tenantId") UUID tenantId, @Param("expirationTime") Long expirationTime);
+    int deleteOutdatedRpcByTenantIdBatch(@Param("tenantId") UUID tenantId,
```
ViacheslavKlimov, line 43: **Low-confidence observation:** Unlike `NotificationRequestRepository.deleteByTenantIdAndCreatedTimeBeforeBatch` which has `@Transactional @Modifying`, this method only has `@Modifying` without `@Transactional`. It works because `JpaRpcDao.deleteOutdatedRpcByTenantIdBatch` provides the transaction, but the inconsistency between the two repositories could be confusing. Consider adding `@Transactional` here for consistency.

application/src/test/java/org/thingsboard/server/service/ttl/NotificationsCleanUpServiceTest.java
```diff
@@ -0,0 +1,139 @@
...
+    @Test
+    public void testNoPartitionsDropped_skipsRequestCleanup() {
+        TopicPartitionInfo myPartition = TopicPartitionInfo.builder().topic("tb_core").myPartition(true).build();
+        when(partitionService.resolve(any(), any(), any())).thenReturn(myPartition);
+        when(partitioningRepository.dropPartitionsBefore(anyString(), anyLong(), anyLong()))
+                .thenReturn(0L);
+
+        cleanUpService.cleanUp();
+
+        verify(notificationRequestDao, never()).removeByTenantIdAndCreatedTimeBeforeBatch(any(), anyLong(), anyInt());
```
ViacheslavKlimov: Bug: `testNoPartitionsDropped_skipsRequestCleanup` is stale after the refactoring commit and will fail.

The refactoring in `NotificationsCleanUpService.cleanUp()` decoupled notification request cleanup from partition dropping — `cleanUpNotificationRequests()` is now called unconditionally. But this test still assumes the old behavior where cleanup was skipped when `dropPartitionsBefore` returned 0.

With the current code:
1. `isSystemTenantPartitionMine()` returns `true` (line 130 stubs all resolves to `myPartition`).
2. `cleanUpNotificationRequests()` is called → `cleanUpByTenant(SYS_TENANT_ID, ...)` calls `removeByTenantIdAndCreatedTimeBeforeBatch` (returns 0 by default).
3. `tenantService.findTenantsIds(...)` is not stubbed → returns `null` → `PageDataIterable` will throw `NullPointerException` during iteration.

Even if the NPE is somehow avoided, the `verify(..., never())` assertion at line 136 would fail because `removeByTenantIdAndCreatedTimeBeforeBatch` is now called for `SYS_TENANT_ID`.

This test should either be removed (since the "skip cleanup when no partitions dropped" behavior no longer exists) or rewritten to test the current behavior — e.g., verifying that request cleanup still runs even when no partitions are dropped.

ViacheslavKlimov (commented):
## Re-review summary

Re-reviewed [Fixed notification requests and RPC cleanup timeout on large datasets](https://github.com/thingsboard/thingsboard/pull/14762) — verified 2 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/14762#pullrequestreview-3908507726).

| Status | Count |
|--------|-------|
| ✅ Resolved | 2 |

All previous findings have been addressed. Also found 1 new issue in the fix commits, commented inline.

### Finding details

- ✅ **application/src/main/java/org/thingsboard/server/service/ttl/NotificationsCleanUpService.java:93** — Notification request cleanup unreachable for non-system tenants in multi-node clusters — *Fixed in code: restructured so each node independently cleans its own tenants.*
- ✅ **dao/src/main/java/org/thingsboard/server/dao/sql/rpc/RpcRepository.java:38** — Missing `@Transactional` annotation for consistency — *Fixed in code: `@Transactional` added.*

---
*This re-review was auto-generated by [Claude](https://claude.ai). Findings may contain errors — please verify before applying changes.*

PR #14775, opened 2026-01-09
Added OTA package data cleanup

AndriiLandiak:
## Pull Request description

https://github.com/thingsboard/thingsboard/pull/14901

merge this as well to master (schema-update.sql)

## General checklist

## Back-End feature checklist

### DIFF
--- a/dao/src/main/java/org/thingsboard/server/dao/ota/BaseOtaPackageService.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/ota/BaseOtaPackageService.java
@@ -234,10 +234,10 @@ public class BaseOtaPackageService extends AbstractCachedEntityService<OtaPackag
         try {
             Integer result = otaPackageDao.unlinkLargeObject(oid);
             if (result != 1) {
-                log.error("[{}][{}] Failed to delete large object (OID: {}). Result code: {}", tenantId, otaPackageId, oid, result);
+                log.warn("[{}][{}] Failed to delete large object (OID: {}). Result code: {}", tenantId, otaPackageId, oid, result);
             }
         } catch (Exception e) {
-            log.error("[{}][{}] Failed to delete large object (OID: {})", tenantId, otaPackageId, oid, e);
+            log.warn("[{}][{}] Failed to delete large object (OID: {})", tenantId, otaPackageId, oid, e);
         }
     }
 


common/data/src/main/java/org/thingsboard/server/common/data/housekeeper/HousekeeperTask.java
```diff
@@ -89,6 +95,10 @@ public static HousekeeperTask deleteJobs(TenantId tenantId, EntityId entityId) {
         return new HousekeeperTask(tenantId, entityId, HousekeeperTaskType.DELETE_JOBS);
     }
 
+    public static HousekeeperTask deleteOtaData(TenantId tenantId, OtaPackageId otaPackageId, Long oid) {
+        return new OtaDataDeletionHousekeeperTask(tenantId, otaPackageId, oid);
```
ViacheslavKlimov: Sorry, but doing it via the Housekeeper is an overkill. lo_unlink just marks the large object for deletion. We can do that right away after deleting the OTA, synchronously

dao/src/main/java/org/thingsboard/server/dao/ota/BaseOtaPackageService.java
```diff
@@ -196,7 +196,13 @@ public void deleteOtaPackage(TenantId tenantId, OtaPackageId otaPackageId) {
         log.trace("Executing deleteOtaPackage [{}]", otaPackageId);
         validateId(otaPackageId, id -> INCORRECT_OTA_PACKAGE_ID + id);
         try {
+            Long oid = getDataOidById(tenantId, otaPackageId);
             otaPackageDao.removeById(tenantId, otaPackageId.getId());
+            Integer result = otaPackageDao.unlinkLargeObject(oid);
+            if (result != 1) {
```
ViacheslavKlimov: Possible NPE

dao/src/main/java/org/thingsboard/server/dao/sql/ota/OtaPackageRepository.java
```diff
@@ -41,4 +41,12 @@ public interface OtaPackageRepository extends JpaRepository<OtaPackageEntity, UU
     @Query("SELECT r.id FROM OtaPackageEntity r WHERE r.tenantId = :tenantId")
     Page<UUID> findIdsByTenantId(@Param("tenantId") UUID tenantId, Pageable pageable);
 
+    // The 'data' column is of type OID (PostgreSQL large object reference), so it returns the OID as Long
+    @Query(value = "SELECT data FROM ota_package WHERE id = :id AND data IS NOT NULL", nativeQuery = true)
+    Long getDataOidById(@Param("id") UUID id);
+
+    @Transactional
+    @Query(value = "SELECT lo_unlink(:oid)", nativeQuery = true)
```
ViacheslavKlimov, line 49: **Low-confidence observation:** The `@Transactional` here uses default propagation (`REQUIRED`), which means it will join an outer transaction if one exists. If a future caller wraps `deleteOtaPackage` in a `@Transactional`, and `lo_unlink` fails (e.g., large object already removed by a concurrent operation), the Spring transaction interceptor will mark the transaction as rollback-only *before* the exception reaches the `catch` block in `unlinkDataIfPresent`. This would cause the row deletion (`removeById`) to be rolled back as well, contradicting the best-effort cleanup intent.

Currently no callers create an outer transaction, so this is not a problem today. But if the contract is that `lo_unlink` failure should never prevent row deletion, consider using `@Transactional(propagation = Propagation.REQUIRES_NEW)` to isolate it.

AndriiLandiak: Thanks for the observation! PostgreSQL's lo_unlink does throw an exception when the OID doesn't exist, but we handle this in the unlinkDataIfPresent method with a try/catch - the exception is logged and not re-thrown, so the row deletion is not affected. 

Currently, no callers wrap deleteOtaPackage in an outer transaction, so the rollback-only scenario doesn't apply. If that changes in the future, we can revisit and add REQUIRES_NEW. 

For now, KISS.

PR #14881, opened 2026-01-23
Support for Execute operation with arguments (OMA LwM2M v1.0.2)

nickAS21:
## Pull Request description

https://thingsboard-portal.atlassian.net/browse/PROD-7544

### Description
Previously, the Execute operation was implemented without support for input arguments. This PR addresses this limitation by adding the ability to perform the Execute operation with arguments, strictly following the OMA-TS-LightweightM2M-V1_0_2 documentation.

The "Execute" operation is used by the LwM2M Server to initiate an action within the LwM2M Client on a specific Resource. With this update, the operation now correctly handles arguments passed as a string in the message payload.

Syntax Rules (per OMA LwM2M Section 5.4.5)
According to the specification, the arguments for the Execute operation must follow these syntax rules:

Format: Arguments are provided as a list of "digit=value" pairs, separated by a comma (,).

Parameter naming: Each argument is identified by a single digit (0-9).

Value: The value is a string.

Syntax: digit'='string[','digit'='string].

Example: 1='v1',2='v2',3 (where 3 is an argument without a specific value).

Changes
Updated the Execute command handler to parse input payloads.

Added validation for the argument string format according to OMA syntax rules.

Ensured backward compatibility for Execute operations without arguments.

## General checklist

## Back-End feature checklist


common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/rpc/DefaultLwM2MRpcRequestHandler.java
```diff
@@ -241,7 +241,13 @@ private void sendDiscoverRequest(LwM2mClient client, TransportProtos.ToDeviceRpc
     }
 
     private void sendExecuteRequest(LwM2mClient client, TransportProtos.ToDeviceRpcRequestMsg requestMsg, String versionedId) {
+        RpcExecuteRequest requestBody = JacksonUtil.fromString(requestMsg.getParams(), RpcExecuteRequest.class);
         TbLwM2MExecuteRequest downlink = TbLwM2MExecuteRequest.builder().versionedId(versionedId).timeout(clientContext.getRequestTimeout(client)).build();
+        if (!requestMsg.getParams().isEmpty()) {
```
ViacheslavKlimov: Two things on this block:

1. `!requestMsg.getParams().isEmpty()` is effectively always true here. By the time `sendExecuteRequest` is called, the dispatcher (line 127) has already JSON-parsed `rpcRequest.getParams()` into a `LwM2MRpcRequestHeader` to extract the object id — an empty params string would have thrown earlier. So the throwaway `downlink` built on the line above is always overwritten. The condition you probably want is `requestBody != null && requestBody.getValue() != null`, since that's what actually distinguishes "Execute with an argument" from "Execute without one".
2. `JacksonUtil.fromString` returns `null` when the input is the literal JSON `null` (e.g. `{"value":null}` would still leave `requestBody.getValue()` null — fine, but if someone ever sent `"params": null` upstream, the line above would NPE before this check fires). Cheap to guard.

Minimum-fix version:

```java
RpcExecuteRequest requestBody = JacksonUtil.fromString(requestMsg.getParams(), RpcExecuteRequest.class);
Object value = requestBody != null ? requestBody.getValue() : null;
TbLwM2MExecuteRequest downlink = TbLwM2MExecuteRequest.builder()
        .versionedId(versionedId)
        .params(value)
        .timeout(clientContext.getRequestTimeout(client))
        .build();
```

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/downlink/DefaultLwM2mDownlinkMsgHandler.java
```diff
@@ -281,16 +284,19 @@ public void sendExecuteRequest(LwM2mClient client, TbLwM2MExecuteRequest request
             } else if (resourceModelExecute.operations.isExecutable()) {
                 ExecuteRequest downlink;
                 if (request.getParams() != null && !resourceModelExecute.multiple) {
-                    downlink = new ExecuteRequest(request.getObjectId(), (String) this.converter.convertValue(request.getParams(),
-                            resourceModelExecute.type, ResourceModel.Type.STRING, new LwM2mPath(request.getObjectId())));
+                    Object params = request.getParams();
+                    ResourceModel.Type resourceModel = resourceModelExecute.type == ResourceModel.Type.NONE ? equalsResourceTypeGetSimpleName(params) : resourceModelExecute.type;
```
ViacheslavKlimov: `equalsResourceTypeGetSimpleName` returns `null` for any type it doesn't recognize (Map, List, JsonNode, etc. — see `LwM2MTransportUtil.java:182-203`). For Execute resources where `resourceModelExecute.type == NONE`, that means an unusual `value` (say a JSON object accidentally sent by a caller) produces `resourceModel == null`, which then flows straight into `this.converter.convertValue(params, null, STRING, ...)` on the next line. Leshan's `LwM2mValueConverterImpl` doesn't handle a `null` source type gracefully — you'll get an obscure NPE that won't be caught by the `InvalidRequestException | InvalidArgumentException` handler at line 299 and will surface as `INTERNAL_SERVER_ERROR` instead of a clean `BAD_REQUEST`.

Suggest defaulting to `ResourceModel.Type.STRING` when the helper returns `null`, or explicitly rejecting the request with a validation error pointing at the offending value type.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/downlink/DefaultLwM2mDownlinkMsgHandler.java
```diff
@@ -281,16 +284,19 @@ public void sendExecuteRequest(LwM2mClient client, TbLwM2MExecuteRequest request
...
+                    String args = (String) this.converter.convertValue(params, resourceModel, ResourceModel.Type.STRING, new LwM2mPath(request.getObjectId()));
+                    Arguments arguments = Arguments.parse(args);
```
ViacheslavKlimov: This is the line that quietly breaks compatibility for existing Execute-with-value RPCs.

Before the PR, the converter output was handed straight to `new ExecuteRequest(path, String)`. Now it goes through `Arguments.parse(args)`, which enforces OMA §5.4.5 syntax (`digit['='string][','digit['='string]]`). Any value that doesn't conform fails — including all multi-digit numeric values. The repurposed test in `RpcLwm2mIntegrationExecuteTest.java` is the smoking gun: it had to change `value = 60` → `value = 5` and also wrap the value as a JSON string, because `Arguments.parse("60")` no longer parses.

Concretely, callers sending `{"method":"Execute","params":{"id":"3/0/4","value":60}}` (the exact "execute reboot after 60 seconds" example the previous test documented, and the kind of payload any existing integration is likely to send) will now get `BAD_REQUEST: Invalid argument...` instead of `CHANGED`.

Options:
- Treat this as a breaking change and document it in the PR description / release notes / upgrade script. Right now the description only mentions backward compatibility for Execute *without* arguments.
- Or fall back: if `Arguments.parse` throws and the value is a plain string/number, send it through the old single-string path so legacy clients keep working.

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
```diff
@@ -90,14 +91,54 @@ public void testExecuteRegistrationUpdateTriggerById_Result_CHANGED() throws Exc
...
     /**
-     * execute_resource_with_parameters (execute reboot after 60 seconds on device)
-     * Execute {"id":"3/0/4","value":60}
+     * execute_resource_with_parameters (execute reboot if digit = 5 on device)
+     * Execute {"id":"3/0/4","value":5}
      * {"result":"CHANGED"}
      */
     @Test
-    public void testExecuteResourceWithParametersById_Result_CHANGED() throws Exception {
+    public void testExecuteResourceWithParametersOnlyOneDigitValueNullById_Result_Ok() throws Exception {
```
ViacheslavKlimov: The original `testExecuteResourceWithParametersById_Result_CHANGED` (with `value = 60` representing "execute reboot after 60 seconds") has been repurposed in place rather than kept. That deletes regression coverage for the canonical Execute-with-numeric-value scenario — the exact case this PR is most likely to break for existing users (see comment on `DefaultLwM2mDownlinkMsgHandler.java:290`).

If you intend numeric values to keep working, please re-add the original test (e.g. `value = "0='60'"` via the new Arguments syntax) so the "reboot after 60 seconds" path is still exercised end-to-end.

Separately, naming nits on the new test:
- `_Result_Ok` is inconsistent with the file's `_Result_CHANGED` / `_Result_BAD_REQUEST` convention — the assertion compares against `ResponseCode.CHANGED.getName()`, so `_Result_CHANGED` is what matches reality. Same applies to the three other new tests on lines 113, 126, 139.
- `OnlyOneDigitValueNull` is hard to parse; `SingleDigitNoValue` reads closer to the OMA spec terminology and would match the test's actual intent.

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
```diff
@@ -90,14 +91,54 @@ public void testExecuteRegistrationUpdateTriggerById_Result_CHANGED() throws Exc
...
+    public void testExecuteResourceWithParametersDigit2_0_ValueLinkById_Result_Ok() throws Exception {
+        String expectedPath = objectInstanceIdVer_3 + "/" + RESOURCE_ID_5;
+        Object expectedValue = "2,0='https://thingsboard.io/docs/reference/lwm2m-api/'";
+        String actualResult = sendRPCExecuteWithValueById(expectedPath, expectedValue);
+        ObjectNode rpcActualResult = JacksonUtil.fromString(actualResult, ObjectNode.class);
+        assertEquals(ResponseCode.CHANGED.getName(), rpcActualResult.get("result").asText());
+    }
+
+    /**
+     * execute_resource_with_parameters (execute Factory Reset after connect with link on device)
```
ViacheslavKlimov: Stale copy-paste from the test above — there's no "link" in this case, the value is just ten bare digits (`0,1,2,3,4,5,6,7,8,9`). Suggest: `execute Factory Reset with ten digit-only arguments (no values)`.

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
```diff
@@ -178,7 +219,7 @@ private String sendRPCExecuteById(String path) throws Exception {
     }
 
     private String sendRPCExecuteWithValueById(String path, Object value) throws Exception {
-        String setRpcRequest = "{\"method\": \"Execute\", \"params\": {\"id\": \"" + path + "\", \"value\": " + value + " }}";
+        String setRpcRequest = "{\"method\": \"Execute\", \"params\": {\"id\": \"" + path + "\", \"value\": \"" + value + "\"}}";
```
ViacheslavKlimov: Wrapping `value` in `\"` is itself a breaking API change. The previous version interpolated the value as raw JSON, so callers could send `{"value": 60}` (number), `{"value": true}` (boolean), `{"value": "foo"}` (string), etc. After this change every Execute RPC client must encode the value as a JSON string. That should at minimum be called out in the PR description and the upgrade docs — currently the description only mentions "backward compatibility for Execute operations without arguments", which gives the opposite impression.

Also, see the comment on `DefaultLwM2mDownlinkMsgHandler.java:290`: even with the string wrapping, multi-digit values that don't match OMA `digit='value'` syntax (like `"60"`) will fail to parse — so the JSON-encoding change alone isn't sufficient to preserve the previous integer-value behavior.

application/src/test/java/org/thingsboard/server/transport/lwm2m/client/SimpleLwM2MDevice.java
```diff
@@ -160,7 +160,14 @@ public ExecuteResponse execute(LwM2mServer identity, int resourceId, Arguments a
         if (!arguments.isEmpty())
             withArguments = " with arguments " + arguments;
         log.info("Execute on Device resource /{}/{}/{} {}", getModel().id, getId(), resourceId, withArguments);
-        return ExecuteResponse.success();
+        switch (resourceId) {
+            case 4:
+                return ExecuteResponse.success();
+            case 5:
```
ViacheslavKlimov, line 165: Minor: `case 4` and `case 5` fall through to identical bodies. Combine to make the intent clearer:

```java
switch (resourceId) {
    case 4:
    case 5:
        return ExecuteResponse.success();
    default:
        return super.execute(identity, resourceId, arguments);
}
```

ViacheslavKlimov (commented):
## Review summary

Reviewed 5 changed files in [Support for Execute operation with arguments (OMA LwM2M v1.0.2)](https://github.com/thingsboard/thingsboard/pull/14881). Left 7 comment(s) inline.

The change adds OMA LwM2M Arguments parsing on the Execute RPC path, which is the right direction. Two things stand out as worth a closer look before merging:

1. The wire format for the `value` field is silently tightening. Before this PR, `{"method":"Execute","params":{"id":"3/0/4","value":60}}` (the canonical "execute reboot after 60 seconds" example that the test suite itself demonstrated) worked. After this PR, the same call fails — Integer 60 → String "60" → `Arguments.parse("60")` throws `InvalidArgumentException` because OMA requires a single digit. The PR description says "Ensured backward compatibility for Execute operations without arguments" but is silent on Execute with arguments, which is the actual scenario this PR is about. Worth either preserving the previous behavior as a fallback or being explicit in the release notes / upgrade docs.
2. `DefaultLwM2MRpcRequestHandler.sendExecuteRequest` builds two `TbLwM2MExecuteRequest` instances guarded by `requestMsg.getParams().isEmpty()`, which is never true at that point — see the inline comment.

There are also a few smaller items in the tests (a repurposed test that drops the previous `value=60` regression coverage, copy-paste Javadoc, naming inconsistencies, and a minor switch simplification).

---

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/downlink/DefaultLwM2mDownlinkMsgHandler.java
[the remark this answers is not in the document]
nickAS21: Ok

nickAS21: Ok

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/downlink/DefaultLwM2mDownlinkMsgHandler.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/lwm2m/client/SimpleLwM2MDevice.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
nickAS21: Regarding naming: Agreed. Renamed the test method to testExecuteResourceWithParametersSingleDigitNoValueById_Result_Ok().
This naming is more precise as it follows the OMA spec terminology: a "Single Digit" represents the Argument ID, and the test confirms that sending an ID without an associated value (No Value) is handled correctly by the parser.

Regarding Backward Compatibility & Arguments.parse:
It is important to note that backward compatibility is not an issue here. Prior to this PR, the sendExecuteRequest method did not support parameters at all — requests were always sent as new ExecuteRequest(path) without any value.

Since the ability to send values is a new feature introduced in this PR, I have implemented it strictly following the OMA LwM2M specification (v1.0.2, Section 5.4.5). According to the spec, to send a value (like "60"), one must provide an Argument ID (e.g., {"value": "2='60'"}).

Using Arguments.parse ensures that we adhere to the standard from the very beginning. Users of existing Execute RPCs won't be affected (as their empty-parameter calls remain unchanged), and new users will be guided to use the standard-compliant format.

ViacheslavKlimov (commented):
## Re-review summary

Re-reviewed [Support for Execute operation with arguments (OMA LwM2M v1.0.2)](https://github.com/thingsboard/thingsboard/pull/14881) — verified 7 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/14881#pullrequestreview-4279675995).

| Status | Count |
|--------|-------|
| ✅ Resolved | 2 |
| 💬 Acknowledged | 3 |
| ❌ Unresolved | 2 |

Also found 3 new issue(s) in the fix commit, commented inline. The most important is in `DefaultLwM2mDownlinkMsgHandler.java:297` — the new `catch (IllegalArgumentException e)` fallback added to address the backward-compatibility concern is unreachable: `Arguments.parse` throws Leshan's `InvalidArgumentException`, which `extends Exception`, not `IllegalArgumentException`. The new BAD_REQUEST tests confirm the fallback never fires.

### Finding details

- ✅ **DefaultLwM2mDownlinkMsgHandler.java** — `equalsResourceTypeGetSimpleName` returning null — *Fixed in code: now throws `InvalidArgumentException` explicitly when the type is unsupported, which is caught by the outer handler and surfaces as BAD_REQUEST.*
- ✅ **SimpleLwM2MDevice.java** — switch fall-through redundancy — *Fixed in code: `case 4` now falls through to `case 5`.*
- 💬 **DefaultLwM2mDownlinkMsgHandler.java** — silently breaking compatibility for Execute-with-value RPCs — *Developer argued the prior `value=60` behavior was already buggy (value was being silently dropped in `DefaultLwM2MRpcRequestHandler.sendExecuteRequest` before this PR, so existing clients sending values never had them honored). That's a fair point. The new behavior is strict OMA compliance, which is a reasonable design call.*
- 💬 **RpcLwm2mIntegrationExecuteTest.java** — dropped `value=60` regression coverage and naming nits — *Developer renamed tests and explained the previous coverage relied on the silent-drop bug. The `_Result_Ok` vs `_Result_CHANGED` consistency nit was not addressed, but tracking it isn't worth blocking on.*
- 💬 **RpcLwm2mIntegrationExecuteTest.java** — wrapping `value` in `\"` is a wire-format change — *Same rationale as above: the previous wire format accepted values that were silently dropped server-side; the new format is consistent with the OMA-compliant parsing.*
- ❌ **DefaultLwM2MRpcRequestHandler.java:246** — `!requestMsg.getParams().isEmpty()` is effectively always true — *Still present, not addressed.*
- ❌ **RpcLwm2mIntegrationExecuteTest.java:135** — Javadoc still says "after connect with link on device" for a test whose value has no link — *Still present, not addressed.*

---

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/rpc/DefaultLwM2MRpcRequestHandler.java
ViacheslavKlimov: **Still unresolved:** `!requestMsg.getParams().isEmpty()` is effectively always true here. The dispatcher at line 127 already calls `JacksonUtil.fromString(rpcRequest.getParams(), LwM2MRpcRequestHeader.class)` and then `getIdFromParameters(client, header)` to extract the `objectId` — if the params string were empty or missing an `id`, those calls would have failed before `sendExecuteRequest` was even reached. The `downlink` built on the line above is therefore always discarded.

The condition you probably want is `requestBody != null && requestBody.getValue() != null`, which is what actually distinguishes "Execute with an argument" from "Execute without one." Minimum fix:

```java
RpcExecuteRequest requestBody = JacksonUtil.fromString(requestMsg.getParams(), RpcExecuteRequest.class);
Object value = requestBody != null ? requestBody.getValue() : null;
TbLwM2MExecuteRequest downlink = TbLwM2MExecuteRequest.builder()
        .versionedId(versionedId)
        .params(value)
        .timeout(clientContext.getRequestTimeout(client))
        .build();
```

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/downlink/DefaultLwM2mDownlinkMsgHandler.java
```diff
@@ -276,21 +279,32 @@ public void sendExecuteRequest(LwM2mClient client, TbLwM2MExecuteRequest request
...
+                    ResourceModel.Type resourceModel = equalsResourceTypeGetSimpleName(params);
+                    if (resourceModel == null) {
+                        throw new InvalidArgumentException("Unsupported parameter type: " + params.getClass().getSimpleName() +
+                                ". Only simple types (String, Integer, Boolean, etc.) are allowed for Execute arguments.");
+                    }
+                    String args = (String) this.converter.convertValue(params, resourceModel, ResourceModel.Type.STRING, new LwM2mPath(request.getObjectId()));
+                    try {
+                        Arguments arguments = Arguments.parse(args);
+                        downlink = new ExecuteRequest(request.getObjectId(), arguments);
+                    } catch (IllegalArgumentException e) {
```
ViacheslavKlimov: This fallback is unreachable as written. `Arguments.parse` throws Leshan's `org.eclipse.leshan.core.request.argument.InvalidArgumentException`, which `extends Exception` — not `IllegalArgumentException`. So when `Arguments.parse("60")` fails, the exception skips this catch and propagates to the outer `catch (InvalidRequestException | InvalidArgumentException e)` at line 311, surfacing as BAD_REQUEST.

The two new BAD_REQUEST tests (`...SingleDigitValueInvalidById_BAD_REQUEST_Error_UintegerBetween_0_And_9_Expected` and `...MultipleArgumentsById_Result_BAD_REQUEST_Error_UnableParseArguments`) confirm this: if the fallback were actually catching parse failures, they'd assert CHANGED, not BAD_REQUEST.

Two options:
- If the BAD_REQUEST behavior is the intent, drop the dead try/catch entirely so the code matches what runs.
- If you do want a fallback to the legacy single-string ExecuteRequest for parse failures (preserving compatibility for callers like `{"value":"60"}`), change the catch to `InvalidArgumentException`. In that case the new BAD_REQUEST tests need to be removed/updated, since they'd now return CHANGED.

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
```diff
@@ -90,26 +91,100 @@ public void testExecuteRegistrationUpdateTriggerById_Result_CHANGED() throws Exc
...
+    public void testExecuteResourceWithParametersMultipleArgumentsIncludingLinkById_Result_Ok() throws Exception {
+        String expectedPath = objectInstanceIdVer_3 + "/" + RESOURCE_ID_5;
+        Object expectedValue = "2,0='https://thingsboard.io/docs/reference/lwm2m-api/'";
+        String actualResult = sendRPCExecuteWithValueById(expectedPath, expectedValue);
+        ObjectNode rpcActualResult = JacksonUtil.fromString(actualResult, ObjectNode.class);
+        assertEquals(ResponseCode.CHANGED.getName(), rpcActualResult.get("result").asText());
+    }
+
+    /**
+     * execute_resource_with_parameters (execute Factory Reset after connect with link on device)
```
ViacheslavKlimov: **Still unresolved:** Stale copy-paste from the test above — the value here is just bare digits (`0,1,2,3,4,5,6,7,8,9`), no link. Suggest: `execute Factory Reset with ten digit-only arguments (no values)`.

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
```diff
@@ -90,26 +91,100 @@ public void testExecuteRegistrationUpdateTriggerById_Result_CHANGED() throws Exc
...
+    }
+
+
+    /**
+     * execute_resource_with_parameters (execute Factory Reset after 60 seconds on device)
+     * Execute {"id":"3/0/5","value":"'60'"}
+
+     */
+    @Test
+    public void testExecuteResourceWithParametersSingleDigitValueInvalidById_BAD_REQUEST_Error_UintegerBetween_0_And_9_Expected() throws Exception {
```
ViacheslavKlimov: Typo in the method name: `UintegerBetween_0_And_9` should be `IntegerBetween_0_And_9` — the assertion message itself says "an integer between 0 and 9 is expected". Also, the surrounding tests use `_Result_BAD_REQUEST_...` while this one drops the `_Result_` prefix; keeping them consistent would help.

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
```diff
@@ -159,7 +234,7 @@ public void testExecuteNonExistingResourceOnNonExistingObjectById_Result_BAD_REQ
      * {"result":"BAD_REQUEST","error":"Specified object id 0 absent in the list supported objects of the client or is security object!"}
      */
     @Test
-    public void testExecuteSecurityObjectById_Result_NOT_FOUND() throws Exception {
+    public void testExecuteSecurityObjectById_Result_BAD_REQUEST_Error_InvalidDigit() throws Exception {
```
ViacheslavKlimov: `Error_InvalidDigit` doesn't match what this test actually verifies. The assertion checks the error `"Specified object id 0 absent in the list supported objects of the client or is security object!"` — there's no "invalid digit" anywhere in the failure path (the digit/argument parser isn't even reached because there's no `value` in the request). Something like `_Error_SecurityObject` or `_Error_SpecifiedObjectAbsent` would describe the assertion accurately.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/rpc/DefaultLwM2MRpcRequestHandler.java
nickAS21: Ok

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/rpc/DefaultLwM2MRpcRequestHandler.java
```diff
@@ -241,7 +241,13 @@ private void sendDiscoverRequest(LwM2mClient client, TransportProtos.ToDeviceRpc
...
+            downlink = TbLwM2MExecuteRequest.builder().versionedId(versionedId)
+                    .params(requestBody.getValue())
+                    .timeout(clientContext.getRequestTimeout(client)).build();
+        }
```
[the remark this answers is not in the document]
nickAS21: Ok

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/rpc/DefaultLwM2MRpcRequestHandler.java
nickAS21: Ok

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/server/downlink/DefaultLwM2mDownlinkMsgHandler.java
nickAS21: Okj

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
```diff
@@ -90,14 +91,54 @@ public void testExecuteRegistrationUpdateTriggerById_Result_CHANGED() throws Exc
...
+    public void testExecuteResourceWithParametersDigit2Value60ById_Result_Ok() throws Exception {
+        String expectedPath = objectInstanceIdVer_3 + "/" + RESOURCE_ID_5;
+        Object expectedValue = "2='60'";
+        String actualResult = sendRPCExecuteWithValueById(expectedPath, expectedValue);
+        ObjectNode rpcActualResult = JacksonUtil.fromString(actualResult, ObjectNode.class);
+        assertEquals(ResponseCode.CHANGED.getName(), rpcActualResult.get("result").asText());
+    }
+
+    /**
+     * execute_resource_with_parameters (execute Factory Reset after connect with link on device)
```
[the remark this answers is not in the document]
nickAS21: Ok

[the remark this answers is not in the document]
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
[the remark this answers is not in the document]
nickAS21: OK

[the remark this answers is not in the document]
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/lwm2m/rpc/sql/RpcLwm2mIntegrationExecuteTest.java
nickAS21: I updated the test helper method to use ObjectNode for JSON construction instead of manual string concatenation with hardcoded quotes. This ensures that the value field preserves its original type (String, Number, or Boolean) without forcing unnecessary quotes, maintaining backward compatibility as requested.

PR #14912, opened 2026-01-27
[LTS-4.2] fix(rule-engine): fix race conditions in calculate delta and math node

dskarzh:
**Issue**

`ConcurrentReferenceHashMap.computeIfAbsent()` does not guarantee atomicity — the mapping function can execute concurrently for the same key, creating multiple `SemaphoreWithTbMsgQueue` instances per entity. This breaks serial message processing in `TbMathNode` and `CalculateDeltaNode`.

**Changes**
- Replace `ConcurrentReferenceHashMap` with Caffeine `LoadingCache` which guarantees exactly-once loading per key
- Configure cache with 1-day expiry after last access
- Add `onPartitionChangeMsg` handler in `CalculateDeltaNode` to evict queues for non-local entities
- Add concurrent access test for `TbMathNode` to verify serial message processing

image

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/metadata/CalculateDeltaNode.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/metadata/CalculateDeltaNode.java
@@ -17,6 +17,8 @@ package org.thingsboard.rule.engine.metadata;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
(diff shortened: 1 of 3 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/metadata/CalculateDeltaNode.java
```diff
@@ -74,7 +78,9 @@ public void init(TbContext ctx, TbNodeConfiguration configuration) throws TbNode
         if (config.isAddPeriodBetweenMsgs() && StringUtils.isBlank(config.getPeriodValueKey())) {
             throw new TbNodeException("Period value key should be specified!", true);
         }
-        locks = new ConcurrentReferenceHashMap<>(16, ConcurrentReferenceHashMap.ReferenceType.WEAK);
+        queues = Caffeine.newBuilder()
+                .expireAfterAccess(Duration.ofDays(1L))
```
ViacheslavKlimov, line 82: Decrease it to 5 mins

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/metadata/CalculateDeltaNode.java
```diff
@@ -91,13 +97,18 @@ public void onMsg(TbContext ctx, TbMsg msg) {
...
         }
-        locks.computeIfAbsent(msg.getOriginator(), SemaphoreWithTbMsgQueue::new)
-                .addToQueueAndTryProcess(msg, ctx, this::processMsgAsync);
+        var processingQueue = queues.get(msg.getOriginator());
+        processingQueue.addToQueueAndTryProcess(msg, ctx, this::processMsgAsync);
+    }
+
+    @Override
+    public void onPartitionChangeMsg(TbContext ctx, PartitionChangeMsg msg) {
+        queues.asMap().keySet().removeIf(entityId -> !ctx.isLocalEntity(entityId));
```
ViacheslavKlimov, line 106: Let's just await invalidation by Caffeine

PR #14987, opened 2026-02-06
sparkplug-unique-device-names

nickAS21:
## Pull Request description

https://thingsboard-portal.atlassian.net/browse/PROD-7630

https://github.com/thingsboard/thingsboard/issues/15492

### Pull Request Description Update
Backward Compatibility & Migration
The migration from short names to full-path names is implicit and occurs automatically upon receiving the first Sparkplug message after the update.

Key migration details:

Device Renaming: Devices previously identified by a short name (e.g., deviceId) are automatically renamed to their full Sparkplug path (e.g., groupId:edgeNode:deviceId).

Label Preservation: To ensure no data loss, the original short name is saved into the Label field. This only happens if the Label field is currently empty, preventing the overwrite of any user-defined labels set via the UI.

Legacy Support: The system maintains a fallback lookup mechanism. If a device is not found by its full path in the cache, the service attempts a secondary lookup using the legacy short name before proceeding with a rename or new device creation.

Action Required:
No manual migration scripts are needed. However, it is recommended to include a note in the Release Docs regarding the naming convention change for Sparkplug devices to inform users who rely on the old naming scheme in their dashboards or rules.

#### Problem

 - Previously, Sparkplug devices in ThingsBoard used only the deviceId as their entity name.

#### Collisions: 
- If a device with the same name existed in a different Group or Node, it was impossible to create or register a new one.
- Ambiguity: It was difficult to identify which Group or Node a device belonged to just by looking at the ThingsBoard device list.

#### Solution

We now enforce a fully qualified naming convention (namespacing) by combining the Sparkplug hierarchy:
- {groupId}:{nodeId}:{deviceId}
- To ensure backward compatibility and a smooth transition, the system now handles legacy devices as follows:
#### Identification: 
- When a device with a legacy name (deviceId only) sends its first message (NBIRTH or DBIRTH), the system identifies it.
- The new device is automatically created with the new format: {groupId}:{nodeId}:{deviceId}.
- Old device with deviceId as entity name - remains unchanged.

#### Changes

- Identity Logic: 
- -- Updated the Sparkplug engine to use the triple-identifier format for all new device registrations.

### Example for node:  groupId = `Sensor-01`, nodeId =  `Node_5`

| Aspect | Before | After |
| :--- | :--- | :--- |
|Thingsboard name:  **Plant_A** | `Plant_A` | `Plant_A` |

or any unique name according to Saprkplug 3.0 requirements

### Example for new device

| Aspect | Before | After |
| :--- | :--- | :--- |
| **Device Name** | `Sensor-01` | `Group_A:Node_5:Sensor-01` |

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -330,84 +332,203 @@ private BasicCredentialsValidationResult validateMqttCredentials(TransportProtos
...
+            log.error("[{}] Error checking relation for device {}", gateway.getId(), existingDevice.getId(), e);
+            return null;
+        }
+
+        // If the device is found but not related to this gateway, it's a security breach
+        if (!isRelated) {
+            log.error("[{}] Security breach attempt! Gateway tried to rename device [{}] without 'Created' relation.",
+                    gateway.getId(), existingDevice.getId());
+            // Throwing exception to halt the entire connection process
+            throw new RuntimeException("Security breach attempt! Unauthorized device rename.");
```
ViacheslavKlimov, line 418: This throw worries me as a behavioral regression. The relation check only proves *this* gateway didn't create the short-named device — but a device sharing that name isn't necessarily an attack. Consider a tenant that already has an unrelated device named `temperature` (created manually, or owned by a different gateway), and a Sparkplug gateway then reports `deviceId=temperature` under some group. The full-path name `group:node:temperature` doesn't exist, so we reach the rename path, find the unrelated `temperature` device, see no `Created` relation, and throw — aborting the entire `GetOrCreateDevice` request (the exception propagates out of `handle()` into the `handlerExecutor.submit` future, failing the gateway connect). The net effect is that this gateway can *never* register its namespaced device, even though `group:node:temperature` is free. Wouldn't the safe behavior here be to fall through to `createNewDevice` (the full path is already confirmed absent) rather than block? 

Separately, if a throw is kept for the genuine-hijack case, a raw `RuntimeException` loses the error-code semantics the codebase uses elsewhere — `onDeviceConnectProto` right in this flow wraps things into `ThingsboardException(BAD_REQUEST_PARAMS)`. A typed exception would let callers distinguish an authorization failure from an unexpected crash.

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -330,84 +332,203 @@ private BasicCredentialsValidationResult validateMqttCredentials(TransportProtos
...
+        }
+
+        // Logic for renaming the device if it's related and no naming conflicts exist
+        boolean changed = false;
+        String newName = requestMsg.getDeviceName();
+
+        if (!newName.equals(existingDevice.getName())) {
+            // Check if the new name is already taken by another device
+            Device conflictDevice = deviceService.findDeviceByTenantIdAndName(gateway.getTenantId(), newName);
```
ViacheslavKlimov, line 427: Is this conflict lookup ever able to return non-null? `newName` is `requestMsg.getDeviceName()`, which `findOrCreateDevice` already looked up at the top (`findDeviceByTenantIdAndName(tenantId, deviceName)`) and got `null` — that's the only reason we reached `tryRenameSparkplugDevice`. The request holds `deviceCreationLocks` keyed on that same name for its whole duration, so no concurrent request can have created it in between. That makes `conflictDevice` always null and this whole branch unreachable. Worth noting the `renameCollisionWhenTargetNameAlreadyExists_Test` doesn't actually exercise it either — it pre-creates the full-path device, so the DBIRTH hits the early `return device` at the top of `findOrCreateDevice` and never gets here.

More broadly, the rename path issues three sequential `findDeviceByTenantIdAndName` DB calls (full path, short name, this conflict re-query). It only runs on cache miss so it isn't hot, but since the conflict result is already known from the first lookup, this third round-trip looks droppable.

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -330,84 +332,203 @@ private BasicCredentialsValidationResult validateMqttCredentials(TransportProtos
...
+            isRelated = relationService.checkRelation(
+                    gateway.getTenantId(),
+                    gateway.getId(),
+                    existingDevice.getId(),
+                    "Created",
+                    RelationTypeGroup.COMMON
+            );
+        } catch (Exception e) {
+            // Log the error from the relation service but return null to allow potential recovery
+            log.error("[{}] Error checking relation for device {}", gateway.getId(), existingDevice.getId(), e);
```
ViacheslavKlimov, line 409: On a transient failure from `checkRelation` (e.g. a DB hiccup), this returns null, which then falls through to `createNewDevice` and creates a brand-new full-path device while the legacy short-named device is left untouched. From the next message onward the full path exists, so you'd permanently end up with two devices for one physical thing. The comment says "allow potential recovery," but swallowing the error and creating a duplicate seems worse than propagating it and letting the gateway retry — was that trade-off intentional?

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/SparkplugNodeSessionHandler.java
```diff
@@ -118,19 +115,26 @@ public void onAttributesTelemetryProto(int msgId, SparkplugBProto.Payload sparkp
...
+                            log.error("[{}] Failed to set birth metrics", finalDeviceName, e);
                             throw new RuntimeException(e);
+                        }
                     }
-                }
-                return ctx;
-            }, MoreExecutors.directExecutor());
+                    return ctx;
+                }, MoreExecutors.directExecutor());
+            }  catch (IllegalArgumentException | DuplicateKeyException e) {
```
ViacheslavKlimov, line 135: This outer `catch (IllegalArgumentException | DuplicateKeyException)` mostly won't fire the way it reads. Those exceptions are thrown inside the `Futures.transform` callback, which runs when `deviceCtx` resolves — for the async device-creation path that's a different stack, so the exception surfaces on the future's executor, not synchronously here. Only the cache-hit (`immediateFuture`) case would resolve in-line and be caught. And the inner try/catch (lines 126-131) already rethrows those as `RuntimeException` anyway, so this outer handler is largely redundant/misleading. Can it be dropped, or is it guarding something I'm missing?

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -330,84 +332,203 @@ private BasicCredentialsValidationResult validateMqttCredentials(TransportProtos
...
+        Device existingDevice = deviceService.findDeviceByTenantIdAndName(gateway.getTenantId(), deviceId);
+
+        if (existingDevice == null) {
+            return null;
+        }
+
+        // Security check: verify that the device was created by this gateway
+        boolean isRelated = false;
+        try {
+            // Security check: verify that the device was originally created by this gateway
```
ViacheslavKlimov, line 399: The "Security check: verify that the device was ... created by this gateway" comment appears twice two lines apart (here and just above the `try`). One of them is redundant.

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -330,84 +332,203 @@ private BasicCredentialsValidationResult validateMqttCredentials(TransportProtos
...
+            isRelated = relationService.checkRelation(
+                    gateway.getTenantId(),
+                    gateway.getId(),
+                    existingDevice.getId(),
+                    "Created",
```
ViacheslavKlimov, line 404: `"Created"` is now a magic string in three places (here and twice in `createNewDevice`'s `saveRelation`). The rename check and the creation must agree on this exact literal for the security check to work, so a shared constant (or reusing an `EntityRelation` constant like the existing `CONTAINS_TYPE`/`MANAGES_TYPE`) would protect that invariant against a future typo in one spot.

ViacheslavKlimov (commented):
## Review summary

Reviewed 17 changed files in [sparkplug-unique-device-names](https://github.com/thingsboard/thingsboard/pull/14987). The fully-qualified naming + legacy-rename approach is reasonable, but the rename path has a behavioral concern worth a closer look: any pre-existing device whose name happens to equal the Sparkplug `deviceId` aborts the whole gateway request, and the collision re-check appears to be unreachable under the request lock. Left 15 comments inline covering those points plus error-handling, dead test code, and a number of cleanliness items.

---

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -330,84 +332,203 @@ private BasicCredentialsValidationResult validateMqttCredentials(TransportProtos
...
+    }
+
+    private Device createNewDevice(GetOrCreateDeviceFromGatewayRequestMsg requestMsg,
+                                   Device gateway,
+                                   DeviceId gatewayId,  String[] topicPath) {
+        TenantId tenantId = gateway.getTenantId();
+        Device device = new Device();
+        device.setTenantId(tenantId);
+        device.setName(requestMsg.getDeviceName());
+        if (requestMsg.getIsSparkplug()) {
```
ViacheslavKlimov, line 455: Two nested single-line `if`s with the inner body on the same line as the condition reads awkwardly and doesn't match the brace style of the rest of this method. `if (requestMsg.getIsSparkplug() && topicPath.length == 3) { device.setLabel(topicPath[2]); }` would be cleaner.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/SparkplugDeviceSessionContext.java
```diff
@@ -128,5 +128,5 @@ public void onToDeviceRpcRequest(UUID sessionId, TransportProtos.ToDeviceRpcRequ
                             rpcRequest.getMethodName() + ". " + e.getMessage());
         }
     }
-
+
```
ViacheslavKlimov, line 131: This is a whitespace-only change (the blank line before the closing brace now has trailing spaces). Worth reverting to keep the file out of the diff.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/util/sparkplug/SparkplugTopic.java
```diff
@@ -329,6 +330,17 @@ public String getNodeDeviceName() {
         return isNode() ? edgeNodeId : deviceId;
     }
 
+    public String getNodeDeviceNameAllPath() {
+        StringBuilder sb = new StringBuilder();
+        if (hostApplicationId == null) {
```
ViacheslavKlimov, line 335: When `hostApplicationId != null` this returns `""` (empty string), which is easy to misuse — `onDeviceConnectSparkplug` feeds the result straight into `devices.get(fullPath)` and ultimately `onDeviceConnect("", ...)`, so a host-application topic reaching this path would silently look up / create a device named empty-string rather than failing loudly. If the precondition is "only group/node/device topics," returning null or throwing would make that explicit.

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -159,52 +187,301 @@ public IMqttToken clientMqttV5ConnectWithNDEATH(long ts, long value, String metr
...
+                    // Both devices must still exist, proving no exception crashed the process
+                    return oldDevice != null && conflictDevice != null;
+                });
+    }
+
+    /**
+     * Coverage: The privilege concern — attempt to rename a device not owned by the gateway.
+     * This test verifies that the original device's ID remains unchanged, meaning it was not hijacked.
+     */
+    protected void unauthorizedRenameAttemptBad_Test() throws Exception {
```
ViacheslavKlimov, line 338: `unauthorizedRenameAttemptBad_Test` and `unauthorizedRenameAttempt_Test` are near-identical (same stranger-device setup, same DBIRTH publish) and differ only in the final assertion. `unauthorizedRenameAttemptBad_Test` also isn't wired to any `@Test`, so it's currently dead test code. Consider folding both assertions into one test, or at least dropping the unused variant.

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -159,52 +187,301 @@ public IMqttToken clientMqttV5ConnectWithNDEATH(long ts, long value, String metr
...
+        Assert.assertNull(device2.get().getLabel());
+    }
+
+    /**
+     * Coverage: Rename when a device with the target full-path name already exists (collision).
+     */
+    protected void renameCollisionWhenTargetNameAlreadyExists_Test() throws Exception {
+        long ts = calendar.getTimeInMillis();
+        String shortName = deviceId + "_1"; // Created in beforeTest
+        String fullPathName = groupId + ":" + edgeNode + ":" + shortName;
```
ViacheslavKlimov, line 307: These new tests rebuild full-path names (`groupId + ":" + edgeNode + ":" + ...`) and DBIRTH topics with raw `":"` / `"/"` literals, even though the rest of this file was just migrated to `DEVICE_NAME_SPLIT_SEPARATOR` / `TOPIC_SPLIT_SEPARATOR`. A small `fullPath(shortName)` / `dbirthTopic(name)` helper would remove the copy-paste and keep the tests in step with the production separators.

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -159,52 +187,301 @@ public IMqttToken clientMqttV5ConnectWithNDEATH(long ts, long value, String metr
...
+
+    /**
+     * Coverage: Concurrent first-message registration with the lock mechanism.
+     */
+    protected void concurrentFirstMessageRegistration_Test() throws Exception {
+        int threadCount = 5;
+        String concurrentDeviceName = "concurrent_device";
+        clientWithCorrectNodeAccessTokenWithNDEATH();
+
+        java.util.concurrent.ExecutorService executor = java.util.concurrent.Executors.newFixedThreadPool(threadCount);
```
ViacheslavKlimov, line 441: `java.util.concurrent.ExecutorService` / `Executors` are referenced fully-qualified inline — worth adding imports and using the short names to match the rest of the file.

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -159,52 +187,301 @@ public IMqttToken clientMqttV5ConnectWithNDEATH(long ts, long value, String metr
...
+                            payload.build().toByteArray(), 0, false);
+                } catch (Exception e) {
+                    log.error("Concurrent publish failed", e);
+                }
+            });
+        }
+
+        String expectedName = groupId + ":" + edgeNode + ":" + concurrentDeviceName;
+        await("Wait for concurrent registration result")
+                .atMost(40, TimeUnit.SECONDS) // Restored to 40s as requested
```
ViacheslavKlimov, line 459: `// Restored to 40s as requested` is a review-conversation artifact rather than something a future reader needs. Same goes for the `// // With a full topic...` double-slash leftover in `connectClientWithCorrectAccessTokenWithNDEATHDevicesCreatingBefore_Test`.

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -98,14 +108,32 @@ public abstract class AbstractMqttV5ClientSparkplugTest extends AbstractMqttInte
     protected static final String metricBirthName_Int32 = "Device Metric int32";
     protected Set<String> sparkplugAttributesMetricNames;
 
-    public void beforeSparkplugTest() throws Exception {
+    public void beforeSparkplugTest(boolean isCreateDevices) throws Exception {
```
ViacheslavKlimov, line 111: A `boolean isCreateDevices` flag forced every existing caller to become `beforeSparkplugTest(false)`, which is opaque at the call site, while only one test passes `true`. Keeping the no-arg `beforeSparkplugTest()` and having the single pre-seeding test call a separately-named helper (e.g. `seedLegacyAndFullPathDevices()`) afterward would avoid touching all the other test classes and make intent obvious.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
```diff
@@ -260,7 +261,7 @@ protected void processOnConnect(MqttPublishMessage msg, String deviceName, Strin
         log.trace("[{}][{}][{}] onDeviceConnect: [{}]", gateway.getTenantId(), gateway.getDeviceId(), sessionId, deviceName);
         int msgId = getMsgId(msg);
         AtomicBoolean ackSent = new AtomicBoolean(false);
-        process(onDeviceConnect(deviceName, deviceType),
+        process(onDeviceConnect(deviceName, deviceType, false),
```
ViacheslavKlimov, line 264: `onDeviceConnect` gained a trailing `boolean isSparkplug`, so the non-Sparkplug callers now pass a bare `false` (here and in the `process(...)` overload) — a positional boolean that's easy to misread. Since the Sparkplug path already has its own `onDeviceConnectSparkplug` entry point, you could keep the original two-arg `onDeviceConnect(name, type)` for these callers and have the Sparkplug entry point call a private overload with the flag, so no existing caller sprouts a magic `false`.

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
nickAS21: Ok

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
nickAS21: Ok

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
nickAS21: Ok

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/SparkplugNodeSessionHandler.java
nickAS21: Ok

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
nickAS21: OK

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
nickAS21: Ok

PR #15043, opened 2026-02-16
Added new linter for css validation to support new css features

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/angular.json
+++ b/ui-ngx/angular.json
@@ -91,6 +91,11 @@
                 "glob": "marker-shadow.png",
                 "input": "node_modules/leaflet/dist/images/",
                 "output": "/"
+              },
+              {
+                "glob": "**/*",
+                "input": "node_modules/ace-linters/build",
+                "output": "/assets/ace-linters"
               }
             ],
             "styles": [
--- a/ui-ngx/angular.json
+++ b/ui-ngx/angular.json
@@ -91,11 +91,6 @@
                 "glob": "marker-shadow.png",
                 "input": "node_modules/leaflet/dist/images/",
                 "output": "/"
-              },
-              {
-                "glob": "**/*",
-                "input": "node_modules/ace-linters/build",
-                "output": "/assets/ace-linters"
               }
             ],
             "styles": [
@@ -139,7 +134,8 @@
             ],
             "allowedCommonJsDependencies": [
               "*"
-            ]
+            ],
+            "webWorkerTsConfig": "tsconfig.worker.json"
           },
           "configurations": {
             "production": {
--- a/ui-ngx/angular.json
+++ b/ui-ngx/angular.json
@@ -134,8 +134,7 @@
             ],
             "allowedCommonJsDependencies": [
               "*"
-            ],
-            "webWorkerTsConfig": "tsconfig.worker.json"
+            ]
           },
           "configurations": {
             "production": {
--- a/ui-ngx/package.json
+++ b/ui-ngx/package.json
@@ -41,6 +41,7 @@
     "@tinymce/tinymce-angular": "^8.0.1",
     "ace-builds": "1.43.6",
     "ace-diff": "^3.3.0",
+    "ace-linters": "^2.0.0",
     "angular-gridster2": "~20.2.4",
     "angular2-hotkeys": "^16.0.1",
     "canvas-gauges": "^2.1.7",
--- a/ui-ngx/src/app/core/worker/css-linter.worker.ts
+++ b/ui-ngx/src/app/core/worker/css-linter.worker.ts
@@ -1,0 +1,37 @@
+///
+/// Copyright © 2016-2026 The Thingsboard Authors
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+
+/// <reference lib="webworker" />
+
+import { ServiceManager } from "ace-linters/build/service-manager";
+import { CssService } from "ace-linters/build/css-service";
+
+const manager = new ServiceManager(self);
+
+manager.registerService("css", {
+  features: {
+    completion: true,
+    completionResolve: true,
+    diagnostics: true,
+    format: true,
+    hover: true,
+    documentHighlight: true,
+    signatureHelp: false,
+  },
+  module: () => Promise.resolve({ CssService }),
+  className: "CssService",
+  modes: "css",
+});
--- a/ui-ngx/src/app/core/worker/css-linter.worker.ts
+++ b/ui-ngx/src/app/core/worker/css-linter.worker.ts
@@ -16,12 +16,12 @@
 
 /// <reference lib="webworker" />
 
-import { ServiceManager } from "ace-linters/build/service-manager";
-import { CssService } from "ace-linters/build/css-service";
+import { ServiceManager } from 'ace-linters/build/service-manager';
+import { CssService } from 'ace-linters/build/css-service';
 
 const manager = new ServiceManager(self);
 
-manager.registerService("css", {
+manager.registerService('css', {
   features: {
     completion: true,
     completionResolve: true,
@@ -32,6 +32,6 @@ manager.registerService("css", {
     signatureHelp: false,
   },
   module: () => Promise.resolve({ CssService }),
-  className: "CssService",
-  modes: "css",
+  className: 'CssService',
+  modes: 'css',
 });
--- a/ui-ngx/src/app/shared/components/css.component.ts
+++ b/ui-ngx/src/app/shared/components/css.component.ts
@@ -119,7 +119,8 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
       (ace) => {
         this.cssEditor = ace.edit(editorElement, editorOptions);
         this.cssEditor.session.setUseWorker(false);
-        this.languageProvider = LanguageProvider.fromCdn(window.location.origin + '/assets/ace-linters');
+        const worker = new Worker(new URL('../../core/worker/css-linter.worker', import.meta.url), { type: 'module' });
+        this.languageProvider = LanguageProvider.create(worker);
         this.languageProvider.registerEditor(this.cssEditor);
         this.cssEditor.session.setUseWrapMode(true);
         this.cssEditor.setValue(this.modelValue ? this.modelValue : '', -1);
@@ -150,9 +151,6 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
   ngOnDestroy(): void {
     if (this.editorResize$) {
       this.editorResize$.disconnect();
-    }
-    if (this.languageProvider && this.cssEditor) {
-      this.languageProvider.unregisterEditor(this.cssEditor);
     }
     if (this.cssEditor) {
       this.cssEditor.destroy();
--- a/ui-ngx/src/app/shared/components/css.component.ts
+++ b/ui-ngx/src/app/shared/components/css.component.ts
@@ -35,6 +35,7 @@ import { UtilsService } from '@core/services/utils.service';
 import { TranslateService } from '@ngx-translate/core';
 import { CancelAnimationFrame, RafService } from '@core/services/raf.service';
 import { beautifyCss } from '@shared/models/beautify.models';
+import { LanguageProvider } from 'ace-linters';
 
 @Component({
     selector: 'tb-css',
@@ -88,6 +89,8 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
 
   private propagateChange = null;
 
+  private languageProvider: LanguageProvider;
+
   constructor(public elementRef: ElementRef,
               private utils: UtilsService,
               private translate: TranslateService,
@@ -115,6 +118,9 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
     getAce().subscribe(
       (ace) => {
         this.cssEditor = ace.edit(editorElement, editorOptions);
+        this.cssEditor.session.setUseWorker(false);
+        this.languageProvider = LanguageProvider.fromCdn(window.location.origin + '/assets/ace-linters');
+        this.languageProvider.registerEditor(this.cssEditor);
         this.cssEditor.session.setUseWrapMode(true);
         this.cssEditor.setValue(this.modelValue ? this.modelValue : '', -1);
         this.cssEditor.setReadOnly(this.disabled);
@@ -144,6 +150,9 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
   ngOnDestroy(): void {
     if (this.editorResize$) {
       this.editorResize$.disconnect();
+    }
+    if (this.languageProvider && this.cssEditor) {
+      this.languageProvider.unregisterEditor(this.cssEditor);
     }
     if (this.cssEditor) {
       this.cssEditor.destroy();
--- a/ui-ngx/src/app/shared/components/css.component.ts
+++ b/ui-ngx/src/app/shared/components/css.component.ts
@@ -27,7 +27,7 @@ import {
 } from '@angular/core';
 import { ControlValueAccessor, UntypedFormControl, NG_VALIDATORS, NG_VALUE_ACCESSOR, Validator } from '@angular/forms';
 import { Ace } from 'ace-builds';
-import { getAce } from '@shared/models/ace/ace.models';
+import { getAce, getLanguageProvider } from '@shared/models/ace/ace.models';
 import { coerceBooleanProperty } from '@angular/cdk/coercion';
 import { Store } from '@ngrx/store';
 import { AppState } from '@core/core.state';
@@ -35,7 +35,9 @@ import { UtilsService } from '@core/services/utils.service';
 import { TranslateService } from '@ngx-translate/core';
 import { CancelAnimationFrame, RafService } from '@core/services/raf.service';
 import { beautifyCss } from '@shared/models/beautify.models';
-import { LanguageProvider } from 'ace-linters';
+import type { LanguageProvider } from 'ace-linters';
+import { Subscription } from 'rxjs';
+import { mergeMap } from 'rxjs/operators';
 
 @Component({
     selector: 'tb-css',
@@ -90,6 +92,8 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
   private propagateChange = null;
 
   private languageProvider: LanguageProvider;
+  private worker: Worker;
+  private aceSubscription: Subscription;
 
   constructor(public elementRef: ElementRef,
               private utils: UtilsService,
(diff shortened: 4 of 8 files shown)


ui-ngx/src/app/shared/components/css.component.ts
```diff
@@ -115,6 +118,10 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
     getAce().subscribe(
       (ace) => {
         this.cssEditor = ace.edit(editorElement, editorOptions);
+        this.cssEditor.session.setUseWorker(false);
```
vvlladd28, line 125: Disabling the built-in Ace worker here means CSS validation no longer comes from Ace's own mode worker. `validate()` keys off `hasErrors`, which is derived from session annotations in the `changeAnnotation` handler below — so that now depends entirely on ace-linters pushing its diagnostics into Ace's annotation list. Did you verify that error highlighting and the form-level invalid state still trigger with the linter? Worth a quick check since the previous validation source is now switched off.

ui-ngx/src/app/shared/components/css.component.ts
```diff
@@ -115,6 +118,10 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
...
+        const worker = new Worker(new URL('../../core/worker/css-linter.worker', import.meta.url), { type: 'module' });
```
vvlladd28: `css.component.ts` and `html.component.ts` are near-identical, and the json/js editor components follow the same shape — this ace-linters wiring (`new Worker` → `LanguageProvider.create` → `registerEditor`, plus the matching teardown) is generic and will likely get copy-pasted into each of them. Might be worth extracting a small helper or shared base that takes an editor plus a worker factory and returns a managed `LanguageProvider`, so the next editor reuses it instead of duplicating this block.

ui-ngx/src/app/shared/components/css.component.ts
```diff
@@ -115,6 +118,10 @@ export class CssComponent implements OnInit, OnDestroy, ControlValueAccessor, Va
...
+        this.languageProvider = LanguageProvider.create(worker);
+        this.languageProvider.registerEditor(this.cssEditor);
```
vvlladd28, line 132: The `Worker` and `LanguageProvider` created here are never torn down — `ngOnDestroy` only disconnects the `ResizeObserver` and calls `cssEditor.destroy()`. `tb-css` is used in ~48 places, many of them dialogs and widget config forms that get mounted/unmounted repeatedly, so each open/close leaks a dedicated worker thread (and its language-service state). Please terminate the worker (`worker.terminate()`) and dispose the language provider in `ngOnDestroy`. Note too that `getAce()` is async — if the component is destroyed before it emits, this block still runs on an already-destroyed component, so the cleanup needs to cover that case as well.

vvlladd28 (commented):
## Review summary

Reviewed 7 changed files in [Added new linter for css validation to support new css features](https://github.com/thingsboard/thingsboard/pull/15043). Left 8 comment(s) inline.

The change cleanly wires `ace-linters` into the CSS editor. The most important thing to address is the worker/`LanguageProvider` lifecycle — they're created per `CssComponent` instance but never disposed, and `tb-css` is used in ~48 places. The rest are smaller maintainability and consistency notes, plus a couple of build-config questions.

---

ui-ngx/angular.json
```diff
@@ -134,7 +134,8 @@
             ],
             "allowedCommonJsDependencies": [
               "*"
-            ]
+            ],
+            "webWorkerTsConfig": "tsconfig.worker.json"
```
vvlladd28: This project builds with the esbuild-based `@angular-builders/custom-esbuild:application` builder, where workers referenced via `new Worker(new URL(...), { type: 'module' })` are discovered and bundled automatically. `webWorkerTsConfig` is the legacy Webpack browser-builder option — is it actually honored by this builder, or a no-op here? If it's ignored, the worker would be type-checked under the main tsconfig rather than `tsconfig.worker.json`. Worth confirming the build actually picks it up as intended.

ui-ngx/src/typings/ace.typings.d.ts
```diff
@@ -0,0 +1,63 @@
...
+///
+
+declare module 'ace-code' {
+  export type EditSession = any;
+  export type Editor = any;
+  export type Point = { row: number; column: number };
+  export type Range = any;
+
+  export namespace Ace {
+    export type EditSession = any;
```
vvlladd28: This stub declares the `ace-code` module almost entirely as `any` (EditSession, Editor, Range, Document, plus `[key: string]: any` on Completion and the ext classes), which effectively turns off type checking for anything touching `ace-code` and will silently drift as ace-linters/ace-code evolve — a future upgrade could change real signatures while this keeps compiling. Could you add a comment explaining why this hand-written shim is needed (it's non-obvious that it exists only to satisfy ace-linters' transitive types), and pull real types from `ace-code`/`ace-builds` where they exist instead of `any`?

ui-ngx/src/app/core/worker/css-linter.worker.ts
```diff
@@ -0,0 +1,37 @@
...
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+
+/// <reference lib="webworker" />
+
+import { ServiceManager } from "ace-linters/build/service-manager";
```
vvlladd28: This file uses double quotes throughout, whereas the rest of the UI codebase (including the sibling css/html components) uses single quotes. Worth aligning with the project convention so it stays consistent and doesn't trip lint.

ui-ngx/src/app/core/worker/css-linter.worker.ts
```diff
@@ -0,0 +1,37 @@
...
+import { CssService } from "ace-linters/build/css-service";
+
+const manager = new ServiceManager(self);
+
+manager.registerService("css", {
+  features: {
```
vvlladd28, line 25: Since this is the first worker in the codebase and the pattern is presumably meant to be repeated for other languages, a short comment on what these feature flags control (and why `signatureHelp` is the only one disabled) would help whoever copies this for html/json next.

ui-ngx/tsconfig.worker.json
```diff
@@ -0,0 +1,15 @@
+/* To learn more about Typescript configuration file: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html. */
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./out-tsc/worker",
+    "lib": [
+      "es2018",
```
vvlladd28: The base tsconfig targets ES2022, but this worker config drops `lib` down to es2018 — an unexplained downgrade that would lose the worker the newer lib typings the rest of the app relies on. Unless ace-linters specifically requires es2018, consider aligning with the project's ES2022 baseline (or adding a brief comment on why es2018 is needed here).

vvlladd28 (commented):
## Review summary

Reviewed 7 changed files in [Added new linter for css validation to support new css features](https://github.com/thingsboard/thingsboard/pull/15043). Left 8 comment(s) inline.

The overall approach is correct — running `ace-linters`' CSS service in a bundled web worker with `LanguageProvider` is the library's documented integration path for bundlers, and the build wiring (`webWorkerTsConfig` + `tsconfig.worker.json`) is the standard Angular way to do it. The main functional concern is the per-instance worker that is never terminated (resource leak). It can also be made noticeably more concise: a single shared worker/provider helper next to `getAce()`, a trimmed `features` block, and type mappings instead of hand-written `any` stubs.

## Additional findings

- **ui-ngx/src/app/modules/home/pages/widget/widget-editor.component.ts:334** — The widget editor's CSS tab (`createAceEditor(this.cssInputElmRef, 'css')`) still uses Ace's built-in csslint worker, so modern CSS features will keep being flagged as errors there — arguably the most prominent place users write widget CSS. If the linter setup is extracted into a shared helper (see inline comments), wiring it up here becomes a one-liner.

---

ui-ngx/src/app/shared/components/css.component.ts
vvlladd28: Every `tb-css` instance spawns its own web worker, and nothing ever tears it down — `ngOnDestroy` destroys the editor and the ResizeObserver, but the worker thread and the `LanguageProvider` stay alive until page reload. `tb-css` appears in many widget/dashboard settings dialogs, so repeatedly opening and closing them accumulates live worker threads, and a settings page can show several editors at once, each with its own worker. There's also a race: `getAce()` is async, so a quickly-closed dialog can create the worker *after* `ngOnDestroy` has already run.

`LanguageProvider` is explicitly designed to serve many editors from one worker via `registerEditor()`, so the more concise fix is a lazily-created module-level singleton next to `getAce()` in `@shared/models/ace/ace.models.ts` (mirroring how the ace module itself is cached); per-editor cleanup then becomes `languageProvider.closeDocument(this.cssEditor.session)` in `ngOnDestroy`. If you keep the per-instance design instead, the provider must be disposed and the worker terminated in `ngOnDestroy` (with a guard for the async-creation race).

ui-ngx/src/app/shared/components/css.component.ts
vvlladd28, line 125: The `hasErrors` → `validate()` flow below relies on `session.getAnnotations()` returning entries with `type === 'error'`, previously produced by the built-in csslint worker. Have you verified that ace-linters publishes its diagnostics as session annotations with the same `error` type for invalid CSS? If the CSS language service reports some problems only as warnings that csslint treated as errors, forms that previously blocked on invalid CSS would now pass silently.

ui-ngx/src/app/shared/components/css.component.ts
```diff
@@ -35,6 +35,7 @@ import { UtilsService } from '@core/services/utils.service';
 import { TranslateService } from '@ngx-translate/core';
 import { CancelAnimationFrame, RafService } from '@core/services/raf.service';
 import { beautifyCss } from '@shared/models/beautify.models';
+import { LanguageProvider } from 'ace-linters';
```
vvlladd28: This static import pulls the ace-linters client (plus its `vscode-languageserver-protocol` dependencies) into the eager main bundle, while the rest of the Ace stack is deliberately lazy-loaded via dynamic imports in `getAce()`. Consider loading `LanguageProvider` the same lazy way — e.g. from a shared helper in `ace.models.ts` — so pages that never open a CSS editor don't pay for it.

ui-ngx/src/app/core/worker/css-linter.worker.ts
```diff
@@ -0,0 +1,37 @@
...
+///
+
+/// <reference lib="webworker" />
+
+import { ServiceManager } from "ace-linters/build/service-manager";
+import { CssService } from "ace-linters/build/css-service";
+
+const manager = new ServiceManager(self);
+
+manager.registerService("css", {
```
vvlladd28: This ServiceManager/worker plumbing is language-agnostic — the JSON and HTML editors in this repo suffer from the same outdated built-in Ace workers and would need exactly this setup with just another `registerService()` call and a different mode. Consider a more generic name (e.g. `ace-linters.worker.ts`), colocating it with the rest of the Ace infrastructure under `@shared/models/ace/` (the literal relative URL in `new Worker(new URL('../../core/worker/...'))` can't use path aliases and breaks only at runtime if either file moves), and extracting the `LanguageProvider.create(...)` + `registerEditor(...)` wiring into a shared helper next to `getAce()`. That would make the next editor migration a one-line change and naturally gives you a single shared worker.

ui-ngx/src/app/core/worker/css-linter.worker.ts
vvlladd28, line 25: ace-linters enables all features by default, so this block could shrink to just the ones you want off — effectively `signatureHelp: false` (which CSS doesn't support anyway). Separately, `format: true` enables the language-service formatter while the component keeps its own Beautify button backed by `beautifyCss()` from `@shared/models/beautify.models` — two formatters with different output styles on the same editor is confusing. Either route the button through `languageProvider.format()` or disable the LSP formatter.

ui-ngx/src/app/core/worker/css-linter.worker.ts
vvlladd28: Double quotes here look pasted from the ace-linters README — the rest of ui-ngx (including the import you added in `css.component.ts` in this same PR) uses single quotes.

ui-ngx/src/typings/ace.typings.d.ts
```diff
@@ -0,0 +1,63 @@
...
+/// limitations under the License.
+///
+
+declare module 'ace-code' {
+  export type EditSession = any;
+  export type Editor = any;
+  export type Point = { row: number; column: number };
+  export type Range = any;
+
+  export namespace Ace {
```
vvlladd28, line 23: These ~60 lines of hand-written `any` stubs exist only to satisfy `ace-linters`' d.ts imports of `ace-code`, and they silently erase type checking at the `LanguageProvider` boundary — `registerEditor()` will accept anything. Since `ace-builds` ships essentially the same Ace type surface as `ace-code` (same upstream project) and tsconfig already has a `paths` block, have you tried mapping `"ace-code": ["node_modules/ace-builds"]` (plus the needed `ace-code/src/*` submodules)? If the mapping doesn't resolve cleanly, at least re-export the concrete types from `ace-builds` instead of `any` — the hand-copied `Completion` shape here duplicates `Ace.Completion`.

ui-ngx/tsconfig.worker.json
vvlladd28: `es2018` is the stale Angular schematic default — the base tsconfig this extends uses lib `es2020` with target `ES2022`, so the worker is artificially compiled against an older lib surface. Suggest aligning to the base (or just adding `webworker` on top of the inherited libs if that resolves cleanly). The scaffold comment on line 1 could also go — other tsconfig files in this repo don't carry it.

vvlladd28 (commented):
## Re-review summary

Re-reviewed [Added new linter for css validation to support new css features](https://github.com/thingsboard/thingsboard/pull/15043) — verified 8 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/15043#pullrequestreview-4619088962).

| Status | Count |
|--------|-------|
| ✅ Resolved | 7 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 1 |

Also found 2 new issue(s) in the fix commits, commented inline — the most important one is that `tsconfig.worker.json` is now referenced by nothing and has no effect on the build.

### Is this the right approach?

Checked against the library, since the question was raised explicitly:

- **The pattern is the documented one.** `ServiceManager` + `registerService` inside a worker, with `LanguageProvider.create(worker)` + `registerEditor()` on the main thread, is exactly ace-linters' "WebWorker" setup from its README. The only more concise alternative the library offers is `LanguageProvider.fromCdn(...)`, which loads the services from unpkg at runtime — not an option for a self-hosted UI. `AceLanguageClient` is for real LSP servers over WebSocket/stdio. Nothing in 2.1.x/2.2.x introduces a shorter path for this; the registering-a-single-service approach is also what keeps the bundle down to just the CSS service instead of all of them.
- **It does fix PROD-7274.** Ran `CssService.doValidation()` from ace-linters 2.0.0 directly against the cases from the ticket — `@container (min-width: 400px) {…}`, `container-type: inline-size`, `50cqw`, native nesting and `:has()` all produce zero diagnostics, where the built-in Ace csslint worker flags them.
- **Form validation still behaves the same.** Genuine syntax errors (`.a { color: red;` → `"} expected"`) come back at LSP severity 1, which ace-linters' `toAnnotations()` maps to Ace `type: "error"`, so `hasErrors`/`validate()` keep blocking. Lint-level problems (unknown property, empty ruleset, unknown at-rule) come back as warnings — which matches the old behaviour, since Ace's csslint worker also emitted its rule messages as warnings and reserved `"error"` for parse errors.

### Finding details

- ✅ **ui-ngx/src/app/shared/components/css.component.ts** — Per-instance worker leaked; nothing torn down in `ngOnDestroy`, plus an async-creation race — *Fixed. The worker and subscription are now held on the instance, `ngOnDestroy` unsubscribes first and then falls back to `worker.terminate()` when the provider never materialised. `closeConnection()` does terminate the worker (`$messageController.closeConnection(() => this.$messageController.$worker.terminate())`), so the per-instance design is now correctly cleaned up.*
- ✅ **ui-ngx/src/app/shared/components/css.component.ts** — Unverified that ace-linters publishes diagnostics as `type === 'error'` annotations — *Verified in this re-review; not an issue. See "Is this the right approach?" above.*
- ✅ **ui-ngx/src/app/shared/components/css.component.ts** — Static `ace-linters` import pulled the client into the eager main bundle — *Fixed. Now `import type` plus a lazy `getLanguageProvider()` in `ace.models.ts` that dynamic-imports `ace-linters` and caches the constructor, matching how `getAce()`/`getAceDiff()` already work.*
- ✅ **ui-ngx/src/app/core/worker/css-linter.worker.ts** — Provider wiring should live in a shared helper next to `getAce()` — *Fixed; `getLanguageProvider()` now lives in `ace.models.ts`.*
- ❌ **ui-ngx/src/app/core/worker/css-linter.worker.ts** — Redundant `features` flags and `format: true` alongside the Beautify button — *Partly still present; see the inline comment, where I've retracted the half of it that doesn't apply.*
- ✅ **ui-ngx/src/app/core/worker/css-linter.worker.ts** — Double-quoted strings — *Fixed.*
- ✅ **ui-ngx/src/typings/ace.typings.d.ts** — `any` stubs erased type checking at the `LanguageProvider` boundary — *Fixed; the `ace-code` shim now re-exports the concrete `ace-builds` types, and the `htmlhint` shim carries a comment explaining why it exists.*
- ✅ **ui-ngx/tsconfig.worker.json** — `lib: es2018` was below the base tsconfig — *Fixed (`es2022`) — though see the inline comment: the file no longer affects the build at all.*

---

ui-ngx/tsconfig.worker.json
```diff
@@ -0,0 +1,15 @@
+/* To learn more about Typescript configuration file: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html. */
+{
+  "extends": "./tsconfig.json",
```
vvlladd28, line 3: This file is now referenced by nothing, so none of it takes effect. The fix commits removed `"webWorkerTsConfig": "tsconfig.worker.json"` from `angular.json`, and `tsconfig.worker` doesn't appear anywhere else in the repo.

That removal was actually correct — the esbuild `application` builder never reads the option. Angular's TS transformer rewrites `new Worker(new URL('...', import.meta.url))` and hands the path to `processWebWorker`, which calls `bundleWebWorker()`; that function spreads the **main** build's esbuild options (`{...build.initialOptions, entryPoints: [workerFile], plugins: undefined}`) and has no notion of a worker tsconfig. `webWorkerTsConfig` survives in the application builder's schema but is only consumed by the karma bridge.

So the net effect today is that `lib: ["es2022", "webworker"]` and `types: []` are never applied, and the worker isn't type-checked by anything either — `src/tsconfig.app.json` has `files: ["main.ts", "polyfills.ts"]` + `include: ["**/*.d.ts"]`, and a `new URL(...)` reference doesn't pull the file into the TS program. esbuild just transpiles it.

Suggest deleting the file, so the next person adding a worker doesn't assume `webworker` lib and `types: []` are configured for them. If you'd rather keep the worker type-checked, that needs a real mechanism (e.g. adding `src/**/*.worker.ts` to a tsconfig that's actually part of a build or lint step), not this file.

ui-ngx/package.json
```diff
@@ -41,6 +41,7 @@
     "@tinymce/tinymce-angular": "^8.0.1",
     "ace-builds": "1.43.6",
     "ace-diff": "^3.3.0",
+    "ace-linters": "^2.0.0",
```
vvlladd28, line 44: `yarn.lock` resolves this to `2.0.0` (published 2026-01-31), but the current `ace-linters` release is `2.2.1` (2026-07-17) — the range already allows it, the lock just predates it. Worth refreshing before merge so a brand-new dependency doesn't enter the tree eight months and six releases stale.

Nothing in 2.1.x/2.2.x changes the CSS setup API or the `CssService` behaviour this PR depends on (the releases are spell-check, XML/PHP services, a signature-tooltip fix), so this is about starting current rather than about a specific fix you're missing.

ui-ngx/src/app/core/worker/css-linter.worker.ts
```diff
@@ -0,0 +1,37 @@
...
+import { CssService } from 'ace-linters/build/css-service';
+
+const manager = new ServiceManager(self);
+
+manager.registerService('css', {
+  features: {
+    completion: true,
+    completionResolve: true,
+    diagnostics: true,
+    format: true,
```
vvlladd28, line 29: **Still unresolved:** — but with a correction, because half of my earlier comment doesn't hold and I don't want you acting on it.

**Retracted:** the "two competing formatters" concern. I checked: `format: true` only advertises the capability on the service; `LanguageProvider` registers no formatting keybinding or command on the editor, and nothing in this component calls `languageProvider.format()`. So the Beautify button backed by `beautifyCss()` stays the only formatter the user can reach. No conflict — leave it as is, or keep `format: true` if you might route the button through the LSP formatter later.

**Still stands, cosmetically:** `ServiceManager.setDefaultFeaturesState()` defaults every feature to `true`, so `completion`, `completionResolve`, `diagnostics`, `format`, `hover` and `documentHighlight` are all no-ops here. The block could be just `features: { signatureHelp: false }`. Purely a readability point — no behavioural difference either way.

PR #15100, opened 2026-02-24
Fixed preservation of rule node execution counter in delay and deduplication nodes

dskarzh:
## Summary
- Delay and deduplication rule nodes were creating brand new `TbMsg` objects when forwarding messages, which reset the `ruleNodeExecCounter` to 0. This allowed bypassing the `maxRuleNodeExecutionsPerMessage` tenant profile limit.
- Fixed both nodes to use `copyWithNewCtx()` which preserves the execution counter while resetting the callback.
- Changed `copyWithNewCtx()` return type from `TbMsg` to `TbMsgBuilder` to allow callers to customize fields (e.g. queue name, message ID) before building.
- Added unit test for `TbMsgDelayNode` and updated existing `TbMsgDeduplicationNode` tests to assert counter preservation and empty callback.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/ruleChain/RuleNodeActorMessageProcessor.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ruleChain/RuleNodeActorMessageProcessor.java
@@ -137,7 +137,7 @@ public class RuleNodeActorMessageProcessor extends ComponentMsgProcessor<RuleNod
                 defaultCtx.tellFailure(msg.getMsg(), e);
             }
         } else {
-            tbMsg.getCallback().onFailure(new RuleNodeException("Message is processed by more then " + maxRuleNodeExecutionsPerMessage + " rule nodes!", ruleChainName, ruleNode));
+            tbMsg.getCallback().onFailure(new RuleNodeException("Message is processed by more than " + maxRuleNodeExecutionsPerMessage + " rule nodes!", ruleChainName, ruleNode));
         }
     }
 
@@ -160,7 +160,7 @@ public class RuleNodeActorMessageProcessor extends ComponentMsgProcessor<RuleNod
                     msg.getCtx().tellFailure(msg.getMsg(), e);
                 }
             } else {
-                tbMsg.getCallback().onFailure(new RuleNodeException("Message is processed by more then " + maxRuleNodeExecutionsPerMessage + " rule nodes!", ruleChainName, ruleNode));
+                tbMsg.getCallback().onFailure(new RuleNodeException("Message is processed by more than " + maxRuleNodeExecutionsPerMessage + " rule nodes!", ruleChainName, ruleNode));
             }
         }
     }
--- a/common/message/src/main/java/org/thingsboard/server/common/msg/TbMsg.java
+++ b/common/message/src/main/java/org/thingsboard/server/common/msg/TbMsg.java
@@ -110,11 +110,10 @@ public final class TbMsg implements Serializable {
                 .build();
     }
 
-    public TbMsg copyWithNewCtx() {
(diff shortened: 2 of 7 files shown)


ViacheslavKlimov (commented):
## Summary

Reviewed 7 changed files in [fix: preserve rule node execution counter in delay and deduplication nodes](https://github.com/thingsboard/thingsboard/pull/15100). The core fix is clean — changing `copyWithNewCtx()` to return a `TbMsgBuilder` lets delay and dedup nodes use `copy()` instead of `newMsg()`, correctly preserving the `TbMsgProcessingCtx` (rule node execution counter). All callers of `copyWithNewCtx()` are updated, and new tests verify the counter preservation. Found 0 issues commented inline and 2 low-confidence observations below.

## Low-Confidence Observations

> These observations might be intentional or incorrect. Verify before acting on them.

- **rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/delay/TbMsgDelayNode.java:65** — The old code used `TbMsg.newMsg()...copyMetaData(...)` which deep-copies the `TbMsgMetaData` (backed by `ConcurrentHashMap`). The new code via `copy()` shares the same `TbMsgMetaData` reference between the pending message and the re-emitted message. Since `TbMsgMetaData` is mutable, any downstream mutation of the metadata on one message would affect the other. In practice this is likely safe (rule nodes don't typically mutate metadata after forwarding), but if defensive copying was intentional, consider calling `.copyMetaData(pendingMsg.getMetaData())` on the builder. Same applies to `TbMsgDeduplicationNode.java:184`.

- **rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/delay/TbMsgDelayNode.java:65** — The old code used `TbMsg.newMsg()` which, when `ts` is 0/unset, defaults to `System.currentTimeMillis()` in the `TbMsg` constructor. The new code via `copy()` preserves the original message's `ts`. This means delayed messages now retain their original timestamp instead of getting a fresh one at re-emission time. This may be the desired behavior (preserving the original event time), but it is a semantic change from the previous behavior. Same applies to `TbMsgDeduplicationNode.java:184`.

---
*This review was auto-generated by [Claude](https://claude.ai). Findings may contain errors — please verify before applying changes.*

dskarzh:
Regarding the first observation (metadata sharing):

This is not an issue because both `TbMsgDelayNode` and `TbMsgDeduplicationNode` forward messages via `ctx.enqueueForTellNext()`. Internally, `enqueueForTellNext` calls `TbMsg.newMsg(source, queueName, ruleChainId, ruleNodeId)` which uses `tbMsg.transform().metaData(tbMsg.getMetaData())` — and `TbMsgTransformer.metaData()` performs a deep copy (`metaData.copy()`). So the metadata is always deep-copied before the message reaches the next rule node.

---
*This comment was generated by [Claude](https://claude.ai).*

dskarzh:
Regarding the second observation (timestamp preservation):

This is not a practical concern. Most downstream rule nodes (telemetry saving, alarm creation, deduplication) use `getMetaDataTs()`, which reads the metadata `"ts"` key first and only falls back to the `TbMsg.ts` field when it's absent. Since both delay and dedup nodes preserve the original metadata, the `TbMsg.ts` field change is irrelevant in the common case.

In the edge case where metadata doesn't contain a `"ts"` entry, preserving the original event time is arguably more correct than stamping it with the current processing time — the message still represents the same event.

Additionally, there are no framework-level message expiration or TTL checks based on `msg.getTs()`. Message validity (`TbMsg.isValid()`) is callback-based, so a delayed message with its original `ts` will not be rejected.

---
*This comment was generated by [Claude](https://claude.ai).*

PR #15102, opened 2026-02-24
Fixed infinite loop when rule chain input node forwards to its own rule chain

smatvienko-tb:
## Summary

- **Bug**: A `TbRuleChainInputNode` configured with `forwardMsgToDefaultRuleChain=true` creates an infinite processing loop when the originator device's default rule chain is the same chain that contains the node itself. The message is re-enqueued to Kafka on every iteration, the return stack (`ctx.stack`) grows unboundedly with identical entries, and the rule-engine microservice reaches 100% CPU usage.

- **Fix — two-layer defence**:
  - **Layer 1** (`TbRuleChainInputNode.onMsg`): if the resolved `targetRuleChainId` equals the rule chain that contains the node (`ctx.getSelf().getRuleChainId()`), `tellFailure()` is called immediately — **zero unnecessary Kafka round-trips**.
  - **Layer 2** (`DefaultTbContext.input`): before pushing a node onto the message's return stack, checks whether that `(ruleChainId, ruleNodeId)` pair is already in the stack. If yes, an indirect cycle (A→B→A) is detected and `tellFailure()` is called. This is the general safety net for cycles that span multiple rule chains.

## Root cause trace

```
Rule Chain A
  └─ Device Profile Node (first node)
       └─ Success/Failure ──► TbRuleChainInputNode "Go to kWh Accumulation"
                                   forwardMsgToDefaultRuleChain = true
                                   device default rule chain    = Rule Chain A  ← same!
                                   ↓
                               [Layer 1] targetRuleChainId == currentRuleChainId → tellFailure ✓
                               (before the fix: ctx.input(msg, RuleChainAId)
                                   pushToStack(currentNode)   ← stack grows
                                   enqueue to Kafka ──► Rule Chain A  ← back to start)
```

The exported Kafka message shows `ruleNodeExecCounter: 17` with 8 identical stack entries, confirming the loop ran for many iterations before the message was observed.

## Why two layers?

| Scenario | Caught by |
|---|---|
| Direct self-loop: node in chain A forwards to chain A | Layer 1 — immediately, no Kafka round-trip |
| Indirect cycle: A→B→A (node in A forwards to B, B forwards back to A) | Layer 2 — on the 2nd pass through A |

## Test plan

- [x] `TbRuleChainInputNodeTest#givenForwardMsgToDefaultIsTrueAndDeviceDefaultRuleChainIsSameAsCurrent_whenOnMsg_thenTellFailureToPreventDirectLoop` — Layer 1, dynamic resolution
- [x] `TbRuleChainInputNodeTest#givenConfiguredRuleChainIsSameAsCurrent_whenOnMsg_thenTellFailureToPreventDirectLoop` — Layer 1, static config
- [x] `DefaultTbContextTest#givenMsgWithCurrentNodeAlreadyInStack_whenInput_thenTellFailureToPreventLoop` — Layer 2, stack-based indirect cycle detection
- [x] `TbMsgProcessingCtxTest` — 7 unit tests for `isAlreadyInStack` edge cases
- [x] All pre-existing tests in `TbRuleChainInputNodeTest` (15→17), `DefaultTbContextTest` (56→57), and `TbMsgProcessingCtxTest` (new, 7) pass

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/flow/TbRuleChainInputNode.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/flow/TbRuleChainInputNode.java
@@ -76,6 +76,13 @@ public class TbRuleChainInputNode implements TbNode {
     public void onMsg(TbContext ctx, TbMsg msg) throws TbNodeException {
         RuleChainId targetRuleChainId = forwardMsgToDefaultRuleChain ?
                 getOriginatorDefaultRuleChainId(ctx, msg).orElse(ruleChainId) : ruleChainId;
+        if (targetRuleChainId.equals(ctx.getSelf().getRuleChainId())) {
+            ctx.tellFailure(msg, new RuntimeException(
+                    "TbRuleChainInputNode in rule chain [" + targetRuleChainId +
+                    "] is configured to forward messages back to the same rule chain it belongs to. " +
+                    "This would cause an infinite loop. Please check the rule chain configuration."));
+            return;
+        }
         ctx.input(msg, targetRuleChainId);
     }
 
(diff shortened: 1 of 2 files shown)


smatvienko-tb:
Related PRs:
https://github.com/thingsboard/thingsboard/pull/15100
https://github.com/thingsboard/thingsboard-pe/pull/4460

PR #15112, opened 2026-02-25
MQTTS metrics and client address logging on exceptionCaught

smatvienko-tb:
## Summary

- Add separate Micrometer gauge metrics for active MQTT vs MQTTS connections (`connections_active` with `protocol=MQTT/MQTTS` tags) — replaces the single `openConnections` gauge
- Extend `TransportService.createGaugeStats` to accept variadic `tags` for metric labeling
- Log the client remote address (IP:port) in `exceptionCaught` alongside the session/device info

## Changes

- `MqttTransportContext`: split `connectionsCounter` into `connectionsActiveCounterMQTT` / `connectionsActiveCounterMQTTS`; pass SSL flag from `MqttTransportHandler`
- `MqttTransportHandler`: add `isSSL()` helper; pass SSL flag to `channelRegistered/channelUnregistered`; include `clientAddr` in exception log messages
- `TransportService`: add `String... tags` vararg to `createGaugeStats`
- `DefaultTransportService`: build the full metric key as `thingsboard.transport.<name>` and include tags in the stats map key

### DIFF
--- a/common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/TransportService.java
+++ b/common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/TransportService.java
@@ -161,5 +161,5 @@ public interface TransportService {
 
     boolean hasSession(SessionInfoProto sessionInfo);
 
-    void createGaugeStats(String openConnections, AtomicInteger connectionsCounter, String... tags);
+    void createGaugeStats(String statsName, AtomicInteger number, String... tags);
 }


smatvienko-tb (commented):
## Code Review

### Summary
This PR adds per-protocol (MQTT/MQTTS) Micrometer gauge metrics and improves exception logging with client address. The changes are conceptually sound, but there are several issues worth addressing.

---

### Issues

#### 1. Metric key prefix — potential double-prefix bug
`DefaultTransportService.createGaugeStats` now manually builds the key as `"thingsboard." + StatsType.TRANSPORT.getName() + "." + statsName`, but the old code simply passed `StatsType.TRANSPORT + "." + statsName` (i.e. `"transport.openConnections"`) to `statsFactory.createGauge`. If `statsFactory` already adds the `thingsboard.` prefix internally (which is common in Micrometer factory wrappers), the new code would produce **`thingsboard.thingsboard.transport.connections_active`**. Please verify whether `statsFactory` adds this prefix.

#### 2. Java naming convention violation — `TagsKey` → `tagsKey`
Java method names must start with a lowercase letter. `TagsKey` should be `tagsKey`, and it should be `private`.

#### 3. `isSSL()` should be `private`
The method is only used within `MqttTransportHandler`. Leaving it package-private leaks internal implementation detail and may cause confusion.

#### 4. Double space in `if  (isSSL())`
Minor style nit: `if  (isSSL())` has two spaces before the parenthesis.

#### 5. `clientAddr` is always computed even when logging is disabled
In `exceptionCaught`, the `clientAddr` computation (including `getAddress(ctx)`) runs unconditionally before checking `log.isDebugEnabled()` or `log.isInfoEnabled()`. For an IOException-heavy path this adds unnecessary overhead when logging is off.

---

### Suggestions (inline via comments below)

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportService.java
```diff
@@ -1257,9 +1257,21 @@ public boolean hasSession(TransportProtos.SessionInfoProto sessionInfo) {
...
+    public void createGaugeStats(String statsName, AtomicInteger number, String... tags) {
+        String key = "thingsboard" + "." + StatsType.TRANSPORT.getName() + "." + statsName;
+        statsFactory.createGauge(key, number, tags);
+        statsMap.put(statsName + TagsKey(tags), number);
+    }
+
+    String TagsKey(String... tags) {
+        if (tags == null || tags.length < 2) return "";
+        StringBuilder sb = new StringBuilder("[");
+        for (int i = 0; i < tags.length; i += 2) {
```
smatvienko-tb, line 1269: **Naming convention**: Java method names must start with lowercase. Also should be `private` since it is only called from within this class.

```suggestion
    private String tagsKey(String... tags) {
```

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -180,16 +180,20 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
         this.rpcAwaitingAck = new ConcurrentHashMap<>();
     }
 
+    boolean isSSL() {
```
smatvienko-tb, line 183: **Visibility**: `isSSL()` is only used within this class — make it `private` to avoid leaking implementation detail.

```suggestion
    private boolean isSSL() {
```

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportContext.java
```diff
@@ -88,20 +88,30 @@ public class MqttTransportContext extends TransportContext {
...
         super.init();
-        transportService.createGaugeStats("openConnections", connectionsCounter);
+        transportService.createGaugeStats("connections_active", connectionsActiveCounterMQTT, "protocol", "MQTT");
+        transportService.createGaugeStats("connections_active", connectionsActiveCounterMQTTS, "protocol", "MQTTS");
     }
 
-    public void channelRegistered() {
-        connectionsCounter.incrementAndGet();
+    public void channelRegistered(boolean isSSL) {
+        if  (isSSL) {
```
smatvienko-tb: **Double space**: Extra space before `(`.

```suggestion
        if (isSSL()) {
```

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportContext.java
```diff
@@ -88,20 +88,30 @@ public class MqttTransportContext extends TransportContext {
...
+            connectionsActiveCounterMQTTS.incrementAndGet();
+        } else {
+            connectionsActiveCounterMQTT.incrementAndGet();
+        }
     }
 
-    public void channelUnregistered() {
-        connectionsCounter.decrementAndGet();
+    public void channelUnregistered(boolean isSSL) {
```
smatvienko-tb, line 109: **Double space** in second `if` as well.

```suggestion
        if (isSSL) {
```

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/TransportService.java
```diff
@@ -161,5 +161,5 @@ void process(LwM2MRequestMsg msg,
 
     boolean hasSession(SessionInfoProto sessionInfo);
 
-    void createGaugeStats(String openConnections, AtomicInteger connectionsCounter);
+    void createGaugeStats(String openConnections, AtomicInteger connectionsCounter, String... tags);
```
ViacheslavKlimov: **Low-confidence observation:** The parameter names `openConnections` and `connectionsCounter` are holdovers from the old single-counter API. Now that this method is general-purpose (accepts any stats name + tags), these names are misleading. Consider renaming to `statsName` and `number` (matching the implementation in `DefaultTransportService`) to avoid confusing future readers.

```java
void createGaugeStats(String statsName, AtomicInteger number, String... tags);
```

PR #15121, opened 2026-02-27
Rule chain notes

vvlladd28:
## Pull Request description

Created support notes in the rule chain

image

image

Implements this discusion https://github.com/thingsboard/thingsboard/discussions/15024

## General checklist

## Back-End feature checklist

### DIFF
--- a/common/data/src/main/java/org/thingsboard/server/common/data/rule/RuleChainDetails.java
+++ b/common/data/src/main/java/org/thingsboard/server/common/data/rule/RuleChainDetails.java
@@ -20,6 +20,7 @@ import io.swagger.v3.oas.annotations.media.Schema;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 
+import java.util.ArrayList;
 import java.util.List;
 
 @Data
@@ -40,7 +41,7 @@ public class RuleChainDetails extends RuleChain {
 
     public RuleChainDetails(RuleChainDetails ruleChainDetails) {
         super(ruleChainDetails);
-        this.notes = ruleChainDetails.getNotes();
+        this.notes = ruleChainDetails.getNotes() != null ? new ArrayList<>(ruleChainDetails.getNotes()) : null;
     }
 
 }
--- a/common/util/src/main/java/org/thingsboard/common/util/JacksonUtil.java
+++ b/common/util/src/main/java/org/thingsboard/common/util/JacksonUtil.java
@@ -243,6 +243,14 @@ public class JacksonUtil {
         }
     }
 
+    public static <T> T treeToValue(JsonNode node, TypeReference<T> type) {
+        try {
+            return OBJECT_MAPPER.treeToValue(node, type);
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Can't convert value: " + node.toString(), e);
+        }
+    }
+
     public static JsonNode toJsonNode(String value) {
         return toJsonNode(value, OBJECT_MAPPER);
     }
--- a/dao/src/main/java/org/thingsboard/server/dao/model/sql/RuleChainDetailsEntity.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/model/sql/RuleChainDetailsEntity.java
@@ -15,6 +15,7 @@
  */
 package org.thingsboard.server.dao.model.sql;
 
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import jakarta.persistence.Column;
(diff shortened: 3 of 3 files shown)


dao/src/main/java/org/thingsboard/server/dao/rule/BaseRuleChainService.java
```diff
@@ -341,22 +344,22 @@ private EntityRelation getRuleChainInputRelation(RuleChainId ruleChainId, RuleNo
     @Override
     public RuleChainMetaData loadRuleChainMetaData(TenantId tenantId, RuleChainId ruleChainId) {
         Validator.validateId(ruleChainId, "Incorrect rule chain id.");
-        RuleChain ruleChain = findRuleChainById(tenantId, ruleChainId);
-        if (ruleChain == null) {
+        RuleChainDetails ruleChainDetails = ruleChainDao.findDetailsById(ruleChainId.getId());
```
ViacheslavKlimov: The original code used `findRuleChainById(tenantId, ruleChainId)` which passes the `tenantId` to the DAO. While the current `JpaAbstractDao.findById` doesn't actually filter by tenant at the SQL level, the `tenantId` parameter is used for logging/tracing and maintains a consistent API contract. The new `findDetailsById(ruleChainId.getId())` drops the `tenantId` entirely.

Consider adding a `findDetailsById(TenantId tenantId, UUID id)` method signature that accepts the tenant ID for consistency with the rest of the DAO API, even if the SQL query itself doesn't filter by it. This follows the existing pattern used by all other `findById` calls in the codebase.

common/data/src/main/java/org/thingsboard/server/common/data/rule/RuleChainNote.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Schema(description = "Horizontal position of the note on the canvas, in pixels")
+    private int x;
+
+    @Schema(description = "Vertical position of the note on the canvas, in pixels")
+    private int y;
+
+    @Schema(description = "Width of the note, in pixels")
+    private int width;
+
```
ViacheslavKlimov, line 43: **Low-confidence observation:** The `content` and `markdownCss` fields have no `@Size` annotations, so there is no server-side validation of their length. The frontend limits these to 65536 characters each (`Validators.maxLength(65536)`), but direct API callers can bypass this and store arbitrarily large content up to the 10MB column limit. Consider adding `@Size(max = 65536)` (or a suitable limit) to `content` and `markdownCss` for defense-in-depth, or at minimum document the expected limits.

ui-ngx/src/app/modules/home/pages/rulechain/rulenote.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (this.applyDefault) {
+      styles.push(RuleNoteComponent.HEADING_STYLE_OVERRIDE);
+      styles.push(RuleNoteComponent.PADDING_STYLE_OVERRIDE);
+    }
+    if (isNotEmptyStr(cssString)) {
+      const cssParser = new cssjs();
+      this.noteClass = 'rule-note-' + hashCode(cssString);
+      cssParser.cssPreviewNamespace = this.noteClass;
+      cssParser.testMode = false;
+      const cssObjects = cssParser.applyNamespacing(cssString);
```
ViacheslavKlimov, line 70: **Low-confidence observation:** The `markdownCss` field allows arbitrary user-provided CSS that is processed through `cssParser.applyNamespacing` and injected into the DOM via `additionalStyles`. While the namespacing provides some protection, CSS-based data exfiltration (e.g., via attribute selectors + `background-image` to external URLs) is still possible in theory. This only affects users who can view the rule chain, so the blast radius is limited. Just flagging for awareness since this is a stored injection vector.

application/src/main/data/upgrade/basic/schema_update.sql
```diff
@@ -80,3 +80,9 @@ DROP TABLE IF EXISTS calculated_field_link;
 ANALYZE calculated_field;
 
 -- REMOVAL OF CALCULATED FIELD LINKS PERSISTENCE END
+
+-- RULE CHAIN NOTES MIGRATION START
+
```
ViacheslavKlimov, line 24: **Low-confidence observation:** The `notes` column is `varchar(10000000)` (~10MB), which stores ALL notes for a rule chain as a single JSON array. There is no server-side limit on the number of notes per rule chain. A user could create thousands of notes via the API, potentially storing several MB in a single row. Consider whether a reasonable upper bound on note count should be enforced in `BaseRuleChainService.saveRuleChainMetaData`.

PR #15127, opened 2026-03-02
Added request body template support for REST API call node

AndriiLandiak:
## Pull Request description

Users need to include secrets and dynamic values in HTTP request bodies (e.g., OAuth token endpoints, API login payloads). Currently, the REST API Call node always uses msg.getData() as the request body - there is no way to construct a custom body with metadata fields or secrets. While secrets and pattern placeholders already work in headers and URL, they don't work in the body. 
  
But, adding secrets to script/transformation nodes would break the security model, as unconstrained script output could leak secret values.
                                                                                                                                                                                                                                                                                                                    
This change adds an optional requestBodyTemplate field to the REST API Call node configuration. When set, the template is used as the request body with ${metadataKey} and $[messageKey] placeholder substitution (same as URL and headers). Secret placeholders (${secret:name;type:TYPE}) are resolved at init time. When the template is empty, the node preserves the existing behaviour using msg.getData().

The parseToPlainText option works with the template as well: when unchecked (default), the template result is parsed as a JSON object; when checked, it is sent as a plain text string.

<img width="723" height="315" alt="image" src="https://github.com/user-attachments/assets/c75784c3-413e-47de-947f-59830401e3e4" />

https://github.com/thingsboard/thingsboard/discussions/15086

## General checklist

## Back-End feature checklist

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
@@ -266,6 +266,11 @@ public class TbHttpClient {
                     });
         } catch (InterruptedException e) {
             log.warn("Timeout during waiting for reply!", e);
+        } catch (Exception e) {
+            if (semaphore != null) {
+                semaphore.release();
+            }
+            onFailure.accept(msg, e);
         }
     }
 
@@ -342,7 +347,14 @@ public class TbHttpClient {
         if (StringUtils.isNotBlank(config.getRequestBodyTemplate())) {
             boolean escapeJson = !config.isParseToPlainText();
             String processedTemplate = TbNodeUtils.processPattern(config.getRequestBodyTemplate(), msg, escapeJson);
-            return config.isParseToPlainText() ? processedTemplate : JacksonUtil.toJsonNode(processedTemplate);
+            if (config.isParseToPlainText()) {
+                return processedTemplate;
+            }
+            try {
+                return JacksonUtil.toJsonNode(processedTemplate);
+            } catch (Exception e) {
+                throw new RuntimeException("Request body template produced invalid JSON: " + processedTemplate, e);
+            }
         }
         return getData(msg, config.isParseToPlainText());
     }
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbRestApiCallNode.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbRestApiCallNode.java
@@ -58,6 +58,7 @@ public class TbRestApiCallNode extends TbAbstractExternalNode {
 
     static final String PARSE_TO_PLAIN_TEXT = "parseToPlainText";
     static final String MAX_IN_MEMORY_BUFFER_SIZE_IN_KB = "maxInMemoryBufferSizeInKb";
+    static final String REQUEST_BODY_TEMPLATE = "requestBodyTemplate";
     static final String TRIM_DOUBLE_QUOTES = "trimDoubleQuotes";
     protected TbHttpClient httpClient;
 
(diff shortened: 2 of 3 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -287,6 +285,14 @@ public URI buildEncodedUri(String endpointUrl) {
         return uri;
     }
 
+    private Object getRequestBody(TbMsg msg) {
+        if (StringUtils.isNotEmpty(config.getRequestBodyTemplate())) {
+            String processedTemplate = TbNodeUtils.processPattern(config.getRequestBodyTemplate(), msg);
+            return config.isParseToPlainText() ? processedTemplate : JacksonUtil.toJsonNode(processedTemplate);
```
ViacheslavKlimov: **Low-confidence observation:** `TbNodeUtils.processPattern` performs raw string substitution — substituted values are not JSON-escaped. When `parseToPlainText` is `false`, the processed template is parsed as JSON via `JacksonUtil.toJsonNode()`. If any substituted metadata or message value contains JSON-special characters (e.g., `"`, `\`, newlines), the result will be malformed JSON and `toJsonNode()` will throw.

For example, template `{"name":"${user}"}` with metadata `user` = `John "Doe"` produces `{"name":"John "Doe""}` → invalid JSON → runtime exception.

This is consistent with how the pattern system works elsewhere (URL, headers), but those paths don't parse the result as JSON. Consider either:
- JSON-escaping substituted values when the result is expected to be valid JSON, or
- Documenting this limitation (e.g., in the UI hint) so users know to avoid JSON-special chars in values when using a JSON template with `parseToPlainText` unchecked.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -340,6 +338,15 @@ private URI buildEncodedUriLegacy(String endpointUrl) {
         return UriComponentsBuilder.fromUriString(endpointUrl).build().encode().toUri();
     }
 
+    private Object getRequestBody(TbMsg msg) {
+        if (StringUtils.isNotBlank(config.getRequestBodyTemplate())) {
+            boolean escapeJson = !config.isParseToPlainText();
+            String processedTemplate = TbNodeUtils.processPattern(config.getRequestBodyTemplate(), msg, escapeJson);
+            return config.isParseToPlainText() ? processedTemplate : JacksonUtil.toJsonNode(processedTemplate);
```
ViacheslavKlimov: If the processed template isn't valid JSON, `JacksonUtil.toJsonNode()` throws a `RuntimeException` that escapes `processMessage()` (which only catches `InterruptedException`). Two consequences:

1. **Unhelpful error** — the user sees a raw Jackson parse error instead of something like "Request body template produced invalid JSON".
2. **Semaphore leak** — when `maxParallelRequestsCount > 0`, the permit acquired at the top of `processMessage()` is never released (the `release()` calls are inside the `subscribe` callbacks). A misconfigured template will hit this on every message, eventually exhausting all permits and deadlocking the node.

The semaphore issue is pre-existing (any exception between `tryAcquire` and `subscribe` leaks), but the body template makes it much more likely — a bad template reliably produces invalid JSON on every invocation.

Consider wrapping this in a try-catch:
```java
try {
    return config.isParseToPlainText() ? processedTemplate : JacksonUtil.toJsonNode(processedTemplate);
} catch (Exception e) {
    throw new RuntimeException("Request body template produced invalid JSON: " + processedTemplate, e);
}
```
And broadening the catch in `processMessage` to release the semaphore on any `Exception`, not just `InterruptedException`.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbRestApiCallNodeConfiguration.java
```diff
@@ -51,6 +51,7 @@ public class TbRestApiCallNodeConfiguration implements NodeConfiguration<TbRestA
     private String proxyPassword;
     private ClientCredentials credentials;
     private boolean ignoreRequestBody;
+    private String requestBodyTemplate;
```
ViacheslavKlimov, line 54: We need the upgrade, sorry

PR #15139, opened 2026-03-03
Add per-format packaging skip flags and replace Spotify dockerfile-maven-plugin

smatvienko-tb:
## Summary

### Why `pkg.disabled` was broken by design

The old `pkg.disabled` property was a single boolean that only controlled `spring-boot-maven-plugin`'s `<skip>` flag — i.e., whether the boot jar (`*-boot.jar`) was produced. It had no effect on the Gradle `.deb`/`.rpm` tasks or the `maven-assembly-plugin` ZIP.

The packaging pipeline has a hard dependency chain:

```
spring-boot:repackage  →  *-boot.jar  →  Gradle buildDeb (packages boot jar into .deb)
   (pkg.disabled)            ↑                    ↑
                        mandatory input      from(mainJar)
```

Gradle's `buildDeb` task copies the boot jar via `from(mainJar)` where `mainJar` resolves to `${project.build.directory}/${project.build.finalName}-boot.jar`. If the boot jar doesn't exist, `buildDeb` fails. This means `pkg.disabled` could never independently skip just the boot jar — it was either "build everything" or "skip the boot jar and hope Gradle doesn't run" (which only worked because non-packaging modules also set `pkg.package.phase=none`).

There was **no way** to express:
- "build the boot jar and `.deb`, but skip `.rpm` and `.zip`" (common on Debian-based CI)
- "build the boot jar only, skip all OS packages" (useful for integration tests)
- "skip just the `.deb`" (e.g., on macOS where `dpkg` isn't available)

The old property was effectively a redundant alias for `pkg.package.phase=none` — it added confusion without adding control.

### No more fat artifacts in `~/.m2/repository`

The old approach declared `.deb` as a Maven dependency with `<type>deb</type>` and used `maven-dependency-plugin:copy` to fetch it from the local Maven repository (`~/.m2/repository`). This meant every `mvn install` wrote ~100 MB+ boot jars and `.deb` files into `~/.m2`, bloating the local repo and slowing down builds.

**Before (old):**
```xml
<!-- msa/tb-node/pom.xml — declared .deb as a Maven dependency -->
<dependency>
    <groupId>org.thingsboard</groupId>
    <artifactId>application</artifactId>
    <classifier>deb</classifier>
    <type>deb</type>
    <scope>provided</scope>
</dependency>
...
<!-- Fetched .deb from ~/.m2 via maven-dependency-plugin:copy -->
<plugin>
    <artifactId>maven-dependency-plugin</artifactId>
    <execution>
        <id>copy-tb-deb</id>
        <phase>package</phase>
        <goal>copy</goal>
        ...
    </execution>
</plugin>
```

**After (new):**
```xml
<!-- No .deb dependency declaration at all — just a plain JAR dep for reactor ordering -->
...
<!-- 1. Enforcer validates .deb exists in upstream target/ (skipped when pkg.skip.deb=true) -->
<plugin>
    <artifactId>maven-enforcer-plugin</artifactId>
    <execution>
        <id>enforce-deb-exists</id>
        <phase>${pkg.deb.phase}</phase>
        <rule>
            <requireFilesExist>
                <file>${main.dir}/application/target/${pkg.name}.deb</file>
            </requireFilesExist>
        </rule>
    </execution>
</plugin>
<!-- 2. Copy .deb directly from upstream module's target/ directory (not ~/.m2) -->
<plugin>
    <artifactId>maven-resources-plugin</artifactId>
    <execution>
        <id>copy-tb-deb</id>
        <phase>${pkg.deb.phase}</phase>
        <goal>copy-resources</goal>
        <directory>${main.dir}/application/target</directory>
        <include>${pkg.name}.deb</include>
    </execution>
</plugin>
```

This change:
- **Eliminates fat artifacts from `~/.m2/repository`** — boot jars and `.deb` files are no longer `install`ed to the local repo
- **Copies `.deb` directly from the upstream module's `target/` directory** — faster, no `.m2` indirection
- **Enforcer check validates `.deb` existence before copy** — clear error message instead of a cryptic `maven-dependency-plugin` resolution failure when the `.deb` is missing
- **Both copy and enforcer are bound to `${pkg.deb.phase}`** — silently skipped when `pkg.skip.deb=true`

### Per-format packaging skip flags (`pkg.skip.*`)

Introduces four independent flags to skip individual packaging artifacts during the build:

| Flag | Artifact | Safe to skip alone? |
|------|----------|---------------------|
| `-Dpkg.skip.bootjar=true` | spring-boot repackage (`*-boot.jar`) | No — also prevents `.deb` build |
| `-Dpkg.skip.deb=true` | Gradle `buildDeb` + Maven `attach-artifact` | Yes |
| `-Dpkg.skip.rpm=true` | Gradle `buildRpm` | Yes |
| `-Dpkg.skip.zip=true` | `maven-assembly-plugin` Windows ZIP | Yes |
| **`-Dpkg.skip=true`** | **All of the above at once** | Yes |

Replaces `pkg.disabled` with a complete set of granular flags. This allows faster dev/test cycles by skipping only the artifacts you don't need:

```bash
# Skip everything (fastest — no boot jars, no .deb, no .rpm, no .zip)
mvn clean install -T6 -DskipTests -Dpkg.skip=true

# Build boot jars + .deb but skip .rpm and .zip
mvn clean install -T6 -DskipTests -Dpkg.skip.rpm=true -Dpkg.skip.zip=true
```

**Key design points:**
- `pkg.deb.phase` property controls the lifecycle phase for `.deb` copy/attach goals — set to `none` when skipping
- `msa/pom.xml` mirrors the `skip-pkg` profile to override its own `<pkg.deb.phase>package</pkg.deb.phase>` (child POM properties have higher priority than parent profile properties in Maven)
- MSA docker modules copy `.deb` files directly from upstream `target/` directories (not `.m2`), bound to `${pkg.deb.phase}`
- `maven-enforcer-plugin` validates `.deb` existence only when not skipping
- Gradle `build.gradle` now receives `skipDeb`/`skipRpm` flags and guards task execution, so individual formats can be skipped even when `pkg.package.phase=package`

### Replace Spotify dockerfile-maven-plugin with exec-maven-plugin

`com.spotify:dockerfile-maven-plugin:1.4.13` is deprecated and broken on modern macOS (Apple Silicon) due to its Java HTTP client being incompatible with current Docker Desktop socket/API.

Replace with `exec-maven-plugin` calling the `docker` CLI directly — the same approach already used by the multiarch `push-docker-amd-arm-images` profiles. All existing Maven property contracts preserved:
- `dockerfile.skip` (default `true`) controls docker build
- `push-docker-image` profile for single-arch push
- `push-docker-amd-arm-images` profiles unchanged (already use exec)

### Fix black-box-tests docker-info dependency

The `docker-info` artifact type was produced by the removed `dockerfile-maven-plugin`. Replace with `pom` type since the upstream docker modules use `pom` packaging. This preserves reactor ordering dependencies.

### Fix enforcer false-negative with non-canonical paths

`msa/*` docker modules used `${basedir}/../..` (non-canonical) for `main.dir`. `maven-enforcer-plugin` 3.5.0's `osIndependentNameMatch()` compares `file.toURI()` vs `file.getCanonicalFile().toURI()` — these differ when the path contains `..`, causing `RequireFilesExist` to report false-negatives. Fixed by replacing with `${maven.multiModuleProjectDirectory}`.

## Files changed

### Commit 1: Packaging skip flags (27 files)
- `pom.xml` — remove `pkg.disabled`, add `pkg.skip.*` properties, `skip-deb`/`skip-pkg` profiles, wire flags into spring-boot/gradle/assembly plugins
- `application/pom.xml`, `edqs/pom.xml`, `monitoring/pom.xml`, `transport/*/pom.xml`, `msa/vc-executor/pom.xml` — `pkg.disabled=false` → `pkg.skip.bootjar=false`
- `msa/pom.xml` — mirror `skip-pkg` profile, set `pkg.deb.phase`
- `msa/*/pom.xml` (docker modules) — remove `type=deb` dependency, replace `maven-dependency-plugin:copy` with `maven-resources-plugin:copy-resources` from upstream `target/`, add `maven-enforcer-plugin` `.deb` existence check
- `packaging/java/build.gradle`, `packaging/js/build.gradle` — honor `skipDeb`/`skipRpm` flags
- `TEST_FAST.md` — document `pkg.skip.*` flags

### Commit 2: Docker plugin replacement (13 files)
- `msa/pom.xml` — remove Spotify extension/pluginManagement, add `exec-maven-plugin` version
- `msa/*/pom.xml` (all docker modules) — replace `dockerfile-maven-plugin` with `exec-maven-plugin`

### Commit 3: black-box-tests fix (1 file)
- `msa/black-box-tests/pom.xml` — `docker-info` → `pom` dependency type

## Test plan
```bash
# Clean state
rm -rf ~/.m2/repository/org/thingsboard
docker image prune -af

# Build with all packaging (default)
mvn -T1C clean install -DskipTests

# Build with Docker images
mvn -T1C clean install -DskipTests -Dpkg.skip.rpm=true -Dpkg.skip.zip=true -Ddockerfile.skip=false

# Verify skip-all works (fastest dev cycle)
mvn -T1C clean install -DskipTests -Dpkg.skip=true
docker image ls  # should show no thingsboard images

# Verify black-box-tests dependency resolution
mvn dependency:resolve -pl msa/black-box-tests -DblackBoxTests.skip=false -DskipTests
```

## Verified (2026-03-19, macOS arm64)

All 13 CE docker images built and pushed successfully. `pkg.skip=true` correctly skips all packaging. `black-box-tests` dependencies resolve without errors.

## Related
- PE PR: https://github.com/thingsboard/thingsboard-pe/pull/4500

 ##  Documentation Impact Assessment                                                                                                                                                                          
                                                                                                                                                                                                           
  No breaking changes to existing docs — I searched the entire site and found:                                                                                                                             
                  
  - pkg.disabled: Zero references in documentation. Nothing to fix.                                                                                                                                        
  - -Ddockerfile.skip=false: Used in 6 places across 3 files. These commands still work because the PR preserves the dockerfile.skip property contract. No fix required.
  - mvn clean install -DskipTests: Used in many places. Still works as before. No fix required.                                                                                                            
                                                                                                                                                                                                           
  Potential enhancement — the new pkg.skip flags are useful for faster dev builds but are not documented anywhere on the site. The PR itself updates TEST_FAST.md in the main repo to recommend mvn clean  
  install -T6 -DskipTests -Dpkg.skip=true.                                                                                                                                                                 
                                                                                                                                                                                                           
  Verdict         

  No documentation updates are strictly required. All existing build commands remain valid. The changes are backward-compatible from a documentation perspective.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,7 +92,6 @@
         <paho.client.version>1.2.5</paho.client.version>
         <paho.mqttv5.client.version>1.2.5</paho.mqttv5.client.version>
         <os-maven-plugin.version>1.7.1</os-maven-plugin.version>
-        <surefire.version>3.5.4</surefire.version>
         <jar-plugin.version>3.5.0</jar-plugin.version>
         <maven-surefire-plugin.version>3.5.4</maven-surefire-plugin.version>
         <maven-install-plugin.version>3.1.4</maven-install-plugin.version>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -35,7 +35,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>thingsboard</pkg.name>
--- a/edqs/pom.xml
+++ b/edqs/pom.xml
@@ -33,7 +33,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>tb-edqs</pkg.name>
--- a/monitoring/pom.xml
+++ b/monitoring/pom.xml
@@ -33,7 +33,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>tb-monitoring</pkg.name>
--- a/msa/js-executor/pom.xml
+++ b/msa/js-executor/pom.xml
@@ -37,7 +37,6 @@
         <docker.name>tb-js-executor</docker.name>
         <pkg.name>tb-js-executor</pkg.name>
         <pkg.type>js</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.linux.dist>${project.build.directory}/package/linux</pkg.linux.dist>
--- a/msa/vc-executor/pom.xml
+++ b/msa/vc-executor/pom.xml
@@ -35,7 +35,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/../..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>tb-vc-executor</pkg.name>
--- a/msa/web-ui/pom.xml
+++ b/msa/web-ui/pom.xml
@@ -39,7 +39,6 @@
         <docker.name>tb-web-ui</docker.name>
         <pkg.name>tb-web-ui</pkg.name>
         <pkg.type>js</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.linux.dist>${project.build.directory}/package/linux</pkg.linux.dist>
--- a/transport/coap/pom.xml
+++ b/transport/coap/pom.xml
@@ -34,7 +34,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/../..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>tb-coap-transport</pkg.name>
--- a/transport/http/pom.xml
+++ b/transport/http/pom.xml
@@ -34,7 +34,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/../..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>tb-http-transport</pkg.name>
--- a/transport/lwm2m/pom.xml
+++ b/transport/lwm2m/pom.xml
@@ -34,7 +34,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/../..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>tb-lwm2m-transport</pkg.name>
--- a/transport/mqtt/pom.xml
+++ b/transport/mqtt/pom.xml
@@ -34,7 +34,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/../..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>tb-mqtt-transport</pkg.name>
--- a/transport/snmp/pom.xml
+++ b/transport/snmp/pom.xml
@@ -36,7 +36,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <main.dir>${basedir}/../..</main.dir>
         <pkg.type>java</pkg.type>
-        <pkg.skip.bootjar>false</pkg.skip.bootjar>
         <pkg.process-resources.phase>process-resources</pkg.process-resources.phase>
         <pkg.package.phase>package</pkg.package.phase>
         <pkg.name>tb-snmp-transport</pkg.name>


smatvienko-tb:
comparison of disk usage

```
rm -rf ~/.m2/repository/org/thingsboard
```

```
mvn clean install -T6 -DskipTests 

du -h ~/.m2/repository/org/thingsboard | tail -1
4.2G
du -h | tail -1
11G
```

```
mvn clean install -T6 -DskipTests -Dpkg.skip.bootjar=true -Dpkg.skip.deb=true -Dpkg.skip.rpm=true -Dpkg.skip.zip=true \
  -pl='!msa/tb-node,!msa/tb,!msa/edqs,!msa/monitoring,!msa/vc-executor-docker,!msa/transport/coap,!msa/transport/http,!msa/transport/lwm2m,!msa/transport/mqtt,!msa/transport/snmp'
  
du -h ~/.m2/repository/org/thingsboard | tail -1
 42M
du -h | tail -1
5.0G   
```

packaging/java/build.gradle
```diff
@@ -173,3 +173,6 @@ buildDeb {
     link("${pkgInstallFolder}/bin/${pkgName}.yml", "${pkgInstallFolder}/conf/${pkgName}.yml")
     link("/etc/${pkgName}/conf", "${pkgInstallFolder}/conf")
 }
+
+buildDeb.onlyIf { !findProperty('skipDeb')?.toBoolean() }
```
smatvienko-tb: Maven always passes property values as strings, so `?.toBoolean()` on a `String` works but adds non-obvious Groovy null-safe chaining. A direct string comparison is clearer and has identical behavior:

```suggestion
buildDeb.onlyIf { findProperty('skipDeb') != 'true' }
buildRpm.onlyIf { findProperty('skipRpm') != 'true' }
```

msa/pom.xml
```diff
@@ -31,6 +31,10 @@
 
     <properties>
         <main.dir>${basedir}/..</main.dir>
+        <!-- msa/* docker modules copy the .deb from target/ directly (not via Maven reactor),
+             so they need the DEB-copy goal to run at the package phase regardless of
+             pkg.package.phase (which is "none" here since these modules don't produce packaging). -->
+        <pkg.deb.phase>package</pkg.deb.phase>
```
smatvienko-tb, line 37: ⚠️ **Property precedence risk.** This child-POM `<properties>` entry may override the `pkg.deb.phase=none` set by the `skip-deb` profile in the root POM when `-Dpkg.skip.deb=true` is active. If testing confirms the profile override does NOT win here, add a mirror profile in this file to be safe:

```xml
    <profiles>
        <profile>
            <!-- Mirror of the root-pom skip-deb profile; ensures pkg.deb.phase=none
                 wins over the <properties> entry above in this module's scope. -->
            <id>skip-deb</id>
            <activation>
                <property>
                    <name>pkg.skip.deb</name>
                    <value>true</value>
                </property>
            </activation>
            <properties>
                <pkg.deb.phase>none</pkg.deb.phase>
            </properties>
        </profile>
    </profiles>
```

msa/tb-node/pom.xml
```diff
@@ -56,45 +54,40 @@
...
                             <goal>copy-resources</goal>
                         </goals>
                         <configuration>
                             <outputDirectory>${project.build.directory}</outputDirectory>
+                            <overwrite>true</overwrite>
                             <resources>
                                 <resource>
-                                    <directory>docker</directory>
-                                    <filtering>true</filtering>
+                                    <directory>${main.dir}/application/target</directory>
```
smatvienko-tb, line 107: 🟡 **Silent failure risk.** Unlike `maven-dependency-plugin:copy` (which fails explicitly when an artifact is missing), `copy-resources` silently succeeds if the source file does not exist, producing an incomplete Docker build context.

Consider adding a `maven-enforcer` check before this execution to fail fast when `pkg.skip.deb` is false but the upstream DEB hasn't been built:

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-enforcer-plugin</artifactId>
    <executions>
        <execution>
            <id>enforce-deb-exists</id>
            <phase>${pkg.deb.phase}</phase>
            <goals><goal>enforce</goal></goals>
            <configuration>
                <rules>
                    <requireFilesExist>
                        <files>
                            <file>${main.dir}/application/target/${pkg.name}.deb</file>
                        </files>
                        <message>DEB artifact not found. Build the 'application' module first or pass -Dpkg.skip.deb=true.</message>
                    </requireFilesExist>
                </rules>
            </configuration>
        </execution>
    </executions>
</plugin>
```

This runs at `${pkg.deb.phase}` so it is automatically skipped (phase `none`) when `-Dpkg.skip.deb=true`.

smatvienko-tb:
## Follow-up: review issues addressed in b80d18b

All items from the code review have been resolved:

### 1. ✅ Maven profile property precedence (was 🔴)

Added a mirror `skip-deb` profile to `msa/pom.xml`. Maven resolves child `<properties>` entries with higher priority than profile properties inherited from a parent POM, so the root-pom profile alone was insufficient. Redeclaring it in `msa/pom.xml` ensures `pkg.deb.phase=none` wins when `-Dpkg.skip.deb=true` is passed.

### 2. ✅ Silent failure mode for missing DEB files (was 🟡)

Added `maven-enforcer-plugin` (`requireFilesExist`, bound to `${pkg.deb.phase}`) to all 10 MSA docker modules. The enforcer is declared before `maven-resources-plugin` so it runs first within the `package` phase and fails immediately with a clear diagnostic message if the upstream DEB hasn't been built. When `-Dpkg.skip.deb=true` is passed, `pkg.deb.phase=none` skips both the enforcer and the copy transparently.

`maven-enforcer-plugin` version `3.5.0` pinned in root `pom.xml` `<pluginManagement>`.

### 3. ✅ Boot jar coupling comment (was 🟡)

Added inline comment near `spring-boot:repackage`'s `<skip>` config in the root POM's `packaging` profile explaining the `pkg.skip.bootjar` / `pkg.skip.deb` coupling.

### 4. ✅ Gradle `onlyIf` clarity (was suggestion)

Simplified `!findProperty('skipDeb')?.toBoolean()` → `findProperty('skipDeb') != 'true'` in both `packaging/java/build.gradle` and `packaging/js/build.gradle`. Maven always passes property values as strings, so a direct string comparison is clearer.

msa/tb-node/pom.xml
```diff
@@ -56,25 +56,21 @@
...
+                <artifactId>maven-enforcer-plugin</artifactId>
                 <executions>
                     <execution>
-                        <id>copy-tb-deb</id>
-                        <phase>package</phase>
-                        <goals>
-                            <goal>copy</goal>
-                        </goals>
+                        <id>enforce-deb-exists</id>
+                        <phase>pre-integration-test</phase>
```
ViacheslavKlimov: The `enforce-deb-exists` execution is hardcoded to `pre-integration-test`, which runs during `mvn install`. When `-Dpkg.skip.deb=true` is passed, the DEB copy execution (at phase `${pkg.deb.phase}` = `none`) is skipped, but the enforcer still runs and should fail because `${project.build.directory}/${pkg.name}.deb` doesn't exist.

This affects all 10 MSA docker modules (`msa/tb-node`, `msa/tb`, `msa/edqs`, `msa/monitoring`, `msa/vc-executor-docker`, `msa/transport/{coap,http,lwm2m,mqtt,snmp}`).

Consider adding `<skip>${pkg.skip.deb}</skip>` to the enforcer `<configuration>` so the check is bypassed when DEB generation is intentionally skipped:

```xml
<configuration>
    <skip>${pkg.skip.deb}</skip>
    <rules>
        ...
    </rules>
</configuration>
```

(The test plan says `-Dpkg.skip.deb=true` completed successfully — if that's the case, I'd appreciate understanding what mechanism prevents the enforcer from failing, since I can't find one in the POM configuration.)

msa/tb/pom.xml
```diff
@@ -55,44 +55,22 @@
...
-                                    <outputDirectory>${project.build.directory}/docker-cassandra</outputDirectory>
-                                </artifactItem>
-                            </artifactItems>
+                            <rules>
+                                <requireFilesExist>
+                                    <files>
+                                        <file>${project.build.directory}/docker-postgres/${pkg.name}.deb</file>
+                                        <file>${project.build.directory}/docker-cassandra/${pkg.name}.deb</file>
+                                    </files>
+                                    <message>DEB artifact not found in ${project.build.directory}/docker-postgres/${pkg.name}.deb — build the upstream module first or pass -Dpkg.skip.deb=true.</message>
```
ViacheslavKlimov: **Low-confidence observation:** The `<message>` only references the `docker-postgres` path, but the `<requireFilesExist>` rule also checks `docker-cassandra`. If only the cassandra file were missing, the error message would be misleading. In practice both files come from the same source (`application/target`), so this is unlikely to cause confusion, but a more accurate message might reference both paths or use a generic phrasing like:

```
DEB artifact(s) not found — build the upstream module first or pass -Dpkg.skip.deb=true.
```

msa/tb-node/pom.xml
smatvienko-tb: Thanks for reviewing! The  execution uses `<phase>${pkg.deb.phase}</phase>` (not a hardcoded literal), where `pkg.deb.phase` is set to `none` by the `skip-deb` profile when `-Dpkg.skip.deb=true` is active — see [msa/pom.xml profile](https://github.com/smatvienko-tb/thingsboard/blob/improvement/per-format-packaging-skip-flags/msa/pom.xml). So the enforcer is silently skipped along with the copy-resources execution when DEB generation is disabled. This was added in the same commit that addressed the earlier self-review notes.

msa/tb/pom.xml
smatvienko-tb: This was already fixed in the follow-up commit: the enforcer now checks only the single source file `${main.dir}/application/target/${pkg.name}.deb` (the upstream build artifact), and the phase was also changed from the hardcoded `pre-integration-test` to `${pkg.deb.phase}` so the check is silently skipped when `-Dpkg.skip.deb=true`. The intermediate state you reviewed (checking both docker-postgres and docker-cassandra target paths) was superseded.

pom.xml
```diff
@@ -179,6 +206,43 @@
...
+        <profile>
+            <id>skip-pkg</id>
+            <activation>
+                <property>
+                    <name>pkg.skip</name>
+                    <value>true</value>
+                </property>
+            </activation>
+            <properties>
+                <pkg.skip.bootjar>true</pkg.skip.bootjar>
```
ViacheslavKlimov, line 237: Same Maven property precedence issue you documented for `pkg.deb.phase` in `msa/pom.xml`. Every packaging module (`application`, `edqs`, `monitoring`, `msa/vc-executor`, `transport/*`) declares `<pkg.skip.bootjar>false</pkg.skip.bootjar>` in its own `<properties>`. Since child POM properties override parent profile properties, this `<pkg.skip.bootjar>true</pkg.skip.bootjar>` from the `skip-pkg` profile never actually reaches those modules.

So `-Dpkg.skip=true` skips .deb, .rpm, .zip, and the attach-artifact (all correct), but `spring-boot:repackage` still runs in every packaging module because they see `pkg.skip.bootjar=false`.

The simplest fix: remove `<pkg.skip.bootjar>false</pkg.skip.bootjar>` from all child modules. The root POM default is already `false`, so the declarations are redundant — and they block the profile override. The same approach used for `pkg.skip.deb`/`pkg.skip.rpm`/`pkg.skip.zip` (only declared in root, not in children) would work here too.

pom.xml
```diff
@@ -70,6 +94,9 @@
         <os-maven-plugin.version>1.7.1</os-maven-plugin.version>
         <surefire.version>3.5.4</surefire.version>
         <jar-plugin.version>3.5.0</jar-plugin.version>
+        <maven-surefire-plugin.version>3.5.4</maven-surefire-plugin.version>
```
ViacheslavKlimov, line 96: `surefire.version` at line 95 is now unreferenced — all usages were migrated to `maven-surefire-plugin.version`. Can be removed to avoid confusion about which property controls the surefire version.

pom.xml
smatvienko-tb: CE review fixes (pushed to ce-fork/improvement/per-format-packaging-skip-flags):                                                                                                                                                               
  - pkg.skip.bootjar precedence bug: Removed redundant <pkg.skip.bootjar>false</pkg.skip.bootjar> from all 11 CE child modules so -Dpkg.skip=true actually skips spring-boot:repackage
  - Unused surefire.version: Removed (superseded by maven-surefire-plugin.version)                                                                                                                                                               
  - Gradle toBoolean() and msa/pom.xml mirror profiles were already correct

PR #15141, opened 2026-03-04
RPC monitoring for the monitoring microservice

smatvienko-tb:
## Summary

- Adds two-way RPC health checking alongside the existing transport telemetry checks
- Probes the full round-trip: REST API → rule engine partition → transport microservice → device → response, with value validation
- Multiple devices per target cover different topic/partition assignments by hash, so partial microservice failures (affecting only a fraction of devices) are detectable
- Starting with MQTT; extensible to other transports via `RpcTransportType` enum

## Design

```
monitoring.rpc.mqtt:
  enabled: true
  request_timeout_ms: 4000
  qos: 1
  targets:
    - base_url: tcp://mqtt.example.com:1883
      device_id: <uuid>        # pre-existing device
      access_token: <token>
      label: shard0            # used in metric keys
    - ...                      # add one per partition shard
```

**Check flow per device:**
1. On startup: persistent MQTT connection, subscribed to `v1/devices/me/rpc/request/+`
2. On each check: `POST /api/rpc/twoway/{deviceId}` with `{"method":"monitoringCheck","params":{"value":"<uuid>"}}`
3. Paho callback thread receives the request, publishes params back to `v1/devices/me/rpc/response/{requestId}`
4. REST call returns the echoed value; checker validates it matches and reports `mqttRpc_<label>RpcRoundTrip` latency

**Key changes to base classes:**
- `BaseHealthChecker.check(WsClient)` — `final` removed to allow RPC override (WsClient-based validation is skipped for RPC)
- `reporter` and `stopWatch` — made `protected` for subclass access

## Test plan

- [ ] `RpcMonitoringTargetTest` — label fallback, UUID parsing
- [ ] `BaseRpcHealthCheckerTest` — happy path, wrong value, null response, REST exception, request body structure, correct device ID
- [ ] `MqttRpcHealthCheckerTest` — connect options, RPC topic subscription, callback echo, idempotent init, destroy

### DIFF
--- a/monitoring/pom.xml
+++ b/monitoring/pom.xml
@@ -44,6 +44,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.thingsboard.common</groupId>
             <artifactId>data</artifactId>
         </dependency>
--- a/monitoring/src/main/java/org/thingsboard/monitoring/config/rpc/MqttRpcMonitoringConfig.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/config/rpc/MqttRpcMonitoringConfig.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright © 2016-2026 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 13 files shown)


smatvienko-tb:
Closing this PR — the experimental shape doesn't fit the rest of the monitoring service and is being superseded.

Filed as a feature request with the proposed redesign in the issues

TL;DR of what changes:

- RPC becomes an opt-in `rpc:` sub-block on each existing `TransportMonitoringTarget`, **not** a parallel `BaseMonitoringService` with its own YAML and its own MQTT client.
- The auto-provisioned device + credentials are reused (no hand-managed `device_id` / `access_token` in YAML).
- Secure transport settings (`ssl://`, client certs) are inherited from the transport target and require no extra config.
- One MQTT session per host per cycle, covering both telemetry uplink and RPC echo on the same client.
- `BaseHealthChecker.check(WsClient)` regains its `final` modifier; `reporter` and `stopWatch` go back to `private`. RPC plugs in via a new `protected void doRpcCheck(...)` template hook.

PR #15150, opened 2026-03-04
feat: support combined TLS PEM cert for Edge gRPC server

smatvienko-tb:
## Summary

- Makes `EDGES_RPC_SSL_PRIVATE_KEY` optional (empty default), enabling combined PEM files (cert chain + private key in one file) as used in `MQTT_SSL_PEM_CERT` / `COAP_DTLS_PEM_CERT`
- Adds `EDGES_RPC_SSL_PEM_KEY_PASSWORD` for encrypted (passphrase-protected) private keys, aligned with `MQTT_SSL_PEM_KEY_PASSWORD` naming
- `setupSsl()` delegates PEM parsing to `SslUtil.readCertFile()` / `SslUtil.readPrivateKey()` from `common-util`, the same battle-tested path used by MQTT/CoAP/LwM2M transports — handles `PEMEncryptedKeyPair`, `PKCS8EncryptedPrivateKeyInfo`, `PEMKeyPair`, `PrivateKeyInfo`, and auto-registers BouncyCastle provider
- `ResourceUtils.getInputStream()` used via `readResourceAsString()` helper to support both classpath and filesystem paths uniformly

## Configuration

| Scenario | `EDGES_RPC_SSL_CERT` | `EDGES_RPC_SSL_PRIVATE_KEY` |
|---|---|---|
| Combined PEM (cert + key in one file) | path to combined `.pem` | _(empty, default)_ |
| Separate files (previous behavior) | path to cert chain | path to private key |
| Encrypted private key | combined or cert file | empty or key file |

Set `EDGES_RPC_SSL_PEM_KEY_PASSWORD` for passphrase-protected keys.

## Test plan

- `EdgeGrpcSslSetupTest` (no Spring context, standalone):
  - RSA-4096 and Ed25519 — combined PEM and separate PEM
  - Encrypted private key — combined and separate layouts via `@ParameterizedTest`
  - Combined PEM without key → expects `IllegalArgumentException`
  - Wrong client CA → TLS handshake fails (`TRANSIENT_FAILURE`)
  - Plaintext client → connection fails
  - Client with no CA cert → handshake fails

### DIFF
--- a/application/src/test/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcSslSetupTest.java
+++ b/application/src/test/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcSslSetupTest.java
@@ -119,7 +119,7 @@ class EdgeGrpcSslSetupTest {
         private static CertKey buildCert(KeyPairGenerator kpg, int keySize, String sigAlg) throws Exception {
             if (keySize > 0) kpg.initialize(keySize);
             KeyPair kp = kpg.generateKeyPair();
-            X500Name subject = new X500Name("CN=edge-grpc-test");
+            X500Name subject = new X500Name("CN=localhost");
             Date now = new Date();
             ContentSigner signer = new JcaContentSignerBuilder(sigAlg).build(kp.getPrivate());
             X509Certificate cert = new JcaX509CertificateConverter().getCertificate(
@@ -149,7 +149,7 @@ class EdgeGrpcSslSetupTest {
                 () -> startServer(combined.toString(), null, null, TestSocketUtils.findAvailableTcpPort()),
                 keyType);
 
-        assertClientConnectivity(server, ck.cert, ConnectivityState.READY);
+        assumeClientConnectivitySupported(server, ck.cert, keyType);
     }
 
     @ParameterizedTest(name = "separatePem_{0}")
@@ -165,7 +165,7 @@ class EdgeGrpcSslSetupTest {
                 () -> startServer(certFile.toString(), keyFile.toString(), null, TestSocketUtils.findAvailableTcpPort()),
                 keyType);
 
-        assertClientConnectivity(server, ck.cert, ConnectivityState.READY);
+        assumeClientConnectivitySupported(server, ck.cert, keyType);
     }
 
     // ======================================================================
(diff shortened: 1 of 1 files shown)


smatvienko-tb:
## Code Review: PR #15150 — Support Combined TLS PEM Cert for Edge gRPC Server

### Overview
This PR aligns the Edge gRPC server's SSL setup with the pattern used by MQTT/CoAP/LwM2M transports: supporting combined PEM files (cert chain + private key in one file), encrypted private keys, and delegating parsing to the shared `SslUtil` utility. The direction is correct and the configuration design is clean.

---

### Production Code (`EdgeGrpcService.java`)

**Positives:**
- Clean `setupSsl()` extraction from inline code in `onStartUp()`
- Reuses `SslUtil` (battle-tested path) instead of reimplementing PEM parsing
- `readResourceAsString()` is a clean, focused helper
- Making `private_key` default to `""` is backwards-compatible via Spring `@Value`

**Issues:**

#### 1. Missing null-check for `privateKey` in the separate-files branch

In the combined-PEM branch there is a null guard, but in the `else` (separate files) branch there is none:

```java
} else {
    privateKey = SslUtil.readPrivateKey(readResourceAsString(privateKeyResource), keyPassword);
    // ← null check missing here
}
```

If `privateKeyResource` points to a file that contains no private key, `privateKey` will be `null` and the method will throw a `NullPointerException` later rather than a meaningful `IllegalArgumentException`. The same guard from the combined-PEM path should be applied here.

#### 2. Unnecessary PEM round-trip

`SslUtil.readPrivateKey()` returns a `PrivateKey`, which is immediately re-serialized to PEM via `JcaPEMWriter`, only to be fed as a `ByteArrayInputStream` to `useTransportSecurity()`. Same for certificates. This is slightly wasteful (decrypt → re-serialize as plain PEM in memory). This may be unavoidable given the `NettyServerBuilder` API, but a brief comment explaining why the round-trip is necessary would help future readers.

#### 3. `keyPassword` stored as `String`

Passwords stored as `String` cannot be zeroed from memory (immutable). This is a common Spring `@Value` limitation, but since encrypted key support is the explicit new feature here, a comment acknowledging this limitation would be appropriate.

---

### Test Code (`EdgeGrpcSslSetupTest.java`)

**Positives:**
- Comprehensive scenario coverage (combined PEM, separate PEM, encrypted keys, wrong CA, plaintext client, no-CA client)
- Clean `@ParameterizedTest` usage with `@EnumSource`
- Proper temp file cleanup in `@AfterEach`
- Good use of `Assumptions.assumeTrue()` for Ed25519 environment differences
- The class-level Javadoc clearly enumerates all test scenarios

#### Critical: Tests don't exercise the production code path

The class Javadoc says:
> "`startServer` mirrors `EdgeGrpcService.setupSsl()`"

But `startServer()` **re-implements PEM parsing** using raw BouncyCastle APIs instead of calling `SslUtil.readCertFile()` / `SslUtil.readPrivateKey()`. The tests verify a parallel implementation, not the actual production `setupSsl()` path. A bug in the `SslUtil` integration would not be caught by these tests.

Suggested fix: extract the cert/key loading logic from `setupSsl()` into a package-accessible static helper, or restructure the test to call `setupSsl()` directly (e.g. via reflection or by making it package-private). This would ensure the tests actually validate what runs in production.

**Minor issues:**

- `startServer()` is ~80 lines. Splitting into `buildCertChainStream()` / `buildPrivateKeyStream()` helpers would improve readability and eliminate the duplicated decrypt logic for combined vs. separate paths.
- `pollForState()` with a 50 ms sleep loop is functional but could be flaky under load in CI. gRPC's `notifyWhenStateChanged` callback would be more reliable.
- The encrypted-key tests are RSA-only; a brief comment explaining why PKCS8-encrypted Ed25519 is not covered would pre-empt the question.

---

### Configuration (`thingsboard.yml`)

No issues. Comments are clear and the configuration table in the PR description is excellent.

---

### Summary

| Area | Status |
|---|---|
| Correctness | **Needs fix** — NPE risk when separate key file contains no private key |
| Test coverage | **Needs fix** — tests exercise a re-implementation, not production `setupSsl()` |
| Security | Minor — `keyPassword` as `String` worth a comment |
| Style | Minor — import ordering in test file; `startServer()` length |
| Performance | Acceptable — PEM round-trip is slightly wasteful but not a blocker |

smatvienko-tb:
closed in favor https://github.com/thingsboard/thingsboard/pull/15205

PR #15166, opened 2026-03-06
Open-api spec improvements: make schema fields ordered

dashevchenko:
## Pull Request description

**Schema property ordering**

  Previously, the order of properties within each schema in the generated OpenAPI spec was non-deterministic — it depended on the order in which Jackson resolved types at startup, which could vary between runs. This made spec diffs noisy and caused unnecessary churn
  in generated client code.

  Properties are now ordered to match the Java class field declaration order. The implementation uses Class.getDeclaredFields() walking the full class hierarchy (superclass fields first) for ordering, and Jackson's BeanDescription to resolve the correct JSON property
   names (respecting @JsonProperty overrides). Getter-only properties with no backing field are appended at the end.

  **Stable example token**

  The credentials-expired error example was using StringUtils.randomAlphanumeric(30) to generate the refresh token value, producing a different spec on every startup. Replaced with a hardcoded example value.

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/config/SwaggerConfiguration.java
+++ b/application/src/main/java/org/thingsboard/server/config/SwaggerConfiguration.java
@@ -306,7 +306,7 @@ public class SwaggerConfiguration {
                         }
                     } else if (schema != null && schema.getProperties() != null && !schema.getProperties().isEmpty()) {
                         try {
-                            var beanDesc = JacksonUtil.OBJECT_MAPPER.getSerializationConfig().introspect(javaType);
+                            var beanDesc = Json.mapper().getSerializationConfig().introspect(javaType);
                             var orderedNames = resolvePropertyOrder(cls, beanDesc);
                             if (!orderedNames.isEmpty()) {
                                 @SuppressWarnings("unchecked")
@@ -320,7 +320,7 @@ public class SwaggerConfiguration {
                                 schema.setProperties(reordered);
                             }
                         } catch (Exception ignored) {
-                            // keep original order if introspection fails
+                            log.trace("Failed to resolve property order for {}", cls.getName(), ignored);
                         }
                     }
                 }


application/src/main/java/org/thingsboard/server/config/SwaggerConfiguration.java
```diff
@@ -301,6 +304,24 @@ ModelConverter mapAwareConverter() {
                                 schema.setProperties(null);
                             }
                         }
+                    } else if (schema != null && schema.getProperties() != null && !schema.getProperties().isEmpty()) {
+                        try {
+                            var beanDesc = JacksonUtil.OBJECT_MAPPER.getSerializationConfig().introspect(javaType);
```
ViacheslavKlimov: **Low-confidence observation:** The schema properties are resolved using Swagger's `Json.mapper()` (line 283), but the introspection here uses `JacksonUtil.OBJECT_MAPPER`. If these two mappers are configured differently (e.g., different mixins, visibility rules, or modules), `beanDesc` could resolve different JSON property names than what the schema contains, leading to some properties falling through to the `putIfAbsent` fallback rather than being placed in field-declaration order.

In practice this is unlikely to cause issues since both are standard ObjectMappers and the fallback handles mismatches gracefully, but using `Json.mapper().getSerializationConfig().introspect(javaType)` would be more correct since that's the mapper the schema was built with.

application/src/main/java/org/thingsboard/server/config/SwaggerConfiguration.java
```diff
@@ -301,6 +304,24 @@ ModelConverter mapAwareConverter() {
...
+                                Map<String, Schema> current = schema.getProperties();
+                                var reordered = new LinkedHashMap<String, Schema>();
+                                for (String name : orderedNames) {
+                                    Schema prop = current.get(name);
+                                    if (prop != null) reordered.put(name, prop);
+                                }
+                                current.forEach((k, v) -> reordered.putIfAbsent(k, v));
+                                schema.setProperties(reordered);
+                            }
+                        } catch (Exception ignored) {
```
ViacheslavKlimov, line 322: **Low-confidence observation:** Silently swallowing all exceptions here could make debugging difficult if introspection fails for unexpected reasons. Consider at least a `log.trace("Failed to resolve property order for {}", cls.getName(), ignored)` to leave a breadcrumb.

application/src/main/java/org/thingsboard/server/config/SwaggerConfiguration.java
dashevchenko: Fixed

application/src/main/java/org/thingsboard/server/config/SwaggerConfiguration.java
dashevchenko: Fixed.

PR #15187, opened 2026-03-09
Fix "No data" overlay not hiding when widget data becomes available after empty state

mtsymbarov-del:
## Pull Request description
Before:

Знімок екрана 2026-03-09 о 13 54 40
After:
Знімок екрана 2026-03-09 о 13 53 15

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/widget.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/widget.component.ts
@@ -681,6 +681,7 @@ export class WidgetComponent extends PageComponent implements OnInit, OnChanges,
   }
 
   private reInitImpl() {
+    this.displayNoData = false;
     this.onDestroy();
     if (!this.typeParameters.useCustomDatasources) {
       this.createDefaultSubscription().subscribe({
@@ -724,7 +725,6 @@ export class WidgetComponent extends PageComponent implements OnInit, OnChanges,
           subscriptionChanged = subscriptionChanged || subscription.onAliasesChanged(aliasIds);
         }
         if (subscriptionChanged && !this.typeParameters.useCustomDatasources) {
-          this.displayNoData = false;
           this.reInit();
         }
       }
@@ -738,7 +738,6 @@ export class WidgetComponent extends PageComponent implements OnInit, OnChanges,
           subscriptionChanged = subscriptionChanged || subscription.onFiltersChanged(filterIds);
         }
         if (subscriptionChanged && !this.typeParameters.useCustomDatasources) {
-          this.displayNoData = false;
           this.reInit();
         }
       }


ui-ngx/src/app/modules/home/components/widget/widget.component.ts
```diff
@@ -589,7 +590,11 @@ export class WidgetComponent extends PageComponent implements OnInit, OnChanges,
             }
           } else {
             this.loadingData = false;
-            this.displayNoData = true;
+            if (!this.reInitInProgress &&
+              !Object.values(this.widgetContext.subscriptions).some(s => s.loadingData)) {
+              this.displayNoData = true;
+            }
+            this.reInitInProgress = false;
```
vvlladd28: `reInitInProgress` is only cleared on the `else` path here — if `displayWidgetInstance()` returns true and we take the `if` branch above, the flag stays `true` indefinitely until the next reInit overwrites it. It happens to work today because every `reInit` path explicitly re-sets the flag to `true` before calling `onInit()`, but the stuck-true state is fragile and confusing to reason about: any future code that reads `reInitInProgress` outside this exact raf callback would see a stale value. I'd move the reset out of the `else` branch so the flag is cleared unconditionally after the check (e.g., right before `this.detectChanges()`), or duplicate the reset into both branches.

A short comment explaining the flag's purpose ("suppress `displayNoData = true` during the reInit window while subscriptions are still loading") would also help — the name alone doesn't convey the intent.

ui-ngx/src/app/modules/home/components/widget/widget.component.ts
```diff
@@ -976,6 +986,11 @@ export class WidgetComponent extends PageComponent implements OnInit, OnChanges,
       dataLoading: (subscription) => {
         if (this.loadingData !== subscription.loadingData) {
           this.loadingData = subscription.loadingData;
+          if (!this.loadingData
+            && !Object.values(this.widgetContext.subscriptions).some(s => s.loadingData)
+            && !this.displayWidgetInstance()) {
+            this.displayNoData = true;
```
vvlladd28: This new branch handles the `loading → not loading` transition (show "No data" if no subscription is loading and the widget isn't displayable), but nothing handles the reverse direction. If the overlay is currently visible and a subscription starts loading again (e.g., new time window, manual refresh), `displayNoData` stays `true` throughout the load. Depending on the intended UX, you may want to also clear `displayNoData = false` when `this.loadingData` transitions from `false → true`, so the overlay doesn't cover the loading state.

Also, minor readability: the `!Object.values(this.widgetContext.subscriptions).some(s => s.loadingData)` check is repeated here and on line 594 — worth extracting to a small helper like `private isAnySubscriptionLoading(): boolean` so the intent is obvious at call sites.

vvlladd28 (commented):
## Review summary

Reviewed 1 changed file in [Fix "No data" overlay not hiding when widget data becomes available after empty state](https://github.com/thingsboard/thingsboard/pull/15187). Left 2 comment(s) inline.

The fix addresses a real bug — setting `displayNoData = false` on `onDataUpdated`/`onLatestDataUpdated` is the right approach to clear a stale overlay when data arrives. The extra guards around setting `displayNoData = true` (during reInit and while subscriptions are still loading) also look correct.

A couple of robustness concerns called out inline, plus some related inconsistencies in unchanged code that surfaced while reading context:

## Additional findings

- **widget.component.ts:573-590** — The `if (this.displayWidgetInstance())` branch in the `onInit()` raf callback never resets `displayNoData = false`. If the flag was already true (e.g., from a prior empty state, or from the `dataLoading` callback that transiently sets it during subscription init before the subscription is added to `widgetContext.subscriptions`), the overlay lingers until the `onDataUpdated` raf fires and clears it. In practice the `onDataUpdated` raf runs in the same frame so it's not visible, but the logic is indirect — it would be safer and clearer to reset `displayNoData = false` inside the `if` branch when the widget is about to render.

- **widget.component.ts:980-985** — `onInitialPageDataChanged` and `forceReInit` call `this.reInit()` without first setting `this.displayNoData = false`, unlike the alias/filter-change handlers at lines 735 and 749. This inconsistency means a reInit triggered by page data changes or a forced reInit can leave the previous "No data" overlay visible during the reInit window. Consider moving `this.displayNoData = false;` into `reInit()` itself so all callers get uniform behavior.

vvlladd28 (approved):

PR #15199, opened 2026-03-10
Fix EntityViewControllerTest MQTT port collision with other test contexts

smatvienko-tb:
## Root Cause

`EntityViewControllerTest` was importing `MQTT_PORT` from `AbstractMqttIntegrationTest`:

```java
import static org.thingsboard.server.transport.mqtt.AbstractMqttIntegrationTest.MQTT_PORT;
import static org.thingsboard.server.transport.mqtt.AbstractMqttIntegrationTest.MQTT_URL;
```

`MQTT_PORT` is a `static final` field initialized **once per JVM** via `TestSocketUtils.findAvailableTcpPort()`. Multiple test classes that share this constant (`EntityViewControllerTest`, `MqttGatewayRateLimitsTest`, `DeviceEdgeTest`) each produce a **different Spring context cache key** (due to different `@TestPropertySource` / `@ContextConfiguration`), so Spring creates separate `ApplicationContext` instances. Each context starts its own `MqttTransportService` and calls `bind(host, MQTT_PORT)` — the second one throws `BindException: Address already in use`.

This is not an OS TOCTOU race — it is a deterministic collision: same JVM fork, same static port constant, multiple Spring contexts all trying to bind the same port.

## Fix

Define `MQTT_PORT` and `MQTT_URL` as class-level statics in `EntityViewControllerTest` itself, so its Spring context gets its own independently allocated port:

```java
static final int MQTT_PORT = TestSocketUtils.findAvailableTcpPort();
static final String MQTT_URL = "tcp://localhost:" + MQTT_PORT;
```

## Test plan
- [x] `EntityViewControllerTest` passes without `BindException` when run in the same Surefire fork alongside `MqttGatewayRateLimitsTest` / `DeviceEdgeTest`
- [x] All existing `EntityViewControllerTest` telemetry/attribute upload tests still pass (MQTT client connects to the new independent port)

### DIFF
--- a/application/src/test/java/org/thingsboard/server/controller/EntityViewControllerTest.java
+++ b/application/src/test/java/org/thingsboard/server/controller/EntityViewControllerTest.java
@@ -97,6 +97,12 @@ import static org.thingsboard.server.dao.model.ModelConstants.NULL_UUID;
 @ContextConfiguration(classes = {EntityViewControllerTest.Config.class})
 @DaoSqlTest
 public class EntityViewControllerTest extends AbstractControllerTest {
+    // Must NOT be imported from AbstractMqttIntegrationTest. That field is a static final initialized
+    // once per JVM. Other test classes (e.g. MqttGatewayRateLimitsTest, DeviceEdgeTest) share the same
+    // constant but produce a different Spring context cache key, so Spring creates a separate
+    // ApplicationContext for each of them. Every context starts its own MqttTransportService and tries
+    // to bind the same port, causing BindException when tests run in the same Surefire JVM fork.
+    // Declaring the port here gives this context its own independently allocated port.
     static final int MQTT_PORT = TestSocketUtils.findAvailableTcpPort();
     static final String MQTT_URL = "tcp://localhost:" + MQTT_PORT;
 


smatvienko-tb:
## Code Review

### Overview
This is a targeted fix for a deterministic `BindException` (MQTT port collision) when `EntityViewControllerTest` runs in the same Surefire JVM fork alongside `MqttGatewayRateLimitsTest` / `DeviceEdgeTest`. The root cause analysis in the PR description is accurate and well-documented.

### Code Quality

**Correctness** ✅
- The fix is correct. By calling `TestSocketUtils.findAvailableTcpPort()` at class-level in `EntityViewControllerTest`, Spring's `DynamicPropertySource` receives a unique port for this context's cache key, preventing the `bind()` collision.
- The existing `log.warn("transport.mqtt.bind_port = {}", MQTT_PORT)` already ensures the allocated port is observable in CI logs — no extra instrumentation needed.

**Minimal footprint** ✅
- 10 additions / 2 deletions. The change is surgical and touches only what is necessary.

**Comment quality** ✅
- The block comment is unusually detailed for test code, but given that this is non-obvious infrastructure behaviour (static-final sharing across context cache keys), the explanation earns its place.

### Potential Concerns

1. **Other callers of `AbstractMqttIntegrationTest.MQTT_PORT` / `MQTT_URL`**
   The PR description names `MqttGatewayRateLimitsTest` and `DeviceEdgeTest` as classes that still import from `AbstractMqttIntegrationTest`. If those two classes produce *different* Spring context cache keys from each other (e.g. different `@ContextConfiguration` or `@TestPropertySource` sets), they can still collide between themselves. This PR fixes the `EntityViewControllerTest` symptom, but a follow-up audit of the remaining importers may be warranted.

2. **TOCTOU gap in `TestSocketUtils.findAvailableTcpPort()`**
   This is the idiomatic Spring approach and acceptable for test code, but worth noting: the port is reserved at class-load time, not at bind time, so a parallel process could theoretically grab it. This is a pre-existing limitation of the pattern, not introduced by this PR.

3. **`TestSocketUtils` deprecation** (minor)
   `TestSocketUtils.findAvailableTcpPort()` was deprecated in Spring Framework 6.x in favour of OS-assigned ports (bind to 0 and retrieve). This is the existing pattern across the test suite, so consistency is fine here; just something to keep in mind for a broader future cleanup.

### Summary

The fix is correct, minimal, and well-explained. The immediate flakiness problem is resolved. Suggest opening a follow-up issue to audit remaining importers of `AbstractMqttIntegrationTest.MQTT_PORT` to rule out residual collisions.

**Verdict: LGTM** — ready to merge with the follow-up note above.

PR #15201, opened 2026-03-10
Fix flaky tests: await async save futures before assertions

smatvienko-tb:
## Summary
- `TimeseriesServiceNoSqlTest.shouldSaveEntryOfEachTypeWithTtl`: `tsService.save()` result was not awaited before querying, causing intermittent `expected:<5> but was:<2>` failures
- `EntityServiceTest.testFindTenantTelemetry`: `timeseriesService.save()` and `attributesService.save()` results were not awaited, causing intermittent `expected ["45.5"] but got [""]` failures

Both failures observed in TeamCity builds (build 94962 for smoke/logic, build 94863 for controller/edge):
- https://builds.thingsboard.io/buildConfiguration/ThingsBoard_TestSmokeLogic/94962
- https://builds.thingsboard.io/buildConfiguration/ThingsBoard_TestServiceControllerEdge_2/94863

## Root cause
The async save futures were not awaited (`.get()` not called), so the immediately following query would sometimes execute before the data was persisted to the database.

## Fix
Call `.get()` on the returned `ListenableFuture` from `tsService.save()`, `timeseriesService.save()`, and `attributesService.save()` to ensure writes complete before assertions.

## Test plan
- [x] Verify `TimeseriesServiceNoSqlTest.shouldSaveEntryOfEachTypeWithTtl` passes consistently
- [x] Verify `EntityServiceTest.testFindTenantTelemetry` passes consistently

### DIFF
--- a/application/src/test/java/org/thingsboard/server/service/entitiy/EntityServiceTest.java
+++ b/application/src/test/java/org/thingsboard/server/service/entitiy/EntityServiceTest.java
@@ -115,6 +115,8 @@ import java.util.Map;
 import java.util.Random;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


smatvienko-tb:
**Review follow-up**

Found and fixed one issue: the two new `.get()` calls in `EntityServiceTest.testFindTenantTelemetry` were unbounded — if a future ever stalled (executor issue, bug), the test would block the build indefinitely.

Fixed in the latest commit by replacing with `.get(TIMEOUT, TimeUnit.SECONDS)` (`TIMEOUT = 30` inherited from `AbstractWebTest`) and adding `TimeoutException` to the `throws` clause, consistent with how `TimeseriesServiceNoSqlTest` already uses `.get(MAX_TIMEOUT, TimeUnit.SECONDS)`.

LGTM otherwise.

PR #15202, opened 2026-03-10
Fix WS reconnect loop and notification spam when session limit is reached

smatvienko-tb:
## Problem

When a tenant or customer hits the WebSocket session limit, the server closes the connection with **code 1008** (Policy Violation) and a reason like *"Max tenant sessions limit reached!"*. The client then entered a tight reconnect loop:

1. **No backoff** — the client reconnected every fixed 2 s, continuously hammering the server while it was still over the limit. The production HAR captured **560+ connections in ~33 minutes** (one every ~2–3 s), each lasting ~400 ms with zero received messages.

2. **Notification spam** — `onOpen()` always resets `isReconnect = false`, so every subsequent `onClose` was treated as the *first* failure in a cycle. This caused a new red error notification to be dispatched on every attempt, resulting in **fast-blinking red alerts** that the user could not dismiss.

The same symptoms appeared for a tenant whose public dashboard was switched back to private: the public WS session began failing immediately, and the aggressive reconnect loop prevented the tenant user from operating the home page.

## Root cause

`onOpen()` unconditionally sets `isReconnect = false`. When the server rate-limits a connection (opens the WS handshake then immediately closes with 1008), `onOpen` runs, resets the flag, and the next `onClose` sees `!isReconnect = true` again — showing another notification and resetting the 2 s timer. This repeats indefinitely.

## Changes (`ui-ngx/src/app/core/ws/websocket.service.ts`)

| | Before | After |
|---|---|---|
| Reconnect interval | Fixed 2 s forever | Exponential backoff: 2 s → 4 s → 8 s … capped at **60 s** |
| Session-limit close (1008) | Same 2 s retry | Immediately jumps to the **60 s** cap |
| Error notification | Shown on **every** `onClose` with error code | Shown **once** per reconnect cycle; suppressed until a productive connection (first message received) |
| On logout / re-login | — | Backoff and notification state fully reset |

Key design decisions:
- `reconnectAttempts` is reset to 0 only in `onMessage()` (connection is productive), **not** in `onOpen()`. This prevents the open→immediately-closed cycle from resetting the counter.
- `reconnectErrorShown` guards `showWsError()` so the blinking notification appears only once per outage.
- For 1008 + non-empty reason, the client presets `reconnectAttempts` to `⌈log₂(60000/2000)⌉ = 5` so the first retry delay hits the 60 s cap immediately.

## Test plan

- [x] Configure tenant profile with WS session limit = 1, open two browser tabs — verify only **one** red notification appears and reconnect waits ~60 s
- [x] Normal network disconnect (disable Wi-Fi briefly) — verify reconnect backs off: 2 s, 4 s, 8 s …
- [x] After successful reconnect + data flows, disconnect again — verify backoff resets to 2 s
- [x] Logout / login — no stale error notification on fresh session
- [x] Public dashboard made private while open — WS notification shows once, not on a loop

### DIFF
--- a/ui-ngx/src/app/modules/home/components/ws-status-banner/ws-status-banner.component.scss
+++ b/ui-ngx/src/app/modules/home/components/ws-status-banner/ws-status-banner.component.scss
@@ -1,18 +1,18 @@
-///
-/// Copyright © 2016-2026 The Thingsboard Authors
-///
-/// Licensed under the Apache License, Version 2.0 (the "License");
-/// you may not use this file except in compliance with the License.
-/// You may obtain a copy of the License at
-///
-///     http://www.apache.org/licenses/LICENSE-2.0
-///
-/// Unless required by applicable law or agreed to in writing, software
-/// distributed under the License is distributed on an "AS IS" BASIS,
-/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-/// See the License for the specific language governing permissions and
-/// limitations under the License.
-///
+/**
+ * Copyright © 2016-2026 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


smatvienko-tb:
@vvlladd28 please, take over this PR

vvlladd28:
Fixed in https://github.com/thingsboard/thingsboard/pull/15219

PR #15205, opened 2026-03-10
Support combined PEM cert+key for Edge gRPC SSL

smatvienko-tb:
## Summary
- Add support for combined PEM files (cert chain + private key in a single file) for Edge gRPC SSL
- Support encrypted private keys via new `key_password` configuration property
- Make `private_key` config optional (default empty) for combined PEM usage

## Approach
Reuses `PemSslCredentials` directly — the class that already handles combined PEM parsing, separate cert/key files, and encrypted keys across MQTT/CoAP/LwM2M transports. Wires it into gRPC via `GrpcSslContexts.configure(SslContextBuilder.forServer(kmf))`.

**No new BouncyCastle imports. No PEM parsing duplication. 10 lines of new code.**

Alternative to #15150 — avoids introducing `JcaPEMWriter` (not used elsewhere in ThingsBoard) and duplicating PEM parsing logic that `PemSslCredentials` already provides.

## Test coverage

`EdgeGrpcSslTest` — 14 parameterized scenarios, all pass:

| Test | RSA-2048 | RSA-4096 | EC P-256 | EC P-384 |
|------|:--------:|:--------:|:--------:|:--------:|
| `separateCertAndKey` | ✅ | ✅ | ✅ | ✅ |
| `combinedPemWithCertAndKey` | ✅ | ✅ | ✅ | ✅ |
| `encryptedPrivateKey` | ✅ | ✅ | — | — |
| `combinedPemWithCertOnly` (error) | ✅ | ✅ | ✅ | ✅ |

> Encrypted EC keys are excluded: BouncyCastle writes them in traditional PEM format which loses public key info on round-trip, causing `PemSslCredentials` to fail. This is a pre-existing limitation, not introduced by this PR.

Each test mirrors `EdgeGrpcService.setupSsl()` exactly — same `PemSslCredentials` + `GrpcSslContexts` wiring — and verifies a real TLS handshake (gRPC client connects with `READY` state).

## Test plan
- [x] Verify Edge gRPC connection with separate cert and key (existing behavior)
- [x] Verify Edge gRPC connection with combined PEM (cert + key)
- [x] Verify Edge gRPC connection with encrypted private key + key_password
- [x] Verify error when combined PEM has no private key and private_key is empty
- [x] Verify ECDSA P-256 and P-384 key support
- [x] Verify RSA-2048 and RSA-4096 key support

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
+++ b/application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
@@ -202,6 +202,21 @@ public class EdgeGrpcService extends EdgeRpcServiceGrpc.EdgeRpcServiceImplBase i
         log.info("Edge RPC service initialized!");
     }
 
+    /**
+     * Configures TLS for the Edge gRPC server.
+     * <p>
+     * Delegates PEM parsing and key management to {@link PemSslCredentials} — the same
+     * class used by MQTT, CoAP, and LwM2M transports — which supports:
+     * <ul>
+     *   <li>Separate certificate and private key files (classic two-file setup)</li>
+     *   <li>Combined PEM: certificate chain + private key in a single {@code cert} file
+     *       ({@code private_key} left empty)</li>
+     *   <li>Encrypted private keys (password supplied via {@code key_password})</li>
+     * </ul>
+     * Path resolution (for both {@code cert} and {@code private_key}) is handled by
+     * {@link org.thingsboard.server.common.data.ResourceUtils#getInputStream ResourceUtils}:
+     * absolute path → relative / working-dir → classpath → {@code classpath:} prefix.
+     */
     void setupSsl(NettyServerBuilder builder) throws Exception {
         PemSslCredentials credentials = new PemSslCredentials();
         credentials.setCertFile(certFileResource);
--- a/application/src/main/resources/thingsboard.yml
+++ b/application/src/main/resources/thingsboard.yml
@@ -1487,12 +1487,16 @@ edges:
     ssl:
       # Enable/disable SSL support
       enabled: "${EDGES_RPC_SSL_ENABLED:false}"
-      # Path to the server certificate PEM file. May contain the full chain and optionally the private key (combined PEM).
-      # For a combined PEM file containing both the certificate chain and the private key, leave private_key empty.
+      # Path to the server certificate file (holds server certificate or certificate chain, may include server private key).
+      # Accepts an absolute filesystem path (e.g. /etc/thingsboard/certChainFile.pem),
+      # a relative path resolved against the working directory first then the classpath,
+      # or a classpath resource with the explicit "classpath:" prefix (e.g. classpath:conf/certChainFile.pem).
       cert: "${EDGES_RPC_SSL_CERT:certChainFile.pem}"
-      # Path to the private key PEM file. Optional when the key is already included in the cert file (combined PEM).
+      # Path to the server certificate private key file. Optional if the private key is already present in the cert file above.
+      # Supports the same path resolution as 'cert': absolute, relative/classpath, or "classpath:" prefix.
+      # Leave empty when using a combined PEM cert that already contains the private key.
       private_key: "${EDGES_RPC_SSL_PRIVATE_KEY:privateKeyFile.pem}"
-      # Password for encrypted private keys. Optional, only needed for password-protected keys.
+      # Server certificate private key password (optional). Leave empty if the key is not encrypted.
       key_password: "${EDGES_RPC_SSL_KEY_PASSWORD:}"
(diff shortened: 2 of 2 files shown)


smatvienko-tb (commented):
## Backward Compatibility Review — Resource Loading (classpath / absolute paths)

I specifically traced how `classpath:`, absolute, and relative paths are resolved before and after this change.

### TL;DR: Safe for existing users. No breaking change in resource resolution.

---

### How resources were loaded before

```java
// EdgeGrpcService.onStartUp()
InputStream certFileIs = ResourceUtils.getInputStream(this, certFileResource);      // this = EdgeGrpcService
InputStream privateKeyFileIs = ResourceUtils.getInputStream(this, privateKeyResource); // this = EdgeGrpcService
builder.useTransportSecurity(certFileIs, privateKeyFileIs);
```

`ResourceUtils.getInputStream(Object classLoaderSource, String filePath)` resolves via `classLoaderSource.getClass().getClassLoader()`:
1. Strip `classpath:` prefix if present
2. Try filesystem: `new File(path).exists()` → `new FileInputStream(file)`
3. Try classpath: `classLoader.getResourceAsStream(path)`
4. Try Guava: `Resources.getResource(path).toURI()`

Both streams opened **unconditionally** — if either file doesn't exist, immediate `RuntimeException("Unable to find resource: <path>")`.

### How resources are loaded after

```java
// EdgeGrpcService.setupSsl()
PemSslCredentials credentials = new PemSslCredentials();
credentials.setCertFile(certFileResource);
credentials.setKeyFile(StringUtils.isEmpty(privateKeyResource) ? null : privateKeyResource);
```

Inside `PemSslCredentials.loadKeyStore()`:
```java
// certFile — always opened via ResourceUtils.getInputStream(this, this.certFile)  // this = PemSslCredentials
// keyFile  — guarded: only opened if resourceExists(this, this.keyFile) returns true
```

### Classloader difference: not a real concern

The `this` passed to `ResourceUtils` changes from `EdgeGrpcService` → `PemSslCredentials`. However, `ResourceUtils.getInputStream()` uses `classLoaderSource.getClass().getClassLoader()`, and in Spring Boot's flat-JAR deployment (both monolith and microservices), all application classes share the same `LaunchedURLClassLoader`. This is also proven by the existing production usage of `PemSslCredentials` in MQTT/CoAP/LwM2M transports — they pass `this` (PemSslCredentials) to ResourceUtils and `classpath:` resources resolve correctly.

**Verdict**: `classpath:` prefixed paths, absolute paths, and relative paths all resolve identically. ✅

### One minor behavioral difference: error reporting for missing key files

| Scenario | Before | After |
|----------|--------|-------|
| Key file path set but file doesn't exist | `RuntimeException("Unable to find resource: /path/to/key.pem")` — immediate, points to the missing file | `PemSslCredentials` silently skips the key file (`resourceExists` returns false), then throws `IllegalArgumentException("Unable to load private key neither from certFile: ... nor from keyFile: ...")` |

The error still happens — server fails to start in both cases. But the **after** message is less specific: it doesn't explicitly say the key file was not found. Users debugging a misconfigured path would need to check file existence themselves.

This is a pre-existing behavior of `PemSslCredentials` (same pattern used by MQTT/CoAP/LwM2M), so it's consistent with the rest of the codebase. Not a blocker — just worth being aware of.

### Suggestion

Consider adding a `log.warn` when the key file is configured but doesn't exist, to preserve debuggability:

```java
void setupSsl(NettyServerBuilder builder) throws Exception {
    PemSslCredentials credentials = new PemSslCredentials();
    credentials.setCertFile(certFileResource);
    if (StringUtils.isEmpty(privateKeyResource)) {
        credentials.setKeyFile(null);
    } else {
        credentials.setKeyFile(privateKeyResource);
        if (!ResourceUtils.resourceExists(this, privateKeyResource)) {
            log.warn("Configured private key resource not found: {}. Will attempt to extract key from cert file.", privateKeyResource);
        }
    }
    credentials.setKeyPassword(keyPassword);
    credentials.init(false);

    SslContext sslContext = GrpcSslContexts.configure(
            SslContextBuilder.forServer(credentials.createKeyManagerFactory())).build();
    builder.sslContext(sslContext);
}
```

Note: Using `ResourceUtils.resourceExists(this, ...)` where `this` is `EdgeGrpcService` would also make the classloader match the original code exactly (though as discussed above, it's the same classloader in practice).

### Summary

- ✅ Absolute paths (`/etc/ssl/certs/edge.pem`) — same `new File(path).exists()` check, identical
- ✅ Relative paths (`certs/edge.pem`) — same CWD-relative resolution, identical
- ✅ Classpath resources (`classpath:certs/edge.pem`) — same classloader in practice, identical
- ✅ Default placeholders (`certChainFile.pem` / `privateKeyFile.pem`) — unchanged defaults, SSL disabled by default
- ✅ New `key_password` property — defaults to empty string, no impact on existing setups
- ⚠️ Missing key file error message — less specific (not a breaking change, just less debuggable)

LGTM with the optional logging suggestion above.

PR #15206, opened 2026-03-10
fix: use Awaitility for async audit log assertion in sysadmin test

smatvienko-tb:
## Summary
- Fix flaky `testAuditLogs_byTenantIdAndEntityId_Sysadmin` test that intermittently fails with `expected:<2> but was:<1>`
- Root cause: audit log persistence is async, and the test queries immediately after the update POST before the second audit log is saved
- Replace direct `Assert.assertEquals` with `Awaitility.await().atMost(10, SECONDS).untilAsserted(...)`, matching the pattern already used in `testAuditLogsSysAdmin` in the same class

## Failed build
https://builds.thingsboard.io/buildConfiguration/ThingsBoard_TestServiceControllerEdge_2/95058

## Test plan
- [ ] Run `AuditLogControllerTest#testAuditLogs_byTenantIdAndEntityId_Sysadmin` multiple times to confirm it no longer flakes
- [ ] Verify other tests in `AuditLogControllerTest` still pass


smatvienko-tb:
duplicated

PR #15207, opened 2026-03-10
fix: await timeseries batch save to prevent flaky NoSQL test

smatvienko-tb:
## Summary
- The `shouldSaveEntryOfEachTypeWithTtl` test in `TimeseriesServiceNoSqlTest` intermittently fails with `expected:<5> but was:<2>` because the batch `tsService.save()` returns a `ListenableFuture` that is not awaited before querying Cassandra
- Added `.get(MAX_TIMEOUT, TimeUnit.SECONDS)` to the save call, consistent with other tests in the same file (lines 89-90, 117) that already await their save futures

## Failed build
https://builds.thingsboard.io/buildConfiguration/ThingsBoard_TestSmokeLogic/94962

## Test plan
- [ ] Run `TimeseriesServiceNoSqlTest#shouldSaveEntryOfEachTypeWithTtl` repeatedly to confirm it no longer flakes


smatvienko-tb:
duplicated

PR #15221, opened 2026-03-11
Fix flaky AuditLogControllerTest.testAuditLogs_byTenantIdAndEntityId_Sysadmin

smatvienko-tb:
## Summary

- `testAuditLogs_byTenantIdAndEntityId_Sysadmin` fails on every ~3rd run on master with \`expected:<2> but was:<1>\`
- Root cause: audit logs are saved asynchronously via \`executor.submit()\` in \`AuditLogServiceImpl\`; the test issued two API calls (CREATE + UPDATE TenantProfile) then immediately asserted 2 log entries existed — but the second log entry was still in-flight
- The same async-save race is already fixed in the sibling \`testAuditLogs()\` method in this file (using \`Awaitility.await().atMost(10, SECONDS).untilAsserted(...)\`) and in the same test on the \`rc\` branch
- Fix: apply the same `Awaitility` pattern to the failing assertion

## Test plan
- [ ] `testAuditLogs_byTenantIdAndEntityId_Sysadmin` passes consistently on CI after this change


smatvienko-tb:
duplicated

PR #15226, opened 2026-03-12
feat: add RPC monitoring to the monitoring microservice

smatvienko-tb:
## Summary

- Adds two-way RPC health checking alongside the existing transport telemetry checks
- Probes the full round-trip: REST API → rule engine partition → transport microservice → device → response, with value validation
- Multiple devices per target cover different topic/partition assignments by hash, so partial microservice failures (affecting only a fraction of devices) are detectable
- Starting with MQTT; extensible to other transports via `RpcTransportType` enum

## Design

```
monitoring.rpc.mqtt:
  enabled: true
  request_timeout_ms: 4000
  qos: 1
  targets:
    - base_url: tcp://mqtt.example.com:1883
      device_id: <uuid>        # pre-existing device
      access_token: <token>
      label: shard0            # used in metric keys
    - ...                      # add one per partition shard
```

**Check flow per device:**
1. On startup: persistent MQTT connection, subscribed to `v1/devices/me/rpc/request/+`
2. On each check: `POST /api/rpc/twoway/{deviceId}` with `{"method":"monitoringCheck","params":{"value":"<uuid>"}}`
3. Paho callback thread receives the request, publishes params back to `v1/devices/me/rpc/response/{requestId}`
4. REST call returns the echoed value; checker validates it matches and reports `mqttRpc_<label>RpcRoundTrip` latency

**Key changes to base classes:**
- `BaseHealthChecker.check(WsClient)` — `final` removed to allow RPC override (WsClient-based validation is skipped for RPC)
- `reporter` and `stopWatch` — made `protected` for subclass access

## Changes vs original PR #15141 (lts-4.3)

- Rebased onto lts-4.2
- Fixed `UnnecessaryStubbingException` in `MqttRpcHealthCheckerTest`: removed unused `isConnected()` stub from `setUp()` — the checker's internal `mqttClient` field is `null` at test start so `isConnected()` is never reached on first `initClient()` call

## Test plan

- [x] `RpcMonitoringTargetTest` — label fallback, UUID parsing
- [x] `BaseRpcHealthCheckerTest` — happy path, wrong value, null response, REST exception, request body structure, correct device ID
- [x] `MqttRpcHealthCheckerTest` — connect options, RPC topic subscription, callback echo, idempotent init, destroy

### DIFF
--- a/monitoring/pom.xml
+++ b/monitoring/pom.xml
@@ -45,6 +45,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.thingsboard.common</groupId>
             <artifactId>data</artifactId>
         </dependency>
--- a/monitoring/src/main/java/org/thingsboard/monitoring/config/rpc/MqttRpcMonitoringConfig.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/config/rpc/MqttRpcMonitoringConfig.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright © 2016-2026 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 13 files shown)


smatvienko-tb:
duplicated

PR #15228, opened 2026-03-12
fix: resolve UnnecessaryStubbingException in MqttRpcHealthCheckerTest

smatvienko-tb:
## Summary

Fixes `UnnecessaryStubbingException` in `MqttRpcHealthCheckerTest` introduced in #15226.

Two stubs in `setUp()` — `when(connectToken.getException()).thenReturn(null)` and `when(mqttClient.connectWithResult(any())).thenReturn(connectToken)` — are consumed by 8 of 9 tests but not by `destroyClient_beforeInit_isNoOp`, which never calls `initClient()`. Mockito strict stubbing (`@ExtendWith(MockitoExtension.class)`) flags them as `UnnecessaryStubbingException`, causing all 9 tests to fail.

Switched both stubs to `lenient()` so they remain shared in `setUp()` without requiring consumption by every test.

**Depends on:** #15226

## Test plan

- [x] All 9 `MqttRpcHealthCheckerTest` tests pass without `UnnecessaryStubbingException`

### DIFF
--- a/monitoring/pom.xml
+++ b/monitoring/pom.xml
@@ -45,6 +45,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.thingsboard.common</groupId>
             <artifactId>data</artifactId>
         </dependency>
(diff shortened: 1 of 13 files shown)


smatvienko-tb:
Closing as orphaned — RPC monitoring feature does not exist on `lts-4.2`, so the test fix has no parent commit on this branch. The fix has been folded into the feature commit on PR #15141 (`lts-4.3`), which is where the test lives. Branch will be deleted.

PR #15253, opened 2026-03-16
Fixed SSRF DNS rebinding bypass, added allow-list

ViacheslavKlimov:
## Pull Request description

Add SsrfSafeAddressResolverGroup that validates resolved IPs at Netty connection time, eliminating the TOCTOU gap where DNS rebinding domains resolve to safe IPs during validation but to private/metadata IPs at connection time. Disable HTTP redirects in TbHttpClient to prevent redirect-based SSRF bypass.

Add allow-list support (SSRF_ALLOWED_HOSTS) to SsrfProtectionValidator so customers with IoT devices on private networks can whitelist specific addresses or CIDR ranges while keeping SSRF protection enabled.

Add SSRF validation to MS Teams webhook, custom OAuth2 mapper, and GitHub OAuth2 mapper endpoints. Log a warning when SSRF protection is disabled.

## General checklist

## Back-End feature checklist

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/SsrfSafeAddressResolverGroup.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/SsrfSafeAddressResolverGroup.java
@@ -26,14 +26,18 @@ import org.thingsboard.common.util.SsrfProtectionValidator;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
-import java.util.stream.Collectors;
+import java.util.Set;
 
 /**
  * Custom Netty {@link AddressResolverGroup} that validates every resolved IP address
  * against the SSRF block-list at connection time. This eliminates the DNS rebinding
  * TOCTOU gap where a hostname resolves to a safe IP during validation but to a
  * private/metadata IP when the actual connection is made.
+ * <p>
+ * Only wired into {@link TbHttpClient} when SSRF protection is enabled.
  */
 public final class SsrfSafeAddressResolverGroup extends AddressResolverGroup<InetSocketAddress> {
 
@@ -76,16 +80,22 @@ public final class SsrfSafeAddressResolverGroup extends AddressResolverGroup<Ine
         @Override
         public Future<InetSocketAddress> resolve(SocketAddress address, Promise<InetSocketAddress> promise) {
             delegate.resolve(address).addListener((Future<InetSocketAddress> future) -> {
-                if (!future.isSuccess()) {
-                    promise.tryFailure(future.cause());
-                    return;
-                }
-                InetSocketAddress resolved = future.getNow();
-                if (SsrfProtectionValidator.isEnabled() && isBlocked(resolved) && !isOriginalHostAllowed(address)) {
-                    promise.tryFailure(new RuntimeException(
-                            "URI is invalid: host '" + resolved.getAddress().getHostAddress() + "' is not allowed"));
-                } else {
-                    promise.trySuccess(resolved);
+                try {
+                    if (!future.isSuccess()) {
+                        promise.tryFailure(future.cause());
+                        return;
+                    }
+                    InetSocketAddress resolved = future.getNow();
+                    if (isOriginalHostAllowed(address)) {
+                        promise.trySuccess(resolved);
+                    } else if (isBlocked(resolved)) {
+                        promise.tryFailure(new RuntimeException(
+                                "URI is invalid: host '" + getHostString(address) + "' is not allowed"));
+                    } else {
+                        promise.trySuccess(resolved);
+                    }
+                } catch (Exception e) {
+                    promise.tryFailure(e);
                 }
             });
             return promise;
@@ -99,24 +109,41 @@ public final class SsrfSafeAddressResolverGroup extends AddressResolverGroup<Ine
         @Override
         public Future<List<InetSocketAddress>> resolveAll(SocketAddress address, Promise<List<InetSocketAddress>> promise) {
             delegate.resolveAll(address).addListener((Future<List<InetSocketAddress>> future) -> {
-                if (!future.isSuccess()) {
-                    promise.tryFailure(future.cause());
-                    return;
-                }
-                List<InetSocketAddress> resolved = future.getNow();
-                if (!SsrfProtectionValidator.isEnabled() || isOriginalHostAllowed(address)) {
-                    promise.trySuccess(resolved);
-                    return;
-                }
-                List<InetSocketAddress> safe = resolved.stream()
-                        .filter(addr -> !isBlocked(addr))
-                        .collect(Collectors.toList());
-                if (safe.isEmpty()) {
-                    String host = address instanceof InetSocketAddress isa ? isa.getHostString() : address.toString();
-                    promise.tryFailure(new RuntimeException(
-                            "URI is invalid: host '" + host + "' is not allowed"));
-                } else {
-                    promise.trySuccess(safe);
+                try {
+                    if (!future.isSuccess()) {
+                        promise.tryFailure(future.cause());
+                        return;
+                    }
+                    List<InetSocketAddress> resolved = future.getNow();
+                    if (isOriginalHostAllowed(address)) {
+                        promise.trySuccess(resolved);
+                        return;
+                    }
+                    Set<InetSocketAddress> blocked = null;
+                    for (InetSocketAddress addr : resolved) {
+                        if (isBlocked(addr)) {
+                            if (blocked == null) {
+                                blocked = new HashSet<>(2);
+                            }
+                            blocked.add(addr);
+                        }
+                    }
+                    if (blocked == null) {
+                        promise.trySuccess(resolved);
+                    } else if (blocked.size() == resolved.size()) {
+                        promise.tryFailure(new RuntimeException(
+                                "URI is invalid: host '" + getHostString(address) + "' is not allowed"));
+                    } else {
+                        List<InetSocketAddress> safe = new ArrayList<>(resolved.size() - blocked.size());
+                        for (InetSocketAddress addr : resolved) {
+                            if (!blocked.contains(addr)) {
+                                safe.add(addr);
+                            }
+                        }
+                        promise.trySuccess(safe);
+                    }
+                } catch (Exception e) {
+                    promise.tryFailure(e);
                 }
             });
             return promise;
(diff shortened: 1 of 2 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -103,6 +103,8 @@ public class TbHttpClient {
                     .build();
 
             HttpClient httpClient = HttpClient.create(connectionProvider)
+                    .resolver(SsrfSafeAddressResolverGroup.INSTANCE)
```
ViacheslavKlimov: Why is it used even if the validation is disabled?

common/util/src/main/java/org/thingsboard/common/util/SsrfProtectionValidator.java
```diff
@@ -179,10 +199,42 @@ private static boolean isIpLiteral(String entry) {
...
+            }
+            if (trimmed.contains("/") || isIpLiteral(trimmed)) {
+                try {
+                    cidrRanges.add(CidrRange.parse(trimmed));
+                } catch (Exception e) {
+                    log.warn("Failed to parse allowed CIDR/IP entry '{}': {}", trimmed, e.getMessage());
+                }
+            } else {
+                hostnames.add(trimmed.toLowerCase());
+            }
```
ViacheslavKlimov: This code is a complete duplicate from setAdditionalBlockedHosts, isn't it?

application/src/main/java/org/thingsboard/server/service/security/auth/oauth2/GithubOAuth2ClientMapper.java
```diff
@@ -62,6 +64,12 @@ private synchronized String getEmail(String emailUrl, String oauth2Token) {
         restTemplateBuilder = restTemplateBuilder.defaultHeader(AUTHORIZATION, "token " + oauth2Token);
 
         RestTemplate restTemplate = restTemplateBuilder.build();
+        try {
+            SsrfProtectionValidator.validateUri(new URI(emailUrl));
```
ViacheslavKlimov: This should be validated when OAuth2 client is created or updated

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/SsrfSafeAddressResolverGroup.java
```diff
@@ -0,0 +1,135 @@
...
+
+        @Override
+        public Future<InetSocketAddress> resolve(SocketAddress address, Promise<InetSocketAddress> promise) {
+            delegate.resolve(address).addListener((Future<InetSocketAddress> future) -> {
+                if (!future.isSuccess()) {
+                    promise.tryFailure(future.cause());
+                    return;
+                }
+                InetSocketAddress resolved = future.getNow();
+                if (SsrfProtectionValidator.isEnabled() && isBlocked(resolved)) {
```
ViacheslavKlimov: Hostname entries in `SSRF_ALLOWED_HOSTS` won't work for rule engine HTTP requests targeting private networks.

When an admin allow-lists a hostname like `my-internal-service.corp`, this is the flow:

1. `TbHttpClient.buildEncodedUri()` calls `SsrfProtectionValidator.validateUri()` — hostname is in the allow-list, so it returns early (passes).
2. Netty opens the connection, `SsrfSafeAddressResolverGroup` resolves the hostname to e.g. `192.168.1.100`.
3. `isBlockedAddress(192.168.1.100)` checks only the **CIDR** allow-list (no hostname match), then `isSiteLocalAddress()` returns true — **blocked**.

The connection fails with "SSRF protection: resolved address 192.168.1.100 is blocked" even though the admin explicitly allowed the hostname.

One fix would be to check the original hostname against the hostname allow-list here before rejecting:

```java
if (SsrfProtectionValidator.isEnabled() && isBlocked(resolved)) {
    String originalHost = address instanceof InetSocketAddress isa ? isa.getHostString() : null;
    if (originalHost != null && SsrfProtectionValidator.isHostnameAllowed(originalHost)) {
        promise.trySuccess(resolved);
        return;
    }
    promise.tryFailure(...);
}
```

where `isHostnameAllowed` is a new public method on `SsrfProtectionValidator`.

Alternatively, the docs/examples could clarify that only CIDR/IP entries work for rule engine requests targeting private networks.

application/src/main/resources/thingsboard.yml
```diff
@@ -509,6 +509,10 @@ actors:
       # Comma-separated list of additional blocked destinations (IPs, CIDR subnets, or hostnames).
       # Example: "198.51.100.0/24,metadata.tencentyun.com,rancher-metadata"
       ssrf_additional_blocked_hosts: "${SSRF_ADDITIONAL_BLOCKED_HOSTS:}"
+      # Comma-separated list of allowed destinations that bypass SSRF blocking (IPs, CIDR subnets, or hostnames).
+      # Use this when your rule chains need to reach devices on private networks (e.g., 192.168.1.0/24).
+      # Example: "192.168.1.0/24,10.0.0.0/8,my-internal-service.corp"
```
ViacheslavKlimov, line 514: The hostname `my-internal-service.corp` in this example won't work for rule chains targeting private-network devices — the Netty-level SSRF resolver only checks CIDR/IP allow-list entries, not hostnames (see comment on `SsrfSafeAddressResolverGroup`). Consider either fixing the hostname propagation or updating this example to only show CIDR/IP entries.

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/rest/SsrfSafeAddressResolverGroupTest.java
```diff
@@ -0,0 +1,150 @@
...
+        assertThat(SsrfProtectionValidator.isBlockedAddress(InetAddress.getByName("8.8.8.8"))).isFalse();
+    }
+
+    @Test
+    void resolvePublicIpSucceeds() throws Exception {
+        EventExecutor executor = eventLoopGroup.next();
+        AddressResolver<InetSocketAddress> resolver = SsrfSafeAddressResolverGroup.INSTANCE.getResolver(executor);
+        Promise<InetSocketAddress> promise = executor.newPromise();
+
+        executor.submit(() -> resolver.resolve(InetSocketAddress.createUnresolved("example.com", 80), promise));
```
ViacheslavKlimov: This resolves `example.com` via real DNS, so it will fail in air-gapped or DNS-restricted CI environments. Could use `InetSocketAddress.createUnresolved("8.8.8.8", 80)` to avoid the external dependency — though that trades the DNS exercise for determinism.

ViacheslavKlimov:
Addressed all review comments in 2f39347:

**1. Resolver used even when validation is disabled** (TbHttpClient.java:106)
The resolver checks `isEnabled()` at resolution time, so when SSRF is disabled it's a no-op pass-through (one volatile read). If SSRF protection gets enabled at runtime, existing `TbHttpClient` instances immediately start validating without rule node restart. `.followRedirect(false)` is also unconditionally useful (prevents redirect-based SSRF bypass).

**2. Duplicate parsing code** (SsrfProtectionValidator.java)
Fixed — extracted `parseHostEntries()` private helper. Both `setAllowedHosts` and `setAdditionalBlockedHosts` now delegate to it.

**3. OAuth2 SSRF validation at runtime vs save-time** (GithubOAuth2ClientMapper.java)
Fixed — moved SSRF validation to `Oauth2ClientDataValidator.validateDataImpl()` so the custom mapper URL is validated when the OAuth2 client is created/updated. Removed runtime checks from `CustomOAuth2ClientMapper` and `GithubOAuth2ClientMapper`. The GitHub `emailUrl` comes from server-side config (`oauth2.githubMapper.emailUrl`), not user input, so it doesn't need SSRF validation.

**4. Hostname allow-list not working at resolver level** (SsrfSafeAddressResolverGroup.java:84)
Fixed — added `isOriginalHostAllowed()` that extracts the original hostname from the `SocketAddress` and checks it against `SsrfProtectionValidator.isHostnameAllowed()` (new public method). If the hostname is in the allow-list, the resolver allows the connection even if the resolved IP is private.

**5. Hostname example in YAML won't work** (thingsboard.yml:514)
Now works with fix #4 above — the hostname allow-list is checked both at URI validation time (fast-fail) and at Netty resolution time (original hostname propagated).

**6. `example.com` in test requires real DNS** (SsrfSafeAddressResolverGroupTest.java:83)
Fixed — replaced with `8.8.8.8` (IP literal, no DNS needed).

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/SsrfSafeAddressResolverGroup.java
ViacheslavKlimov: **Still unresolved:** Hostname entries in `SSRF_ALLOWED_HOSTS` won't work end-to-end for rule engine HTTP requests.

When an admin allow-lists `my-internal-service.corp`:
1. `validateUri()` — hostname is in the allow-list, passes.
2. Netty resolves to e.g. `192.168.1.100`.
3. `isBlockedAddress(192.168.1.100)` only checks CIDR allow-list entries (not hostnames) → `isSiteLocalAddress()` returns true → **blocked**.

The original `address` parameter is available here and contains the hostname (`isa.getHostString()`). One fix:

```java
if (SsrfProtectionValidator.isEnabled() && isBlocked(resolved)) {
    String originalHost = address instanceof InetSocketAddress isa ? isa.getHostString() : null;
    if (originalHost != null && SsrfProtectionValidator.isHostnameAllowed(originalHost)) {
        promise.trySuccess(resolved);
        return;
    }
    promise.tryFailure(...);
}
```

The same issue exists in `resolveAll` (line 112).

application/src/main/resources/thingsboard.yml
ViacheslavKlimov, line 514: **Still unresolved:** The example `my-internal-service.corp` won't work for rule engine requests targeting private networks — the Netty-level SSRF resolver only checks CIDR/IP allow-list entries, not hostnames. Either fix the hostname propagation (see comment on `SsrfSafeAddressResolverGroup`) or update this example to only show CIDR/IP entries.

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/rest/SsrfSafeAddressResolverGroupTest.java
ViacheslavKlimov: **Still unresolved:** This resolves `example.com` via real DNS, so it will fail in air-gapped or DNS-restricted CI environments. Using `InetSocketAddress.createUnresolved("8.8.8.8", 80)` would avoid the external dependency.

PR #15254, opened 2026-03-16
Added configurable security headers and env-var-backed CORS configuration

ViacheslavKlimov:
## Pull Request description

- Add configurable X-Frame-Options and CSP headers (disabled by default)
- Add X-Content-Type-Options and Referrer-Policy headers (enabled by default)
- Make CORS allowed-origin-patterns configurable via TB_CORS_* env vars

Root cause: ThingsboardSecurityConfiguration called .disable() on the entire HeadersConfigurer, which removed ALL security headers including Cache-Control. Fix uses defaultsDisabled() + selective header enablement via a new HttpSecurityHeadersCustomizer component.

Both Spring Boot (tb-node) and Express.js (web-ui) share the same SECURITY_HEADERS_* environment variables for consistent configuration across monolith and microservice deployments.

## General checklist

## Back-End feature checklist

### DIFF
--- a/msa/web-ui/server.ts
+++ b/msa/web-ui/server.ts
@@ -61,22 +61,22 @@ let connections: Socket[] = [];
         server = http.createServer(app);
 
         // Build security headers map once at startup.
-        // Headers enabled by default use '!== false' so they stay on unless explicitly disabled.
-        // Headers disabled by default use simple truthiness checks.
+        // node-config passes env var overrides as strings, so enabled can be boolean or string.
+        const isEnabled = (val: any) => val === true || val === 'true';
         const securityHeaders: Record<string, string> = {};
         const hc: any = config.get('security.headers');
-        if (hc['x-content-type-options']?.enabled !== false) {
+        if (isEnabled(hc['x-content-type-options']?.enabled)) {
             securityHeaders['X-Content-Type-Options'] = 'nosniff';
         }
-        if (hc['referrer-policy']?.enabled !== false) {
+        if (isEnabled(hc['referrer-policy']?.enabled)) {
             securityHeaders['Referrer-Policy'] = hc['referrer-policy']?.value || 'strict-origin-when-cross-origin';
         }
-        if (hc['x-frame-options']?.enabled) {
+        if (isEnabled(hc['x-frame-options']?.enabled)) {
             securityHeaders['X-Frame-Options'] = hc['x-frame-options']?.value || 'SAMEORIGIN';
         }
-        if (hc['content-security-policy']?.enabled && hc['content-security-policy']?.value) {
+        if (isEnabled(hc['content-security-policy']?.enabled) && hc['content-security-policy']?.value) {
             const csp = hc['content-security-policy'];
-            const name = csp['report-only']
+            const name = isEnabled(csp['report-only'])
                 ? 'Content-Security-Policy-Report-Only' : 'Content-Security-Policy';
             securityHeaders[name] = csp.value;
         }


msa/web-ui/config/default.yml
```diff
@@ -25,6 +25,20 @@ thingsboard:
...
+    referrer-policy:
+      enabled: true
+      value: "strict-origin-when-cross-origin"
+    x-frame-options:
+      enabled: false
+      value: "SAMEORIGIN"
+    content-security-policy:
+      enabled: false
+      value: ""
+      reportOnly: false
```
ViacheslavKlimov: This uses `reportOnly` (camelCase) but `thingsboard.yml` uses `report-only` (kebab-case) for the same logical property. Spring Boot handles both via relaxed binding, but `node-config` does exact key matching — so an operator who creates a local web-ui config override using kebab-case (copying the pattern from `thingsboard.yml`) will get silent misconfiguration.

Consider aligning on the same casing in both files, or adding a comment in `thingsboard.yml` noting the web-ui uses camelCase.

application/src/main/java/org/thingsboard/server/config/HttpSecurityHeadersCustomizer.java
```diff
@@ -0,0 +1,59 @@
...
+        if (properties.getReferrerPolicy().isEnabled()) {
+            headers.addHeaderWriter(new StaticHeadersWriter("Referrer-Policy", properties.getReferrerPolicy().getValue()));
+        }
+
+        if (properties.getXFrameOptions().isEnabled()) {
+            String value = properties.getXFrameOptions().getValue();
+            if ("DENY".equalsIgnoreCase(value)) {
+                headers.frameOptions(HeadersConfigurer.FrameOptionsConfig::deny);
+            } else {
+                headers.frameOptions(HeadersConfigurer.FrameOptionsConfig::sameOrigin);
```
ViacheslavKlimov, line 49: Unrecognized values (typos like `"deny "` with trailing space, deprecated `ALLOW-FROM`, etc.) silently fall through to `sameOrigin`. Since the YAML comment documents only DENY and SAMEORIGIN as valid, a `log.warn` here for the non-DENY case that isn't SAMEORIGIN either would help operators catch misconfigurations.

application/src/main/java/org/thingsboard/server/config/HttpSecurityHeadersProperties.java
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.config;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
```
ViacheslavKlimov: Minor: `@Configuration` creates a CGLIB proxy subclass (for `@Bean` method interception), which isn't needed on a pure properties holder with no `@Bean` methods. `@Component` would give the same component-scanning behavior without the proxy overhead.

msa/web-ui/server.ts
```diff
@@ -60,6 +60,39 @@ let connections: Socket[] = [];
         const app = express();
         server = http.createServer(app);
 
+        // Build security headers map once at startup
+        const securityHeaders: Record<string, string> = {};
+        if (config.has('security.headers')) {
+            const hc: any = config.get('security.headers');
+            if (hc['x-content-type-options']?.enabled !== false) {
```
ViacheslavKlimov: The enabled checks use two different patterns: `!== false` here and on line 70 (default-on headers) vs simple truthiness on lines 73 and 76 (default-off headers). The logic is correct, but the implicit contract — `!== false` means "enabled unless explicitly disabled" — is easy to miss. A brief comment explaining the pattern (or a small helper like `isEnabledByDefault(flag)`) would make it clearer.

PR #15268, opened 2026-03-19
Refactor SSRF protection: injectable config replaces global static state

smatvienko-tb:
## Summary

- Introduces immutable `SsrfProtectionConfig` record to replace shared mutable `static volatile` fields in `SsrfProtectionValidator`
- Passes config through the call chain: `ActorSystemContext` → `TbContext` → `TbRestApiCallNode` → `TbHttpClient` → `SsrfSafeAddressResolverGroup`
- Eliminates test-pollution that caused intermittent `TbRestApiCallNodeTest.deleteRequestWithoutBody` failures under parallel test execution
- Deprecated static methods now delegate to config-based overloads, removing ~100 lines of duplicated validation logic

## Problem

`SsrfProtectionValidator` held 3 shared mutable static volatile fields (`enabled`, `allowedHosts`, `additionalBlocked`) set once at startup via `ActorSystemContext.@PostConstruct` and read from `TbHttpClient`, `SsrfSafeAddressResolverGroup`, and `Oauth2ClientDataValidator`.

**Concrete bug**: Under JUnit 5 parallel execution (ForkJoinPool), `SsrfSafeAddressResolverGroupTest.@BeforeEach` called `SsrfProtectionValidator.setEnabled(true)`, which leaked into `TbRestApiCallNodeTest.deleteRequestWithoutBody` running concurrently. `TbHttpClient.init()` read `isEnabled() == true`, attached the SSRF-safe DNS resolver, which blocked the test's `localhost` HTTP server → CountDownLatch timeout → `AssertionFailedError: Server handled request ==> expected: <true> but was: <false>`.

**Design issues**:
1. **Test pollution** — any test calling `setEnabled(true)` poisons all other tests in the JVM
2. **Hidden coupling** — `TbHttpClient` and `SsrfSafeAddressResolverGroup` silently depend on global state set in `ActorSystemContext`
3. **Non-injectable / non-mockable** — can't give different configs to different node instances
4. **Initialization order fragility** — if `TbHttpClient` is constructed before `ActorSystemContext.init()`, SSRF is silently off

### New file
- **`SsrfProtectionConfig`** (`common/util`) — Immutable config holding `enabled`, allowed/blocked CIDR ranges and hostnames. Factory method `SsrfProtectionConfig.of(boolean, List, List)` parses raw property strings. `DISABLED` constant for tests. Package-private constructor (callers use `of()` factory).

### Modified (production)
- **`SsrfProtectionValidator`** — Added config-accepting overloads: `validateUri(URI, SsrfProtectionConfig)`, `isBlockedAddress(InetAddress, SsrfProtectionConfig)`, `isHostnameAllowed(String, SsrfProtectionConfig)`. Extracted `isIntrinsicallyBlocked()` helper. Deprecated static methods now delegate to config-based overloads via `snapshotGlobalConfig()`, eliminating ~100 lines of duplicated validation code. Made `CidrRange` public.
- **`SsrfSafeAddressResolverGroup`** — No longer a singleton. Takes `SsrfProtectionConfig` in constructor, passes to validator. Created per `TbHttpClient` instance.
- **`TbHttpClient`** — New constructor param `SsrfProtectionConfig`. Uses local config for resolver setup and URI validation instead of static calls.
- **`TbRestApiCallNode`** — Passes `ctx.getSsrfProtectionConfig()` to `TbHttpClient`.
- **`TbContext`** — New method `getSsrfProtectionConfig()`.
- **`DefaultTbContext`** — Implements by delegating to `ActorSystemContext`.
- **`ActorSystemContext`** — Builds `SsrfProtectionConfig` in `@PostConstruct`, exposes via getter. Still populates legacy static state for unmigrated callers.

### Modified (tests)
- **`SsrfSafeAddressResolverGroupTest`** — Uses local `SsrfProtectionConfig` instances instead of global `setEnabled(true)`/`setAllowedHosts()`. No more `@ResourceLock`, `@BeforeEach`/`@AfterEach` state management, or risk of test pollution.
- **`TbRestApiCallNodeTest`** — Stubs `ctx.getSsrfProtectionConfig()` to return `DISABLED`. No `@ResourceLock` needed.
- **`TbHttpClientTest`** — Passes `SsrfProtectionConfig.DISABLED` to constructor; sets field via `ReflectionTestUtils` for mock-based tests.

## Impact

| Aspect | Before (static) | After (local config) |
|---|---|---|
| Test isolation | `@ResourceLock` + manual `setEnabled(false)` per class | Pass `SsrfProtectionConfig.DISABLED` — no shared state |
| Configurability | One policy per JVM | Per-node instance |
| Coupling | Hidden global reads | Explicit constructor dependency |
| Thread safety | Volatile fields | Immutable record |

## Caller not yet migrated (uses deprecated static API)
- `Oauth2ClientDataValidator.validateUri()` — reads global `enabled` flag
- `MicrosoftTeamsNotificationChannel.validateUri()` — reads global `enabled` flag

These continue to work via the deprecated wrappers that delegate to the config-based API. Migration can follow in a separate PR.

## Test plan
- [x] `SsrfSafeAddressResolverGroupTest` — 7 tests pass (no global state touched)
- [x] `TbRestApiCallNodeTest` — 4 tests pass (including `deleteRequestWithoutBody`)
- [x] `TbHttpClientTest` — 8 tests pass
- [x] `SsrfProtectionValidatorTest` — 63 tests pass (legacy API unchanged)
- [ ] CI full build


smatvienko-tb:
Closing in favor of #15454, which addresses the same flake (`TbRestApiCallNodeTest` vs `SsrfSafeAddressResolverGroupTest`) with a 2-line `@ResourceLock("SsrfProtectionValidator")` annotation.

## Why the smaller fix is preferred here

1. **Sufficient for the reported bug.** A project-wide survey (see #15454 description) confirms that all four SSRF-sensitive test classes now carry the same resource lock:
   - `common/util/SsrfProtectionValidatorTest` — already had the lock
   - `rule-engine-components/SsrfSafeAddressResolverGroupTest` — already had the lock
   - `rule-engine-components/TbHttpClientTest` — already had the lock
   - `rule-engine-components/TbRestApiCallNodeTest` — added in #15454

   `application/` and `dao/` don't enable JUnit parallel execution, so the static-state race is not reachable there.

2. **Scope fit for LTS.** This PR changes ~290 lines across 11 files including production API (`TbContext.getSsrfProtectionConfig()`, new `TbHttpClient` constructor arg, `DefaultTbContext` wiring). That is an architectural refactor, not a flake fix, and an LTS branch is the wrong vehicle for it.

3. **Incomplete by own acknowledgement.** Two callers (`Oauth2ClientDataValidator`, `MicrosoftTeamsNotificationChannel`) stay on the deprecated static API, which means \`SsrfProtectionValidator\` would ship with both the config-based API and a \`snapshotGlobalConfig()\` bridge indefinitely.

4. **Currently conflicting.** Branch is out of date with \`lts-4.2\` (\`mergeable: CONFLICTING\` since 2026-03-31) — rebasing for a hotfix-only purpose is wasted work if the lock is sufficient.

## Not lost — future-scope

The motivation behind this PR (remove global mutable state, make the config injectable so future tests don't need to remember \`@ResourceLock\`) is still valid and explicitly called out in #15454's description as optional hardening. A follow-up PR against \`master\` — finishing the migration of all callers including \`Oauth2ClientDataValidator\` and \`MicrosoftTeamsNotificationChannel\` — would be welcome as a standalone cleanup, decoupled from any active flake.

Successor: #15454

PR #15286, opened 2026-03-23
OR conditions support for dashboard filters

### LINKED ISSUE #9379: [Feature Request] OR logic between Filter Keys
Hi

We can really use OR statement between FIlter keys, otherwise they are not very useful to create good user experience filtering/finding devices.

image

Complex condition is only available for each Key and not between keys.

ViacheslavKlimov:
## Pull Request description

image

With env option to forbid `or` usage for DB performance reasons (`SQL_QUERY_KEY_FILTERS_OR_CONDITIONS_ENABLED`):
image

## General checklist

## Back-End feature checklist


dao/src/main/java/org/thingsboard/server/dao/sql/query/EntityKeyMapping.java
```diff
@@ -283,8 +284,31 @@ private String getAliasByEntityKeyAndType(String key, EntityType entityType) {
...
-            String keyAlias = (entityKey.getType().equals(EntityKeyType.ENTITY_FIELD) && getEntityKeyColumn() != null) ? "e" : alias;
+            String keyAlias;
+            if (entityKey.getType().equals(EntityKeyType.ENTITY_FIELD) && getEntityKeyColumn() != null) {
+                if (outerContext) {
+                    // In the middle layer (OR relocation), entity field columns are available
+                    // by their alias name from the inner subquery SELECT (e.g., "alias2" from
+                    // "cast(e.name as varchar) as alias2"). Temporarily null out entityKeyColumn
+                    // so buildSimplePredicateQuery uses the alias directly as the field.
+                    String savedColumn = this.entityKeyColumn;
+                    this.entityKeyColumn = null;
```
ViacheslavKlimov: Temporarily nulling out `entityKeyColumn` and restoring it is fragile — it mutates shared mutable state on the mapping object. If this code is ever called from a parallel stream or if `buildKeyQuery` throws an exception, `entityKeyColumn` stays null.

A safer approach would be to pass a flag or override to `buildKeyQuery` that tells it to use the alias directly, rather than mutating instance state:

```java
// Instead of null-swap, pass useAliasDirectly=true to buildKeyQuery
List<String> predicates = keyFilters.stream()
    .map(keyFilter -> this.buildKeyQuery(ctx, alias, keyFilter, filterType, /* useAliasDirectly */ true))
    .collect(Collectors.toList());
return predicates.stream();
```

Alternatively, at minimum wrap in try/finally:
```java
String savedColumn = this.entityKeyColumn;
try {
    this.entityKeyColumn = null;
    // ...
} finally {
    this.entityKeyColumn = savedColumn;
}
```

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
```diff
@@ -93,6 +95,10 @@ public class DefaultEntityQueryService implements EntityQueryService {
     @Value("${server.ws.max_entities_per_alarm_subscription:1000}")
     private int maxEntitiesPerAlarmSubscription;
 
+    @Setter
```
ViacheslavKlimov: The `@Setter` annotation here exists solely to support `entityQueryService.setKeyFiltersOrConditionsEnabled(false)` in the test. This exposes a public setter on a production service bean, which could be invoked accidentally at runtime.

Consider using `@TestPropertySource(properties = "sql.query.key-filters-or-conditions.enabled=false")` on a dedicated test class, or a package-private setter annotated with `@VisibleForTesting`, to avoid leaking test concerns into the production API.

application/src/test/java/org/thingsboard/server/controller/EntityQueryControllerTest.java
```diff
@@ -1729,4 +1733,791 @@ private static EntityDataQuery buildDeviceQuery(String deviceName) {
         return new EntityDataQuery(filter, new EntityDataPageLink(1, 0, null, null), Collections.emptyList(), null, null);
     }
 
+    // --- OR conditions tests ---
```
ViacheslavKlimov, line 1738: Good range of test scenarios. One observation: every test uses `Thread.sleep()` for attribute propagation. While this is consistent with existing tests in the file, consider whether the existing `awaitility`-based waiting (or a retry-with-timeout pattern) would make these more resilient in slow CI environments.

Also, the `testFindEntityDataWithOrPagination` test (line 1877) uses `int temp = (i <= 2) ? 60 + i : (i <= 4) ? 5 - i : 25;` which produces temperatures `[61, 62, 2, 1, 25]`. Devices 3 and 4 get negative-ish values (`5-3=2`, `5-4=1`) which are < 10 — that works, but the comment says "Device 3,4: temperature < 10" while the actual values are 2 and 1, which could be confusing. Minor readability nit.

ViacheslavKlimov (commented):
## Review summary

Reviewed 44 changed files in [OR conditions support for dashboard filters](https://github.com/thingsboard/thingsboard/pull/15286). Left 3 comment(s) inline.

The overall design is solid — the `ComplexOperation` enum unification, the three-layer SQL generation (inner → middle → outer) for OR, the EDQS filter evaluation refactor, and the feature toggle are all well-structured. Backward compatibility is properly maintained with null-defaulting-to-AND throughout.

## Test coverage analysis

The added tests are well-written — they cover the core OR scenarios with clear setups and assertions. Each test is focused on a single concern (basic OR, same-key OR, mixed entity-field + attribute OR, pagination, zero matches, backward compatibility, feature toggle). The EDQS unit tests complement the integration tests nicely.

However, there are some coverage gaps worth noting:

- **No test for OR with `textSearch`** — The SQL generation in `findEntityDataByQuery` has special handling for combining `textSearch` with OR-mode filters (lines 499-504), but no test exercises this path. A query with both `pageLink.textSearch` and `keyFiltersOperation=OR` would validate the `" where "` vs `" and "` conditional logic.
- **No integration test for TIME_SERIES key filters with OR** — All controller tests use `SHARED_SCOPE` attributes. The SQL generation for time-series JOINs is slightly different (different table, different ON clause), so an integration test with time-series key filters under OR would increase confidence.
- **No EDQS test for STRING/entity-field value type with OR** — The `RepositoryUtilsTest` OR tests only cover `NUMERIC` (via `getTemperatureFilter`/`getHumidityFilter`). A test with string entity-field filters under OR would verify the entity-field special case in `evaluateSingleFilter` (where null string + entity field = match).
- **No test for OR with `ComplexFilterPredicate` within a key filter** — A key filter can contain nested AND/OR predicates via `ComplexFilterPredicate`. Testing a key-level complex predicate combined with query-level OR would verify there's no interaction between the two levels of OR.

---

ViacheslavKlimov (commented):
## Re-review summary

Re-reviewed [OR conditions support for dashboard filters](https://github.com/thingsboard/thingsboard/pull/15286) — verified 3 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/15286#pullrequestreview-3998341782).

| Status | Count |
|--------|-------|
| ✅ Resolved | 3 |

### Finding details

- ✅ **dao/src/main/java/org/thingsboard/server/dao/sql/query/EntityKeyMapping.java:299** — Fragile null-swap of `entityKeyColumn` without exception safety — *Fixed in code: wrapped in try/finally.*
- ✅ **application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java:98** — `@Setter` exposing production setter for test purposes — *Fixed in code: `@Setter` and field removed; validation moved to `BaseEntityService`; test uses `ReflectionTestUtils` instead.*
- ✅ **application/src/test/java/org/thingsboard/server/controller/EntityQueryControllerTest.java:1738** — `Thread.sleep()` for attribute propagation and confusing temperature values — *Fixed in code: all `Thread.sleep()` replaced with `await().atMost(...).untilAsserted(...)` pattern; temperature comment clarified with explicit computed values.*

The fix commits also add substantial new test coverage addressing the gaps noted in the previous review body: `textSearch` + OR, time-series key filters with OR, string entity-field filters in EDQS, `ComplexFilterPredicate` within key filter + query-level OR, and the entity-field null string special case.

---

vvlladd28 (commented):
## Review summary

Reviewed 50 changed files in [OR conditions support for dashboard filters](https://github.com/thingsboard/thingsboard/pull/15286). Left 5 comment(s) inline.

Well-structured PR overall — the SQL OR generation strategy (force LEFT JOINs + middle-layer WHERE with OR) is sound, the EDQS in-memory evaluation is clean, and the feature toggle with validation is a good safety net. The Angular modernization (typed forms, `takeUntilDestroyed`, control-flow syntax) is a nice bonus.

dao/src/main/java/org/thingsboard/server/dao/sql/query/EntityKeyMapping.java
```diff
@@ -283,8 +284,34 @@ private String getAliasByEntityKeyAndType(String key, EntityType entityType) {
...
         if (hasFilter()) {
-            String keyAlias = (entityKey.getType().equals(EntityKeyType.ENTITY_FIELD) && getEntityKeyColumn() != null) ? "e" : alias;
+            String keyAlias;
+            if (entityKey.getType().equals(EntityKeyType.ENTITY_FIELD) && getEntityKeyColumn() != null) {
+                if (outerContext) {
+                    // In the middle layer (OR relocation), entity field columns are available
+                    // by their alias name from the inner subquery SELECT (e.g., "alias2" from
+                    // "cast(e.name as varchar) as alias2"). Temporarily null out entityKeyColumn
+                    // so buildSimplePredicateQuery uses the alias directly as the field.
+                    String savedColumn = this.entityKeyColumn;
```
vvlladd28: Temporarily nulling out `entityKeyColumn` to change how a downstream method behaves is fragile — it couples the caller's correctness to the internal implementation of `buildKeyQuery`. Consider passing the desired alias/column reference as a parameter to `buildKeyQuery` (or a new overload) instead of mutating shared instance state. Today it's safe because the call is single-threaded and wrapped in `finally`, but this pattern is easy to break during future refactoring.

ui-ngx/src/app/modules/home/components/filter/key-filter-list.component.scss
```diff
@@ -13,30 +13,45 @@
...
-
-:host ::ng-deep {
-  .tb-filter-preview {
-    .tb-filter-text {
-      max-height: 200px;
+    display: flex;
+    justify-content: center;
+    margin-top: -14px;
+    &-container {
+      background-color: white;
```
vvlladd28, line 45: `background-color: white` is hardcoded here. This will look wrong in dark-theme mode — the operation label container will render as a bright white rectangle. Consider using a theme-aware variable or `var(--mdc-dialog-container-color)` (or whatever the project's surface-color token is).

application/src/main/java/org/thingsboard/server/service/subscription/TbAlarmCountSubCtx.java
```diff
@@ -130,7 +130,7 @@ public void doFetchAlarmCount() {
     private EntityDataQuery buildEntityDataQuery() {
         EntityDataPageLink edpl = new EntityDataPageLink(maxEntitiesPerAlarmSubscription, 0, null,
                 new EntityDataSortOrder(new EntityKey(EntityKeyType.ENTITY_FIELD, ModelConstants.CREATED_TIME_PROPERTY)));
-        return new EntityDataQuery(query.getEntityFilter(), edpl, null, null, query.getKeyFilters());
+        return new EntityDataQuery(query.getEntityFilter(), edpl, null, null, query.getKeyFilters(), query.getKeyFiltersOperation());
```
vvlladd28: This uses `query.getKeyFiltersOperation()` (raw, nullable), while `DefaultEntityQueryService.buildEntityDataQuery(AlarmCountQuery)` uses `query.getKeyFiltersOperationOrDefault()` (normalizes null → AND) for the same purpose. Both work because `getKeyFiltersOperationOrDefault()` is called downstream during SQL/EDQS generation, but it would be clearer to normalize consistently at the point of construction. Same applies to `TbAlarmDataSubCtx.buildEntityDataQuery()`.

ui-ngx/src/app/modules/home/components/filter/key-filter-list.component.scss
```diff
@@ -13,30 +13,45 @@
...
+    font-size: 16px;
+  }
+  .filter-list {
     overflow: auto;
     max-height: 300px;
-    .no-data-found {
-      height: 50px;
+
+    &-divider {
+      border-top: 1px solid rgba(0, 0, 0, 0.12);
```
vvlladd28, line 33: Same dark-theme concern: `rgba(0, 0, 0, 0.12)` is a hardcoded light-theme divider color. Consider using a theme variable or the project's standard divider token so it adapts to dark mode.

ui-ngx/src/app/modules/home/components/filter/filter-dialog.component.scss
```diff
@@ -14,15 +14,4 @@
  * limitations under the License.
  */
 :host {
```
vvlladd28: The `:host {}` block is now empty after removing the `.tb-editable-switch` styles. You could remove the file entirely (and the `styleUrls` reference in the component decorator) to avoid leaving an empty stylesheet.

ui-ngx/src/app/modules/home/components/filter/key-filter-list.component.scss
```diff
@@ -13,30 +13,45 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@import "../scss/constants";
```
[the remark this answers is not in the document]
vvlladd28: this is correct because resolv is from the ui-ngx directory

PR #15312, opened 2026-03-26
Fix race condition in notification deduplication check

smatvienko-tb:
Jira: [PROD-7951](https://thingsboard-portal.atlassian.net/browse/PROD-7951)

## Summary
- The `alreadyProcessed()` method in `DefaultNotificationDeduplicationService` used separate `get()` and `put()` calls on the local cache, allowing concurrent threads in the notification executor pool to both read `null` and bypass deduplication, creating duplicate notifications.
- Replace with a single `compute()` call that atomically checks and updates the cache entry.
- Discard external cache timestamps more than 1 hour in the future (clock skew protection).
- Avoid reading back from the SOFT ref local cache when writing to external cache (GC could null it out between `get()` and `put()`).

## GC safety analysis for `compute()` with external cache IO inside lambda

`ConcurrentReferenceHashMap` overrides `compute()` via its `Segment.doTask()` mechanism. `Segment extends ReentrantLock`, and `doTask()` acquires the segment lock before invoking the lambda:

```
Segment.doTask() {
    lock();          // ReentrantLock
    try {
        // resolve entry, invoke lambda (including external cache IO)
        // store returned value as new SOFT ref
    } finally {
        unlock();
    }
}
```

**What happens if GC triggers during the IO-blocked external cache call inside the lambda:**

1. **Atomicity is preserved** — the segment lock prevents any concurrent `get()`/`put()`/`compute()` on the same segment from observing intermediate state. Other threads block until the lambda completes.
2. **GC CAN clear SOFT refs concurrently** (GC doesn't respect Java locks), but no application thread can act on the cleared state while the lock is held by our lambda.
3. **The current entry's key and value are held by strong references** on the stack inside `doTask()` — they won't be collected during lambda execution.
4. **The value returned by the lambda** is stored as a new SOFT ref entry under the same lock, before unlock.
5. **After unlock** — the new entry is a fresh SOFT ref. If GC collects it later (between two separate `alreadyProcessed()` calls), that's acceptable best-effort dedup behavior.

**Conclusion**: the `compute()` call is segment-locked and GC-safe. The external cache IO inside the lambda does hold the segment lock longer, but notification dedup is low-contention (few keys, infrequent triggers), so this is acceptable.

## Test plan
- [x] New unit test `DefaultNotificationDeduplicationServiceTest` (6 tests, all pass < 1s):
  - `testFirstTriggerIsNotDeduplicated` — basic: first trigger passes through
  - `testSecondTriggerIsDeduplicated` — basic: repeat trigger is blocked
  - `testTriggerPassesAfterDeduplicationWindowExpires` — window expiry works
  - `testFutureTimestampFromExternalCacheIsDiscarded` — clock skew guard (>1h future timestamp ignored)
  - `testValidTimestampFromExternalCacheIsDeduplicated` — external cache fallback works
  - `testConcurrentTriggersProduceExactlyOneNonDeduplicated` — 10 threads via CyclicBarrier, exactly 1 passes
- [x] `NotificationRuleApiTest#testNotificationsDeduplication_resourcesShortage` passes locally
- [x] `NotificationRuleApiTest#testNotificationsResourcesShortage_whenThresholdChangeToMatchingFilter_thenSendNotification` passes locally
- [x] CI build passes

### DIFF
--- a/common/queue/src/main/java/org/thingsboard/server/queue/notification/DefaultNotificationDeduplicationService.java
+++ b/common/queue/src/main/java/org/thingsboard/server/queue/notification/DefaultNotificationDeduplicationService.java
@@ -32,6 +32,7 @@ import org.thingsboard.server.queue.util.PropertyUtils;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
 
 import static org.springframework.util.ConcurrentReferenceHashMap.ReferenceType.SOFT;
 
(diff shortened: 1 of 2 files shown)


ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [Fix race condition in notification deduplication check](https://github.com/thingsboard/thingsboard/pull/15312). No issues found.

The race condition fix is correct — replacing separate `get()`/`put()` with an atomic `compute()` on `ConcurrentReferenceHashMap` (which uses segment-level `ReentrantLock` in Spring 6.2+) eliminates the TOCTOU window that allowed concurrent threads to bypass deduplication. The clock skew guard (discarding external cache timestamps >1h in the future) and the switch to writing `now` directly to the external cache (avoiding a stale SOFT-ref read-back) are both sound improvements. Test coverage is thorough, including a `CyclicBarrier`-based concurrency test that validates exactly-once pass-through.

---

PR #15326, opened 2026-03-31
Added dedicated alarm rules API

ViacheslavKlimov:
## Pull Request description

Alarm rules are currently managed through the Calculated Fields API, requiring users to set `type = ALARM` — an implementation detail that shouldn't be exposed. This PR introduces a standalone Alarm Rules API that hides the underlying calculated field mechanics.

### Changes

**New REST API** (`/api/alarm/rule`, `/api/alarm/rules`):
- Create, read, update, delete alarm rules
- List alarm rules with filtering by entity type
- Get alarm rule names
- Debug events and TBEL expression testing

**New DTOs:**
- `AlarmRuleDefinition` — request/response model without the `type` field
- `AlarmRuleDefinitionInfo` — extends with entity name for list responses

**UI:**
- New `AlarmRulesService` pointing to the new endpoints
- All alarm rule components switched from `CalculatedFieldsService` to `AlarmRulesService`

**Tests:**
- `AlarmRulesTest` updated to use the new API

The existing Calculated Fields API is unchanged — alarm rules can still be accessed through it with explicit `type=ALARM` filtering.

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/controller/CalculatedFieldController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/CalculatedFieldController.java
@@ -88,7 +88,7 @@ public class CalculatedFieldController extends BaseController {
 
     public static final String CALCULATED_FIELD_ID = "calculatedFieldId";
 
-    static final String TEST_SCRIPT_EXPRESSION =
+    private static final String TEST_SCRIPT_EXPRESSION =
             "Execute the Script expression and return the result. The format of request: \n\n"
             + MARKDOWN_CODE_BLOCK_START
             + "{\n" +
@@ -150,13 +150,13 @@ public class CalculatedFieldController extends BaseController {
     @PreAuthorize("hasAnyAuthority('TENANT_ADMIN')")
     @GetMapping(value = "/{entityType}/{entityId}/calculatedFields", params = {"pageSize", "page"})
     public PageData<CalculatedField> getCalculatedFieldsByEntityIdV1(@PathVariable("entityType") String entityType,
-                                                                   @PathVariable("entityId") String entityIdStr,
-                                                                   @RequestParam int pageSize,
-                                                                   @RequestParam int page,
-                                                                   @RequestParam(required = false) CalculatedFieldType type,
-                                                                   @RequestParam(required = false) String textSearch,
-                                                                   @RequestParam(required = false) String sortProperty,
-                                                                   @RequestParam(required = false) String sortOrder) throws ThingsboardException {
+                                                                     @PathVariable("entityId") String entityIdStr,
+                                                                     @RequestParam int pageSize,
+                                                                     @RequestParam int page,
+                                                                     @RequestParam(required = false) CalculatedFieldType type,
+                                                                     @RequestParam(required = false) String textSearch,
+                                                                     @RequestParam(required = false) String sortProperty,
+                                                                     @RequestParam(required = false) String sortOrder) throws ThingsboardException {
         PageLink pageLink = createPageLink(pageSize, page, textSearch, sortProperty, sortOrder);
         checkParameter("entityId", entityIdStr);
         EntityId entityId = EntityIdFactory.getByTypeAndUuid(entityType, entityIdStr);
--- a/common/data/src/main/java/org/thingsboard/server/common/data/device/profile/DeviceProfileData.java
+++ b/common/data/src/main/java/org/thingsboard/server/common/data/device/profile/DeviceProfileData.java
@@ -35,6 +35,7 @@ public class DeviceProfileData implements Serializable {
     private DeviceProfileTransportConfiguration transportConfiguration;
     @Schema(description = "JSON object of provisioning strategy type per device profile")
     private DeviceProfileProvisionConfiguration provisionConfiguration;
+    @Valid
     @Schema(hidden = true)
     private List<DeviceProfileAlarm> alarms;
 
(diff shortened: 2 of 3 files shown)


application/src/main/java/org/thingsboard/server/controller/AlarmRuleController.java
```diff
@@ -0,0 +1,327 @@
...
+
+@RestController
+@TbCoreComponent
+@RequestMapping("/api")
+@RequiredArgsConstructor
+@Slf4j
+public class AlarmRuleController extends BaseController {
+
+    private final TbCalculatedFieldService tbCalculatedFieldService;
+    private final CalculatedFieldController calculatedFieldController;
```
ViacheslavKlimov: Injecting one controller into another is an anti-pattern — it couples their lifecycles and makes testing harder. The two methods used from `CalculatedFieldController` (`checkReferencedEntities` and `getLatestTimestamp`) are pure business logic that doesn't depend on controller state. Could these be extracted into a shared service or a static utility so both controllers can use them independently?

application/src/main/java/org/thingsboard/server/controller/AlarmRuleController.java
```diff
@@ -0,0 +1,271 @@
...
+        return tbCalculatedFieldService.executeTestScript(getTenantId(), inputParams);
+    }
+
+    private void checkReferencedEntities(CalculatedFieldConfiguration calculatedFieldConfig) throws ThingsboardException {
+        Set<EntityId> referencedEntityIds = calculatedFieldConfig.getReferencedEntities();
+        for (EntityId referencedEntityId : referencedEntityIds) {
+            EntityType refEntityType = referencedEntityId.getEntityType();
+            switch (refEntityType) {
+                case TENANT -> {
+                    return;
```
ViacheslavKlimov: This `return` exits the entire `checkReferencedEntities` method, not just the current iteration. If the set contains a TENANT entity followed by a CUSTOMER or DEVICE entity, the authorization check for those subsequent entities is skipped entirely.

Should this be `continue` instead? The same issue exists in `CalculatedFieldController` — worth fixing in both places.

application/src/main/java/org/thingsboard/server/controller/AlarmRuleController.java
```diff
@@ -0,0 +1,271 @@
...
+    @PreAuthorize("hasAuthority('TENANT_ADMIN')")
+    @PostMapping("/alarm/rule/testScript")
+    public JsonNode testAlarmRuleScript(
+            @io.swagger.v3.oas.annotations.parameters.RequestBody(description = "Test alarm rule TBEL condition expression. The expression must return a boolean value.")
+            @RequestBody JsonNode inputParams) throws ThingsboardException {
+        checkParameter("expression", inputParams.has("expression") ? inputParams.get("expression").asText() : null);
+        return tbCalculatedFieldService.executeTestScript(getTenantId(), inputParams);
+    }
+
+    private void checkReferencedEntities(CalculatedFieldConfiguration calculatedFieldConfig) throws ThingsboardException {
```
ViacheslavKlimov: This `checkReferencedEntities` method is identical to the one in `CalculatedFieldController` (lines 291–303). Consider extracting it to `BaseController` or a shared utility to avoid the duplication.

application/src/main/java/org/thingsboard/server/service/entitiy/cf/DefaultTbCalculatedFieldService.java
```diff
@@ -89,6 +112,75 @@ public void delete(CalculatedField calculatedField, SecurityUser user) {
...
+                    args[i] = arg;
+                }
+            }
+
+            JsonNode json = engine.executeJsonAsync(args).get(TIMEOUT, TimeUnit.SECONDS);
+            output = JacksonUtil.toString(json);
+        } catch (Exception e) {
+            log.error("Error evaluating expression", e);
+            Throwable rootCause = ExceptionUtils.getRootCause(e);
+            errorText = ObjectUtils.firstNonNull(rootCause.getMessage(), e.getMessage(), e.getClass().getSimpleName());
```
ViacheslavKlimov: `ExceptionUtils.getRootCause(e)` returns `null` when the exception has no nested cause (e.g. a simple `IllegalArgumentException`). In that case, `rootCause.getMessage()` will throw an NPE — which is caught by the surrounding `catch`, but then `ExceptionUtils.getRootCause(npe)` also returns null, causing an infinite-NPE-catch loop that eventually returns an unhelpful error message.

Consider:
```java
Throwable rootCause = ObjectUtils.firstNonNull(ExceptionUtils.getRootCause(e), e);
errorText = ObjectUtils.firstNonNull(rootCause.getMessage(), e.getClass().getSimpleName());
```

application/src/main/java/org/thingsboard/server/controller/CalculatedFieldController.java
```diff
@@ -305,84 +284,20 @@ public JsonNode getLatestCalculatedFieldDebugEvent(@Parameter @PathVariable(CALC
...
-                long ts = singleValueArg.getTs();
-                lastUpdateTimestamp = Math.max(lastUpdateTimestamp, ts);
-            } else if (entry instanceof TbelCfTsRollingArg tsRollingArg) {
-                long maxTs = tsRollingArg.getValues().stream().mapToLong(TbelCfTsDoubleVal::getTs).max().orElse(-1);
-                lastUpdateTimestamp = Math.max(lastUpdateTimestamp, maxTs);
-            }
-        }
-        return lastUpdateTimestamp == -1 ? System.currentTimeMillis() : lastUpdateTimestamp;
+            @RequestBody JsonNode inputParams) throws ThingsboardException {
+        return tbCalculatedFieldService.executeTestScript(getTenantId(), inputParams);
```
ViacheslavKlimov, line 289: The `AlarmRuleController.testAlarmRuleScript` validates the `expression` parameter before calling `executeTestScript`, but this endpoint doesn't — so a request missing `"expression"` will NPE inside the service (`inputParams.get("expression").asText()`). Consider adding the same `checkParameter` guard here, or moving the validation into `executeTestScript` itself so both callers are protected.

PR #15333, opened 2026-03-31
Performance and reliability improvements for Efento message processing

dashevchenko:
## Pull Request description

1. Updated efento proto files according to the latest release
2. Added support for new modem types.
3. Fixed bug related to the CONFIGURATION efento message processing.
4. "measured_at" is not saved anymore as redundant telemetry
5. Data point optimization: general parameters like "serial""battery","next_transmission_at", "signal" are saved once per message with the timestamp of the first sample of the first channel
6. "measurement_interval" telemetry is updated to "measurement_interval_" + channel and is saved once per channel (not for every sample)

## General checklist

## Back-End feature checklist

### DIFF
--- a/common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/efento/CoapEfentoTransportResource.java
+++ b/common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/efento/CoapEfentoTransportResource.java
@@ -53,7 +53,6 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.TreeMap;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
@@ -277,7 +276,7 @@ public class CoapEfentoTransportResource extends AbstractCoapTransportResource {
 
             // measurements per channel
             JsonObject tsValues = valuesMap.computeIfAbsent(startTimestampMillis, k -> new JsonObject());
-            tsValues.addProperty("measurement_interval_" + channel, measurementPeriod);
+            tsValues.addProperty("measurement_interval_" + (channel + 1), measurementPeriod);
 
             for (int i = 0; i < sampleOffsetsList.size(); i++) {
                 int sampleOffset = sampleOffsetsList.get(i);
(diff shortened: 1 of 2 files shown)


common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/efento/CoapEfentoTransportResource.java
```diff
@@ -258,6 +258,9 @@ List<EfentoTelemetry> getEfentoMeasurements(MeasurementsProtos.ProtoMeasurements
         }
 
         Map<Long, JsonObject> valuesMap = new TreeMap<>();
+        // general measurements per message
+        valuesMap.put(TimeUnit.SECONDS.toMillis(channelsList.get(0).getTimestamp()), CoapEfentoUtils.setDefaultMeasurements(serialNumber, batteryStatus, nextTransmissionAtMillis, signal));
```
mateusz-efento, line 262: Continuous sensors should be fine, but I'm concerned about mixed/binary types. The code batches them using the first channel's timestamp before doing a detailed check. This works when timestamps are identical, but in mixed/binary sensors, they can drift apart.

dashevchenko: Previously, serialNumber, batteryStatus, nextTransmissionAtMillis, and signal were duplicated for each data point, even though their values were identical within a single message. I consolidated them to appear only once per message. For the timestamp, I used the value from the first channel. If needed, I can duplicate values for each channel with an appropriate timestamp.

mateusz-efento: Thanks for the explanation, it is ok.

common/transport/coap/src/main/java/org/thingsboard/server/transport/coap/efento/CoapEfentoTransportResource.java
```diff
@@ -476,41 +477,86 @@ private EfentoTelemetry getEfentoDeviceInfo(DeviceInfoProtos.ProtoDeviceInfo pro
...
+        values.addProperty("sim_card_identification", modem.getSimCardIdentification());
+        values.addProperty("firmware_version", modem.getFirmwareVersion().toString());
+        values.addProperty("modem_identification", modem.getModemIdentification());
+        if (modem.getModemStatisticsCount() >= 4) {
+            values.addProperty("modem_transmissions_count", modem.getModemStatistics(0));
+            values.addProperty("modem_time_since_last_devinfo", modem.getModemStatistics(1));
+            values.addProperty("modem_total_psm_time", modem.getModemStatistics(2));
+            values.addProperty("modem_total_active_time", modem.getModemStatistics(3));
+        }
+        switch (modemType) {
```
ViacheslavKlimov, line 491: The `modemStatisticsCount >= 4` guard on line 485 is a good defensive pattern. Could a similar check on `getParametersCount()` be added before each branch here? BC660 expects 22 parameters, SHARED_MODEM expects 4, and the default (BC66) branch expects 34. If a device with older firmware (or a modem type with a different parameter layout) sends fewer parameters than a branch expects, `getParameters(N)` will throw `IndexOutOfBoundsException` and kill the entire device info processing for that message.

Something like:
```java
case MODEM_TYPE_BC660:
    if (modem.getParametersCount() >= 22) {
        // ... parameter access
    }
    break;
```

PR #15334, opened 2026-03-31
Fixed REST API Call node blocking actor thread and semaphore permit leak

smatvienko-tb:
## Problem

The REST API Call rule node had two issues:

1. **Actor thread blocked**: `semaphore.tryAcquire(timeout, MILLISECONDS)` was called directly on the rule-engine actor thread, parking it for up to the configured timeout on every request when the concurrency limit was reached.
2. **Semaphore permit leak**: an exception thrown between `tryAcquire()` and `subscribe()` left the permit permanently taken, silently dropping all subsequent messages for that node instance.

## Solution

Replace the semaphore-only guard with a **bounded FIFO queue + semaphore** pattern, closely following `SemaphoreWithTbMsgQueue`:

- When `maxParallelRequestsCount=0` (default), no semaphore or queue is created — `processMessage` calls `doHttpCall` directly, identical to the old behavior.
- When a concurrency limit is configured, every incoming message is enqueued into a `LinkedBlockingQueue` (capacity = `maxPendingRequests`, default unbounded). `offer()` is non-blocking — it returns `false` immediately when the queue is full, which triggers `onFailure`.
- `tryProcess()` acquires one semaphore slot and fires the next valid queued task. Stale tasks (`msg.isValid() == false` — batch pack deadline expired) are dropped with `onFailure` and the slot is immediately reused for the next candidate in the same loop pass.
- A `doFinally` hook (not inline callback code) calls `semaphore.release()` + `tryProcess()` exactly once after any terminal signal — success, error, or cancel — preventing double-release and permit leaks.
- `publishOn(Schedulers.fromExecutor(ctx.getExternalCallExecutor()))` moves all callbacks off the reactor-netty I/O threads onto the rule-engine external executor.

## How it works

```
 processMessage(ctx, msg)
         │
         ▼
 ┌────────────────────────────────────────────────────────────────┐
 │  LinkedBlockingQueue  (capacity: maxPendingRequests, unbounded) │
 │                                                           │
 │  [task₁][task₂][task₃] ... [task₁₀₀₀]  ← offer()        │
 │                                           returns false   │
 │                              if full ──→  onFailure()     │
 └───────────────────────────────────────────────────────────┘
         │
         │  offer() succeeded → tryProcess()
         ▼
 ┌─────────────────────────────────────────────────────────────────┐
 │  tryProcess()                                                   │
 │                                                                 │
 │  ┌──────────────────────────────────────────────┐              │
 │  │  Semaphore  (permits: maxParallelRequests=3)  │              │
 │  │                                              │              │
 │  │  tryAcquire() ──── no free permit ──────────►│ return       │
 │  │       │                                      │ (a callback  │
 │  │       │ permit acquired                      │  will retry) │
 │  └───────┼──────────────────────────────────────┘              │
 │          │                                                      │
 │          ▼                                                      │
 │     queue.poll()                                                │
 │          │                                                      │
 │          ├── null (queue empty) ──► semaphore.release() return  │
 │          │                                                      │
 │          ▼                                                      │
 │     msg.isValid()?   ◄── TbMsgCallback.isMsgValid()            │
 │          │                (batch pack deadline expired?)        │
 │          │                                                      │
 │          ├── false ──► semaphore.release()                      │
 │          │             onFailure("no longer valid")  ┐          │
 │          │             continue loop  ◄──────────────┘          │
 │          │             (drain all consecutive stale tasks)      │
 │          │                                                      │
 │          └── true ──► doHttpCall(task)   return                 │
 └─────────────────────────────────────────────────────────────────┘
                              │
                              │  WebClient  (async, non-blocking)
                              ▼
                    ┌─────────────────┐
                    │  HTTP Request   │  (1 of up to 3 in-flight)
                    │  in-flight      │
                    └────────┬────────┘
                             │
                    publishOn(externalCallExecutor)
                             │
              ┌──────────────┴──────────────┐
              │  2xx                         │  error / non-2xx
              ▼                             ▼
        onSuccess(msg)              onFailure(msg, t)
              │                             │
              └──────────┬──────────────────┘
                         │
                         ▼
                  doFinally hook            ← fires exactly once (onComplete/onError/cancel)
                  semaphore.release()       ← slot freed
                  tryProcess()             ← drive queue forward
```

**Semaphore lifecycle with maxParallel=3:**

```
initial:    [●●●]  3 free permits

msg₁ arrives → tryProcess → acquire → [●●○]  2 free  → HTTP in-flight
msg₂ arrives → tryProcess → acquire → [●○○]  1 free  → HTTP in-flight
msg₃ arrives → tryProcess → acquire → [○○○]  0 free  → HTTP in-flight
msg₄ arrives → tryProcess → tryAcquire fails → queued
msg₅ arrives → tryProcess → tryAcquire fails → queued

msg₁ completes → release → [●○○] → tryProcess → acquire → [○○○] → msg₄ fires
msg₂ completes → release → [●○○] → tryProcess → acquire → [○○○] → msg₅ fires
```

**Invalidation (stale task drain):**

```
Rule engine batch deadline
        │
        ▼
TbMsgPackCallback.isMsgValid() → false
        │
        │  msg sits in queue → tryProcess polls it
        ▼
msg.isValid() == false
        │
        ├── semaphore.release()          (slot immediately reused)
        ├── onFailure("no longer valid") (rule chain failure path)
        └── continue loop               (next task checked in same pass)
```

Stale tasks never hold a semaphore slot — the slot is released *before* `onFailure` is called, so the next valid task can be picked up in the same `tryProcess` loop pass.

## `isValid()` behaviour vs `forceAck`

`msg.isValid()` resolves to `TbMsgCallback.isMsgValid()`. The callback instance depends on whether `forceAck` is enabled:

| `forceAck` | callback on enqueued task | `isValid()` while queued |
|------------|--------------------------|--------------------------|
| `false` | original `TbMsgPackCallback` — tied to the batch pack deadline | can become `false` when the batch expires |
| `true` | `TbMsgCallback.EMPTY` (set explicitly by `copyWithNewCtx()`) | always `true` |

When `forceAck=true`, `ackIfNeeded` calls `ctx.ack(msg)` to release the original batch pack, then returns `msg.copyWithNewCtx().build()`. `copyWithNewCtx()` hardcodes `.callback(TbMsgCallback.EMPTY)`, fully detaching the new task from any deadline:

```java
// TbMsg.java
public TbMsgBuilder copyWithNewCtx() {
    return copy()
            .ctx(ctx.copy())
            .callback(TbMsgCallback.EMPTY);  // no expiry, always valid
}
```

This means:
- **`forceAck=false`**: `isValid()` is the meaningful guard. If a message waits in the queue past its batch deadline, `tryProcess()` drops it and routes it to the rule chain failure path.
- **`forceAck=true`**: `isValid()` is always `true` — the task has no expiry and **must** eventually be processed. It is the sole carrier of that work (the original was already acked). The only bound on how long it can wait is `maxPendingRequests`.

## Configuration

| Field | Default | Meaning |
|-------|---------|---------|
| `maxParallelRequestsCount` | `0` (unlimited) | Max concurrent in-flight HTTP calls |

`maxParallelRequestsCount` is a pure concurrency throttle — it never causes failures on its own. Only a full pending queue triggers `onFailure`.

### Request pipeline layers

Every message passes through these layers in order:

```
actor thread (single-threaded)
  → pendingQueue          (maxPendingRequests  — ThingsBoard queue layer)
    → semaphore           (maxParallelRequests — ThingsBoard throttle layer)
      → WebClient
        → connection pool (poolMaxConnections  — reactor-netty transport layer)
          → remote server
```

`maxPendingRequests` and `maxParallelRequests` operate at the ThingsBoard level — full visibility, drop logic, and anomaly logging. `poolMaxConnections` is a transport-level concern: exhausting it causes reactor-netty to internally queue requests, invisible to ThingsBoard.

### System-level safety caps

In SaaS deployments, per-tenant rule node settings are untrusted. Operator-level ceilings are declared in `thingsboard.yml` (YAML comments are auto-generated into documentation) and flow through a Spring component → `ActorSystemContext` → `TbContext` chain — no UI changes, no upgrade scripts, just redeploy:

| `thingsboard.yml` key | Env var | What it caps | Behaviour when set |
|-----------------------|---------|--------------|--------------------|
| `actors.rule.external.http_client.max_parallel_requests` | `ACTORS_RULE_EXTERNAL_HTTP_CLIENT_MAX_PARALLEL_REQUESTS` | In-flight HTTP slots per node | `userConfig=0` → forced to `N`; `userConfig > 0` → `min(user, N)` |
| `actors.rule.external.http_client.max_pending_requests` | `ACTORS_RULE_EXTERNAL_HTTP_CLIENT_MAX_PENDING_REQUESTS` | Queue depth per node | same logic |
| `actors.rule.external.http_client.pool_max_connections` | `ACTORS_RULE_EXTERNAL_HTTP_CLIENT_POOL_MAX_CONNECTIONS` | TCP connection pool size per node | `0` = reactor-netty default: `max(availableProcessors, 8) * 2` (e.g. 16 on 8-core) |

`RestApiCallNodeSettingsComponent` is annotated `@TbRuleEngineComponent` so it is only instantiated in `monolith` and `tb-rule-engine` service types.

### What's in this PR

`TbHttpClient` now carries five `AtomicLong` counters (`dispatched`, `success`, `failure`, `droppedQueueFull`, `droppedStale`) and a `tenantId`/`nodeId` pair passed from `TbRestApiCallNode.init()`.

**Logging strategy:**
- Individual drops (queue full, stale) → **DEBUG** — verbose but off by default; searchable in ELK/Kibana when needed.
- Periodic anomaly summary → **WARN** at most once per 60 s, only when drops have occurred since the last report. Each line includes `tenantId`, `nodeId`, all five counters, queue depth, queue remaining capacity, and `semaphorePermits` (permits stuck at `0` while `queueDepth > 0` and counts stop growing indicates a leak).
- Node lifecycle → **WARN** on `destroy()` if any anomalies happened during the lifetime; **DEBUG** otherwise.

Example WARN (periodic):
```
[<tenantId>][<nodeId>] REST API call node anomalies: droppedQueueFull=42, droppedStale=7
  (dispatched=1205, success=1180, failure=25, queueDepth=318, queueRemaining=682, semaphorePermits=0).
```

This lets ops search for any tenant experiencing drops (`droppedQueueFull > 0` or `droppedStale > 0`) with a single Kibana query, without flooding logs on healthy nodes.

### Follow-up (not in this PR)

- **JMX MBean** — expose the five counters and live `queueSize` / `availablePermits` as a `TbHttpClientMXBean` per node instance (object name tagged with `tenantId` + `nodeId`); register in constructor, unregister in `destroy()`. No Spring dependency needed.
- **Micrometer gauges + counters** — requires `StatsFactory getStatsFactory()` to be added to the `TbContext` API. Once available, register Micrometer `Gauge` for `pendingQueue.size()` and `semaphore.availablePermits()` (polled by Prometheus), plus `StatsCounter` for each of the five outcomes. Unregister in `destroy()` to avoid reference leaks.
- **Grafana panels** — queue fill % by tenant (alert when approaching `maxPendingRequests`), in-flight utilization, drop rate, p95 HTTP RTT via timer.

## Backward compatibility

Updating the docker image is sufficient to deliver this fix — no configuration changes, no upgrade scripts, no UI changes required.

| Concern | Status |
|---------|--------|
| `TbRestApiCallNodeConfiguration` fields | **Unchanged** — existing persisted node configs load without migration |
| Node version | **Still `3`** — no upgrade script triggered |
| `maxParallelRequestsCount=0` (default) | **Identical behavior** — `semaphore == null` → direct `doHttpCall`, same path as before |
| `maxParallelRequestsCount > 0` (existing users) | **Intentional improvement** — old code blocked the actor thread for `readTimeoutMs` then failed with `"Timeout during waiting for reply!"`; new code enqueues non-blocking and processes the message as soon as a slot opens |
| `TB_RE_HTTP_CLIENT_POOL_MAX_CONNECTIONS` (old env var) | **Honored** via yaml fallback: `${ACTORS_RULE_EXTERNAL_HTTP_CLIENT_POOL_MAX_CONNECTIONS:${TB_RE_HTTP_CLIENT_POOL_MAX_CONNECTIONS:0}}` |
| New yaml keys (`actors.rule.external.http_client.*`) | **All default to `0`** = no system-level restriction |
| Non-rule-engine services (`tb-core`, `tb-transport`, etc.) | **Unaffected** — `RestApiCallNodeSettingsComponent` is `@TbRuleEngineComponent` (not instantiated); `ActorSystemContext` field is `@Autowired(required = false)` |

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
@@ -33,6 +33,7 @@ import org.springframework.web.reactive.function.client.WebClientResponseExcepti
 import org.springframework.web.util.UriComponentsBuilder;
 import org.thingsboard.common.util.JacksonUtil;
 import org.thingsboard.common.util.SsrfProtectionValidator;
+import org.thingsboard.rule.engine.api.TbHttpClientSettings;
 import org.thingsboard.rule.engine.api.TbContext;
 import org.thingsboard.rule.engine.api.TbNodeException;
 import org.thingsboard.rule.engine.api.util.TbNodeUtils;
@@ -42,6 +43,7 @@ import org.thingsboard.rule.engine.credentials.CredentialsType;
 import org.thingsboard.server.common.data.StringUtils;
 import org.thingsboard.server.common.msg.TbMsg;
 import org.thingsboard.server.common.msg.TbMsgMetaData;
+import reactor.core.scheduler.Schedulers;
 import reactor.netty.http.client.HttpClient;
 import reactor.netty.resources.ConnectionProvider;
 import reactor.netty.transport.ProxyProvider;
@@ -53,8 +55,11 @@ import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
@@ -84,17 +89,48 @@ public class TbHttpClient {
 
     public static final String MAX_IN_MEMORY_BUFFER_SIZE_IN_KB = "tb.http.maxInMemoryBufferSizeInKb";
 
+    private static final long ANOMALY_REPORT_INTERVAL_MS = 60_000;
+
     private final TbRestApiCallNodeConfiguration config;
+    private final String tenantId;
+    private final String nodeId;
+    private final TbHttpClientSettings settings;
 
     private EventLoopGroup eventLoopGroup;
     private WebClient webClient;
     private Semaphore semaphore;
+    private BlockingQueue<PendingTask> pendingQueue;
+
+    private final AtomicLong dispatchedCount   = new AtomicLong();
+    private final AtomicLong successCount      = new AtomicLong();
+    private final AtomicLong failureCount      = new AtomicLong();
+    private final AtomicLong droppedFullCount  = new AtomicLong();
+    private final AtomicLong droppedStaleCount = new AtomicLong();
+    private volatile long lastAnomalyReportAt  = 0;
+
+    private record PendingTask(
+            TbContext ctx,
+            TbMsg msg,
+            Consumer<TbMsg> onSuccess,
+            BiConsumer<TbMsg, Throwable> onFailure,
+            long enqueuedNanos) {}
 
     TbHttpClient(TbRestApiCallNodeConfiguration config, EventLoopGroup eventLoopGroupShared) throws TbNodeException {
+        this(config, eventLoopGroupShared, "n/a", "n/a", TbHttpClientSettings.DEFAULT);
+    }
+
+    TbHttpClient(TbRestApiCallNodeConfiguration config, EventLoopGroup eventLoopGroupShared,
+                 String tenantId, String nodeId, TbHttpClientSettings settings) throws TbNodeException {
         try {
             this.config = config;
-            if (config.getMaxParallelRequestsCount() > 0) {
-                semaphore = new Semaphore(config.getMaxParallelRequestsCount());
+            this.tenantId = tenantId;
+            this.nodeId = nodeId;
+            this.settings = settings;
+            int effectiveParallel = effectiveMax(config.getMaxParallelRequestsCount(), settings.getMaxParallelRequests());
+            if (effectiveParallel > 0) {
+                semaphore = new Semaphore(effectiveParallel);
+                int effectivePending = effectiveMax(0, settings.getMaxPendingRequests());
+                pendingQueue = effectivePending > 0 ? new LinkedBlockingQueue<>(effectivePending) : new LinkedBlockingQueue<>();
             }
 
             ConnectionProvider connectionProvider = ConnectionProvider
@@ -155,16 +191,20 @@ public class TbHttpClient {
         }
     }
 
-    private int getPoolMaxConnections() {
-        String poolMaxConnectionsEnv = System.getenv("TB_RE_HTTP_CLIENT_POOL_MAX_CONNECTIONS");
+    /**
+     * Returns the effective limit: {@code min(userMax, systemMax)} when both are positive,
+     * {@code systemMax} when only the system ceiling is set, or {@code userMax} otherwise.
+     * A value of {@code 0} means unlimited.
+     */
+    private static int effectiveMax(int userMax, int systemMax) {
+        if (systemMax <= 0) return userMax;
+        if (userMax <= 0)   return systemMax;
+        return Math.min(userMax, systemMax);
+    }
 
-        int poolMaxConnections;
-        if (poolMaxConnectionsEnv != null) {
-            poolMaxConnections = Integer.parseInt(poolMaxConnectionsEnv);
-        } else {
-            poolMaxConnections = ConnectionProvider.DEFAULT_POOL_MAX_CONNECTIONS;
-        }
-        return poolMaxConnections;
+    private int getPoolMaxConnections() {
+        int configured = settings.getPoolMaxConnections();
+        return configured > 0 ? configured : ConnectionProvider.DEFAULT_POOL_MAX_CONNECTIONS;
     }
 
     private void validateMaxInMemoryBufferSize(TbRestApiCallNodeConfiguration config) throws TbNodeException {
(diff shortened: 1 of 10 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/rest/TbHttpClient.java
```diff
@@ -212,49 +224,86 @@ void destroy() {
...
-
-                        onFailure.accept(processException(msg, throwable), processThrowable(throwable));
                     });
-        } catch (InterruptedException e) {
-            log.warn("Timeout during waiting for reply!", e);
+        } catch (Exception e) {
+            task.onFailure().accept(processException(task.msg(), e), processThrowable(e));
+            if (semaphore != null) {
+                semaphore.release();
+                tryProcess();
```
ViacheslavKlimov: This recursive `tryProcess()` call can cause a `StackOverflowError` when many queued tasks fail synchronously. If the URL pattern is misconfigured (e.g., empty or missing scheme), every task in the queue will throw in `buildEncodedUri()`, and the call chain becomes:

```
tryProcess() → doHttpCall() → [throws] → catch → tryProcess() → doHttpCall() → [throws] → catch → ...
```

With `maxPendingRequests=0` (unbounded, the default), a sustained message burst to a misconfigured node could build up thousands of queued tasks, each adding ~3-4 stack frames — enough to overflow.

`SemaphoreWithTbMsgQueue` (the stated pattern reference) avoids this by handling synchronous failures iteratively with `continue` inside its `while` loop, never recursing through the HTTP call method. Consider a similar approach here — for example, have `doHttpCall` return a boolean indicating whether it started an async operation, and loop in `tryProcess()` on synchronous failure:

```java
private void tryProcess() {
    while (true) {
        if (!semaphore.tryAcquire()) return;
        PendingTask next = pendingQueue.poll();
        if (next == null) { semaphore.release(); return; }
        if (!next.msg().isValid()) {
            semaphore.release();
            log.warn("...");
            next.onFailure().accept(next.msg(), ...);
            continue;
        }
        if (doHttpCall(next)) return; // async in-flight → callback will drive the queue
        // doHttpCall returned false → synchronous failure, semaphore already released, loop
    }
}

// returns true if async call was started, false if failed synchronously
private boolean doHttpCall(PendingTask task) {
    try {
        // ... build and fire request ...
        return true;
    } catch (Exception e) {
        task.onFailure().accept(...);
        if (semaphore != null) { semaphore.release(); }
        return false; // caller loops instead of recursing
    }
}
```

ViacheslavKlimov (commented):
## Review summary

Reviewed 4 changed files in [Fix REST API Call node blocking actor thread and semaphore permit leak](https://github.com/thingsboard/thingsboard/pull/15334). Left 1 comment(s) inline.

The approach is solid — replacing the blocking `semaphore.tryAcquire(timeout)` with a non-blocking bounded FIFO queue + semaphore pattern cleanly solves both the actor-thread blocking and the permit leak. The `publishOn` to move callbacks off reactor-netty I/O threads is a nice improvement. Tests cover the key scenarios well (forceAck path, synchronous failure + semaphore release, queue overflow behavior).

---

PR #15336, opened 2026-04-01
Added audit logging for sysadmin-level settings

zzzeebra:
## Summary
- Add audit log entries for Admin settings (General, Mail/SMS), Security settings, JWT settings, and 2FA platform settings in `AdminController` and `TwoFactorAuthConfigController`
- Add audit logging for Queue create/update/delete operations in `DefaultTbQueueService` via `logEntityActionService`
- Add tests for all new audit log entries in `AdminControllerTest`, `TwoFactorAuthConfigTest`, and `BaseQueueControllerTest`

## Test plan
- [x] `AdminControllerTest` — 11/11 tests pass (3 new audit log tests)
- [x] `TwoFactorAuthConfigTest` — 21/21 tests pass (1 new audit log test)
- [x] `BaseQueueControllerTest` — 7/7 tests pass (2 new audit log tests)

## Manual tests (sysadmin)
- [x] System settings → General → Save → audit log `Settings updated` with `settingsKey=general`
- [x] Security settings → General → Save → audit log `Settings updated` with `settingsKey=securitySettings`
- [x] Security settings → JWT security settings section → Save → audit log `Settings updated` with `settingsKey=jwtSettings`
- [x] Security settings → Platform 2FA settings → Save → audit log `Settings updated` with `settingsKey=twoFaSettings`
- [x] Queues → Add new queue → audit log `Added` on `QUEUE` entity
- [x] Queues → Delete queue → audit log `Deleted` on `QUEUE` entity

Screenshot 2026-05-15 at 11 59 41

-----

Screenshot 2026-05-15 at 11 59 52


application/src/test/java/org/thingsboard/server/controller/AdminControllerTest.java
```diff
@@ -181,4 +189,51 @@ public void testCreateJwtSettings() throws Exception {
...
+    }
+
+    private void awaitAuditLog(ActionType expectedAction) {
+        Awaitility.await("Await audit log: " + expectedAction)
+                .atMost(TIMEOUT, TimeUnit.SECONDS)
+                .until(() -> doGetTypedWithTimePageLink(
+                        "/api/audit/logs/entity/USER/" + currentUserId.getId() + "?",
+                        new TypeReference<PageData<AuditLog>>() {},
+                        new TimePageLink(100)).getData().stream()
+                        .anyMatch(log -> log.getActionType() == expectedAction)
```
ViacheslavKlimov: This helper only checks for the existence of *any* `UPDATED` audit log on the user entity — it doesn't verify which settings key was changed. Since all three audit-log tests in this class (`testSaveAdminSettingsAuditLog`, `testSaveSecuritySettingsAuditLog`, `testSaveJwtSettingsAuditLog`) assert the same `ActionType.UPDATED` on the same user, the second and third tests could pass immediately off a leftover audit log from an earlier test if the database isn't cleaned between methods.

Consider filtering on the audit log's action data (additional info) to match the specific settings key (e.g. `"general"`, `"securitySettings"`, `"jwtSettings"`), or at minimum recording a timestamp before the action and filtering to logs created after that timestamp.

application/src/main/java/org/thingsboard/server/controller/TwoFactorAuthConfigController.java
```diff
@@ -257,7 +258,17 @@ public PlatformTwoFaSettings getPlatformTwoFaSettings() throws ThingsboardExcept
     @PreAuthorize("hasAnyAuthority('SYS_ADMIN')")
     public PlatformTwoFaSettings savePlatformTwoFaSettings(@Parameter(description = "Settings value", required = true)
                                                            @RequestBody PlatformTwoFaSettings twoFaSettings) throws ThingsboardException {
-        return twoFaConfigManager.savePlatformTwoFaSettings(getTenantId(), twoFaSettings);
+        SecurityUser securityUser = getCurrentUser();
+        try {
+            PlatformTwoFaSettings savedSettings = twoFaConfigManager.savePlatformTwoFaSettings(getTenantId(), twoFaSettings);
+            auditLogService.logEntityAction(securityUser.getTenantId(), securityUser.getCustomerId(), securityUser.getId(), securityUser.getName(),
```
ViacheslavKlimov: The full `auditLogService.logEntityAction(securityUser.getTenantId(), securityUser.getCustomerId(), securityUser.getId(), securityUser.getName(), securityUser.getId(), securityUser, ...)` call is duplicated between the success path (here) and the error path (line 268). `AdminController` avoids this with a `logSettingsAction` helper — might be worth doing the same here for consistency and to reduce the duplication.

application/src/main/java/org/thingsboard/server/service/entitiy/queue/DefaultTbQueueService.java
```diff
@@ -47,26 +50,40 @@ public class DefaultTbQueueService extends AbstractTbEntityService implements Tb
...
-            oldQueue = null;
-        } else {
-            oldQueue = queueService.findQueueById(queue.getTenantId(), queue.getId());
-        }
+    public Queue saveQueue(Queue queue, User user) {
+        ActionType actionType = queue.getId() == null ? ActionType.ADDED : ActionType.UPDATED;
+        TenantId tenantId = queue.getTenantId();
+        try {
+            Queue oldQueue;
+            if (queue.getId() == null) {
```
ViacheslavKlimov: Minor: `queue.getId() == null` is checked here and also on line 54 (to determine `actionType`). You could reuse the `actionType` variable instead — `if (actionType == ActionType.ADDED)` — to avoid the redundant null check.

ViacheslavKlimov (commented):
## Review summary

Reviewed 8 changed files in [Add audit logging for Sysadmin-level settings](https://github.com/thingsboard/thingsboard/pull/15336). Left 3 comment(s) inline.

Clean, well-structured PR. The audit logging pattern is consistent with how other entity services handle it, and the tests cover create/update/delete paths. A few observations below.

---

application/src/test/java/org/thingsboard/server/controller/AdminControllerTest.java
zzzeebra: Good catch. Added timestamp filtering — tests now record `startTs` before the action and use `TimePageLink(100, 0, null, null, startTs, null)` to only match audit logs created after that moment. This eliminates any false positives from leftover logs.

application/src/main/java/org/thingsboard/server/controller/TwoFactorAuthConfigController.java
zzzeebra: Good point about consistency. In this case there's only a single method (`savePlatformTwoFaSettings`) with the try/catch audit pattern, so extracting a helper for just 2 lines felt like overengineering. In `AdminController` the helper was justified by 3 methods × 2 calls = 6 duplicates. If more audit-logged methods appear in this controller later, we'll extract a helper then.

application/src/main/java/org/thingsboard/server/service/entitiy/queue/DefaultTbQueueService.java
zzzeebra: Good catch, fixed — now reusing `actionType == ActionType.ADDED` instead of the redundant null check.

ViacheslavKlimov (commented):
## Re-review summary

Re-reviewed [Add audit logging for Sysadmin-level settings](https://github.com/thingsboard/thingsboard/pull/15336) — verified 3 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/15336#pullrequestreview-4049446908).

| Status | Count |
|--------|-------|
| ✅ Resolved | 2 |
| 💬 Acknowledged | 1 |

### Finding details

- ✅ **AdminControllerTest.java:awaitAuditLog** — Tests could pass off leftover audit logs from earlier tests — *Fixed in code: `startTs` timestamp filtering added to all audit-log tests and the helper method.*
- 💬 **TwoFactorAuthConfigController.java:264** — Duplicated `auditLogService.logEntityAction(...)` call between success and error paths — *Developer: "Only 1 method uses this pattern here, so a helper would be overengineering. AdminController's helper was justified by 3 methods × 6 call sites. Will extract if more audit-logged methods appear."*
- ✅ **DefaultTbQueueService.java:58** — Redundant `queue.getId() == null` check when `actionType` already captures that — *Fixed in code: now uses `actionType == ActionType.ADDED`.*

---

PR #15343, opened 2026-04-01
[Openapi] Response schema improvements

dashevchenko:
## Pull Request description

Added missed response schemas for some endpoints in DashboardController, deleted redundant required mode for some entity fields, added appropriate media type for some api methods.
PE: https://github.com/thingsboard/thingsboard-pe/pull/4575

## General checklist

## Back-End feature checklist

### DIFF
--- a/common/data/src/main/java/org/thingsboard/server/common/data/Device.java
+++ b/common/data/src/main/java/org/thingsboard/server/common/data/Device.java
@@ -176,7 +176,7 @@ public class Device extends BaseDataWithAdditionalInfo<DeviceId> implements HasL
         this.label = label;
     }
 
-    @Schema(description = "JSON object with Device Profile Id. If not provided, the default device profile will be used.")
+    @Schema(description = "JSON object with Device Profile Id. If not provided, the type will be used to determine the profile. If neither deviceProfileId nor type is specified, the default device profile will be used.")
     public DeviceProfileId getDeviceProfileId() {
         return deviceProfileId;
     }
--- a/common/data/src/main/java/org/thingsboard/server/common/data/asset/AssetProfileInfo.java
+++ b/common/data/src/main/java/org/thingsboard/server/common/data/asset/AssetProfileInfo.java
@@ -23,6 +23,7 @@ import lombok.ToString;
 import lombok.Value;
 import org.thingsboard.server.common.data.EntityInfo;
 import org.thingsboard.server.common.data.EntityType;
+import org.thingsboard.server.common.data.id.AssetProfileId;
 import org.thingsboard.server.common.data.id.DashboardId;
 import org.thingsboard.server.common.data.id.EntityId;
 import org.thingsboard.server.common.data.id.EntityIdFactory;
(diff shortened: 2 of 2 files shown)


common/data/src/main/java/org/thingsboard/server/common/data/Device.java
```diff
@@ -176,7 +176,7 @@ public void setLabel(String label) {
         this.label = label;
     }
 
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED, description = "JSON object with Device Profile Id.")
+    @Schema(description = "JSON object with Device Profile Id. If not provided, the default device profile will be used.")
```
ViacheslavKlimov: Are you sure? Not device profile resolved from `type` field??

application/src/main/java/org/thingsboard/server/controller/RpcV1Controller.java
```diff
@@ -50,7 +52,8 @@ public class RpcV1Controller extends AbstractRpcController {
     public DeferredResult<ResponseEntity> handleOneWayDeviceRPCRequestV1(
             @Parameter(description = DEVICE_ID_PARAM_DESCRIPTION)
             @PathVariable("deviceId") String deviceIdStr,
-            @Parameter(description = "A JSON value representing the RPC request.")
+            @io.swagger.v3.oas.annotations.parameters.RequestBody(description = "A JSON object representing the RPC request.",
+                    content = @Content(mediaType = "text/plain", schema = @Schema(type = "string")))
```
ViacheslavKlimov, line 56: The description says "A JSON object representing the RPC request" but the media type is `text/plain`. Clients typically send `Content-Type: application/json` to these endpoints, and the body *is* a JSON object — it's just received as a raw `String` by Spring.

Using `text/plain` in the OpenAPI spec could mislead API consumers and cause generated clients to set the wrong Content-Type. Consider `application/json` with `schema = @Schema(type = "object")` (or keep `type = "string"` if you want to emphasize the freeform nature), which would more accurately reflect how these endpoints are actually called.

This same pattern is applied across `RpcV2Controller`, `RuleEngineController`, and `TelemetryController` — so worth deciding on the right media type once and applying it consistently.

common/data/src/main/java/org/thingsboard/server/common/data/DeviceProfileInfo.java
```diff
@@ -75,4 +76,10 @@ public DeviceProfileInfo(DeviceProfile profile) {
                 profile.getType(), profile.getTransportType());
     }
 
+    @Override
+    @Schema(implementation = DeviceProfileId.class, description = "JSON object with the Device Profile Id.")
```
ViacheslavKlimov, line 80: Good fix — without this, the OpenAPI schema for `DeviceProfileInfo.id` would show the generic `EntityId` type from the parent class.

`AssetProfileInfo` has the same structure (extends `EntityInfo`, stores an `AssetProfileId`) but doesn't have this override yet. Was that intentional, or should it get the same treatment for consistency?

PR #15346, opened 2026-04-02
SNMP: defer querying tasks until transport session is registered

volodymyr-babak:
## Pull Request description

Fixes a race condition in the SNMP transport where querying tasks could start polling a device before its transport session was fully registered, leading to dropped responses and noisy No values in the response warnings with no actionable context.

Scope of changes

SnmpTransportContext.establishDeviceSession() — no longer creates querying tasks immediately after putting the session into the sessions map. Tasks are now created later, once registration is confirmed.

SnmpTransportContext.onSuccess() (ValidateDeviceCredentialsResponse callback) — moved snmpTransportService.createQueryingTasks(sessionContext) to run after registerTransportSession(...) and the STARTED lifecycle event have completed. Added an info log to make the ordering visible in production logs.

SnmpTransportService.createQueryingTasks() — the scheduled task now checks both sessionContext.isActive() and sessionContext.isConnected() before sending a request, preventing requests on sessions that are active but not yet/no longer connected.

SnmpTransportService.processResponse() — enriched the No values in the response warning to include the communication spec, response PDUs, and the response mapping count, so the warning is actually diagnosable.

Why

Previously, createQueryingTasks was invoked synchronously inside establishDeviceSession — before the asynchronous validateDeviceCredentials / registerTransportSession callback chain completed. The first poll could fire against a session that was not yet registered with the transport service, producing empty responses and confusing logs.

## General checklist

## Back-End feature checklist

### DIFF
--- a/common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
+++ b/common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
@@ -164,7 +164,7 @@ public class SnmpTransportService implements TbTransportService, CommandResponde
                     ScheduledTask scheduledTask = new ScheduledTask();
                     scheduledTask.init(() -> {
                         try {
-                            if (sessionContext.isActive()) {
+                            if (sessionContext.isActive() && sessionContext.isConnected()) {
                                 return sendRequest(sessionContext, repeatingCommunicationConfig);
                             }
                         } catch (Exception e) {
@@ -390,7 +390,9 @@ public class SnmpTransportService implements TbTransportService, CommandResponde
 
         JsonObject responseData = responseDataMappers.get(requestContext.getCommunicationSpec()).map(response, requestContext);
         if (responseData.size() == 0) {
-            log.warn("[{}] No values in the response", sessionContext.getDeviceId());
+            log.warn("[{}] No values in the response for spec {}. Response PDUs: {}, Mappings count: {}",
+                    sessionContext.getDeviceId(), requestContext.getCommunicationSpec(),
+                    response, requestContext.getResponseMappings() != null ? requestContext.getResponseMappings().size() : 0);
             throw new IllegalArgumentException("No values in the response");
         }
 
--- a/common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
+++ b/common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
@@ -390,9 +390,11 @@ public class SnmpTransportService implements TbTransportService, CommandResponde
 
         JsonObject responseData = responseDataMappers.get(requestContext.getCommunicationSpec()).map(response, requestContext);
         if (responseData.size() == 0) {
-            log.warn("[{}] No values in the response for spec {}. Response PDUs: {}, Mappings count: {}",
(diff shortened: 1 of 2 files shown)


common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
```diff
@@ -390,7 +390,9 @@ private void processResponse(DeviceSessionContext sessionContext, List<PDU> resp
 
         JsonObject responseData = responseDataMappers.get(requestContext.getCommunicationSpec()).map(response, requestContext);
         if (responseData.size() == 0) {
-            log.warn("[{}] No values in the response", sessionContext.getDeviceId());
+            log.warn("[{}] No values in the response for spec {}. Response PDUs: {}, Mappings count: {}",
+                    sessionContext.getDeviceId(), requestContext.getCommunicationSpec(),
+                    response, requestContext.getResponseMappings() != null ? requestContext.getResponseMappings().size() : 0);
```
ViacheslavKlimov: Logging the full `response` (a `List<PDU>`) at WARN level is a fairly heavy footprint for what is fundamentally a routine "didn't match any mapping" condition. A misconfigured device profile can fire this on every poll, and SNMP PDUs include the variable bindings — i.e., real device-side values — which then end up in production WARN logs. Two suggestions worth considering:

- Move the PDU dump to DEBUG/TRACE and keep the WARN minimal (deviceId, spec, pdu count, mappings count). The summary alone is usually enough to localize the misconfig, and the detail is available when the log level is bumped during investigation.
- If you do want the PDU contents at WARN, consider logging just sizes (`response.size()`, per-PDU variable-binding counts) rather than `response.toString()`, to avoid leaking arbitrary device payload into INFO/WARN-tier log sinks.

ViacheslavKlimov: Minor: the `requestContext.getResponseMappings() != null` guard here is inconsistent with how the same field is used elsewhere in this class — `onToDeviceRpcRequest` does `communicationConfig.getAllMappings().stream()` (line 263) and the default response data mapper at line 417-419 passes the value straight to `pduService.processPdus(pdus, requestContext.getResponseMappings())`, neither of which null-checks. If `responseMappings` can legitimately be null, those sites would NPE before reaching this log line; if it cannot, the null check here is dead. Probably worth dropping the ternary and just calling `.size()` for consistency.

ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [fix(snmp): defer querying tasks until transport session is registered](https://github.com/thingsboard/thingsboard/pull/15346). The core fix — moving `createQueryingTasks` out of `establishDeviceSession` and into the auth `onSuccess` callback after `registerTransportSession` has run — looks correct and addresses the described race. The added `isConnected()` guard is a sensible defensive belt-and-braces for the same scenario. Left 2 inline comments, both about the enriched warning log in `processResponse`.

---

common/transport/snmp/src/main/java/org/thingsboard/server/transport/snmp/service/SnmpTransportService.java
volodymyr-babak: updated

volodymyr-babak: warn minimal - the rest move to debug

PR #15363, opened 2026-04-03
Fixed resolve stale WebSocket cmdId after reconnect causing lost subscription updates

deaflynx:
## Summary

Fixes an issue where dashboard widgets stop receiving data after a WebSocket
reconnect (commonly triggered on mobile when the screen turns off/on, but also
reproducible on desktop via network interruption or server restart).

## Root cause

When the WebSocket connection drops and reconnects, `subscribe()` assigns **new
cmdIds** to the subscriber's `subscriptionCommands`. However, other command
objects held by `entity-data-subscription` (e.g. `this.dataCommand`) retain the
**old cmdId** from the previous WS session.

When `update()` is called (e.g. time window refresh on resume, or user changing
the date range), it sends commands with the stale cmdId. The server responds
using that cmdId, but `subscribersMap` only contains the new cmdId — so the
response is silently dropped and the widget spinner never stops.

A secondary issue: `update()` is blocked by `if (!this.isReconnect)` during the
reconnect window. If an update is triggered between `onClose` and `onOpen`
(common on Android where `onWindowFocus` fires before the reconnect timer), the
update is silently dropped and never replayed.

## Changes

**`websocket.service.ts`**
- Added `pendingUpdates` set to queue updates requested during reconnect
- Added `processPendingUpdates()` — replays queued updates after reconnect
  completes and subscribers are re-registered with fresh cmdIds
- `pendingUpdates` cleared on full reset (logout) but preserved during reconnect

**`telemetry-websocket.service.ts`**
- Added `syncCommandId()` — when `update()` detects a command with a cmdId not
  present in `subscribersMap`, it resolves the subscriber's actual registered
  cmdId via reverse lookup and corrects the command before sending
- `update()` now queues to `pendingUpdates` when `isReconnect` is true instead
  of silently dropping the update

## Test plan

- [ ] Open a dashboard with a timeseries widget on mobile
- [ ] Turn screen off, wait a few seconds until WebSocket closes (code 1006)
- [ ] Turn screen on — dashboard should display data without stuck spinner
- [ ] Change the date range — widget should load new data normally
- [ ] Verify normal desktop WebSocket behavior is unaffected
- [ ] Verify reconnect after server restart works correctly

### DIFF
--- a/ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
+++ b/ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
@@ -81,6 +81,12 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
         const cmdId = this.nextCmdId();
         if (!(subscriptionCommand instanceof MarkAsReadCmd) && !(subscriptionCommand instanceof MarkAllAsReadCmd)) {
           this.subscribersMap.set(cmdId, subscriber);
+          let cmdIds = this.subscriberCmdIds.get(subscriber);
+          if (!cmdIds) {
+            cmdIds = new Set<number>();
+            this.subscriberCmdIds.set(subscriber, cmdIds);
+          }
+          cmdIds.add(cmdId);
         }
         subscriptionCommand.cmdId = cmdId;
         this.cmdWrapper.cmds.push(subscriptionCommand);
@@ -141,6 +147,13 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
           const cmdId = subscriptionCommand.cmdId;
           if (cmdId) {
             this.subscribersMap.delete(cmdId);
+            const cmdIds = this.subscriberCmdIds.get(subscriber);
+            if (cmdIds) {
+              cmdIds.delete(cmdId);
+              if (cmdIds.size === 0) {
+                this.subscriberCmdIds.delete(subscriber);
+              }
+            }
           }
         }
       );
@@ -183,11 +196,9 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
 
   private syncCommandId(subscriptionCommand: WebsocketCmd, subscriber: TelemetrySubscriber) {
     if (!this.subscribersMap.has(subscriptionCommand.cmdId)) {
-      for (const [cmdId, sub] of this.subscribersMap.entries()) {
-        if (sub === subscriber) {
-          subscriptionCommand.cmdId = cmdId;
-          break;
-        }
+      const cmdIds = this.subscriberCmdIds.get(subscriber);
+      if (cmdIds?.size) {
+        subscriptionCommand.cmdId = cmdIds.values().next().value;
       }
     }
   }
--- a/ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
+++ b/ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
@@ -195,7 +195,7 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
   }
 
   private syncCommandId(subscriptionCommand: WebsocketCmd, subscriber: TelemetrySubscriber) {
-    if (!this.subscribersMap.has(subscriptionCommand.cmdId)) {
+    if (this.subscribersMap.get(subscriptionCommand.cmdId) !== subscriber) {
       const cmdIds = this.subscriberCmdIds.get(subscriber);
       if (cmdIds?.size) {
         subscriptionCommand.cmdId = cmdIds.values().next().value;
--- a/ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
+++ b/ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
@@ -50,6 +50,7 @@ import {
   UnreadCountSubCmd,
   UnreadSubCmd,
   UnsubscribeCmd,
+  WebsocketCmd,
   WebsocketDataMsg
 } from '@app/shared/models/telemetry/telemetry.models';
 import { Store } from '@ngrx/store';
@@ -94,11 +95,14 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
       subscriber.subscriptionCommands.forEach(
         (subscriptionCommand) => {
           if (subscriptionCommand.cmdId && (subscriptionCommand instanceof EntityDataCmd || subscriptionCommand instanceof UnreadSubCmd)) {
+            this.syncCommandId(subscriptionCommand, subscriber);
             this.cmdWrapper.cmds.push(subscriptionCommand);
           }
         }
       );
       this.publishCommands();
+    } else {
+      this.pendingUpdates.add(subscriber);
     }
   }
 
@@ -173,6 +177,17 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
       subscriber = this.subscribersMap.get(message.subscriptionId) as TelemetrySubscriber;
       if (subscriber) {
         subscriber.onData(new SubscriptionUpdate(message));
+      }
+    }
+  }
+
+  private syncCommandId(subscriptionCommand: WebsocketCmd, subscriber: TelemetrySubscriber) {
+    if (!this.subscribersMap.has(subscriptionCommand.cmdId)) {
+      for (const [cmdId, sub] of this.subscribersMap.entries()) {
+        if (sub === subscriber) {
+          subscriptionCommand.cmdId = cmdId;
+          break;
+        }
       }
     }
   }
--- a/ui-ngx/src/app/core/ws/websocket.service.ts
+++ b/ui-ngx/src/app/core/ws/websocket.service.ts
@@ -52,6 +52,7 @@ export abstract class WebsocketService<T extends WsSubscriber> implements WsServ
   subscribersMap = new Map<number, TelemetrySubscriber | NotificationSubscriber>();
 
   reconnectSubscribers = new Set<WsSubscriber>();
+  pendingUpdates = new Set<T>();
 
   wsUri: string;
 
@@ -140,6 +141,7 @@ export abstract class WebsocketService<T extends WsSubscriber> implements WsServ
     if (close) {
       this.reconnectAttempts = 0;
       this.lastShownCloseCode = null;
+      this.pendingUpdates.clear();
       this.closeSocket();
     }
   }
@@ -223,6 +225,7 @@ export abstract class WebsocketService<T extends WsSubscriber> implements WsServ
         }
       );
       this.reconnectSubscribers.clear();
+      this.processPendingUpdates();
     } else {
       this.publishCommands();
     }
@@ -297,5 +300,14 @@ export abstract class WebsocketService<T extends WsSubscriber> implements WsServ
     this.store.dispatch(new ActionNotificationShow({
       message, type: notificationType
     }));
+  }
+
+  private processPendingUpdates() {
+    if (this.pendingUpdates.size > 0) {
+      this.pendingUpdates.forEach((subscriber) => {
+        this.update(subscriber);
+      });
+      this.pendingUpdates.clear();
+    }
   }
 }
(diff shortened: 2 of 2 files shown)


ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
```diff
@@ -177,4 +181,15 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
     }
   }
 
+  private syncCommandId(subscriptionCommand: WebsocketCmd, subscriber: TelemetrySubscriber) {
+    if (!this.subscribersMap.has(subscriptionCommand.cmdId)) {
+      for (const [cmdId, sub] of this.subscribersMap.entries()) {
+        if (sub === subscriber) {
```
vvlladd28: This reverse lookup finds the **first** cmdId in `subscribersMap` that maps to the given subscriber and assigns it to the command. If a subscriber has multiple entries in the map (i.e., was subscribed with multiple commands), this will assign the same cmdId to every command processed by `update()`, which is incorrect — each command should get its own corresponding cmdId.

Today this works because the `reconnect$` handler in `entity-data-subscription.ts` reduces `subscriptionCommands` to a single-element array before re-subscribing, so each subscriber ends up with exactly one entry in the map after reconnect. But this invariant is implicit and fragile — nothing enforces it, and a future change to the reconnect handler could silently break the sync.

A safer approach would be to track the mapping explicitly. For example, `subscribe()` could store the old-to-new cmdId mapping during reconnect, and `syncCommandId` could use that map instead of the reverse lookup. Alternatively, at minimum, add a comment documenting the single-entry invariant so future maintainers know not to break it.

ui-ngx/src/app/core/ws/websocket.service.ts
```diff
@@ -223,6 +225,7 @@ export abstract class WebsocketService<T extends WsSubscriber> implements WsServ
         }
       );
       this.reconnectSubscribers.clear();
+      this.processPendingUpdates();
```
vvlladd28, line 230: Could `pendingUpdates` contain a subscriber that is NOT in `reconnectSubscribers`? If so, `processPendingUpdates` would call `update()` for a subscriber that hasn't been re-subscribed — its commands would have stale cmdIds and no entry in `subscribersMap`, so `syncCommandId` would fail to find a match and silently send the stale cmdId to the server.

I believe this can't happen today because pending updates only accumulate while `isReconnect` is true, and all subscribers from the previous session are collected into `reconnectSubscribers` in `onClose`. But it might be worth adding a guard in `processPendingUpdates` — e.g., skip subscribers that aren't in `subscribersMap` — to be defensive.

ui-ngx/src/app/core/ws/websocket.service.ts
```diff
@@ -223,6 +225,7 @@ export abstract class WebsocketService<T extends WsSubscriber> implements WsServ
...
     } else {
```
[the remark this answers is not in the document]
deaflynx: Not a bug — `isReconnect` is set to `false` (line 220) before `processPendingUpdates()` is called (line 229). Additionally, `processPendingUpdates()` now copies the set into a local array and clears the set before iterating, so re-queuing during iteration is impossible.

ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
```diff
@@ -177,4 +181,15 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
...
+          subscriptionCommand.cmdId = cmdId;
+          break;
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
deaflynx: Fixed — replaced the O(n) reverse lookup with a dedicated `subscriberCmdIds: Map<WsSubscriber, Set<number>>` reverse index, maintained in `subscribe()`, `unsubscribe()`, and `reset()`. `syncCommandId()` now does an O(1) lookup via `this.subscriberCmdIds.get(subscriber)`.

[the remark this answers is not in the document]
deaflynx: Fixed — `subscriberCmdIds` reverse index provides O(1) lookup. See previous reply.

ui-ngx/src/app/core/ws/websocket.service.ts
```diff
@@ -298,4 +301,13 @@ export abstract class WebsocketService<T extends WsSubscriber> implements WsServ
       message, type: notificationType
     }));
   }
+
+  private processPendingUpdates() {
+    if (this.pendingUpdates.size > 0) {
+      this.pendingUpdates.forEach((subscriber) => {
+        this.update(subscriber);
+      });
+      this.pendingUpdates.clear();
```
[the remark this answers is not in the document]
deaflynx: Not needed — after mobile resume there are typically 1–2 pending updates. `publishCommands()` already has its own `MAX_PUBLISH_COMMANDS` batching built in. The suggested `setTimeout` batching adds complexity and latency for no practical benefit.

ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
deaflynx: Fixed — replaced the O(n) reverse lookup with a dedicated `subscriberCmdIds: Map<WsSubscriber, Set<number>>` reverse index maintained in `subscribe()`, `unsubscribe()`, and `reset()`. `syncCommandId()` now does `this.subscriberCmdIds.get(subscriber)` — O(1) lookup, no iteration over `subscribersMap`.

ui-ngx/src/app/core/ws/websocket.service.ts
deaflynx: Fixed — `processPendingUpdates()` now skips subscribers that have no entry in `subscriberCmdIds` (meaning they were not re-subscribed during reconnect).

ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
```diff
@@ -177,4 +194,13 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
     }
   }
 
+  private syncCommandId(subscriptionCommand: WebsocketCmd, subscriber: TelemetrySubscriber) {
+    if (!this.subscribersMap.has(subscriptionCommand.cmdId)) {
+      const cmdIds = this.subscriberCmdIds.get(subscriber);
+      if (cmdIds?.size) {
+        subscriptionCommand.cmdId = cmdIds.values().next().value;
```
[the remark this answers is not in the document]
deaflynx: Good catch — changed `!this.subscribersMap.has(subscriptionCommand.cmdId)` to `this.subscribersMap.get(subscriptionCommand.cmdId) !== subscriber` to check ownership instead of just existence. This handles the case where a stale cmdId collides with a new one assigned to a different subscriber after `lastCmdId` resets during reconnect.

Kept `.values().next().value` rather than the `.find()` suggestion since all cmdIds in `subscriberCmdIds` are guaranteed to map to that subscriber by construction.

ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
```diff
@@ -177,4 +194,13 @@ export class TelemetryWebsocketService extends WebsocketService<TelemetrySubscri
     }
   }
 
+  private syncCommandId(subscriptionCommand: WebsocketCmd, subscriber: TelemetrySubscriber) {
+    if (this.subscribersMap.get(subscriptionCommand.cmdId) !== subscriber) {
+      const cmdIds = this.subscriberCmdIds.get(subscriber);
+      if (cmdIds?.size) {
+        subscriptionCommand.cmdId = cmdIds.values().next().value;
```
vvlladd28, line 201: `cmdIds.values().next().value` picks whichever cmdId was inserted first in the Set, which is only a correct mapping when each subscriber owns exactly one tracked cmdId. The data structure (`Set<number>`) and the `subscribe()` loop above (which calls `cmdIds.add(cmdId)` per element of `subscriber.subscriptionCommands`) both suggest a subscriber can have more than one — but if that ever happens, all stale commands passed through `syncCommandId` will silently get rewritten to the same first cmdId, collapsing distinct subscriptions onto one and routing server responses to the wrong command.

In current usage `subscriber.subscriptionCommands` holds a single tracked command (entries like `[targetCommand]` / `[subsCommand]` / `[dataCommand]` in `entity-data-subscription`), so this works in practice. Worth either tightening the type to `Map<WsSubscriber, number>` to make the 1:1 invariant explicit, or — if multi-command subscribers are intentionally supported — matching the original command to its corresponding fresh cmdId rather than always taking the first.

vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Fixed resolve stale WebSocket cmdId after reconnect causing lost subscription updates](https://github.com/thingsboard/thingsboard/pull/15363). The fix correctly addresses the described stale-cmdId problem and the missed-update-during-reconnect window — `syncCommandId` properly checks ownership (not just existence) of the cmdId, `pendingUpdates` is preserved across reconnect but cleared on logout, and `processPendingUpdates` guards against subscribers that no longer have a fresh registration. Left 1 comment inline.

## Additional findings

No automated tests were added for either the cmdId resync path or the pending-updates queue. The reconnect flow is timing-sensitive and subtle — cmdIds can be remapped while user-driven `update()` calls are in flight, and the regression being fixed here was exactly that kind of issue. Unit coverage of `syncCommandId` (stale cmdId not in map, stale cmdId pointing at a different subscriber, subscriber missing from `subscriberCmdIds`) and `processPendingUpdates` (duplicate `update()` calls coalesced, unsubscribed-during-reconnect skipped) would help guard against regressions in this code path.

ui-ngx/src/app/core/ws/telemetry-websocket.service.ts
deaflynx: `syncCommandId` is only called from `update()`, which runs after reconnect completes. At that point `subscriberCmdIds` has already been cleared by `reset(false)` and repopulated by the reconnect `subscribe()` call. Since `entity-data-subscription.ts`'s `reconnect$` handler reduces `subscriptionCommands` to a single-element array before re-subscribing, each subscriber ends up with exactly one entry in the Set — so `values().next().value` is always the correct cmdId.

Tightening to `Map<WsSubscriber, number>` would break the normal (non-reconnect) `subscribe()` path where a subscriber legitimately holds multiple cmdIds. The 1:1 guarantee here comes from the reconnect flow, not the data structure itself.

Agreed it is worth keeping documented — the invariant is implicit and a future change to the reconnect handler could silently break it.

vvlladd28 (commented):
## Re-review summary

Re-reviewed [Fixed resolve stale WebSocket cmdId after reconnect causing lost subscription updates](https://github.com/thingsboard/thingsboard/pull/15363) — verified 1 finding from [previous review](https://github.com/thingsboard/thingsboard/pull/15363#pullrequestreview-4198084992).

| Status | Count |
|--------|-------|
| ✅ Resolved | 0 |
| 💬 Acknowledged | 1 |

All previous findings have been addressed. There are no new commits since the previous review; this pass re-read the full change with two independent review lenses and found no new issues that survived verification.

### Finding details

- 💬 **ui-ngx/src/app/core/ws/telemetry-websocket.service.ts:201** — `syncCommandId` takes the first cmdId out of the subscriber's `Set`, which is only a correct mapping under an implicit 1:1 invariant — *Developer: "`syncCommandId` is only called from `update()`, which runs after reconnect completes… Since `entity-data-subscription.ts`'s `reconnect$` handler reduces `subscriptionCommands` to a single-element array before re-subscribing, each subscriber ends up with exactly one entry in the Set. Tightening to `Map<WsSubscriber, number>` would break the normal (non-reconnect) `subscribe()` path where a subscriber legitimately holds multiple cmdIds."*

That reasoning checks out against the current code on every point I could test: `syncCommandId` has exactly one caller; `reset(false)` in `onClose` clears `subscriberCmdIds` before `isReconnect` is set, and `onOpen` repopulates it through `subscribe()`; and `entity-data-subscription.ts:449`/`:451` does collapse `subscriptionCommands` to a single element in the `reconnect$` handler. The multi-cmdId case on the normal path is real as well — `device-check-connectivity-dialog.component.ts:176` pushes a second `AttributesSubscriptionCmd` onto a subscriber built by `createEntityAttributesSubscription` — but those are `SubscriptionCmd`s, which `update()` filters out before `syncCommandId` ever sees them. So the 1:1 assumption holds where it needs to.

One loose end from that thread rather than a new objection: you agreed the invariant was worth documenting, and no comment has landed above `syncCommandId` yet. Since the guarantee comes from the reconnect flow in a different file, a one-line note there is the thing most likely to save the next reader.

---

PR #15382, opened 2026-04-08
Added enable/disable toggle for calculated fields and alarm rules

zzzeebra:
## Summary

Currently, if a user wants to pause a calculated field or alarm rule, they must delete the configuration, losing all settings. This PR adds the ability to temporarily disable and re-enable calculated fields and alarm rules without deleting them.

**How it works:**

When a CF is disabled via the API (or UI toggle), the backend treats it as a logical delete — the CF is removed from in-memory actor maps and stops processing telemetry. Links and scheduled reevaluations  are also removed. The CF remains in the database and is visible in the UI.
                                                                                                                                                                                                               
  When a CF is re-enabled, it goes through the same initialization path as a newly created CF — context is created, links are registered, and the CF recalculates using the latest available telemetry.

  On server restart, disabled CFs are filtered out during actor initialization, so the disabled state persists across restarts without requiring any additional flags or state tracking.                       
   
  The enabled field is a boolean column with DEFAULT true, so all existing CFs remain enabled after migration with no impact on current behavior.                                                              
                  
  Alarm rules use the same mechanism — AlarmRuleDefinition maps the enabled field to/from the underlying CalculatedField entity via toCalculatedField()/fromCalculatedField(), so all backend actor logic  
(disable = logical delete, re-enable = full init, restart filtering) works automatically. The UI toggle in the Alarm Rules table follows the same pattern as the Calculated Fields table.

## Automated tests
- **Controller tests** (`CalculatedFieldControllerTest`): default enabled state, disable/enable persistence and versioning, disabled CF visible in list, delete disabled CF
- **Integration tests** (`CalculatedFieldIntegrationTest`): disabled CF does not process telemetry, re-enabled CF resumes, multiple enable/disable cycles
- **MSA tests** (`CalculatedFieldTest`): disable in running system, enable/disable cycles with recalculation verification on re-enable
- **MSA tests** (`AlarmRulesTest`): disabled alarm rule does not create alarms, re-enable resumes alarm creation, enable/disable cycles

1. New CF is created with enabled state by default
<img width="799" height="760" alt="test1" src="https://github.com/user-attachments/assets/adcc0842-9e81-467b-a2f9-3ff36637d58a" />


2. Enabled CF processes telemetry — output key appears in Latest telemetry
<img width="799" height="760" alt="test2" src="https://github.com/user-attachments/assets/1d533188-2861-4d8d-8078-4b344943b83e" />


3. Toggle disables CF — icon changes to OFF state
<img width="799" height="760" alt="test3" src="https://github.com/user-attachments/assets/87a27fdf-eda7-414b-8900-92a9e9f48778" />


4. Disabled CF does not process new telemetry — output value unchanged
<img width="799" height="760" alt="test4" src="https://github.com/user-attachments/assets/5e8c40dd-03a9-4e24-8188-ff5a3dee2acc" />


5. Re-enabled CF resumes processing — output recalculated
<img width="799" height="760" alt="test5" src="https://github.com/user-attachments/assets/1acddbaa-43fc-4ccd-83b6-61e0fc84b293" />


6. Disabled CF remains disabled after server restart
<img width="799" height="760" alt="test6" src="https://github.com/user-attachments/assets/11680502-5f0c-492b-9dfe-a556b7e3b500" />

7. Toggle disables alarm rule — icon changes to OFF state, alarm is not triggered
<img width="793" height="753" alt="test7" src="https://github.com/user-attachments/assets/11e20944-a2b4-4274-aa57-f6de83b361e3" />


8. Re-enabled alarm rule resumes creating alarms
9. Disabled alarm rule remains disabled after server restart


ViacheslavKlimov (commented):
## Review summary

Reviewed 12 changed files in [Add enable/disable support for calculated fields](https://github.com/thingsboard/thingsboard/pull/15382). Left 2 comment(s) inline.

The overall design is clean — disabling reuses the existing `onCfDeleted` path and re-enabling goes through `onCfCreated`, which avoids duplicating lifecycle logic. Schema migration, entity mapping, copy constructor, and test coverage all look solid.

---

ui-ngx/src/app/modules/home/components/calculated-fields/calculated-fields-table-config.ts
```diff
@@ -378,6 +387,16 @@ export class CalculatedFieldsTableConfig extends EntityTableConfig<CalculatedFie
...
   }
 
+  private toggleEnabled($event: Event, entity: CalculatedFieldsTableEntity): void {
+    $event?.stopPropagation();
+    this.calculatedFieldsService.getCalculatedFieldById(entity.id.id).pipe(
+      switchMap(cf => this.calculatedFieldsService.saveCalculatedField({...cf, enabled: !cf.enabled}, {ignoreLoading: true}))
+    ).subscribe((saved) => {
+        entity.enabled = saved.enabled;
+        this.getTable().detectChanges();
+      });
```
ViacheslavKlimov, line 401: This subscription is missing `catchError` and `takeUntilDestroyed(this.destroyRef)` — the `onDebugConfigChanged` method right below (line 400) uses both. Without `catchError`, a failed save (network error, version conflict) will silently swallow the error and leave the UI showing a stale toggle state. Without `takeUntilDestroyed`, the subscription leaks if the component is destroyed mid-flight.

common/data/src/main/java/org/thingsboard/server/common/data/cf/CalculatedField.java
```diff
@@ -67,6 +67,8 @@ public static boolean isSupportedRefEntity(EntityId entity) {
 
     private TenantId tenantId;
     private EntityId entityId;
+    @Schema(description = "Enable/disable the calculated field.", example = "true")
+    private boolean enabled = true;
```
ViacheslavKlimov, line 71: Nit: `toString()` (line 143) doesn't include `enabled`. Since this field directly controls whether a CF processes telemetry, having it in log output would make debugging enable/disable flows easier.

ui-ngx/src/app/modules/home/components/calculated-fields/calculated-fields-table-config.ts
zzzeebra: Fixed. Added `catchError` and `takeUntilDestroyed` to match the pattern in `onDebugConfigChanged`. Also wrapped the subscribe body with a null check since `catchError` now returns `of(null)`.

common/data/src/main/java/org/thingsboard/server/common/data/cf/CalculatedField.java
zzzeebra: Fixed. Added `enabled` to `toString()`.

ViacheslavKlimov, line 71: Will the import of calculated field JSONs without `enabled` property work fine? Needs some testing

application/src/main/java/org/thingsboard/server/actors/calculatedField/CalculatedFieldManagerMessageProcessor.java
```diff
@@ -805,6 +808,10 @@ public void initCalculatedFields() {
         PageDataIterable<CalculatedField> cfs = new PageDataIterable<>(pageLink -> cfDaoService.findCalculatedFieldsByTenantId(tenantId, pageLink), cfSettings.getInitTenantFetchPackSize());
         cfs.forEach(cf -> {
             log.trace("Processing calculated field record: {}", cf);
+            if (!cf.isEnabled()) {
+                log.debug("[{}] Skipping disabled calculated field [{}]", tenantId, cf.getId());
```
ViacheslavKlimov: Wasteful to fetch them in the first place. Filter in SQL

ViacheslavKlimov:
Where is the "enabled" gate for telemetry processing? And why no changes in the DefaultCalculatedFieldCache and DefaultCalculatedFieldQueueService? Why do we at all send any updates for disabled calculated fields/alarm rules? NOTHING must be processed for the disabled calculated fields/alarm rules, any processing must be discarded in the earliest point possible.

vvlladd28 (approved):

application/src/main/java/org/thingsboard/server/actors/calculatedField/CalculatedFieldManagerMessageProcessor.java
zzzeebra: Done. `initCalculatedFields` now fetches via `findCalculatedFieldsByTenantIdAndEnabled(tenantId, true, pageLink)` — disabled fields are filtered in SQL, the in-memory `isEnabled()` skip is removed.

common/data/src/main/java/org/thingsboard/server/common/data/cf/CalculatedField.java
zzzeebra: Added `CalculatedFieldEnabledDeserializationTest`. JSON without the `enabled` property deserializes to `enabled = true` for both `CalculatedField` and `AlarmRuleDefinition` — Jackson keeps the field initializer when the property is absent — and `enabled: false` is preserved. So importing pre-feature exports keeps them enabled.

zzzeebra:
Added the gate in `DefaultCalculatedFieldCache`: `addCalculatedField` now skips disabled fields, and the native startup query (`CF_QUERY`/`CF_COUNT_QUERY`) loads only `enabled = true`. So disabled fields never enter the cache.

`DefaultCalculatedFieldQueueService` is intentionally left unchanged — it reads exclusively from the cache (`hasCalculatedFields`, `getCalculatedFieldCtxsByEntityId`/`ByType`, links). With disabled fields absent from the cache, `checkEntityForCalculatedFields` returns `false` and nothing is pushed to the queue. Telemetry for disabled fields is therefore discarded at the earliest point, before the queue.

PR #15403, opened 2026-04-09
Made user activation link expiration configurable

zzzeebra:
## Summary

Currently, the User Activation Link TTL is hardcoded to a maximum of 24 hours. This is not enough for some deployment scenarios where users may not activate their accounts immediately.

This PR makes the maximum allowed TTL configurable via `thingsboard.yml` with a default of 720 hours (30 days).

## How it works

**Server configuration (`thingsboard.yml`):**
A new parameter `security.user_activation_link_max_ttl` defines the maximum TTL value that a System Administrator can set in Security Settings UI. Default is 720 hours (30 days). Can be overridden via environment variable `SECURITY_USER_ACTIVATION_LINK_MAX_TTL`.

**Backend validation:**
The hardcoded `@Max(24)` annotation is removed from `SecuritySettings.userActivationTokenTtl`. Instead, programmatic validation in `DefaultSecuritySettingsService` enforces the configurable limit on save.

**Exposing the limit to the UI:**
The configured maximum is returned as a read-only `maxActivationLinkTtl` field of `SystemParams` (`GET /api/system/params`), which the UI already loads on login. It is set for System Administrators only — the Security Settings page is available to them only. `SecuritySettings` itself is not touched, so no config-derived value ends up in the persisted, user-editable settings.

**Frontend:**
The Security Settings page reads `maxActivationLinkTtl` from the auth state and applies it as the `max` validator of the User Activation Link TTL field. Error messages use `{{max}}` interpolation so the displayed limit always matches the server configuration.

**What is NOT changed:**
- Password Reset Link TTL — remains hardcoded at max 24 hours
- Default TTL value (24 hours) — unchanged, this is what new installations get in Security Settings

## Automated tests

- `testActivationLinkMaxTtlExposedInSystemParams` — verifies `/api/system/params` returns 720 (default) for a System Administrator
- `testActivationLinkTtlAtMaxIsAllowed` — boundary: TTL == max → 200 OK
- `testActivationLinkTtlExceedsMax` — TTL > max → 400 Bad Request

1. Set User Activation Link TTL to 720 hours → saves successfully
test11

2. Set User Activation Link TTL to 721 hours → validation error is shown with correct max value
Screenshot 2026-04-09 at 15 58 26

3. Create a new user with TTL set to 720 hours → activation link dialog shows correct remaining time (~30 days)
Screenshot 2026-04-09 at 16 02 19


dao/src/main/java/org/thingsboard/server/dao/settings/DefaultSecuritySettingsService.java
```diff
@@ -34,6 +36,9 @@ public class DefaultSecuritySettingsService implements SecuritySettingsService {
 
     private final AdminSettingsService adminSettingsService;
 
+    @Value("${security.user_activation_link_max_ttl:720}")
+    private int maxActivationLinkTtl;
```
ViacheslavKlimov, line 40: There's no validation on `maxActivationLinkTtl` itself. If an operator misconfigures `SECURITY_USER_ACTIVATION_LINK_MAX_TTL=0` (or a negative value), the server starts fine but every `saveSecuritySettings` call then fails with `"User activation link TTL must not exceed 0 hours"` — even for the well-known default of 24 that the UI initializes with. Worth failing fast at startup (e.g. a `@PostConstruct` check that throws if `maxActivationLinkTtl < 1`) so a bad env var is surfaced immediately rather than at the next save attempt.

dao/src/main/java/org/thingsboard/server/dao/settings/DefaultSecuritySettingsService.java
```diff
@@ -59,10 +64,18 @@ public SecuritySettings getSecuritySettings() {
...
+    @Override
+    public int getMaxActivationLinkTtl() {
+        return maxActivationLinkTtl;
+    }
+
     @CacheEvict(cacheNames = SECURITY_SETTINGS_CACHE, key = "'securitySettings'")
     @Override
     public SecuritySettings saveSecuritySettings(SecuritySettings securitySettings) {
         ConstraintValidator.validateFields(securitySettings);
+        if (securitySettings.getUserActivationTokenTtl() > maxActivationLinkTtl) {
```
ViacheslavKlimov, line 71: This only enforces the cap on the *write* path. If an admin lowers `SECURITY_USER_ACTIVATION_LINK_MAX_TTL` below the value already persisted in `securitySettings.userActivationTokenTtl`, `UserServiceImpl.generateUserActivationToken` will keep using the old (now over-cap) value until a sysadmin opens the settings page and re-saves. The frontend will surface the violation visually, but the backend silently honors the over-cap value. If "config wins" is the intended contract, consider clamping at read time (in `getSecuritySettings()` or in `generateUserActivationToken`) so a config change takes effect immediately on restart.

ui-ngx/src/app/modules/home/pages/admin/security-settings.component.html
```diff
@@ -55,12 +55,10 @@
...
-                   max="24"/>
-            <mat-error *ngIf="securitySettingsFormGroup.get('userActivationTokenTtl').hasError('min')">
-              {{ 'admin.user-activation-token-ttl-range' | translate }}
-            </mat-error>
-            <mat-error *ngIf="securitySettingsFormGroup.get('userActivationTokenTtl').hasError('max')">
-              {{ 'admin.user-activation-token-ttl-range' | translate }}
+                   [max]="maxActivationLinkTtl"/>
+            <mat-error *ngIf="securitySettingsFormGroup.get('userActivationTokenTtl').hasError('min') ||
+                              securitySettingsFormGroup.get('userActivationTokenTtl').hasError('max')">
+              {{ 'admin.user-activation-token-ttl-range' | translate: {max: maxActivationLinkTtl} }}
```
ViacheslavKlimov: If `getActivationLinkMaxTtl()` fails (the `catchError(() => of(null))` path in the .ts), `maxActivationLinkTtl` stays `undefined`. The `Validators.max(...)` is then *not* added, so the user can only ever trigger the `min` branch — but this same `mat-error` is shown for both, and the message still interpolates `{{max}}`, which would render as `"User activation link TTL must be in range from 1 to  hours"`. Consider falling back to a sensible default in the template (e.g. `maxActivationLinkTtl ?? 24`) or splitting the `min` and `max` errors back out so the `min` message doesn't depend on an unloaded value.

ui-ngx/src/app/modules/home/pages/admin/security-settings.component.ts
```diff
@@ -66,9 +67,17 @@ export class SecuritySettingsComponent extends PageComponent implements HasConfi
...
+    this.adminService.getActivationLinkMaxTtl().pipe(
+      catchError(() => of(null)),
+      tap(maxTtl => {
+        if (maxTtl != null) {
+          this.maxActivationLinkTtl = maxTtl;
+          this.securitySettingsFormGroup.get('userActivationTokenTtl')
+            .addValidators(Validators.max(maxTtl));
+        }
+      }),
+      switchMap(() => this.adminService.getSecuritySettings())
```
ViacheslavKlimov: Before, `getSecuritySettings()` had its own subscription, and a failure was simply unhandled. Now the chain is `getActivationLinkMaxTtl → switchMap(getSecuritySettings)`, so an error from the inner call also short-circuits the whole chain — same end state, but it's a good moment to add a `catchError` on the inner observable too (or attach an error handler to `subscribe`) so a failed `getSecuritySettings()` doesn't leave the form silently stuck on the placeholder defaults of `[24, ...]` with no feedback to the user.

application/src/main/java/org/thingsboard/server/controller/AdminController.java
```diff
@@ -180,6 +180,15 @@ public SecuritySettings saveSecuritySettings(
         return securitySettings;
     }
 
+    @ApiOperation(value = "Get maximum allowed TTL for user activation link (getActivationLinkMaxTtl)",
+            notes = "Returns the maximum allowed TTL in hours for user activation link, configured on the server." + SYSTEM_AUTHORITY_PARAGRAPH)
+    @PreAuthorize("hasAuthority('SYS_ADMIN')")
+    @GetMapping(value = "/securitySettings/activationLinkMaxTtl")
+    public int getActivationLinkMaxTtl() throws ThingsboardException {
+        accessControlService.checkPermission(getCurrentUser(), Resource.ADMIN_SETTINGS, Operation.READ);
+        return securitySettingsService.getMaxActivationLinkTtl();
```
ViacheslavKlimov: Style nit: every other endpoint in this controller returns either a domain object (`SecuritySettings`, `JwtSettings`, `AdminSettings`) or `void`. Returning a raw primitive `int` is a bit of an outlier and makes future extension awkward — e.g. the moment you also want to expose `passwordResetTokenTtl`'s max, or surface other limits, you'd have to either add another single-int endpoint or break the contract. A small response object (e.g. `SecurityLimits { maxActivationLinkTtl }` or even `Map.of("maxTtl", value)`) would be more consistent and forward-compatible. Up to you — the simpler shape works fine for the immediate need.

ViacheslavKlimov (commented):
## Review summary

Reviewed 24 changed files in [Made user activation link expiration configurable](https://github.com/thingsboard/thingsboard/pull/15403). Left 5 comment(s) inline.

Overall the change is straightforward and well-scoped: hardcoded `@Max(24)` is replaced by a configurable server-side cap with programmatic validation, a thin SYS_ADMIN endpoint exposes the cap to the UI, and locale strings are updated for the parameterized message. Tests cover the happy path, the boundary, the over-limit case, and the AuthZ failures.

Main things worth a second look:
- No sanity check on the configured `maxActivationLinkTtl` value itself — a misconfigured `0`/negative value silently wedges the save flow.
- When `getActivationLinkMaxTtl()` fails on the client, the error message still tries to interpolate `{{max}}`, which renders as an empty string.
- A drop in the configured cap is not retroactively enforced against the value already persisted in `SecuritySettings`, so `generateUserActivationToken` keeps using the old (now over-cap) TTL until a sysadmin resaves.

---

dao/src/main/java/org/thingsboard/server/dao/settings/DefaultSecuritySettingsService.java
zzzeebra: Skipping — the neighboring numeric params in `thingsboard.yml` don't do startup sanity checks either, so adding one here would break the pattern. With `=0` the operator gets a clear error message at first save.

dao/src/main/java/org/thingsboard/server/dao/settings/DefaultSecuritySettingsService.java
zzzeebra: Intentional. The cap is a write-time validation boundary (consistent with how `@Max` works elsewhere in the codebase), not a runtime hard limit. Clamping the persisted value at read time would be a silent behavior change. If a sysadmin lowers the cap, the over-cap value is shown red on settings open and they fix it.

ui-ngx/src/app/modules/home/pages/admin/security-settings.component.html
zzzeebra: Fixed — fallback to `24` in the template when the value isn't loaded.

ui-ngx/src/app/modules/home/pages/admin/security-settings.component.ts
zzzeebra: Pre-existing behavior — `getSecuritySettings()` had no explicit error handler before either; failures are surfaced by the global HTTP interceptor. Not a regression introduced by this PR.

application/src/main/java/org/thingsboard/server/controller/AdminController.java
zzzeebra: Skipping — YAGNI. Will refactor to a DTO when another limit needs to be exposed.

application/src/main/java/org/thingsboard/server/controller/AdminController.java
```diff
@@ -180,6 +180,15 @@ public SecuritySettings saveSecuritySettings(
         return securitySettings;
     }
 
+    @ApiOperation(value = "Get maximum allowed TTL for user activation link (getActivationLinkMaxTtl)",
+            notes = "Returns the maximum allowed TTL in hours for user activation link, configured on the server." + SYSTEM_AUTHORITY_PARAGRAPH)
+    @PreAuthorize("hasAuthority('SYS_ADMIN')")
+    @GetMapping(value = "/securitySettings/activationLinkMaxTtl")
+    public int getActivationLinkMaxTtl() throws ThingsboardException {
```
ViacheslavKlimov: Why is there a separate endpoint for this? Discussed with @vvlladd28 - UI already fetches the security settings

dao/src/main/java/org/thingsboard/server/dao/settings/SecuritySettingsService.java
```diff
@@ -23,4 +23,6 @@ public interface SecuritySettingsService {
 
     SecuritySettings saveSecuritySettings(SecuritySettings securitySettings);
 
+    int getMaxActivationLinkTtl();
```
ViacheslavKlimov: A dedicated getter in too much. Let the caller do getSecuritySettings

ui-ngx/src/app/modules/home/pages/admin/security-settings.component.ts
```diff
@@ -66,9 +67,17 @@ export class SecuritySettingsComponent extends PageComponent implements HasConfi
...
+    ).subscribe(securitySettings => this.processSecuritySettings(securitySettings));
```
vvlladd28: There is no need to make an additional `getActivationLinkMaxTtl()` request here (and to introduce the separate `/securitySettings/activationLinkMaxTtl` endpoint at all). Since `maxActivationLinkTtl` is a server-side config value that is only used together with security settings, it should be returned as a read-only field of the `SecuritySettings` object from the existing `getSecuritySettings()` API (and ignored on save).

Then the constructor can stay as it was, with a single request:

```ts
this.adminService.getSecuritySettings().subscribe(
  securitySettings => this.processSecuritySettings(securitySettings)
);
```

and the max TTL should be processed in `processSecuritySettings()` — it is also called after `save()`, so the validator will always stay in sync with the server value:

```ts
private processSecuritySettings(securitySettings: SecuritySettings) {
  this.securitySettings = securitySettings;
  if (isDefinedAndNotNull(securitySettings.maxActivationLinkTtl)) {
    this.maxActivationLinkTtl = securitySettings.maxActivationLinkTtl;
    const userActivationTokenTtl = this.securitySettingsFormGroup.get('userActivationTokenTtl');
    userActivationTokenTtl.setValidators([Validators.required, Validators.min(1), Validators.max(this.maxActivationLinkTtl)]);
    userActivationTokenTtl.updateValueAndValidity({emitEvent: false});
  }
  this.securitySettingsFormGroup.reset(this.securitySettings);
}
```

This removes the extra HTTP round-trip and the `catchError`/`switchMap` chaining, and the form is not left without the `max` validator if the extra request fails.

application/src/main/java/org/thingsboard/server/controller/AdminController.java
zzzeebra: Removed the endpoint. `maxActivationLinkTtl` is now a read-only field of `SecuritySettings`, populated in `getSecuritySettings()` and ignored on save.

dao/src/main/java/org/thingsboard/server/dao/settings/SecuritySettingsService.java
zzzeebra: Removed. The value is set on the `SecuritySettings` object in `getSecuritySettings()`, so callers no longer need a dedicated getter.

ui-ngx/src/app/modules/home/pages/admin/security-settings.component.ts
zzzeebra: Done. `maxActivationLinkTtl` is now a read-only field of `SecuritySettings` (ignored on save); single `getSecuritySettings()` request, validator applied in `processSecuritySettings()`. Removed the extra endpoint and the `catchError`/`switchMap` chain.

common/data/src/main/java/org/thingsboard/server/common/data/security/model/SecuritySettings.java
```diff
@@ -43,10 +44,14 @@ public class SecuritySettings implements Serializable {
     @Schema(description = "Mobile secret key length")
     private Integer mobileSecretKeyLength;
 
-    @NotNull @Min(1) @Max(24)
-    @Schema(description = "TTL in hours for user activation link", minimum = "1", maximum = "24", requiredMode = Schema.RequiredMode.REQUIRED)
+    @NotNull @Min(1)
+    @Schema(description = "TTL in hours for user activation link", minimum = "1", requiredMode = Schema.RequiredMode.REQUIRED)
     private Integer userActivationTokenTtl;
 
+    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
```
ViacheslavKlimov: Marking `maxActivationLinkTtl` as `READ_ONLY` means it serializes on output but is dropped on input — so any typed Java consumer that GETs the settings into `SecuritySettings` sees `null` for it, including the shipped `RestClient.getSecuritySettings()` (which deserializes into `SecuritySettings.class`). The new `testActivationLinkMaxTtlExposedInSecuritySettings` test even has to work around this by reading the raw `JsonNode` rather than the typed object. So a value the server clearly sends over HTTP is invisible to the canonical SDK client. Would a plain (non-read-only) field that the service simply overwrites on read and ignores on save give SDK consumers a value they can actually read? `testActivationLinkMaxTtlIsReadOnlyOnSave` already proves the server keeps its configured cap regardless of what the client supplies, so the read-only-ness isn't buying much on the input side.

dao/src/main/java/org/thingsboard/server/dao/settings/DefaultSecuritySettingsService.java
```diff
@@ -72,7 +81,9 @@ public SecuritySettings saveSecuritySettings(SecuritySettings securitySettings)
         adminSettings.setJsonValue(JacksonUtil.valueToTree(securitySettings));
         AdminSettings savedAdminSettings = adminSettingsService.saveAdminSettings(TenantId.SYS_TENANT_ID, adminSettings);
         try {
-            return JacksonUtil.convertValue(savedAdminSettings.getJsonValue(), SecuritySettings.class);
+            SecuritySettings savedSecuritySettings = JacksonUtil.convertValue(savedAdminSettings.getJsonValue(), SecuritySettings.class);
+            savedSecuritySettings.setMaxActivationLinkTtl(maxActivationLinkTtl);
```
ViacheslavKlimov: `maxActivationLinkTtl` now gets stamped onto the DTO in two separate spots — here and in `getSecuritySettings` (line 64) — so any future read path has to remember to do the same, which is easy to miss. Also worth noting: because `READ_ONLY` only suppresses deserialization, `valueToTree(securitySettings)` on line 81 still writes a `maxActivationLinkTtl` entry into the persisted admin-settings JSON (as `null`, since the incoming value is dropped) that then gets overwritten on every read — a config-derived value leaking into the persisted, user-editable model. Might be cleaner to keep the max out of the persisted model entirely and inject it in a single place on the way out.

ui-ngx/src/app/modules/home/pages/admin/security-settings.component.html
```diff
@@ -55,12 +55,10 @@
...
-                   max="24"/>
-            <mat-error *ngIf="securitySettingsFormGroup.get('userActivationTokenTtl').hasError('min')">
-              {{ 'admin.user-activation-token-ttl-range' | translate }}
-            </mat-error>
-            <mat-error *ngIf="securitySettingsFormGroup.get('userActivationTokenTtl').hasError('max')">
-              {{ 'admin.user-activation-token-ttl-range' | translate }}
+                   [max]="maxActivationLinkTtl"/>
+            <mat-error *ngIf="securitySettingsFormGroup.get('userActivationTokenTtl').hasError('min') ||
+                              securitySettingsFormGroup.get('userActivationTokenTtl').hasError('max')">
+              {{ 'admin.user-activation-token-ttl-range' | translate: {max: maxActivationLinkTtl ?? 24} }}
```
ViacheslavKlimov: The `?? 24` fallback is a stale magic number — it's the old hardcoded cap and no longer matches the real server default (720). It's also effectively unreachable for the `max` case: the `Validators.max` validator is only attached inside the `isDefinedAndNotNull(maxActivationLinkTtl)` guard in the component, so when `maxActivationLinkTtl` is undefined there's no `max` error to display anyway. If you want a defensive default here, could it reference a shared named constant (also used for the initial `[24, ...]` form value in the `.ts`) so the message can't contradict the actual limit?

application/src/test/java/org/thingsboard/server/controller/AuthControllerTest.java
```diff
@@ -294,6 +294,42 @@ public void testActivationLinkTtl() throws Exception {
...
+        doPost("/api/admin/securitySettings", securitySettings).andExpect(status().isBadRequest());
+    }
+
+    @Test
+    public void testActivationLinkMaxTtlIsReadOnlyOnSave() throws Exception {
+        loginSysAdmin();
+        SecuritySettings securitySettings = doGet("/api/admin/securitySettings", SecuritySettings.class);
+        // a client-supplied max is ignored: the server keeps its configured cap, so a TTL within that cap is accepted
+        securitySettings.setMaxActivationLinkTtl(1);
+        securitySettings.setUserActivationTokenTtl(100);
```
ViacheslavKlimov: These tests persist non-default `userActivationTokenTtl` values (720 and 100) into the shared, DB-backed security settings, but `tearDown` (line 59) only restores the password-policy fields — it never resets the activation-token TTL. Since the settings row is a singleton shared across the class, that leaves mutated global state behind. The pre-existing `testActivationLinkTtl` has the same gap, so this isn't new, but it might be worth resetting `userActivationTokenTtl` in `tearDown` for consistency with how the password policy is already restored.

ViacheslavKlimov (commented):
## Review summary

Reviewed 22 changed files in [Made user activation link expiration configurable](https://github.com/thingsboard/thingsboard/pull/15403). Left 4 comments inline.

The change is small, well-scoped, and well-tested — I found no correctness or safety issues. The unboxing in `saveSecuritySettings` is guarded by `validateFields`, raising the cap to 720h stays well within the `long` expiry math, the field is backward compatible, and all 15 locale files that carry the range key were updated consistently. The inline comments are about API usability of the new read-only field, some duplicated/leaky backend wiring, and minor frontend/test cleanup.

---

common/data/src/main/java/org/thingsboard/server/common/data/security/model/SecuritySettings.java
zzzeebra: Removed the `@JsonProperty(READ_ONLY)` — `maxActivationLinkTtl` is now a plain field that typed clients (incl. `RestClient`) can read. The service overwrites it on read and drops it before persisting, so save still ignores client input. Tests now read the typed `SecuritySettings` object instead of the raw `JsonNode`.

dao/src/main/java/org/thingsboard/server/dao/settings/DefaultSecuritySettingsService.java
zzzeebra: Both stamp sites now go through a single `applyMaxActivationLinkTtl(...)` helper. And `maxActivationLinkTtl` is set to `null` before `valueToTree`, so the config-derived value is no longer written into the persisted settings JSON.

ui-ngx/src/app/modules/home/pages/admin/security-settings.component.html
zzzeebra: Removed the `?? 24` fallback — the message binds directly to `maxActivationLinkTtl`, which is always populated from `getSecuritySettings()`. The `max` validator is only attached once the value is loaded, so the interpolation can't render empty. Kept the single combined `-range` key to stay consistent with the sibling fields in this form.

application/src/test/java/org/thingsboard/server/controller/AuthControllerTest.java
zzzeebra: `tearDown` now resets `userActivationTokenTtl` to the default (24) alongside the password policy, so the shared settings row isn't left mutated.

dao/src/main/java/org/thingsboard/server/dao/settings/DefaultSecuritySettingsService.java
```diff
@@ -56,26 +61,37 @@ public SecuritySettings getSecuritySettings() {
...
+        if (securitySettings.getUserActivationTokenTtl() > maxActivationLinkTtl) {
+            throw new DataValidationException("User activation link TTL must not exceed " + maxActivationLinkTtl + " hours");
+        }
         AdminSettings adminSettings = adminSettingsService.findAdminSettingsByKey(TenantId.SYS_TENANT_ID, "securitySettings");
         if (adminSettings == null) {
             adminSettings = new AdminSettings();
             adminSettings.setTenantId(TenantId.SYS_TENANT_ID);
             adminSettings.setKey("securitySettings");
         }
+        securitySettings.setMaxActivationLinkTtl(null); // config-derived, must not be persisted
```
ViacheslavKlimov: If it's read-only for UI - this is not the place. Add to SystemParams instead

zzzeebra: Moved to `SystemParams` — `maxActivationLinkTtl` is no longer a field of `SecuritySettings`, so nothing config-derived is stamped on the DTO or stripped before persisting. `SystemInfoController` sets it for SYS_ADMIN only; the UI reads it from the auth state and applies the `max` validator statically.

PR #15411, opened 2026-04-13
Fixed server crash on upload of large images

zzzeebra:
## Summary

- Large resolution images (e.g. 17869×12802, 25MB file) cause OOM during upload because the decoded `BufferedImage` occupies ~915MB of heap
- Preview generation is now skipped when decoded image size exceeds tenant's `maxResourceSize`, saving a 1×1 transparent GIF placeholder instead
- Original image is still saved to the database as-is

## How it works

Before generating a preview thumbnail, the server reads image dimensions from metadata via `ImageReader` (without decoding the full image into memory). If the decoded size (`width × height × 4 bytes`) exceeds the tenant's `Maximum resource file size` limit, preview generation is skipped and a placeholder is saved instead.

SVG images are not affected — they use a separate processing path. For system tenants (`maxResourceSize=0`), preview is always generated as before.

**Changed files:**
- `ImageUtils.java` — added `getImageDimensions()` and `PLACEHOLDER_PREVIEW` constant
- `ResourceDataValidator.java` — added `getMaxResourceSize()` public method
- `BaseImageService.java` — added decoded size check before calling `ImageUtils.processImage()`

## Automated tests

Added `testSkipPreviewForLargeImage` in `ImageControllerTest` — sets `maxResourceSize` to 50KB, uploads a PNG whose decoded size (200×160×4 = 128KB) exceeds the limit, and verifies the preview is a placeholder GIF with dimensions 1×1.

Existing tests cover the normal path (`maxResourceSize=0`, unlimited) and remain unchanged.

Adding 5Mb limit 
Screenshot 2026-04-13 at 09 18 20

Uploading the image with size 2.2Mb and decoded size: 4096 × 1844 × 4 = 30,212,096b (~30Mb). The image is uploaded but without preview.
Screenshot 2026-04-13 at 09 32 49

### DIFF
--- a/application/src/test/java/org/thingsboard/server/controller/ImageControllerTest.java
+++ b/application/src/test/java/org/thingsboard/server/controller/ImageControllerTest.java
@@ -333,25 +333,26 @@ public class ImageControllerTest extends AbstractControllerTest {
         });
         loginTenantAdmin();
 
-        String filename = "large_decoded_image.png";
-        TbResourceInfo imageInfo = uploadImage(HttpMethod.POST, "/api/image", filename, "image/png", PNG_IMAGE);
+        try {
+            String filename = "large_decoded_image.png";
+            TbResourceInfo imageInfo = uploadImage(HttpMethod.POST, "/api/image", filename, "image/png", PNG_IMAGE);
 
-        ImageDescriptor imageDescriptor = imageInfo.getDescriptor(ImageDescriptor.class);
-        assertThat(imageDescriptor.getWidth()).isEqualTo(200);
-        assertThat(imageDescriptor.getHeight()).isEqualTo(160);
+            ImageDescriptor imageDescriptor = imageInfo.getDescriptor(ImageDescriptor.class);
+            assertThat(imageDescriptor.getWidth()).isEqualTo(200);
+            assertThat(imageDescriptor.getHeight()).isEqualTo(160);
 
-        ImageDescriptor previewDescriptor = imageDescriptor.getPreviewDescriptor();
-        assertThat(previewDescriptor.getMediaType()).isEqualTo("image/gif");
-        assertThat(previewDescriptor.getWidth()).isEqualTo(1);
-        assertThat(previewDescriptor.getHeight()).isEqualTo(1);
+            ImageDescriptor previewDescriptor = imageDescriptor.getPreviewDescriptor();
+            assertThat(previewDescriptor.getMediaType()).isEqualTo("image/gif");
+            assertThat(previewDescriptor.getWidth()).isEqualTo(1);
+            assertThat(previewDescriptor.getHeight()).isEqualTo(1);
 
-        assertThat(downloadImagePreview("tenant", filename)).containsExactly(ImageUtils.PLACEHOLDER_PREVIEW);
-
-        // Reset maxResourceSize
-        loginSysAdmin();
-        updateDefaultTenantProfileConfig(tenantProfileConfig -> {
-            tenantProfileConfig.setMaxResourceSize(0);
-        });
+            assertThat(downloadImagePreview("tenant", filename)).containsExactly(ImageUtils.PLACEHOLDER_PREVIEW);
+        } finally {
+            loginSysAdmin();
+            updateDefaultTenantProfileConfig(tenantProfileConfig -> {
+                tenantProfileConfig.setMaxResourceSize(0);
+            });
+        }
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


dao/src/main/java/org/thingsboard/server/dao/util/ImageUtils.java
```diff
@@ -75,6 +81,27 @@ public static String fileExtensionToMediaType(String extension) {
         return new MimeType("image", subtype).toString();
     }
 
+    public static int[] getImageDimensions(byte[] data, String mediaType) {
+        if (mediaTypeToFileExtension(mediaType).equals("svg")) {
+            return null;
+        }
+        try (ImageInputStream iis = ImageIO.createImageInputStream(new ByteArrayInputStream(data))) {
+            var readers = ImageIO.getImageReaders(iis);
```
ViacheslavKlimov, line 92: `ImageIO.createImageInputStream(...)` can return `null` (e.g. when no suitable `ImageInputStreamSpi` is registered for the input). If that happens, `ImageIO.getImageReaders(null)` throws `IllegalArgumentException`, which escapes the `catch (IOException)` block and propagates up uncaught.

A null-check on `iis` before calling `getImageReaders` would make this defensive:
```java
if (iis == null) {
    return null;
}
```
In practice this is unlikely with a `ByteArrayInputStream` input, but it would match the defensiveness of the rest of the method.

application/src/test/java/org/thingsboard/server/controller/ImageControllerTest.java
```diff
@@ -322,6 +323,37 @@ public void testGetImageUploadSpecs() throws Exception {
...
+        assertThat(previewDescriptor.getMediaType()).isEqualTo("image/gif");
+        assertThat(previewDescriptor.getWidth()).isEqualTo(1);
+        assertThat(previewDescriptor.getHeight()).isEqualTo(1);
+
+        assertThat(downloadImagePreview("tenant", filename)).containsExactly(ImageUtils.PLACEHOLDER_PREVIEW);
+
+        // Reset maxResourceSize
+        loginSysAdmin();
+        updateDefaultTenantProfileConfig(tenantProfileConfig -> {
+            tenantProfileConfig.setMaxResourceSize(0);
```
ViacheslavKlimov: If any assertion above this line fails, the `maxResourceSize` reset block never runs, leaving a non-zero limit for subsequent tests. Consider wrapping the assertion + cleanup in a try-finally, or resetting in an `@AfterEach` method, to avoid flaky downstream tests.

dao/src/main/java/org/thingsboard/server/dao/resource/BaseImageService.java
```diff
@@ -138,7 +138,8 @@ public TbResourceInfo saveImage(TbResource image) {
         resourceValidator.validate(image, TbResourceInfo::getTenantId);
 
         ImageDescriptor descriptor = image.getDescriptor(ImageDescriptor.class);
-        Pair<ImageDescriptor, byte[]> result = processImage(image.getData(), descriptor);
+        long maxDecodedImageSize = resourceValidator.getMaxResourceSize(image.getTenantId());
```
ViacheslavKlimov: The variable name `maxDecodedImageSize` suggests this is a limit on the *decoded* size, but `getMaxResourceSize` returns the tenant's max *resource file* size — the same value that caps the uploaded file size. Reusing it as a decoded-size threshold is a pragmatic choice (the PR description explains the rationale well), but the variable name here creates a false impression that there's a separate "max decoded size" configuration.

Something like `maxResourceSize` (matching the getter name) would be less misleading, and the comparison at line 161 (`decodedSize > maxResourceSize`) would then read as "skip preview when the decoded footprint exceeds the resource size budget" — which is what's actually happening.

ViacheslavKlimov (commented):
## Review summary

Reviewed 4 changed files in [Skip preview generation for images with large decoded size to prevent OOM](https://github.com/thingsboard/thingsboard/pull/15411). Left 3 comment(s) inline.

Clean, well-scoped change. The approach of reading dimensions via `ImageReader` metadata (without decoding) and comparing the projected decoded size against the tenant limit is solid. A few things to tighten up below.

---

dao/src/main/java/org/thingsboard/server/dao/util/ImageUtils.java
zzzeebra: Fixed. Added null-check for ImageInputStream before calling getImageReaders().

application/src/test/java/org/thingsboard/server/controller/ImageControllerTest.java
zzzeebra: Fixed. Wrapped assertions and cleanup in try-finally so maxResourceSize is always reset.

dao/src/main/java/org/thingsboard/server/dao/resource/BaseImageService.java
zzzeebra: Fixed. Renamed maxDecodedImageSize to maxResourceSize to match the getter name.

PR #15412, opened 2026-04-13
Fixed SSRF vulnerability in AI model provider URLs

zzzeebra:
## Summary

Closes SSRF vulnerability in AI model endpoints (`/api/ai/model/chat` and `/api/ai/model`).

A tenant admin could specify an internal network address (e.g. `http://172.17.0.1:22`) as a provider base URL and use "Check connectivity" to scan internal services. The server would make the request and return the response — including service banners like `SSH-2.0-OpenSSH_9.6p1`.

### What changed

Applied the existing `SsrfProtectionValidator.validateUri()` to the three AI providers that accept user-supplied URLs:

| Provider | URL field | 
|----------|-----------|
| OpenAI | `baseUrl` |
| Azure OpenAI | `endpoint` |
| Ollama | `baseUrl` |

Other providers (Google AI, Mistral, Anthropic, AWS Bedrock, GitHub Models) have URLs hardcoded in their SDKs — no user-controlled URL, no SSRF vector.

Validation is added in two places (defence-in-depth, matching the existing pattern used for OAuth2 and webhooks):

1. **At execution time** in `Langchain4jChatModelConfigurerImpl` — blocks the HTTP request before it's made
2. **At save time** in `AiModelDataValidator` — rejects models with blocked URLs on save (same pattern as `Oauth2ClientDataValidator`)

Both are controlled by the global `actors.rule.external.ssrf_protection_enabled` flag (default: `false`). When enabled, private/internal IPs, localhost, link-local, and cloud metadata endpoints are blocked.

## Automated tests

**`Langchain4jChatModelConfigurerImplTest`** — 4 unit tests for execution-time validation:
- OpenAI with private IP (`172.17.0.1`) — blocked
- OpenAI with localhost — blocked
- Azure OpenAI with private IP (`10.0.0.1`) — blocked
- Ollama with private IP (`192.168.1.100`) — blocked

**`AiModelControllerTest#saveAiModel_whenBaseUrlIsPrivateIp_shouldReturnBadRequest`** — integration test for save-time validation: saving an AI model with a private IP as base URL returns HTTP 400.

## Manual tests

1. OpenAI — Save with private IP → blocked
2. OpenAI — Check connectivity with private IP → blocked
3. Azure OpenAI — Save with localhost → blocked
4. Azure OpenAI — Check connectivity with localhost → blocked
5. Ollama — Save with private IP → blocked
6. Ollama — Check connectivity with private IP → blocked
7. OpenAI — Save with normal URL (`https://api.openai.com/v1`) → saves successfully
8. OpenAI — Check connectivity with normal URL → request goes through (401 auth error)


application/src/main/java/org/thingsboard/server/service/ai/Langchain4jChatModelConfigurerImpl.java
```diff
@@ -69,6 +71,7 @@ class Langchain4jChatModelConfigurerImpl implements Langchain4jChatModelConfigur
 
     @Override
     public ChatModel configureChatModel(OpenAiChatModelConfig chatModelConfig) {
+        validateBaseUrl(chatModelConfig.providerConfig().baseUrl());
```
ViacheslavKlimov, line 74: `validateBaseUrl` is called synchronously inside `sendChatRequestAsync` (via the `configure(chatModelConfigurer)` call), before the `FluentFuture` is created. In the controller's `sendChatRequest`, the `.catching(Throwable.class, ex -> new TbChatResponse.Failure(...))` only catches failures within the future chain — it won't catch a synchronous `RuntimeException` thrown here.

So when SSRF protection blocks a URL at execution time, instead of returning the structured `TbChatResponse.Failure` envelope the chat endpoint normally uses, the user gets a raw 500 from the generic `@ExceptionHandler`.

The save-time path (`AiModelDataValidator`) is fine — it properly wraps the exception in `DataValidationException`. But for the execution-time path, you might want to either:
1. Wrap the `configure()` call in `AiChatModelServiceImpl.sendChatRequestAsync` so it returns a failed future instead of throwing, or
2. Catch and wrap the exception in the controller before the future chain.

ViacheslavKlimov (commented):
## Review summary

Reviewed 4 changed files in [fix: add SSRF protection to AI model provider URLs](https://github.com/thingsboard/thingsboard/pull/15412). Left 1 comment inline.

Clean, well-scoped security fix. The defence-in-depth approach (validate at save time + execution time) and the test coverage are solid.

## Additional findings

- **`application/src/main/java/org/thingsboard/server/service/ai/Langchain4jChatModelConfigurerImpl.java:184-185`** — Copy-paste bug in `configureChatModel(GoogleVertexAiGeminiChatModelConfig)`: both the `if` condition and the value passed to `setPresencePenalty()` use `chatModelConfig.frequencyPenalty()` instead of `chatModelConfig.presencePenalty()`. This means presence penalty is always ignored for Vertex AI Gemini models, and frequency penalty is applied twice.

---

application/src/main/java/org/thingsboard/server/service/ai/Langchain4jChatModelConfigurerImpl.java
zzzeebra: Good catch — confirmed, `configure()` is called synchronously in `AiChatModelServiceImpl.sendChatRequestAsync` before the future is built, so a sync throw bypasses the controller's `.catching()` and surfaces as a 500 instead of a `TbChatResponse.Failure` envelope.

Went with option 1 — wrapped the `configure()` call in `AiChatModelServiceImpl.sendChatRequestAsync` so any exception is converted to `Futures.immediateFailedFuture(t)`. Also added an integration test `sendChatRequest_whenBaseUrlBlockedAtRuntime_shouldReturnFailureEnvelope` that enables SSRF protection at runtime and verifies `/api/ai/chat` returns a `Failure` envelope (not a 500).

Save-time path left untouched — it already correctly wraps the exception in `DataValidationException`.

dao/src/main/java/org/thingsboard/server/dao/service/validator/AiModelDataValidator.java
```diff
@@ -64,6 +70,26 @@ protected void validateDataImpl(TenantId tenantId, AiModel model) {
         if (!tenantService.tenantExists(tenantId)) {
             throw new DataValidationException("AI model reference a non-existent tenant!");
         }
+
+        // provider URL SSRF validation
+        if (model.getConfiguration() != null) {
+            AiProviderConfig providerConfig = model.getConfiguration().providerConfig();
+            String url = null;
+            if (providerConfig instanceof OpenAiProviderConfig c) {
```
ViacheslavKlimov, line 78: Since `AiProviderConfig` is a sealed interface, this `if/else if` chain could be a switch expression so the compiler enforces exhaustiveness:

```java
String url = switch (providerConfig) {
    case OpenAiProviderConfig c -> c.baseUrl();
    case AzureOpenAiProviderConfig c -> c.endpoint();
    case OllamaProviderConfig c -> c.baseUrl();
    case GoogleAiGeminiProviderConfig ignored -> null;
    case GoogleVertexAiGeminiProviderConfig ignored -> null;
    case MistralAiProviderConfig ignored -> null;
    case AnthropicProviderConfig ignored -> null;
    case AmazonBedrockProviderConfig ignored -> null;
    case GitHubModelsProviderConfig ignored -> null;
};
```

The same set of providers is enumerated again in `Langchain4jChatModelConfigurerImpl` (where each one is configured separately), so adding a new URL-bearing provider in the future means remembering to update both call sites — and silently skipping SSRF on save if you forget here. Switching to an exhaustive `switch` would turn that into a compile error. Even simpler: add a default `default String getUrl() { return null; }` (or similar) on `AiProviderConfig` and override in the three URL-bearing records — then this method body shrinks to one line and there's no enumeration to keep in sync.

application/src/test/java/org/thingsboard/server/controller/AiModelControllerTest.java
```diff
@@ -136,6 +143,68 @@ public void saveAiModel_whenUpdatingExistingModelAsTenantAdmin_shouldSucceed() t
         assertThat(updatedModel.getExternalId()).isNull();
     }
 
+    @Test
+    public void saveAiModel_whenBaseUrlIsPrivateIp_shouldReturnBadRequest() throws Exception {
+        // GIVEN
+        loginTenantAdmin();
+        SsrfProtectionValidator.setEnabled(true);
```
ViacheslavKlimov, line 150: These two tests unconditionally call `setEnabled(true)` on entry and `setEnabled(false)` in `finally`, so they leak `false` into whatever runs next instead of restoring the previous value. `SsrfProtectionValidatorTest` already shows the safer pattern:

```java
boolean original = SsrfProtectionValidator.isEnabled();
SsrfProtectionValidator.setEnabled(true);
try {
    ...
} finally {
    SsrfProtectionValidator.setEnabled(original);
}
```

It's safe today because the default is `false` and the application module isn't running tests in parallel, but if either of those changes (e.g. someone enables the flag in `application-test.properties`, or a future `junit-platform.properties` enables parallelism) this becomes flaky. Also worth noting: `Langchain4jChatModelConfigurerImplTest` got an `@ResourceLock("SsrfProtectionValidator")` to guard against parallel runs but `AiModelControllerTest` didn't — and JUnit 4 (which `AbstractControllerTest` uses) wouldn't honor that annotation anyway.

application/src/test/java/org/thingsboard/server/controller/AiModelControllerTest.java
```diff
@@ -136,6 +143,68 @@ public void saveAiModel_whenUpdatingExistingModelAsTenantAdmin_shouldSucceed() t
...
+                    null,
+                    new TbUserMessage(List.of(new TbContent.TbTextContent("hi"))),
+                    modelConfig);
+
+            // WHEN
+            TbChatResponse response = doPostAsync("/api/ai/model/chat", chatRequest, TbChatResponse.class, status().isOk());
+
+            // THEN
+            assertThat(response).isInstanceOf(TbChatResponse.Failure.class);
+            assertThat(((TbChatResponse.Failure) response).errorDetails()).contains("URI is invalid");
```
ViacheslavKlimov, line 202: Asserting on the user-facing error string (`"URI is invalid"`) couples this integration test to the wording inside `SsrfProtectionValidator`. If anyone reworks that exception message later, this test breaks for a reason that has nothing to do with the controller. Asserting `errorDetails()` is non-null/non-blank, or matching on a stable substring like the host (`"10.0.0.1"`), would make this less brittle while still proving the SSRF block fired.

application/src/test/java/org/thingsboard/server/service/ai/Langchain4jChatModelConfigurerImplTest.java
```diff
@@ -41,6 +53,72 @@ class Langchain4jChatModelConfigurerImplTest {
...
     private final Langchain4jChatModelConfigurerImpl configurer = new Langchain4jChatModelConfigurerImpl();
 
+    @BeforeEach
+    void enableSsrfProtection() {
+        SsrfProtectionValidator.setEnabled(true);
+    }
+
+    @AfterEach
+    void disableSsrfProtection() {
+        SsrfProtectionValidator.setEnabled(false);
```
ViacheslavKlimov, line 63: Same state-preservation concern as `AiModelControllerTest`: `disableSsrfProtection()` always resets to `false` rather than restoring the original. Since this class is the one that explicitly added `@ResourceLock("SsrfProtectionValidator")` (presumably anticipating parallel execution), it's worth getting this right here too:

```java
private boolean originalSsrfState;

@BeforeEach
void enableSsrfProtection() {
    originalSsrfState = SsrfProtectionValidator.isEnabled();
    SsrfProtectionValidator.setEnabled(true);
}

@AfterEach
void restoreSsrfProtection() {
    SsrfProtectionValidator.setEnabled(originalSsrfState);
}
```

ViacheslavKlimov (commented):
## Review summary

Reviewed 5 changed files in [Fixed SSRF vulnerability in AI model provider URLs](https://github.com/thingsboard/thingsboard/pull/15412). Left 4 comment(s) inline.

The fix correctly applies `SsrfProtectionValidator.validateUri()` to the three providers that accept user-supplied URLs (OpenAI, Azure OpenAI, Ollama) — and the PR description's claim that the other providers don't have user-controlled URLs checks out against the provider config records. Validation is layered at both save time and runtime, mirroring the OAuth2/webhook pattern. Most of the comments are about future-maintainability and test isolation rather than bugs.

## Additional findings

These observations are about behavior of the wider system, not the diff itself.

- **TOCTOU / DNS rebinding gap.** `validateUri` resolves the host once at `configureChatModel` time, but the actual HTTP request is made later by the langchain4j HTTP client, which doesn't go through `SsrfProtectionValidator`. A malicious DNS server could return a public IP at validation time and a private IP at request time. The codebase already has `SsrfSafeAddressResolverGroup` for the rule-engine REST node — wiring an equivalent into the langchain4j HTTP client (or using `pinHostAddress`-style trick) would close this gap. Out of scope for this fix, but worth a follow-up issue.
- **Defaults-off note.** `SSRF_PROTECTION_ENABLED` defaults to `false` (existing global flag), so deployments that don't flip the env var remain vulnerable after this PR. The release notes / CVE advisory should call this out so admins know they need to opt in (and configure allow-lists for legit private Ollama endpoints).

---

dao/src/main/java/org/thingsboard/server/dao/service/validator/AiModelDataValidator.java
zzzeebra: Going to leave as-is.

Java 17 (`<release>17</release>` in root pom) — pattern-matching switch is preview only, not used in prod. A default method on `AiProviderConfig` would push an SSRF-shaped concern into a domain interface in `common/data` for one validator's call site, and doesn't fully prevent the bug either since new permitted records can still rely on the null default without compile error.

Adding a new AI provider requires per-type wiring in `Langchain4jChatModelConfigurerImpl` anyway, so anyone adding one naturally surfaces both files.

application/src/test/java/org/thingsboard/server/controller/AiModelControllerTest.java
zzzeebra: Going to leave as-is.

The save & restore pattern would be defensive against scenarios that aren't in play: `SsrfProtectionValidator.isEnabled()` defaults to `false`, and `AbstractControllerTest` is JUnit 4 (no parallel runs, ignores `@ResourceLock`). So `setEnabled(false)` in `finally` is behaviorally equivalent to `setEnabled(originalState)` here.

If/when the application module migrates to JUnit 5 with parallel execution, that's its own effort with a broader fixture audit — better to do test-isolation patterns uniformly at that point than spot-fix here.

application/src/test/java/org/thingsboard/server/controller/AiModelControllerTest.java
zzzeebra: Going to leave as-is.

The point of this test is to prove that `/api/ai/model/chat` returns a `Failure` envelope **specifically when SSRF blocked the URL**, not when something unrelated failed. Asserting on the host substring (`10.0.0.1`) wouldn't help here — connection-refused errors from langchain4j's HTTP client also include the host, so the test would silently pass if `setEnabled(true)` ever no-op'd and the call reached the network. `non-blank` is even weaker.

`URI is invalid` is the only string unique to the `SsrfProtectionValidator.validateUri` path. Coupling to it is intentional — for a security regression test, brittleness against changes to that wording is a feature: it forces whoever changes the SSRF message to confirm every test that exercises the path is still correct.

application/src/test/java/org/thingsboard/server/service/ai/Langchain4jChatModelConfigurerImplTest.java
zzzeebra: Going to leave as-is.

Same reasoning as the `AiModelControllerTest` thread. `@ResourceLock` serializes this class against any other test holding the same lock, and within that window `@BeforeEach`/`@AfterEach` toggle `false → true → false`, which matches the global default. There's no real leak today.

If the default ever changes or a broader parallel-readiness audit happens, save & restore can be applied uniformly across fixtures touching `SsrfProtectionValidator`.

PR #15425, opened 2026-04-15
Added config property to control null ordering in dashboards

dashevchenko:
## Pull Request description

Previously, there were two issues:

1.) Backend: the PostgreSQL query handled rows with null values incorrectly, excepting null values as 0. Old query:

select * from (select entities.*, (coalesce(cast(alias5.bool_v as varchar), '') || coalesce(alias5.str_v, '') || coalesce(cast(alias5.long_v as varchar), '') || coalesce(cast(alias5.dbl_v as varchar), '') || coalesce(cast(alias5.json_v as varchar), '')) as alias5_value, alias5.ts as alias5_ts, coalesce(alias5.dbl_v, cast(alias5.long_v as double precision), **(case when alias5.bool_v then 1 else 0 end)**) alias5_value_so_num,coalesce(alias5.str_v, cast(alias5.json_v as varchar), '') alias5_value_so_varchar from (select e.id id, 'DEVICE' entity_type, cast(e.name as varchar) as alias2, cast(e.label as varchar) as alias3, cast(e.additional_info as varchar) as alias4 from device e where e.tenant_id=:permissions_tenant_id and (e.id in (:entity_filter_entity_ids))) entities left join ts_kv_latest alias5 ON alias5.entity_id=entities.id AND alias5.key = (select key_id from key_dictionary where key = :alias5_key_id)  ) result  order by alias5_value_so_num asc, alias5_value_so_varchar asc, result.id asc limit 10 offset 0

When both dbl_v and long_v are NULL (rows with no value), the **CASE WHEN alias5.bool_v THEN 1 ELSE 0 END** evaluates to 0 instead of NULL, because CASE WHEN NULL THEN 1 ELSE 0 END returns 0 in PostgreSQL.

2.) UI: incorrect sorting for empty values

3.) For SQL-backed entity data queries added a configurable nulls order strategy (default, nulls_first, nulls_last)

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/test/java/org/thingsboard/server/service/entitiy/EdqsEntityServiceTest.java
+++ b/application/src/test/java/org/thingsboard/server/service/entitiy/EdqsEntityServiceTest.java
@@ -16,19 +16,27 @@
 package org.thingsboard.server.service.entitiy;
 
 import com.google.common.collect.Lists;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
 import org.junit.Before;
 import org.junit.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.test.context.TestPropertySource;
+import org.thingsboard.server.common.data.Device;
 import org.thingsboard.server.common.data.EntityType;
 import org.thingsboard.server.common.data.asset.Asset;
 import org.thingsboard.server.common.data.id.CustomerId;
 import org.thingsboard.server.common.data.id.IdBased;
+import org.thingsboard.server.common.data.kv.TimeseriesSaveResult;
 import org.thingsboard.server.common.data.page.PageData;
+import org.thingsboard.server.common.data.query.DeviceTypeFilter;
 import org.thingsboard.server.common.data.query.EntityCountQuery;
 import org.thingsboard.server.common.data.query.EntityData;
+import org.thingsboard.server.common.data.query.EntityDataPageLink;
 import org.thingsboard.server.common.data.query.EntityDataQuery;
+import org.thingsboard.server.common.data.query.EntityDataSortOrder;
+import org.thingsboard.server.common.data.query.EntityKey;
 import org.thingsboard.server.common.data.query.EntityKeyType;
 import org.thingsboard.server.common.data.query.RelationsQueryFilter;
 import org.thingsboard.server.common.data.relation.EntitySearchDirection;
@@ -43,10 +51,13 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
+import static org.thingsboard.server.common.data.query.EntityKeyType.ENTITY_FIELD;
 
 @DaoSqlTest
 @TestPropertySource(properties = {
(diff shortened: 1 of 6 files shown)


ui-ngx/src/app/shared/models/page/page-link.ts
```diff
@@ -84,11 +84,23 @@ export function sortItems(item1: any, item2: any, property: string, asc: boolean
...
-      if (item1Value && !item2Value) {
+    } else if (item1Type === 'boolean' && item2Type === 'boolean') {
+      result = item1Value ? 1 : -1;
+    } else if (item1Type !== item2Type) {
+      const item1Empty = item1Value === null || item1Value === undefined || item1Value === '';
+      const item2Empty = item2Value === null || item2Value === undefined || item2Value === '';
+      if (!item1Empty && item2Empty) {
         result = 1;
-      } else if (!item1Value && item2Value) {
+      } else if (item1Empty && !item2Empty) {
```
vvlladd28, line 94: This returns `-1` when only `item1` is empty, so in ASC order rows with empty values come **first**. The companion backend fix in `EntityKeyMapping.java` now produces `NULL` for value-less rows, and Postgres' default `ORDER BY ... ASC` treats NULL as **larger** than non-null (NULLS LAST) — so rows sort last on the server. Since `sortItems` is invoked on top of already-paginated backend results (`entities-table-widget.component.ts:937`, `alarms-table-widget.component.ts:1333`), the UI can flip the position of empty rows away from where the backend placed them. Was empty-first for ASC intentional, or should this mirror the backend (empty last for ASC, first for DESC)?

ui-ngx/src/app/shared/models/page/page-link.ts
```diff
@@ -84,11 +84,23 @@ export function sortItems(item1: any, item2: any, property: string, asc: boolean
...
         result = -1;
+      } else if (!item1Empty && !item2Empty) {
+        const num1 = Number(item1Value);
```
vvlladd28, line 97: The `item1Empty`/`item2Empty` checks above only treat the exact string `''` as empty, but `Number(' ')`, `Number('\t')`, and `Number('\n')` all coerce to `0`. A whitespace-only string compared against a real number would therefore be pulled into the numeric branch and sorted as `0` rather than falling through to `localeCompare`. If such values can appear in telemetry/attribute strings, consider trimming before the `isNaN` check (or using `Number.isFinite(Number(item1Value.toString().trim()))` style detection) so whitespace-only strings are compared lexically like any other string.

vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [[Entity data queries] Fixed telemetry/attribute sorting for rows with null values](https://github.com/thingsboard/thingsboard/pull/15425). The backend SQL fix is straightforward and correct — explicit `IS NULL` handling prevents the CASE expression from collapsing NULL `bool_v` to `0`. On the UI side the new cross-type branch is a real improvement over the previous truthiness-based check, but it introduces an ordering asymmetry with the backend worth confirming, plus a minor edge case around whitespace strings. Left 2 comments inline.

## Additional findings

These observations are about surrounding/overall concerns rather than specific diff lines.

- **Tests** — neither change has a regression test. `EntityKeyMappingTest.java` currently only covers `splitByCommaWithoutQuotes`, but the SQL-builder bug (CASE returning `0` for NULL `bool_v`) is exactly the kind of subtle issue a focused unit test on `buildTimeSeriesOrAttrSelection` could pin down. Similarly, `sortItems` has no spec coverage, so the new empty/cross-type logic has no safety net for future refactors. The PR checklist still shows the test items unchecked — worth adding at least a `sortItems` unit test given the behavioral change across several branches.

vvlladd28 (dismissed):

ViacheslavKlimov (commented):
## Review summary

Reviewed 6 changed files in [[Entity data queries] Fixed telemetry/attribute sorting for rows with null values](https://github.com/thingsboard/thingsboard/pull/15425). Left 4 comment(s) inline.

The core fix in `EntityKeyMapping.java` looks correct — the extra `case when bool_v is null then null` branch is exactly what's needed to stop PostgreSQL from treating NULL booleans as 0. The configurable `nullsOrderStrategy` is a nice escape hatch for users who want deterministic behavior across `NULL` and non-`NULL` rows.

A few things worth a second look: the EDQS path's default null ordering is now the opposite of the SQL default, invalid config values silently degrade to default, and the tests mutate a shared Spring bean via reflection. Details inline.

---

application/src/test/java/org/thingsboard/server/service/entitiy/EdqsEntityServiceTest.java
```diff
@@ -103,6 +114,100 @@ public void testCountHierarchicalEntitiesByMultiRootQuery() throws InterruptedEx
         assetService.deleteAssetsByTenantId(tenantId);
     }
 
+    // edqs has no nulls order strategies, always returns NULLs first for ASC and NULLs last for DESC
```
ViacheslavKlimov, line 117: This comment documents a real discrepancy that deserves to be visible outside the test file: EDQS always orders NULLs first for ASC / last for DESC, while the new SQL default is the opposite (NULLS LAST for ASC, NULLS FIRST for DESC, per the yml description). That means a tenant who toggles EDQS on/off will see a different null ordering for the same dataset, and the `sql.entity_data_query_nulls_order_strategy` knob silently has no effect on the EDQS path.

A few options worth considering:
- Align EDQS's default to match PostgreSQL's default so the `default` strategy is truly consistent.
- Plumb the strategy through to EDQS so all three options work there too.
- At minimum, call out the EDQS limitation in the `entity_data_query_nulls_order_strategy` yml comment so ops users aren't surprised.

dao/src/main/java/org/thingsboard/server/dao/sql/query/DefaultEntityQueryRepository.java
```diff
@@ -525,6 +530,14 @@ public PageData<EntityData> findEntityDataByQuery(TenantId tenantId, CustomerId
         });
     }
 
+    private String resolveNullsOrder() {
+        return switch (nullsOrderStrategy) {
```
ViacheslavKlimov, line 550: Invalid values for `SQL_ENTITY_DATA_QUERY_NULLS_ORDER_STRATEGY` fall into the `default` branch silently — a typo like `nulls-first`, `first`, or `NULLS_FIRST` (uppercase) will look accepted but will actually use the PostgreSQL default. Consider validating the value at startup (e.g. in a `@PostConstruct`) and either logging a warning or failing fast. A small enum or `Set<String>` of accepted values would also prevent the same literals from drifting out of sync between this `switch` and the yml description.

application/src/test/java/org/thingsboard/server/service/entitiy/EntityServiceTest.java
```diff
@@ -1750,6 +1753,117 @@ public void testFindEntityDataByQueryWithTimeseries() throws ExecutionException,
...
+                    List.of("true", "false", "false", "", ""),
+                    devices.size());
+        } finally {
+            deviceService.deleteDevicesByTenantId(tenantId);
+        }
+    }
+
+    private void assertNullsOrdering(String strategy, List<String> expectedAsc, List<String> expectedDesc, int deviceSize) {
+        String originalStrategy = entityQueryRepository.getNullsOrderStrategy();
+        ReflectionTestUtils.setField(entityQueryRepository, "nullsOrderStrategy", strategy);
```
ViacheslavKlimov, line 1836: Mutating a Spring-managed singleton's field via reflection inside a test couples these tests to global state. It's fine today because application tests run sequentially, but any future move toward parallel test execution (or a second test in the same module that reads/writes the same field) will race. Two cleaner options:
- Split each strategy into its own test class with `@TestPropertySource(properties = "sql.entity_data_query_nulls_order_strategy=...")`.
- Extract the sort clause construction to a method that takes the strategy as a parameter so it can be unit-tested without touching the bean.

Also worth noting: this relies on the injected field not being `final`. If Lombok/Spring conventions in this module ever shift toward constructor injection with `final` fields, this test will silently stop doing what it looks like it's doing.

ui-ngx/src/app/shared/models/page/page-link.ts
```diff
@@ -84,11 +84,25 @@ export function sortItems(item1: any, item2: any, property: string, asc: boolean
...
+    } else if (item1Type === 'boolean' && item2Type === 'boolean') {
+      result = item1Value ? 1 : -1;
+    } else if (item1Type !== item2Type) {
+      const item1Empty = item1Value === null || item1Value === undefined || item1Value === '';
+      const item2Empty = item2Value === null || item2Value === undefined || item2Value === '';
+      if (!item1Empty && item2Empty) {
+        return asc ? 1 : -1;
+      } else if (item1Empty && !item2Empty) {
+        return asc ? -1 : 1;
+      } else if (!item1Empty && !item2Empty) {
```
ViacheslavKlimov, line 96: This mixed-type branch is a larger behavior change than it might look. Before, any two mixed-type values used simple truthy/falsy comparison; now it's empty-check → numeric parse → `localeCompare`. A few cases that shift:

- `true` vs `'hello'`: old → tie (both truthy, `result = 0`); new → `'true'.localeCompare('hello')` → negative.
- `1` (number) vs `'5'` (string): old → tie (both truthy); new → `1 - 5 = -4` via numeric branch.
- `0` (number) vs `'anything'`: old → `!item1Value && item2Value` → `result = -1`; new → numeric `0` vs NaN → falls to string compare.

For telemetry-style data the new behavior is clearly more useful, but any widget that happened to rely on the old truthy/falsy ordering (entities table, alarms table are the two callers) may shift. Worth spot-checking those two widgets with a mixed-type column to make sure no existing sort is visibly broken.

Minor: the early `return asc ? 1 : -1` at lines 93/95 short-circuits the final `asc ? result : result * -1` at line 109, while the `result = ...` assignments in this block go through it. That asymmetry works but is easy to get wrong on a future edit — consider also assigning `result` here for uniformity.

dao/src/main/java/org/thingsboard/server/dao/sql/query/DefaultEntityQueryRepository.java
dashevchenko: fixed

ui-ngx/src/app/shared/models/page/page-link.ts
dashevchenko: Fixed.

PR #15429, opened 2026-04-15
Fixed incorrect aggregated values in Entity Table widget data export

zzzeebra:
## Summary
- Added REST endpoint `POST /api/entitiesQuery/find/aggHistory` — REST equivalent of the WebSocket `AggHistoryCmd`. For each matched entity returns one aggregated `TsValue` per key over `[startTs, endTs]`, with optional previous window (`previousStartTs/previousEndTs`) in the `aggLatest` field.
- Motivation: PE UI exports Entity Table widget data via REST. The widget rendered aggregated values (e.g. SUM) but CSV export returned raw latest because no REST aggregated-history endpoint existed. With this endpoint PE UI can fetch the same aggregated data the WS channel provides.
- Pages the entity set consistently with `/api/entitiesQuery/find` (no separate page-size cap), validates `endTs >= startTs`. On per-entity fetch failure `aggLatest` is filled with `ComparisonTsValue(EMPTY, EMPTY)` for a predictable response contract.

## Test plan
- [x] `EntityQueryControllerTest#testFindEntityDataAggHistoryByQuery` — 20 points × SUM → current=600, previous=300 (10 points)
- [x] `testFindEntityDataAggHistoryByQuery_previousValueOnly` — `previousValueOnly=true` → current=null, previous=sum
- [x] `testFindEntityDataAggHistoryByQuery_noTelemetry` — empty window → current="0" (SUM aggregator behavior), previous=null
- [x] `testFindEntityDataAggHistoryByQuery_multipleKeysMixedAggregation` — MIN=10, MAX=50, AVG=30 on same key with three `AggKey` entries

### DIFF
--- a/application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
@@ -17,6 +17,7 @@ package org.thingsboard.server.controller;
 
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.Schema;
+import lombok.Data;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.prepost.PreAuthorize;
@@ -49,6 +50,7 @@ import org.thingsboard.server.config.annotations.ApiOperation;
 import org.thingsboard.server.queue.util.TbCoreComponent;
 import org.thingsboard.server.service.query.EntityQueryService;
 import org.thingsboard.server.service.security.permission.Operation;
+import org.thingsboard.server.service.ws.telemetry.cmd.v2.AggHistoryCmd;
 
 import static org.thingsboard.server.controller.ControllerConstants.ALARM_DATA_QUERY_DESCRIPTION;
 import static org.thingsboard.server.controller.ControllerConstants.ATTRIBUTES_SCOPE_DESCRIPTION;
@@ -147,6 +149,36 @@ public class EntityQueryController extends BaseController {
             pageLink.setPageSize(MAX_PAGE_SIZE);
         }
         return entityQueryService.getKeysByQuery(getCurrentUser(), tenantId, query, isTimeseries, isAttributes, scope);
+    }
+
+    @ApiOperation(value = "Find Aggregated Historical Entity Data by Query",
+            notes = "Runs the entity data query and, for each matched entity, fetches a single aggregated value per key over [startTs, endTs] " +
+                    "with per-key aggregation function. Optional previousStartTs/previousEndTs per key add a comparison window. " +
+                    "REST equivalent of the WebSocket AggHistoryCmd. The aggregated values are returned in the 'aggLatest' field of each EntityData.")
+    @PreAuthorize("hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN', 'CUSTOMER_USER')")
+    @PostMapping("/entitiesQuery/find/aggHistory")
+    public DeferredResult<PageData<EntityData>> findEntityDataAggHistoryByQuery(
+            @Parameter(description = "A JSON value representing the entity data query and aggregated history command.")
+            @RequestBody EntityDataAggHistoryRequest request) throws ThingsboardException {
+        checkNotNull(request);
+        checkNotNull(request.getQuery());
+        checkNotNull(request.getAggHistoryCmd());
+        AggHistoryCmd cmd = request.getAggHistoryCmd();
+        if (cmd.getEndTs() < cmd.getStartTs()) {
+            throw new IllegalArgumentException("endTs must be >= startTs");
+        }
+        EntityDataPageLink pageLink = request.getQuery().getPageLink();
+        if (pageLink != null && pageLink.getPageSize() > MAX_PAGE_SIZE) {
+            pageLink.setPageSize(MAX_PAGE_SIZE);
+        }
+        resolveQuery(request.getQuery());
+        return entityQueryService.findEntityDataAggHistoryByQuery(getCurrentUser(), request.getQuery(), cmd);
+    }
+
+    @Data
+    public static class EntityDataAggHistoryRequest {
+        private EntityDataQuery query;
+        private AggHistoryCmd aggHistoryCmd;
     }
 
     @PreAuthorize("hasAnyAuthority('SYS_ADMIN')")
--- a/application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
@@ -155,7 +155,8 @@ public class EntityQueryController extends BaseController {
     @ApiOperation(value = "Find Aggregated Historical Entity Data by Query",
             notes = "Runs the entity data query and, for each matched entity, fetches a single aggregated value per key over [startTs, endTs] " +
                     "with per-key aggregation function. Optional previousStartTs/previousEndTs per key add a comparison window. " +
-                    "REST equivalent of the WebSocket AggHistoryCmd. The aggregated values are returned in the 'aggLatest' field of each EntityData.")
+                    "REST equivalent of the WebSocket AggHistoryCmd. The aggregated values are returned in the 'aggLatest' field of each EntityData. " +
+                    "Note: a separate timeseries aggregation query is issued per matched entity, so a large pageSize fans out proportionally - callers should page responsibly.")
     @PreAuthorize("hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN', 'CUSTOMER_USER')")
     @PostMapping("/entitiesQuery/find/aggHistory")
     public DeferredResult<PageData<EntityData>> findEntityDataAggHistoryByQuery(
--- a/application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
@@ -168,10 +168,6 @@ public class EntityQueryController extends BaseController {
         if (cmd.getEndTs() < cmd.getStartTs()) {
             throw new ThingsboardException("endTs must be >= startTs", ThingsboardErrorCode.BAD_REQUEST_PARAMS);
         }
-        EntityDataPageLink pageLink = request.getQuery().getPageLink();
-        if (pageLink != null && pageLink.getPageSize() > MAX_PAGE_SIZE) {
-            pageLink.setPageSize(MAX_PAGE_SIZE);
-        }
         resolveQuery(request.getQuery());
         return entityQueryService.findEntityDataAggHistoryByQuery(getCurrentUser(), request.getQuery(), cmd);
     }
--- a/application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
@@ -32,6 +32,7 @@ import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.context.request.async.DeferredResult;
 import org.thingsboard.server.common.data.edqs.EdqsState;
 import org.thingsboard.server.common.data.edqs.ToCoreEdqsRequest;
+import org.thingsboard.server.common.data.exception.ThingsboardErrorCode;
 import org.thingsboard.server.common.data.exception.ThingsboardException;
 import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.common.data.id.UserId;
@@ -165,7 +166,7 @@ public class EntityQueryController extends BaseController {
         checkNotNull(request.getAggHistoryCmd());
         AggHistoryCmd cmd = request.getAggHistoryCmd();
         if (cmd.getEndTs() < cmd.getStartTs()) {
-            throw new IllegalArgumentException("endTs must be >= startTs");
+            throw new ThingsboardException("endTs must be >= startTs", ThingsboardErrorCode.BAD_REQUEST_PARAMS);
         }
         EntityDataPageLink pageLink = request.getQuery().getPageLink();
         if (pageLink != null && pageLink.getPageSize() > MAX_PAGE_SIZE) {
--- a/application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
+++ b/application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
@@ -36,14 +36,12 @@ import org.thingsboard.server.common.data.EntityType;
 import org.thingsboard.server.common.data.id.EntityId;
 import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.common.data.kv.AttributeKvEntry;
-import org.thingsboard.server.common.data.kv.BaseReadTsKvQuery;
 import org.thingsboard.server.common.data.kv.ReadTsKvQuery;
 import org.thingsboard.server.common.data.kv.ReadTsKvQueryResult;
 import org.thingsboard.server.common.data.page.PageData;
 import org.thingsboard.server.common.data.query.AlarmCountQuery;
 import org.thingsboard.server.common.data.query.AlarmData;
 import org.thingsboard.server.common.data.query.AlarmDataQuery;
-import org.thingsboard.server.common.data.query.ComparisonTsValue;
 import org.thingsboard.server.common.data.query.ComplexFilterPredicate;
 import org.thingsboard.server.common.data.query.DynamicValue;
 import org.thingsboard.server.common.data.query.EntityCountQuery;
@@ -57,7 +55,6 @@ import org.thingsboard.server.common.data.query.FilterPredicateType;
 import org.thingsboard.server.common.data.query.KeyFilter;
 import org.thingsboard.server.common.data.query.KeyFilterPredicate;
 import org.thingsboard.server.common.data.query.SimpleKeyFilterPredicate;
-import org.thingsboard.server.common.data.query.TsValue;
 import org.thingsboard.server.dao.alarm.AlarmService;
 import org.thingsboard.server.dao.attributes.AttributesService;
 import org.thingsboard.server.dao.entity.EntityService;
@@ -68,14 +65,13 @@ import org.thingsboard.server.queue.util.TbCoreComponent;
 import org.thingsboard.server.service.executors.DbCallbackExecutorService;
 import org.thingsboard.server.service.security.AccessValidator;
 import org.thingsboard.server.service.security.model.SecurityUser;
+import org.thingsboard.server.service.subscription.AggregationQueryUtils;
 import org.thingsboard.server.service.subscription.ReadTsKvQueryInfo;
 import org.thingsboard.server.service.ws.telemetry.cmd.v2.AggHistoryCmd;
-import org.thingsboard.server.service.ws.telemetry.cmd.v2.AggKey;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -358,19 +354,7 @@ public class DefaultEntityQueryService implements EntityQueryService {
             return response;
         }
         TenantId tenantId = securityUser.getTenantId();
-        Map<Integer, ReadTsKvQueryInfo> queries = new HashMap<>();
-        for (AggKey key : cmd.getKeys()) {
-            if (key.getPreviousValueOnly() == null || !key.getPreviousValueOnly()) {
-                var q = new BaseReadTsKvQuery(key.getKey(), cmd.getStartTs(), cmd.getEndTs(),
-                        cmd.getEndTs() - cmd.getStartTs(), 1, key.getAgg());
-                queries.put(q.getId(), new ReadTsKvQueryInfo(key, q, false));
-            }
-            if (key.getPreviousStartTs() != null && key.getPreviousEndTs() != null && key.getPreviousEndTs() >= key.getPreviousStartTs()) {
-                var q = new BaseReadTsKvQuery(key.getKey(), key.getPreviousStartTs(), key.getPreviousEndTs(),
-                        key.getPreviousEndTs() - key.getPreviousStartTs(), 1, key.getAgg());
-                queries.put(q.getId(), new ReadTsKvQueryInfo(key, q, true));
-            }
-        }
+        Map<Integer, ReadTsKvQueryInfo> queries = AggregationQueryUtils.buildAggHistoryQueries(cmd.getKeys(), cmd.getStartTs(), cmd.getEndTs());
         List<ReadTsKvQuery> queryList = queries.values().stream().map(ReadTsKvQueryInfo::getQuery).collect(Collectors.toList());
         Map<EntityData, ListenableFuture<List<ReadTsKvQueryResult>>> fetchResultMap = new LinkedHashMap<>();
         entityDataList.forEach(entityData -> fetchResultMap.put(entityData,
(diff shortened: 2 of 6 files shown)


application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
```diff
@@ -149,6 +151,36 @@ public DeferredResult<ResponseEntity> findEntityTimeseriesAndAttributesKeysByQue
...
+    @PostMapping("/entitiesQuery/find/aggHistory")
+    public DeferredResult<PageData<EntityData>> findEntityDataAggHistoryByQuery(
+            @Parameter(description = "A JSON value representing the entity data query and aggregated history command.")
+            @RequestBody EntityDataAggHistoryRequest request) throws ThingsboardException {
+        checkNotNull(request);
+        checkNotNull(request.getQuery());
+        checkNotNull(request.getAggHistoryCmd());
+        AggHistoryCmd cmd = request.getAggHistoryCmd();
+        if (cmd.getEndTs() < cmd.getStartTs()) {
+            throw new IllegalArgumentException("endTs must be >= startTs");
```
ViacheslavKlimov: Throwing a raw `IllegalArgumentException` here ends up in the catch-all branch of `ThingsboardErrorResponseHandler` (line 162) and returns **500 INTERNAL_SERVER_ERROR**, not 400. The convention elsewhere in this codebase (see e.g. `BaseController.java:514`, `:520`, `:532`) is `throw new ThingsboardException("endTs must be >= startTs", ThingsboardErrorCode.BAD_REQUEST_PARAMS);` which the handler maps to 400. Worth fixing so clients get a proper bad-request response.

application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
```diff
@@ -149,6 +151,36 @@ public DeferredResult<ResponseEntity> findEntityTimeseriesAndAttributesKeysByQue
...
+                    "with per-key aggregation function. Optional previousStartTs/previousEndTs per key add a comparison window. " +
+                    "REST equivalent of the WebSocket AggHistoryCmd. The aggregated values are returned in the 'aggLatest' field of each EntityData.")
+    @PreAuthorize("hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN', 'CUSTOMER_USER')")
+    @PostMapping("/entitiesQuery/find/aggHistory")
+    public DeferredResult<PageData<EntityData>> findEntityDataAggHistoryByQuery(
+            @Parameter(description = "A JSON value representing the entity data query and aggregated history command.")
+            @RequestBody EntityDataAggHistoryRequest request) throws ThingsboardException {
+        checkNotNull(request);
+        checkNotNull(request.getQuery());
+        checkNotNull(request.getAggHistoryCmd());
```
ViacheslavKlimov, line 167: Validation here checks the request envelope but not `cmd.getKeys()`. If a client sends `aggHistoryCmd` with `keys = null` or `[]`, the service silently short-circuits (DefaultEntityQueryService.java:356) and returns the page data with no aggregation in `aggLatest`. That's indistinguishable from "no data found" and quietly masks client bugs. Consider rejecting empty/null keys with `BAD_REQUEST_PARAMS` here, the same way `endTs < startTs` is rejected.

ViacheslavKlimov (commented):
## Review summary

Reviewed 4 changed files in [Fixed incorrect aggregated values in Entity Table widget data export](https://github.com/thingsboard/thingsboard/pull/15429). The new `/api/entitiesQuery/find/aggHistory` endpoint mirrors the WebSocket `AggHistoryCmd` handler and the test coverage for the happy paths is solid. Left 5 inline comments — the most consequential are the error-mapping for invalid request parameters, the missing `RestClient` method (PR checklist item), and a latent NPE/`UnsupportedOperationException` risk on the EDQS path because `EntityData.aggLatest` is not always a mutable map.

## Additional findings

- **`rest-client/src/main/java/org/thingsboard/rest/client/RestClient.java`** — The PR checklist (`pull_request_template.md`) requires updating `RestClient.java` whenever a new REST endpoint is introduced. There's already a `findEntityDataByQuery` method around line 1705 — please add the `findEntityDataAggHistoryByQuery` counterpart so external clients can use the new endpoint.
- **`application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java` vs `application/src/main/java/org/thingsboard/server/service/subscription/DefaultTbEntityDataSubscriptionService.java`** — The new `findEntityDataAggHistoryByQuery` duplicates ~30 lines of logic from `handleAggHistoryCmd` (query construction, the previous-window guard, the per-entity fetch loop, the `(EMPTY, EMPTY)` fallback). They will drift over time. Consider extracting a shared helper (e.g. `AggHistoryFetcher`) so the WS and REST paths stay in lockstep when behaviour evolves.
- **`common/data/src/main/java/org/thingsboard/server/common/data/query/EntityData.java:34`** — `aggLatest` has no default initialization. Callers in both the WS and REST paths assume `getAggLatest()` returns a mutable map and call `computeIfAbsent`/`putIfAbsent` on it. The SQL adapter (`EntityDataAdapter.java:59`) initialises it to `new HashMap<>()`, but the EDQS adapter (`QueryResult.toOldEntityData()`) initialises it to `Collections.emptyMap()`. Initialising it lazily in the getter, or in `toOldEntityData`, would make this safer.

---

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
```diff
@@ -339,4 +348,69 @@ public void onFailure(Throwable t) {
...
+            @Override
+            public void onSuccess(@Nullable List<List<ReadTsKvQueryResult>> ignored) {
+                fetchResultMap.forEach((entityData, future) -> {
+                    try {
+                        List<ReadTsKvQueryResult> queryResults = future.get();
+                        if (queryResults != null) {
+                            for (ReadTsKvQueryResult queryResult : queryResults) {
+                                ReadTsKvQueryInfo info = queries.get(queryResult.getQueryId());
+                                ComparisonTsValue comparisonTsValue = entityData.getAggLatest()
+                                        .computeIfAbsent(info.getKey().getId(), k -> new ComparisonTsValue());
```
ViacheslavKlimov: When EDQS is enabled and `validForEdqs(query)` is true, `BaseEntityService.findEntityDataByQuery` returns `EntityData` instances built by `QueryResult.toOldEntityData()` — and that constructs them with `aggLatest = Collections.emptyMap()`, which is **immutable**. `computeIfAbsent` and the `putIfAbsent` calls a few lines below will throw `UnsupportedOperationException` on every entity in that case.

The same pattern exists in `DefaultTbEntityDataSubscriptionService.handleAggHistoryCmd`, so this isn't a regression introduced by this PR — but it's worth confirming the EDQS path actually works end-to-end here (the new tests use the standard SQL path). The cleanest fix is to make `EntityData.aggLatest` mutable by default, or fix `QueryResult.toOldEntityData()` to use `new HashMap<>()`.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
```diff
@@ -339,4 +348,69 @@ public void onFailure(Throwable t) {
...
+        }
+        TenantId tenantId = securityUser.getTenantId();
+        Map<Integer, ReadTsKvQueryInfo> queries = new HashMap<>();
+        for (AggKey key : cmd.getKeys()) {
+            if (key.getPreviousValueOnly() == null || !key.getPreviousValueOnly()) {
+                var q = new BaseReadTsKvQuery(key.getKey(), cmd.getStartTs(), cmd.getEndTs(),
+                        cmd.getEndTs() - cmd.getStartTs(), 1, key.getAgg());
+                queries.put(q.getId(), new ReadTsKvQueryInfo(key, q, false));
+            }
+            if (key.getPreviousStartTs() != null && key.getPreviousEndTs() != null && key.getPreviousEndTs() >= key.getPreviousStartTs()) {
```
ViacheslavKlimov: If a client sends `previousEndTs < previousStartTs` (or only one of the two), the previous-window query is silently skipped and the response just reports `previous = EMPTY`. Same behaviour as the WS path, but for a REST contract it would be friendlier to validate this in the controller and return `BAD_REQUEST_PARAMS` with a clear message — otherwise misconfigured callers will see "empty" without knowing why.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
```diff
@@ -339,4 +348,69 @@ public void onFailure(Throwable t) {
...
+                var q = new BaseReadTsKvQuery(key.getKey(), key.getPreviousStartTs(), key.getPreviousEndTs(),
+                        key.getPreviousEndTs() - key.getPreviousStartTs(), 1, key.getAgg());
+                queries.put(q.getId(), new ReadTsKvQueryInfo(key, q, true));
+            }
+        }
+        List<ReadTsKvQuery> queryList = queries.values().stream().map(ReadTsKvQueryInfo::getQuery).collect(Collectors.toList());
+        Map<EntityData, ListenableFuture<List<ReadTsKvQueryResult>>> fetchResultMap = new LinkedHashMap<>();
+        entityDataList.forEach(entityData -> fetchResultMap.put(entityData,
+                timeseriesService.findAllByQueries(tenantId, entityData.getEntityId(), queryList)));
```
ViacheslavKlimov, line 361: This issues one `findAllByQueries` call per entity in the page. With `pageSize` capped at 100 and N keys (current + optional previous) per entity, that's up to ~100 × 2N parallel timeseries queries per request. The WS handler does the same, but the WS path runs out of `wsCallBackExecutor` and is fire-and-forget, while this is a synchronous REST call competing for `dbCallbackExecutor`. Worth thinking about whether large pages will cause timeouts or DB-pool starvation in practice — at minimum the API note on line 154 should probably mention the per-entity fan-out.

zzzeebra:
Addressed the 500-vs-400 issue in 6f7d86f. The remaining items are intentionally deferred:

- **RestClient method**: requires relocating `AggHistoryCmd`/`AggKey` from `application` to `common/data` since `rest-client` only depends on `common/data`. Out of scope for this bugfix on `lts-4.2`; the actual consumer (PE UI export) calls the endpoint via Angular HTTP. Will track as a follow-up on master.
- **Shared `AggHistoryFetcher` helper**: speculative refactor; declining for now.
- **`EntityData.aggLatest` immutability on EDQS path**: pre-existing in `DefaultTbEntityDataSubscriptionService.handleAggHistoryCmd`, not introduced by this PR. Will track separately.

application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
zzzeebra: Fixed in 6f7d86f — switched to `ThingsboardException(BAD_REQUEST_PARAMS)`.

application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
zzzeebra: Declining. Empty/null `keys` returns a valid response (entity list without `aggLatest`) and matches WS-handler behavior. Adding controller-side validation here would diverge from the WS path without functional benefit.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
zzzeebra: Confirmed pre-existing: `DefaultTbEntityDataSubscriptionService:341,352` does the same `computeIfAbsent`/`putIfAbsent` on the same `getAggLatest()`. Not introduced by this PR. Tracking as a separate issue rather than fixing under this PR.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
zzzeebra: Declining. Previous-window is opt-in per key; silently skipping a malformed prev-window matches WS-handler semantics and still returns the current window. Adding REST-only validation would diverge.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
zzzeebra: Declining for now. Mirrors WS-handler fan-out; `pageSize` is already capped at `MAX_PAGE_SIZE=100`. Will revisit if benchmarks show DB-pool starvation in practice.

zzzeebra: Cap removed in fc8ddede16 — `/find/aggHistory` is the paired endpoint for the PE entity-table export, which pages entities up to 1000 via `/find`. Capping aggregated values at 100 while the entity set goes to 1000 silently truncated `aggLatest` to a subset, breaking the export.

The fan-out cost remains real. Two mitigators: per-entity `findAllByQueries` calls are dispatched in parallel and joined via `Futures.allAsList`, not serial on `dbCallbackExecutor`. And the WS path runs the same per-page fan-out uncapped on dashboards with 1000+ entities. If real DB-pool starvation shows up, the right knob is the export-side `pageSize` (callers' choice), not a controller-level cap that desynchronises with `/find`.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
```diff
@@ -339,4 +348,69 @@ public void onFailure(Throwable t) {
         }, dbCallbackExecutor);
     }
 
+    @Override
+    public DeferredResult<PageData<EntityData>> findEntityDataAggHistoryByQuery(SecurityUser securityUser, EntityDataQuery query, AggHistoryCmd cmd) {
+        DeferredResult<PageData<EntityData>> response = new DeferredResult<>();
+        PageData<EntityData> pageData = findEntityDataByQuery(securityUser, query);
+        List<EntityData> entityDataList = pageData.getData();
```
ViacheslavKlimov, line 351: Why is there no removals from other classes? Please get rid of the duplicated code and refactor the DefaultTbEntityDataSubscriptionService to use common method.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
```diff
@@ -339,4 +348,69 @@ public void onFailure(Throwable t) {
...
+        Futures.addCallback(Futures.allAsList(fetchResultMap.values()), new FutureCallback<>() {
+            @Override
+            public void onSuccess(@Nullable List<List<ReadTsKvQueryResult>> ignored) {
+                fetchResultMap.forEach((entityData, future) -> {
+                    try {
+                        List<ReadTsKvQueryResult> queryResults = future.get();
+                        if (queryResults != null) {
+                            for (ReadTsKvQueryResult queryResult : queryResults) {
+                                ReadTsKvQueryInfo info = queries.get(queryResult.getQueryId());
+                                ComparisonTsValue comparisonTsValue = entityData.getAggLatest()
```
ViacheslavKlimov: When EDQS is enabled, `findEntityDataByQuery` returns `EntityData` built by `QueryResult.toOldEntityData()` (`common/data/.../edqs/query/QueryResult.java:36`), where `aggLatest` is set to `Collections.emptyMap()` — an immutable map. `computeIfAbsent` here (and the `putIfAbsent` below on lines 396/400) would then throw `UnsupportedOperationException`. The SQL path (`EntityDataAdapter.java:59`) uses a mutable `HashMap`, which is why the new `@DaoSqlTest` tests pass — they never exercise the EDQS path. The pre-existing WS `handleAggCmd` shares this exact pattern, but since this endpoint is new and EDQS is the default in many deployments, it's worth confirming against an EDQS-enabled setup (and possibly guarding by ensuring `aggLatest` is a mutable map before populating it).

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
```diff
@@ -339,4 +348,69 @@ public void onFailure(Throwable t) {
...
+        DeferredResult<PageData<EntityData>> response = new DeferredResult<>();
+        PageData<EntityData> pageData = findEntityDataByQuery(securityUser, query);
+        List<EntityData> entityDataList = pageData.getData();
+        if (entityDataList.isEmpty() || cmd.getKeys() == null || cmd.getKeys().isEmpty()) {
+            response.setResult(pageData);
+            return response;
+        }
+        TenantId tenantId = securityUser.getTenantId();
+        Map<Integer, ReadTsKvQueryInfo> queries = new HashMap<>();
+        for (AggKey key : cmd.getKeys()) {
```
ViacheslavKlimov: This whole method is a near-verbatim copy of the WebSocket `handleAggHistoryCmd` + `handleAggCmd` in `DefaultTbEntityDataSubscriptionService` (the query-build loop, the per-entity `findAllByQueries`, the `queryResult -> ComparisonTsValue` current/previous mapping, and the `EMPTY/EMPTY` fallback are all identical down to variable names). Two copies of this aggregation logic will drift over time — could the shared part be extracted into a helper both paths call? Relatedly, the `new BaseReadTsKvQuery(..., endTs - startTs, 1, agg)` arguments (interval = the whole window, limit = 1) are non-obvious magic values; a short comment or named factory explaining "single bucket spanning the full window → one aggregated point" would help the next reader avoid "fixing" them.

ViacheslavKlimov (commented):
## Review summary

Reviewed 4 changed files in [Fixed incorrect aggregated values in Entity Table widget data export](https://github.com/thingsboard/thingsboard/pull/15429). Left 5 comment(s) inline.

The new `/entitiesQuery/find/aggHistory` endpoint cleanly mirrors the WebSocket `AggHistoryCmd` and is well covered by tests on the SQL path. The main thing worth verifying before merge is behavior when EDQS is enabled (the `aggLatest` map is immutable on that path), plus a few quality/consistency points around duplication with the WS handler, an unbounded per-entity fan-out, and the request DTO shape.

---

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
```diff
@@ -339,4 +348,69 @@ public void onFailure(Throwable t) {
...
+            }
+            if (key.getPreviousStartTs() != null && key.getPreviousEndTs() != null && key.getPreviousEndTs() >= key.getPreviousStartTs()) {
+                var q = new BaseReadTsKvQuery(key.getKey(), key.getPreviousStartTs(), key.getPreviousEndTs(),
+                        key.getPreviousEndTs() - key.getPreviousStartTs(), 1, key.getAgg());
+                queries.put(q.getId(), new ReadTsKvQueryInfo(key, q, true));
+            }
+        }
+        List<ReadTsKvQuery> queryList = queries.values().stream().map(ReadTsKvQueryInfo::getQuery).collect(Collectors.toList());
+        Map<EntityData, ListenableFuture<List<ReadTsKvQueryResult>>> fetchResultMap = new LinkedHashMap<>();
+        entityDataList.forEach(entityData -> fetchResultMap.put(entityData,
```
ViacheslavKlimov, line 360: One `findAllByQueries` is issued per entity and fanned out for the entire page at once. Unlike `/find/keys`, which caps the page size at `MAX_PAGE_SIZE` (100) before its lookups, this endpoint applies no cap (it was intentionally removed per the commit history), so a caller requesting a large `pageSize` with several keys triggers `pageSize × keys` timeseries aggregation queries in a single burst. Given the CSV-export use case is likely to request large pages, is it worth bounding the page size (or entity count) here the way the WS path bounds entities via `maxEntitiesPerDataSubscription`?

application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
```diff
@@ -149,6 +152,32 @@ public DeferredResult<ResponseEntity> findEntityTimeseriesAndAttributesKeysByQue
...
+        AggHistoryCmd cmd = request.getAggHistoryCmd();
+        if (cmd.getEndTs() < cmd.getStartTs()) {
```
ViacheslavKlimov, line 169: The `endTs >= startTs` guard lives in the controller, but the service method is what actually consumes `startTs`/`endTs` to build the queries and is what the tests and any future caller exercise directly. Moving the validation into the service (where the values are used) would keep it from being silently bypassed if the service method ever gains another caller.

application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
```diff
@@ -149,6 +152,32 @@ public DeferredResult<ResponseEntity> findEntityTimeseriesAndAttributesKeysByQue
...
+            throw new ThingsboardException("endTs must be >= startTs", ThingsboardErrorCode.BAD_REQUEST_PARAMS);
+        }
+        resolveQuery(request.getQuery());
+        return entityQueryService.findEntityDataAggHistoryByQuery(getCurrentUser(), request.getQuery(), cmd);
+    }
+
+    @Data
+    public static class EntityDataAggHistoryRequest {
```
ViacheslavKlimov, line 177: Every sibling endpoint in this controller takes the query object directly as the `@RequestBody` and passes extras as `@RequestParam` (e.g. `/find/keys` with `timeseries`/`attributes`/`scope`). This one instead introduces a nested wrapper DTO as a controller-inner static class, which breaks that consistency and is awkward to reference from `rest-client` and to document in Swagger. Worth aligning with the sibling endpoints' shape, or at least promoting this to a top-level named DTO.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
zzzeebra: Done. Extracted the query-building and aggLatest-population blocks into AggregationQueryUtils (buildAggHistoryQueries / populateAggLatest); both DefaultTbEntityDataSubscriptionService (WS AggHistoryCmd handler) and the REST endpoint now call it. The duplicated blocks are removed from the subscription service.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
zzzeebra: Done. Logic extracted to AggregationQueryUtils and shared by the WS and REST paths. The interval=window, limit=1 query is now a named singleBucketQuery(...) with a comment explaining the single-bucket-over-full-window intent.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
zzzeebra: Not reproducible on the EDQS path. The result crosses a JSON boundary: DefaultEdqsApiService.processRequest serializes the request and deserializes the response via JacksonUtil, so toOldEntityData's Collections.emptyMap() never reaches this code. EntityData is a plain @Data (no @JsonInclude), so aggLatest serializes as {} and deserializes into a Jackson-built mutable, non-null map on the tb-core side. computeIfAbsent/putIfAbsent work fine. So no change here.

application/src/main/java/org/thingsboard/server/service/query/DefaultEntityQueryService.java
zzzeebra: Keeping it uncapped. /find/aggHistory must page in lockstep with /find (up to 1000) — a controller-level cap would silently truncate aggLatest to a subset and re-break the export this PR fixes. Per-entity calls fan out in parallel via Futures.allAsList, and the WS path already runs the same per-page fan-out uncapped. The effective throttle is the caller's pageSize; added a note about the per-entity fan-out to the endpoint API docs.

application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
zzzeebra: Keeping the validation in the controller. Request validation at the controller layer is the convention across this controller, and there is a single caller. Moving it into the service would require adding throws ThingsboardException to the EntityQueryService interface for marginal benefit.

application/src/main/java/org/thingsboard/server/controller/EntityQueryController.java
zzzeebra: Leaving as-is for now. Both query and aggHistoryCmd are complex request bodies, so a wrapper is unavoidable here — unlike the sibling endpoints whose extras are primitives passed as @RequestParam. rest-client cannot reference this type until AggHistoryCmd/AggKey move from application to common/data, which is tracked together with the RestClient method on master. Inner vs top-level is identical on the wire, so I will promote it alongside that follow-up.

PR #15436, opened 2026-04-16
Prevented internal network access via Microsoft Teams notifications

zzzeebra:
## Summary
- Disable HTTP redirect following in Microsoft Teams notification channel to prevent SSRF bypass via attacker-controlled 3xx redirects to internal hosts
- Sanitize all error paths to return a generic failure message without leaking status codes, response bodies, or target URLs
- Switch to `HttpComponentsClientHttpRequestFactory` with `HttpClientBuilder.disableRedirectHandling()` to keep HttpClient5 connection pooling while blocking redirects
- Extract shared `post()` method to ensure both Adaptive Card and Message Card paths go through the same validation and error handling

## Automated testing
- Successful POST (200) does not throw — both Message Card (`useOldApi=true`) and Adaptive Card (`useOldApi=false`) paths
- HTTP error response (e.g. 404) returns generic message, does not leak status code or response body
- Connection error returns generic message, does not leak internal IP/port
- Redirect response (302) is rejected with generic message, does not leak Location header or status code

## Manual testing
- **Info disclosure blocked** — Workflow URL pointing to internal server returning HTML body → generic error in UI, no URL/status/body leaked                                                               
  - **Redirect not followed** — Workflow URL pointing to server returning 302 to internal host → generic error in UI, internal server got no request
  
Screenshot 2026-04-16 at 09 56 45
                                                          
  - **Happy path** — Workflow URL pointing to mock server returning 200 → sent successfully, both old and new API
  
Screenshot 2026-04-16 at 10 02 02

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/notification/channels/MicrosoftTeamsNotificationChannel.java
+++ b/application/src/main/java/org/thingsboard/server/service/notification/channels/MicrosoftTeamsNotificationChannel.java
@@ -18,6 +18,7 @@ package org.thingsboard.server.service.notification.channels;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.base.Strings;
+import jakarta.annotation.PreDestroy;
 import lombok.RequiredArgsConstructor;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
@@ -46,6 +47,7 @@ import org.thingsboard.server.common.data.notification.template.MicrosoftTeamsDe
 import org.thingsboard.server.service.notification.NotificationProcessingContext;
 import org.thingsboard.server.service.security.system.SystemSecurityService;
 
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.Duration;
@@ -63,21 +65,31 @@ public class MicrosoftTeamsNotificationChannel implements NotificationChannel<Mi
 
     private final SystemSecurityService systemSecurityService;
 
-    @Setter
-    private RestTemplate restTemplate = buildRestTemplate();
+    private final CloseableHttpClient httpClient = buildHttpClient();
 
-    private static RestTemplate buildRestTemplate() {
+    @Setter
+    private RestTemplate restTemplate = new RestTemplate(new HttpComponentsClientHttpRequestFactory(httpClient));
+
+    private static CloseableHttpClient buildHttpClient() {
         // Teams webhook endpoints (*.webhook.office.com, *.logic.azure.com) respond directly without 3xx redirects.
         // Disabling redirect handling prevents SSRF bypass via attacker-controlled 3xx -> internal host.
         RequestConfig requestConfig = RequestConfig.custom()
                 .setConnectTimeout(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)
                 .setResponseTimeout(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)
                 .build();
-        CloseableHttpClient httpClient = HttpClientBuilder.create()
+        return HttpClientBuilder.create()
                 .disableRedirectHandling()
                 .setDefaultRequestConfig(requestConfig)
                 .build();
-        return new RestTemplate(new HttpComponentsClientHttpRequestFactory(httpClient));
+    }
+
+    @PreDestroy
+    public void shutdown() {
+        try {
+            httpClient.close();
+        } catch (IOException e) {
+            log.warn("Failed to close Microsoft Teams HTTP client", e);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/thingsboard/server/service/notification/channels/MicrosoftTeamsNotificationChannel.java
```diff
@@ -143,13 +159,35 @@ private void sendTeamsMessageCard(MicrosoftTeamsNotificationTargetConfig targetC
...
-        HttpEntity<String> request = new HttpEntity<>(JacksonUtil.toString(teamsMessageCard), headers);
-        restTemplate.postForEntity(webhookUri, request, String.class);
+        HttpEntity<String> request = new HttpEntity<>(JacksonUtil.toString(body), headers);
+
+        ResponseEntity<String> response;
+        try {
+            response = restTemplate.postForEntity(webhookUri, request, String.class);
+        } catch (RestClientResponseException e) {
+            log.warn("Microsoft Teams webhook returned non-success status {}", e.getStatusCode());
+            log.debug("Microsoft Teams webhook error response for [{}]", webhookUri, e);
```
ViacheslavKlimov: Microsoft Teams webhook URLs are bearer credentials — the URL itself authorizes posting to the channel (Microsoft's docs explicitly tell users to treat the webhook URL as a secret). Logging `webhookUri` here, on line 182, and on lines 187–188 writes that credential into log files / log aggregators (ELK, Splunk, etc.), which is contrary to the PR description's stated goal of "does not leak status codes, response bodies, or target URLs" — it's protected from the UI but not from logs.

Debug level doesn't really change the threat model: customers commonly enable debug for troubleshooting, and the URL doesn't need to be in the message to make it diagnosable. Consider logging `webhookUri.getHost()` (or host + path with the secret token portion stripped) instead of the full URI in all three debug statements.

application/src/main/java/org/thingsboard/server/service/notification/channels/MicrosoftTeamsNotificationChannel.java
```diff
@@ -42,21 +49,36 @@
...
+        // Teams webhook endpoints (*.webhook.office.com, *.logic.azure.com) respond directly without 3xx redirects.
+        // Disabling redirect handling prevents SSRF bypass via attacker-controlled 3xx -> internal host.
+        RequestConfig requestConfig = RequestConfig.custom()
+                .setConnectTimeout(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)
+                .setResponseTimeout(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)
+                .build();
+        CloseableHttpClient httpClient = HttpClientBuilder.create()
+                .disableRedirectHandling()
+                .setDefaultRequestConfig(requestConfig)
+                .build();
```
ViacheslavKlimov, line 83: The previous `RestTemplateBuilder`-built RestTemplate didn't own a closeable resource, but this one does — `CloseableHttpClient` holds a connection pool that should be released on bean shutdown. There's no `@PreDestroy` to call `httpClient.close()`, so the pool leaks at app shutdown, and (more relevantly) every time a test calls `setRestTemplate(mock)` the original client and its pool are orphaned. `DefaultNotificationSchedulerService` in this same package has the `@PreDestroy` pattern if you want a local reference.

application/src/test/java/org/thingsboard/server/service/notification/channels/MicrosoftTeamsNotificationChannelTest.java
```diff
@@ -0,0 +1,138 @@
...
+        Throwable thrown = catchThrowable(() -> channel.sendNotification(targetConfig, template, null));
+
+        assertThat(thrown).isInstanceOf(RuntimeException.class);
+        assertThat(thrown.getMessage())
+                .isEqualTo(GENERIC_FAILURE)
+                .doesNotContain("172.17.0.1");
+    }
+
+    @Test
+    void redirectResponseIsRejectedWithoutLeak() {
```
ViacheslavKlimov, line 125: The test name suggests it verifies that redirects are rejected, but because `RestTemplate` is mocked, the underlying `HttpClient5` config (`disableRedirectHandling()`) is never exercised — this only tests the `is2xxSuccessful()` check in `post()`. If someone refactors `buildRestTemplate()` and accidentally drops `disableRedirectHandling()`, this test would still pass while the SSRF bypass becomes possible again.

Given that disabling redirects is the *primary* mitigation the PR is built around, it's worth backing it with a real test — e.g., spin up a local `HttpServer` (or WireMock) that returns a 302, hit it with the real `RestTemplate` produced by `buildRestTemplate()`, and assert no follow-up request to the redirect target was made.

ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [Prevented internal network access via Microsoft Teams notifications](https://github.com/thingsboard/thingsboard/pull/15436). Left 3 comment(s) inline.

The SSRF mitigation looks correct: `disableRedirectHandling()` on the underlying `HttpClient5` plus the explicit `is2xxSuccessful()` check is good defense-in-depth, and routing both the Adaptive Card and Message Card paths through a single `post()` method removes the duplication that made the original easy to miss. Main concerns are around the debug-level logging of `webhookUri` (Teams webhook URLs are themselves the auth credential), missing `CloseableHttpClient` cleanup, and that the redirect test exercises the response-status check rather than the HTTP-client-level redirect-disable config.

---

application/src/main/java/org/thingsboard/server/service/notification/channels/MicrosoftTeamsNotificationChannel.java
zzzeebra: Fixed in 867b04c — all three debug logs now pass `webhookUri.getHost()` instead of the full URI, and dropped `response.getHeaders()` from the non-2xx log too since the `Location` header in a 302 response could itself point at an attacker-chosen internal target.

application/src/main/java/org/thingsboard/server/service/notification/channels/MicrosoftTeamsNotificationChannel.java
zzzeebra: Fixed in 867b04c. Pulled the `CloseableHttpClient` out into its own field, added `@PreDestroy public void shutdown()` that closes it (matching `DefaultNotificationSchedulerService`).

application/src/test/java/org/thingsboard/server/service/notification/channels/MicrosoftTeamsNotificationChannelTest.java
zzzeebra: Added in 867b04c — `disableRedirectHandlingPreventsFollowUpToInternalHost` spins up two local `com.sun.net.httpserver.HttpServer` instances (one returning 302 with `Location` pointing at the second), uses the channel's real `RestTemplate`, and asserts the redirect-target listener got 0 hits. Catches the regression you described.

PR #15439, opened 2026-04-16
Fixed alarm rule crash on duration source change

zzzeebra:
## Summary
- During REINIT, `init()` was called before `fetchArguments()`, so when a DURATION alarm rule with active tracking (`firstEventTs > 0`) tried to resolve a newly added dynamic argument, the arguments map was empty → `IllegalArgumentException: Argument 'X' is missing`
- Moved `fetchArguments()` + `state.update()` before `state.init(false)` in `initState()`

## Automated testing
- `testChangeDurationConditionFromStaticToDynamic` — create alarm with static duration → trigger condition so `firstEventTs > 0` → switch duration to dynamic with new attribute argument → verify alarm is created without "Argument is missing" error

## Manual testing
  - Created alarm rule with static duration (5s) on a device                                                                                                                                                  
  - Posted telemetry (temperature=50) to trigger condition and set firstEventTs > 0                                                                                                                           
  - Edited alarm rule: added server attribute argument durationThreshold, switched duration from static to dynamic                                                                                            
  - Saved — no Argument missing error in logs (this is where it crashed before the fix)                                                                                                                       
  - Posted server attribute durationThreshold=3000 — existing alarm preserved, no errors

### DIFF
--- a/application/src/test/java/org/thingsboard/server/cf/AlarmRulesTest.java
+++ b/application/src/test/java/org/thingsboard/server/cf/AlarmRulesTest.java
@@ -403,6 +403,54 @@ public class AlarmRulesTest extends AbstractControllerTest {
     }
 
     @Test
+    public void testChangeDurationConditionFromStaticToDynamic() throws Exception {
+        Argument temperatureArgument = new Argument();
+        temperatureArgument.setRefEntityKey(new ReferencedEntityKey("temperature", ArgumentType.TS_LATEST, null));
+        temperatureArgument.setDefaultValue("0");
+        Map<String, Argument> arguments = new HashMap<>(Map.of(
+                "temperature", temperatureArgument
+        ));
+
+        long staticDurationMs = 5000L;
+        Map<AlarmSeverity, Condition> createRules = Map.of(
+                AlarmSeverity.CRITICAL, new Condition("return temperature >= 50;", null, staticDurationMs)
+        );
+
+        CalculatedField calculatedField = createAlarmCf(deviceId, "High Temperature Alarm",
+                arguments, createRules, null);
+
+        // post telemetry to trigger condition, so that firstEventTs > 0 in AlarmRuleState
+        postTelemetry(deviceId, "{\"temperature\":50}");
+        Thread.sleep(1000);
+
+        // update CF: add attribute argument and switch duration from static to dynamic
+        AlarmCalculatedFieldConfiguration configuration =
+                (AlarmCalculatedFieldConfiguration) calculatedField.getConfiguration();
+
+        Argument durationArgument = new Argument();
+        durationArgument.setRefEntityKey(new ReferencedEntityKey("durationThreshold",
+                ArgumentType.ATTRIBUTE, AttributeScope.SERVER_SCOPE));
+        durationArgument.setDefaultValue("-1");
+        configuration.getArguments().put("durationThreshold", durationArgument);
+
+        DurationAlarmCondition durationCondition = (DurationAlarmCondition)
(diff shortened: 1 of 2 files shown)


application/src/test/java/org/thingsboard/server/cf/AlarmRulesTest.java
```diff
@@ -402,6 +402,54 @@ public void testCreateAlarm_durationCondition_defaultValue() {
...
+        Map<AlarmSeverity, Condition> createRules = Map.of(
+                AlarmSeverity.CRITICAL, new Condition("return temperature >= 50;", null, staticDurationMs)
+        );
+
+        CalculatedField calculatedField = createAlarmCf(deviceId, "High Temperature Alarm",
+                arguments, createRules, null);
+
+        // post telemetry to trigger condition, so that firstEventTs > 0 in AlarmRuleState
+        postTelemetry(deviceId, "{\"temperature\":50}");
+        Thread.sleep(1000);
```
ViacheslavKlimov: This test only exercises the bug path if the temperature telemetry is processed by the CF entity actor (so `firstEventTs > 0` is set on the rule state) before `saveCalculatedField` triggers REINIT. With a fixed `Thread.sleep(1000)`, on a slow CI box that may not happen — and if it doesn't, `ruleState.isEmpty()` returns true during REINIT, `reevalNeeded` stays false, the buggy `reeval()` call is never made, and the alarm still ends up created later via the normal eval path (when the `durationThreshold` attribute arrives, `evalDuration` sets `firstEventTs` itself and the scheduled reeval fires the alarm 3s later). The test would then pass without actually catching a regression of this bug.

Would it be possible to wait for a deterministic signal that the rule state has become non-empty (e.g., poll for a debug event with `NOT_YET_TRUE`/non-zero `conditionDuration`, or await on `getLatestAlarmResult` returning a non-error result for the static-duration phase) before invoking the config update? That way the test is guaranteed to drive REINIT through the failing branch.

ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [Fixed alarm rule crash on duration change](https://github.com/thingsboard/thingsboard/pull/15439). The fix is correct and minimal: during REINIT, `state.reset()` clears the `arguments` map but preserves `createRuleStates` (with `firstEventTs > 0`), so calling `state.init(false)` first caused `AlarmCalculatedFieldState.init()` to invoke `reeval()` → `getRequiredDurationInMs()` → `resolveValue()`, which looked up the dynamic-duration argument in an empty map and threw. Moving `init(false)` after `update(arguments, ctx)` guarantees `arguments` is populated before any rule reevaluation. The reorder is safe for all other state types (`Simple`/`Script` don't override `init`; `Propagation`/`RelatedEntitiesAggregation`/`EntityAggregation` only do work when `restored=true`; `Geofencing` is handled separately above).

Left 1 inline comment on the test.

---

application/src/test/java/org/thingsboard/server/cf/AlarmRulesTest.java
zzzeebra: Fixed in 7131e4018d. Replaced `Thread.sleep(1000)` with `await().until(() -> getDebugEvents(cfId, 1), events -> !events.isEmpty() && !events.get(0).getId().equals(latestEventId))`. The poll waits for a new debug event after `postTelemetry`, which guarantees the static-phase eval has run and set `firstEventTs > 0` before `saveCalculatedField` triggers REINIT.

Verified locally: with the fix reverted, the test now fails deterministically with `RuntimeException: Argument 'durationThreshold' is missing`; with the fix in place, it passes.

PR #15451, opened 2026-04-17
Fixed app hanging on MQTT port conflict at startup

zzzeebra:
## Summary

If the MQTT transport fails to start (for example, because `transport.mqtt.bind_port` is already in use), the JVM currently hangs instead of exiting. `MqttTransportService.init()` creates two `NioEventLoopGroup` instances before calling `bind(...).sync()`. When `bind` throws `BindException`, Spring treats the bean as failed and does not invoke the `@PreDestroy` `shutdown()` method, so the non-daemon Netty threads stay alive and prevent JVM termination. IDEA/systemd have to SIGKILL the process on timeout.

## Fix

Wrap the bootstrap/bind logic in `init()` in a `try/catch`. On any exception, close the plain and SSL server channels if they were bound, then gracefully shut down both worker and boss groups, and rethrow the original exception so Spring marks the bean as failed. `@PreDestroy` logic is unchanged (still handles the happy-path shutdown).

## Automated tests

New `MqttTransportServiceTest` with two unit tests that occupy a TCP port via `ServerSocket` and invoke `init()` via reflection:

- `whenPlainBindFails_thenInitThrowsAndReleasesNettyResources` — plain MQTT port is occupied. Verifies `BindException` is rethrown, both event loop groups enter `isShuttingDown()` and reach `isTerminated()`.
- `whenSslBindFailsAfterPlainBound_thenInitThrowsAndClosesPlainChannelAndReleasesNettyResources` — plain bind succeeds, SSL bind fails. Verifies the plain `serverChannel` is closed (`!isOpen()`) and both groups are shut down and terminated.

## Manual tests

- Normal start/stop: TB started and stopped cleanly (exit 130 on SIGINT).
- MQTT port 1883 occupied (via a helper script holding a dual-stack listener) — **before the fix**: TB printed `BindException` but JVM never exited; thread dumps showed the `nioEventLoopGroup-*` non-daemon thread alive, `DestroyJavaVM` waiting for it; the process had to be SIGKILL'd by the IDE/timeout. **After the fix**: same scenario prints `Failed to start MQTT transport, releasing resources` plus the `BindException`, and the JVM exits on its own with code 1 within seconds, no SIGKILL needed.

### DIFF
--- a/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
+++ b/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
@@ -78,9 +78,9 @@ public class MqttTransportService implements TbTransportService {
         ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.valueOf(leakDetectorLevel.toUpperCase()));
 
         log.info("Starting MQTT transport...");
-        bossGroup = new NioEventLoopGroup(bossGroupThreadCount);
-        workerGroup = new NioEventLoopGroup(workerGroupThreadCount);
         try {
+            bossGroup = new NioEventLoopGroup(bossGroupThreadCount);
+            workerGroup = new NioEventLoopGroup(workerGroupThreadCount);
             ServerBootstrap b = new ServerBootstrap();
             b.group(bossGroup, workerGroup)
                     .channel(NioServerSocketChannel.class)
@@ -96,7 +96,7 @@ public class MqttTransportService implements TbTransportService {
                         .childOption(ChannelOption.SO_KEEPALIVE, keepAlive);
                 sslServerChannel = b.bind(sslHost, sslPort).sync().channel();
             }
-        } catch (Throwable e) {
+        } catch (Exception e) {
             log.error("Failed to start MQTT transport, releasing resources", e);
             if (e instanceof InterruptedException) {
                 Thread.currentThread().interrupt();
@@ -108,16 +108,17 @@ public class MqttTransportService implements TbTransportService {
                 if (sslServerChannel != null) {
                     sslServerChannel.close().sync();
                 }
-            } catch (InterruptedException ie) {
-                Thread.currentThread().interrupt();
+            } catch (Exception suppressed) {
+                e.addSuppressed(suppressed);
             } finally {
-                workerGroup.shutdownGracefully();
-                bossGroup.shutdownGracefully();
+                if (workerGroup != null) {
+                    workerGroup.shutdownGracefully();
+                }
+                if (bossGroup != null) {
+                    bossGroup.shutdownGracefully();
+                }
             }
-            if (e instanceof Exception) {
-                throw (Exception) e;
-            }
-            throw (Error) e;
+            throw e;
         }
         log.info("Mqtt transport started!");
     }
--- a/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
+++ b/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
@@ -96,17 +96,28 @@ public class MqttTransportService implements TbTransportService {
                         .childOption(ChannelOption.SO_KEEPALIVE, keepAlive);
                 sslServerChannel = b.bind(sslHost, sslPort).sync().channel();
             }
-        } catch (Exception e) {
+        } catch (Throwable e) {
             log.error("Failed to start MQTT transport, releasing resources", e);
-            if (serverChannel != null) {
-                serverChannel.close();
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
             }
-            if (sslServerChannel != null) {
-                sslServerChannel.close();
+            try {
+                if (serverChannel != null) {
+                    serverChannel.close().sync();
+                }
+                if (sslServerChannel != null) {
+                    sslServerChannel.close().sync();
+                }
+            } catch (InterruptedException ie) {
+                Thread.currentThread().interrupt();
+            } finally {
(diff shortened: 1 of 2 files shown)


common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
```diff
@@ -80,20 +80,44 @@ public void init() throws Exception {
         log.info("Starting MQTT transport...");
         bossGroup = new NioEventLoopGroup(bossGroupThreadCount);
         workerGroup = new NioEventLoopGroup(workerGroupThreadCount);
```
ViacheslavKlimov: Minor follow-up: these two `NioEventLoopGroup` allocations are still outside the new `try` block. If `new NioEventLoopGroup(workerGroupThreadCount)` throws (e.g., resource exhaustion / `OutOfMemoryError`) after `bossGroup` has already been created, `bossGroup` is leaked — the catch below never runs and `bossGroup` keeps non-daemon threads alive, which is the exact failure mode this PR is fixing.

Moving both allocations into the `try` (and guarding `shutdownGracefully()` with null checks in the cleanup) would close that gap. Probability is very low in practice, so feel free to leave it — just flagging since the rest of the change is precisely about this category of leak.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
```diff
@@ -80,20 +80,44 @@ public void init() throws Exception {
...
+                Thread.currentThread().interrupt();
+            }
+            try {
+                if (serverChannel != null) {
+                    serverChannel.close().sync();
+                }
+                if (sslServerChannel != null) {
+                    sslServerChannel.close().sync();
+                }
+            } catch (InterruptedException ie) {
```
ViacheslavKlimov: This catch only handles `InterruptedException`, but `ChannelFuture.sync()` can rethrow the close failure cause via Netty's sneaky-throw (`PlatformDependent.throwException`). If `serverChannel.close().sync()` (or the SSL one) fails for any reason other than interruption, that exception escapes this catch, the `finally` still runs `shutdownGracefully()`, and the new exception then propagates out of `init()` instead of the original `BindException`. The original cause is still in `log.error` above, but Spring's bean-failure log will report the wrong root cause.

Safer pattern: wrap each close in its own `try { ... } catch (Throwable suppressed) { e.addSuppressed(suppressed); }` so the original `e` is preserved and the close failure is attached as a suppressed exception.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
```diff
@@ -80,20 +80,44 @@ public void init() throws Exception {
...
+                Thread.currentThread().interrupt();
+            } finally {
+                workerGroup.shutdownGracefully();
+                bossGroup.shutdownGracefully();
+            }
+            if (e instanceof Exception) {
+                throw (Exception) e;
+            }
+            throw (Error) e;
```
ViacheslavKlimov: `Throwable` is a concrete class and can technically have direct subclasses that are neither `Exception` nor `Error`, so this cast is theoretically unsafe (would throw `ClassCastException` and mask the real failure). In practice every JVM-thrown throwable is one or the other, so this works — but a couple of cleaner alternatives:

1. `} catch (Exception e) { ... throw e; }` — `Error`s would propagate naturally without the catch, which is usually what you want for `OutOfMemoryError` / `LinkageError` etc.
2. If catching `Throwable` is intentional (to also clean up on `Error`), reorder as: `if (e instanceof Error err) { throw err; } throw (Exception) e;` — same brittleness in theory but reads more explicitly, and lines up with the `instanceof InterruptedException` style used a few lines above.

ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [Fixed app hanging on MQTT port conflict at startup](https://github.com/thingsboard/thingsboard/pull/15451). Left 3 comment(s) inline.

The fix correctly addresses the reported hang: on bind failure, channels are closed and Netty event loops are shut down so non-daemon threads no longer keep the JVM alive. Tests cover both the plain-bind-fails and ssl-bind-fails-after-plain-bind paths.

A few smaller points worth considering:
- The two `NioEventLoopGroup` allocations are still outside the new `try` block, so a (rare) failure between them could re-introduce a smaller version of the same leak.
- The cleanup-block catch is `InterruptedException` only — any other exception from `close().sync()` will replace the original `BindException` propagated to Spring.
- The `throw (Error) e` cast at the tail of the catch block is technically brittle; a slightly cleaner shape exists.

---

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
zzzeebra: Moved both NioEventLoopGroup allocations inside the try and added null guards on the shutdownGracefully calls.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
zzzeebra: Switched the inner cleanup catch to attach the close failure via addSuppressed so the original BindException is preserved.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportService.java
zzzeebra: Narrowed the outer catch to Exception (BindException is the actual failure mode), removing the (Error) cast.

PR #15456, opened 2026-04-17
Incident feature for tb-monitoring

dealnat:
## Pull Request description                                                                                                                                                                                                       
                                                            
  Adds an "Incident" grouping feature to `tb-monitoring`: alerts that fire within a short window are threaded under a single Slack message whose header tracks affected services in real time. The incident auto-resolves after a   
  quiet period with a final summary.
                                                                                                                                                                                                                                    
  ### Flow                                                  
  - **New incident**: `PREFIX :rotating_light: | :red_circle: <svc> (<count>)`
  - **Update header** (every minute or on new events): adds duration, updates counts, moves recovered services to green, high latencies to yellow                                                                                   
  - **Resolution** (no alerts for `resolution_timeout_s`): `PREFIX :white_check_mark: (<duration>) | <final summary>`                                                                                                               
                                                                                                                                                                                                                                    
  ### Slack modes                                                                                                                                                                                                                   
  - `bot_token` + `channel_id` set ? Slack API mode (threaded, grouped).                                                                                                                                                            
  - Otherwise ? existing webhook mode (one message per event), fully backward compatible.                                                                                                                                           
  - If both are set, Slack API takes precedence (no duplicate messages).                                                                                                                                                            
  - If `incident.enabled=true` but the Slack bot config is missing, a warn is logged and the channel falls back to webhook mode.                                                                                                    
                                                                                                                                                                                                                                    
  ### Config (`monitoring.notifications.*`)                                                                                                                                                                                         
  | Key | Env | Default |                                                                                                                                                                                                           
  |---|---|---|                                                                                                                                                                                                                     
  | `incident.enabled` | `INCIDENT_ENABLED` | `false` |
  | `incident.resolution_timeout_s` | `INCIDENT_RESOLUTION_TIMEOUT_S` | `90` |                                                                                                                                                      
  | `incident.tag_channel` | `INCIDENT_TAG_CHANNEL` | `false` |                                                                                                                                                                     
  | `slack.bot_token` | `SLACK_BOT_TOKEN` | `""` |                                                                                                                                                                                  
  | `slack.channel_id` | `SLACK_CHANNEL_ID` | `""` |                                                                                                                                                                                
                                                                                                                                                                                                                                    
  ### Implementation notes                                  
  - Notifications expose structured `AffectedService { name, status, failureCount }` via `Notification.getAffectedServices()`, so the incident layer doesn't parse formatted alert text.                                            
  - `IncidentManager` is Slack-agnostic; the Slack binding is a thin `SlackIncidentTransport` over `SlackApiClient`.                                                                                                                
  - PE (and other downstream service-key types) can plug a friendly display name by implementing the small `ShortNameProvider` interface; `TransportInfo` already implements it.                                                    
  - `BaseMonitoringService` splits login / WS connect / WS subscribe into distinct service keys, uses `catch(Exception)`, and wraps `WsClient` in try-with-resources.                                                               
  - HTTP failure messages strip HTML response bodies (Nginx error pages etc.) so alerts stay readable.                                                                                                                              
  - Concurrency: all state mutation and transport I/O run under the manager's monitor. The Slack client has explicit 5s call/read/write timeouts, so the hold time is bounded. The Slack client is closed on `PreDestroy`.          
                                                                                                                                                                                                                                    
  ### Screenshots                                                                                                                                                                                                                   
  <img width="1420" height="492" alt="Incident thread" src="https://github.com/user-attachments/assets/f2d5ee94-f05b-4ea8-aa05-27852d9e5d9a" />                                                                                     
                                                                                                                                                                                                                                    
  Webhook mode (unchanged):
  <img width="1261" height="207" alt="Webhook" src="https://github.com/user-attachments/assets/b85d1417-2ce6-4cc7-9da9-c7487166a783" />                                                                                             
                                                                                                                                                                                                                                    
  ### Tests
  `monitoring/src/test/java/.../IncidentManagerTest.java` + `ServiceFailureNotificationTest.java` ? 15 tests covering incident lifecycle, status transitions, duration formatting, HTML body stripping, and the `ShortNameProvider` 
  dispatch.

## General checklist

## Back-End feature checklist

### DIFF
--- a/monitoring/pom.xml
+++ b/monitoring/pom.xml
@@ -80,6 +80,10 @@
             <artifactId>httpclient</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.slack.api</groupId>
+            <artifactId>slack-api-client</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.eclipse.leshan</groupId>
             <artifactId>leshan-client-cf</artifactId>
             <version>${leshan.version}</version>
@@ -117,6 +121,11 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 
--- a/monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackApiClient.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackApiClient.java
@@ -1,0 +1,125 @@
+/**
+ * Copyright © 2016-2026 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.monitoring.notification.channels.impl;
+
+import com.slack.api.Slack;
+import com.slack.api.SlackConfig;
+import com.slack.api.methods.MethodsClient;
+import com.slack.api.methods.SlackApiTextResponse;
+import com.slack.api.methods.request.chat.ChatPostMessageRequest;
+import com.slack.api.methods.request.chat.ChatUpdateRequest;
+import com.slack.api.methods.response.chat.ChatPostMessageResponse;
+import com.slack.api.methods.response.chat.ChatUpdateResponse;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class SlackApiClient {
+
+    private static final int DEFAULT_CALL_TIMEOUT_MS = 5000;
+
+    private final Slack slack;
+    private final String botToken;
+
+    public SlackApiClient(String botToken) {
+        this(botToken, DEFAULT_CALL_TIMEOUT_MS);
+    }
+
+    public SlackApiClient(String botToken, int callTimeoutMs) {
+        this.botToken = botToken;
+        SlackConfig config = new SlackConfig();
+        config.setHttpClientCallTimeoutMillis(callTimeoutMs);
+        config.setHttpClientReadTimeoutMillis(callTimeoutMs);
+        config.setHttpClientWriteTimeoutMillis(callTimeoutMs);
+        this.slack = Slack.getInstance(config);
+    }
+
+    public String postMessage(String channelId, String text) {
+        ChatPostMessageRequest request = ChatPostMessageRequest.builder()
+                .channel(channelId)
+                .text(text)
+                .build();
+        ChatPostMessageResponse response = sendRequest(request);
+        return response.getTs();
+    }
+
+    public String postThreadReply(String channelId, String threadTs, String text) {
+        ChatPostMessageRequest request = ChatPostMessageRequest.builder()
+                .channel(channelId)
+                .text(text)
+                .threadTs(threadTs)
+                .build();
+        ChatPostMessageResponse response = sendRequest(request);
+        return response.getTs();
+    }
+
+    public void close() {
+        try {
+            slack.close();
+        } catch (Exception e) {
+            log.warn("Failed to close Slack client", e);
+        }
+    }
+
+    public void updateMessage(String channelId, String ts, String text) {
+        ChatUpdateRequest request = ChatUpdateRequest.builder()
+                .channel(channelId)
+                .ts(ts)
+                .text(text)
+                .build();
+        MethodsClient client = slack.methods(botToken);
+        ChatUpdateResponse response;
+        try {
+            response = client.chatUpdate(request);
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to update Slack message: " + e.getMessage(), e);
+        }
+        checkResponse(response);
+    }
+
+    private ChatPostMessageResponse sendRequest(ChatPostMessageRequest request) {
+        MethodsClient client = slack.methods(botToken);
+        ChatPostMessageResponse response;
+        try {
+            response = client.chatPostMessage(request);
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to send Slack message: " + e.getMessage(), e);
+        }
+        checkResponse(response);
+        return response;
+    }
+
+    private void checkResponse(SlackApiTextResponse response) {
+        if (response.isOk()) {
+            return;
+        }
+        String error = response.getError();
+        if (error != null) {
+            switch (error) {
+                case "missing_scope" -> {
+                    String neededScope = response.getNeeded();
+                    error = "bot token scope '" + neededScope + "' is needed";
+                }
+                case "not_in_channel" -> error = "app needs to be added to the channel";
+            }
+        } else if (response.getWarning() != null) {
+            error = "warning: " + response.getWarning();
+        } else {
+            error = "unknown error";
+        }
+        throw new RuntimeException("Slack API error: " + error);
+    }
+
+}
--- a/monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackNotificationChannel.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackNotificationChannel.java
@@ -16,13 +16,16 @@
 package org.thingsboard.monitoring.notification.channels.impl;
 
 import jakarta.annotation.PostConstruct;
+import jakarta.annotation.PreDestroy;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.stereotype.Component;
 import org.springframework.web.client.RestTemplate;
+import org.thingsboard.monitoring.data.notification.Notification;
 import org.thingsboard.monitoring.notification.channels.NotificationChannel;
+import org.thingsboard.monitoring.notification.incident.IncidentManager;
 
 import java.time.Duration;
 import java.util.Map;
@@ -35,19 +38,73 @@ public class SlackNotificationChannel implements NotificationChannel {
     @Value("${monitoring.notifications.slack.webhook_url}")
     private String webhookUrl;
 
+    @Value("${monitoring.notifications.slack.bot_token:}")
+    private String botToken;
+
+    @Value("${monitoring.notifications.slack.channel_id:}")
+    private String channelId;
+
+    @Value("${monitoring.notifications.incident.enabled:}")
+    private boolean incidentEnabled;
+
+    @Value("${monitoring.notifications.incident.resolution_timeout_s:}")
+    private long resolutionTimeoutSeconds;
+
+    @Value("${monitoring.notifications.incident.tag_channel:}")
+    private boolean tagChannel;
+
+    @Value("${monitoring.notifications.message_prefix:}")
+    private String messagePrefix;
+
     private RestTemplate restTemplate;
+    private SlackApiClient slackApiClient;
+    private IncidentManager incidentManager;
 
     @PostConstruct
     private void init() {
-        restTemplate = new RestTemplateBuilder()
-                .setConnectTimeout(Duration.ofSeconds(5))
-                .setReadTimeout(Duration.ofSeconds(2))
-                .build();
+        boolean hasBotConfig = botToken != null && !botToken.isEmpty() && channelId != null && !channelId.isEmpty();
+        if (hasBotConfig) {
+            slackApiClient = new SlackApiClient(botToken);
+            log.info("Slack API mode enabled (channel: {})", channelId);
+            if (incidentEnabled) {
+                incidentManager = new IncidentManager(new SlackIncidentTransport(slackApiClient, channelId),
+                        resolutionTimeoutSeconds, messagePrefix, tagChannel);
+                log.info("Incident grouping enabled via Slack (resolution timeout: {}s)", resolutionTimeoutSeconds);
+            }
+        } else {
+            if (incidentEnabled) {
+                log.warn("Incident grouping is enabled but Slack bot_token/channel_id are not set; " +
+                        "falling back to plain webhook mode without incident support");
+            }
+            restTemplate = new RestTemplateBuilder()
+                    .setConnectTimeout(Duration.ofSeconds(5))
+                    .setReadTimeout(Duration.ofSeconds(2))
+                    .build();
+            log.info("Slack webhook mode enabled");
+        }
     }
 
     @Override
-    public void sendNotification(String message) {
-        restTemplate.postForObject(webhookUrl, Map.of("text", message), String.class);
+    public void sendNotification(String message, Notification notification) {
+        if (incidentManager != null && notification.isIncident()) {
+            // Pass the raw notification text: IncidentManager already puts the prefix into the
+            // incident header, so pre-prefixing the thread reply would double it up.
+            incidentManager.sendAlert(notification.getText(), notification.getAffectedServices());
+        } else if (slackApiClient != null) {
+            slackApiClient.postMessage(channelId, message);
+        } else {
+            restTemplate.postForObject(webhookUrl, Map.of("text", message), String.class);
+        }
+    }
+
+    @PreDestroy
+    private void destroy() {
+        if (incidentManager != null) {
+            incidentManager.shutdown();
+        }
+        if (slackApiClient != null) {
+            slackApiClient.close();
+        }
     }
 
 }
--- a/monitoring/src/main/java/org/thingsboard/monitoring/notification/incident/IncidentManager.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/notification/incident/IncidentManager.java
@@ -1,0 +1,279 @@
+/**
+ * Copyright © 2016-2026 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.monitoring.notification.incident;
+
+import lombok.extern.slf4j.Slf4j;
+import org.thingsboard.monitoring.data.notification.AffectedService;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Thread-safety: all public entry points and scheduled callbacks are {@code synchronized} on the
+ * manager instance. Transport I/O is performed while the monitor is held. This is safe under the
+ * assumptions that (a) the transport enforces a short per-call timeout, and (b) notification
+ * producers are single-threaded; see the Slack client (default 5s) for the Slack-based transport.
+ */
+@Slf4j
+public class IncidentManager {
+
+    private final IncidentTransport transport;
+    private final long resolutionTimeoutSeconds;
+    private final String messagePrefix;
+    private final boolean tagChannel;
+    private final ScheduledExecutorService scheduler;
+
+    private String activeIncidentThreadId;
+    private ScheduledFuture<?> resolutionTask;
+    private ScheduledFuture<?> durationUpdateTask;
+    private Instant incidentStartTime;
+    private Instant lastAlertTime;
+    private final Map<String, Integer> failingServices = new LinkedHashMap<>();
+    private final Set<String> recoveredServices = new LinkedHashSet<>();
+    private final Set<String> highLatencyServices = new LinkedHashSet<>();
+
+    public IncidentManager(IncidentTransport transport, long resolutionTimeoutSeconds,
+                           String messagePrefix, boolean tagChannel) {
+        this.transport = transport;
+        this.resolutionTimeoutSeconds = resolutionTimeoutSeconds;
+        this.messagePrefix = messagePrefix;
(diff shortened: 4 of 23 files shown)


monitoring/src/main/resources/tb-monitoring.yml
```diff
@@ -126,6 +126,17 @@ monitoring:
       enabled: '${SLACK_NOTIFICATION_CHANNEL_ENABLED:false}'
       # Slack webhook url
       webhook_url: '${SLACK_WEBHOOK_URL:}'
+      # Slack Bot OAuth token (xoxb-...) for API-based messaging with incident support, requires chat:write:bot scope
+      bot_token: '${SLACK_BOT_TOKEN:}'
+      # Slack channel ID (e.g. C01234ABCDE) - required when incident feature is enabled
+      channel_id: '${SLACK_CHANNEL_ID:}'
+      incident:
```
smatvienko-tb: please, refactor this by moving incident feature under Notifications. You can mention the the feature requires Slack bot token set, but logically the incident feature is a separate from one of many possible communication channel (currently a single Slack channel implemented)

monitoring/pom.xml
```diff
@@ -79,6 +79,10 @@
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.slack.api</groupId>
+            <artifactId>slack-api-client</artifactId>
```
smatvienko-tb, line 84: should be added to the dependency management. You can literally ask do add to the dependency management

monitoring/src/main/java/org/thingsboard/monitoring/service/BaseMonitoringService.java
```diff
@@ -117,14 +117,37 @@ public final void runChecks() {
...
+                accessToken = tbClient.logIn();
+                reporter.reportLatency(Latencies.LOG_IN, stopWatch.getTime());
+                reporter.serviceIsOk(MonitoredServiceKey.LOGIN);
+            } catch (Throwable e) {
+                reporter.serviceFailure(MonitoredServiceKey.LOGIN, e);
+                return;
+            }
+
+            WsClient wsClient;
+            try {
```
smatvienko-tb, line 133: I believe the WsClient is Closeable, so you can do try (resource) here.

monitoring/src/main/java/org/thingsboard/monitoring/service/BaseMonitoringService.java
```diff
@@ -117,14 +117,37 @@ public final void runChecks() {
...
+                wsClient = wsClientFactory.createClient(accessToken);
+                reporter.serviceIsOk(MonitoredServiceKey.WS_CONNECT);
+            } catch (Throwable e) {
+                reporter.serviceFailure(MonitoredServiceKey.WS_CONNECT, e);
+                return;
+            }
+
+            try (wsClient) {
```
smatvienko-tb: good call

monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackIncidentManager.java
```diff
@@ -0,0 +1,302 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+@Slf4j
+public class SlackIncidentManager {
```
smatvienko-tb: IncidentManager is kind of a separate responsibility. It is better to decouple generic Incident Management from NotificationChannel functionality as much as possible.

smatvienko-tb (changes requested):
Great job! The output is awesome!

Please, find out my comments to improve some boring coding stuff

smatvienko-tb (commented):
Follow-up review notes — skipping points already raised (dependency management, YAML layout under `notifications.incident.*`, try-with-resources, decoupling `IncidentManager` from `NotificationChannel`).

### Correctness

- **`SlackIncidentManager.extractServiceNames` reverse-engineers formatted alert text with regexes.** The source notifications (`ServiceFailureNotification`, `ServiceRecoveryNotification`, `HighLatencyNotification`) already carry `serviceKey` / `Collection<Latency>` as structured fields — the incident layer is turning them into strings in `getText()` and then parsing them back out. Any future tweak to wording silently breaks service-name / recovery detection (e.g. a failure message whose `errorMsg` contains `" is OK"` or `" - Failure:"` will misclassify). Please expose service name(s), failure count, and a recovery flag on the `Notification` interface (or a dedicated `IncidentEvent` produced by the reporter) and drop the regexes entirely.
- **`BaseMonitoringService.runChecks` catches `Throwable` in three places.** That swallows `Error` (OOM, StackOverflow). Use `Exception` unless there's a concrete reason to include errors.

### Configuration

- **Default drift:** `@Value("${...resolution_timeout_s:60}")` defaults to 60s in Java, `${SLACK_INCIDENT_RESOLUTION_TIMEOUT_S:90}` defaults to 90s in YAML, and the PR description says 90s. Spring resolves YAML first so the Java default is dead, but it's a trap for anyone reading the class in isolation. Drop the Spring default; YAML is authoritative.

### Concurrency

- **Blocking Slack I/O inside `synchronized` blocks.** `sendAlert`, `updateDuration`, and `resolveIncident` all hold the monitor while calling `slackApiClient.postMessage` / `chatUpdate`. A slow or hanging Slack API will stall both the notification executor thread and the scheduler thread against each other. Mutate state under the lock, release it, then perform the network call. At minimum configure an explicit request timeout on the `Slack` client (the default is long).

### Tests

- ~400 new lines with a state machine, regex parsing, scheduler, and thread safety — and zero tests. Please add unit coverage for `extractServiceNames`, `extractFailureCount`, `formatDuration`, and an incident-lifecycle test (start → add failing service → recover one → resolve) with `SlackApiClient` mocked. This feature cannot be safely refactored without them.

### Cleanup (minor)

- `SlackIncidentManager`: unused imports `ZoneId`, `DateTimeFormatter`, `Collectors`.
- `SlackIncidentManager.sendDirectMessage` is dead — no caller.
- `buildIncidentMessage` just delegates to `buildOngoingMessageText` — inline it.
- `buildOngoingMessageText` only calls `formatDuration` when `elapsed.toMinutes() >= 1`, so the `"<1m"` branch in `formatDuration` is unreachable. Remove one or the other.
- In `sendAlert`, the failure branch sets `changed = true` unconditionally per iteration — should only flip when the map actually changed (avoids no-op `chat.update` calls when the same service reports the same failure count again).
- `BaseHealthChecker.check` no longer calls `reporter.serviceIsOk(MonitoredServiceKey.GENERAL)` on success, but `BaseMonitoringService` still reports `GENERAL` on outer failures and `MonitoringReporter.serviceIsOk` still special-cases `GENERAL`. Please confirm `GENERAL` is still meaningful under the new split, or clean it up.

monitoring/pom.xml
dealnat: slack-api already present in pom.xml dependecyManagement since 2024 and bumped to 1.39 since 2024 (by you). Latest SDK release is 1.48. We use 1.39 - I believe we can bump to latest but it's up to you to decide dependency lifecycle etc

monitoring/src/main/java/org/thingsboard/monitoring/notification/incident/IncidentManager.java
```diff
@@ -0,0 +1,301 @@
...
+            if (plan.messageToPost != null) {
+                try {
+                    transport.updateIncident(threadId, plan.messageToPost);
+                } catch (Exception e) {
+                    log.error("Failed to update incident message", e);
+                }
+            }
+        }
+
+        transport.postThreadReply(threadId, message);
```
ViacheslavKlimov: There's a race between `sendAlert` and the scheduled `resolveIncident`: the lock is released at line 70, then `threadId` is captured (either from `plan.threadId` or set after `postIncident`), and the outgoing `transport.updateIncident` / `transport.postThreadReply` calls happen without the lock held. If the resolution timer fires in that window — which is realistic near the `resolution_timeout_s` boundary when a new alert arrives right as the timer is about to pop — `resolveIncident` will acquire the lock, write the `:white_check_mark:` resolution message, clear `activeIncidentThreadId`, and release the lock. Then this thread continues and either overwrites the resolution message (`updateIncident` on the old thread id) or posts a reply under a just-resolved incident.

The `NotificationService` executor is single-threaded today, but the scheduler is a separate thread, so this isn't prevented by current wiring. A couple of options: (a) re-check `activeIncidentThreadId == threadId` under the lock immediately before each transport call and skip if it changed, (b) serialize all incident I/O through the `IncidentManager`'s own scheduler so `sendAlert` and `resolveIncident` can't interleave, or (c) simply hold `stateLock` across the transport calls (OK given single-threaded producers and short-timeout HTTP). Worth at least documenting the intended concurrency model.

monitoring/src/main/java/org/thingsboard/monitoring/notification/incident/IncidentManager.java
```diff
@@ -0,0 +1,301 @@
...
+
+        synchronized (stateLock) {
+            lastAlertTime = Instant.now();
+            resetResolutionTimer();
```
ViacheslavKlimov: If `transport.postThreadReply` on line 94 throws, we never reach `lastAlertTime = Instant.now()` or `resetResolutionTimer()`. In the `newIncident` branch this is particularly sticky: `activeIncidentThreadId` is already set and the 60s duration updater is already running, but no resolution timer is ever scheduled, so the incident stays "ongoing" indefinitely until the next successful `sendAlert` — and `lastAlertTime` stays null, which would NPE `buildResolutionMessage` on line 246 if `resolveIncident` were ever triggered manually before a successful reply. Consider scheduling the resolution timer / updating `lastAlertTime` even if the reply fails (e.g. in a `finally`).

monitoring/src/main/java/org/thingsboard/monitoring/notification/incident/IncidentManager.java
```diff
@@ -0,0 +1,301 @@
...
+        }
+        if (messagePrefix != null && !messagePrefix.isEmpty()) {
+            sb.append("*").append(messagePrefix).append("*");
+        }
+        sb.append(" :rotating_light:");
+        Duration elapsed = Duration.between(incidentStartTime, Instant.now());
+        if (elapsed.toMinutes() >= 1) {
+            sb.append(" (").append(formatDuration(elapsed)).append(")");
+        }
+        sb.append(" | ");
```
ViacheslavKlimov: `" | "` is appended unconditionally, and then line 163 only appends the affected-services list when `hasAffected()`. When there are no affected services (possible during an active incident with only state-less `isIncident()=true` notifications), the header ends with a dangling ` | ` followed by nothing. Either gate the `" | "` on `hasAffected()` too, or drop the trailing separator.

monitoring/src/main/java/org/thingsboard/monitoring/notification/incident/IncidentManager.java
```diff
@@ -0,0 +1,301 @@
...
+                    if (failingServices.remove(name) != null) {
+                        recoveredServices.add(name);
+                        changed = true;
+                    }
+                }
+                case HIGH_LATENCY -> {
+                    if (highLatencyServices.add(name)) {
+                        changed = true;
+                    }
+                }
```
ViacheslavKlimov, line 128: A service that once reported `HIGH_LATENCY` stays yellow in the incident header until the incident resolves — there's no path to remove it from `highLatencyServices` when latency returns to normal (and `HighLatencyNotification.getAffectedServices` never emits a `RECOVERED` entry for latency keys). Is that intentional? If a spike was brief, the header will misrepresent the current state for the remainder of the incident.

Related: the key used here is `latency.getKey()` (e.g. `logInLatency`), while `FAILING`/`RECOVERED` services use `shortName(serviceKey)` (e.g. `Login`). Mixed naming will look a bit odd in the header (`:red_circle: Login (2), :large_yellow_circle: logInLatency`) — worth normalising.

monitoring/src/main/java/org/thingsboard/monitoring/notification/incident/IncidentManager.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    private boolean applyAffectedServices(List<AffectedService> affectedServices) {
+        boolean changed = false;
+        for (AffectedService service : affectedServices) {
+            String name = service.name();
+            switch (service.status()) {
+                case FAILING -> {
+                    Integer prev = failingServices.put(name, service.failureCount());
+                    if (prev == null || prev != service.failureCount()) {
```
ViacheslavKlimov: `prev != service.failureCount()` auto-unboxes `prev` to `int`, which is fine here because of the `prev == null` short-circuit — but worth noting this relies on the unboxing. A bit more explicit as `!Integer.valueOf(service.failureCount()).equals(prev)` or just `prev.intValue() != service.failureCount()` after the null guard.

monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackNotificationChannel.java
```diff
@@ -35,19 +38,63 @@ public class SlackNotificationChannel implements NotificationChannel {
...
     @PostConstruct
     private void init() {
-        restTemplate = new RestTemplateBuilder()
-                .setConnectTimeout(Duration.ofSeconds(5))
-                .setReadTimeout(Duration.ofSeconds(2))
-                .build();
+        if (botToken != null && !botToken.isEmpty() && channelId != null && !channelId.isEmpty()) {
+            slackApiClient = new SlackApiClient(botToken);
+            log.info("Slack API mode enabled (channel: {})", channelId);
+            if (incidentEnabled) {
```
ViacheslavKlimov, line 69: If `monitoring.notifications.incident.enabled=true` but `bot_token` / `channel_id` are blank, we silently fall into the `else` branch and run in webhook mode with incidents disabled — the only log line the operator sees is `"Slack webhook mode enabled"`. This is a common misconfiguration worth calling out: log a warn when `incidentEnabled && (botToken/channelId missing)` so the silent downgrade is visible.

monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackNotificationChannel.java
```diff
@@ -35,19 +38,63 @@ public class SlackNotificationChannel implements NotificationChannel {
...
+            log.info("Slack webhook mode enabled");
+        }
     }
 
     @Override
-    public void sendNotification(String message) {
-        restTemplate.postForObject(webhookUrl, Map.of("text", message), String.class);
+    public void sendNotification(String message, Notification notification) {
+        if (incidentManager != null && notification.isIncident()) {
+            incidentManager.sendAlert(message, notification.getAffectedServices());
```
ViacheslavKlimov: `incidentManager.sendAlert(message, ...)` passes the already-prefixed `message` from `NotificationService` (it prepends `message_prefix` before calling the channel). The incident header built in `buildOngoingMessageText` also wraps `messagePrefix` into the header. Net effect: every thread reply starts with the prefix (e.g. `tbqa CoAP - Failure: …`) and the header above it also starts with `*tbqa*`. Not broken, just a slight redundancy worth noting — you could pass `notification.getText()` directly to `IncidentManager` and let it own prefixing end-to-end.

ViacheslavKlimov (commented):
## Review summary

Reviewed 21 changed files in [Incident feature for tb-monitoring](https://github.com/thingsboard/thingsboard/pull/15456). Nice refactor pulling the incident grouping behind a transport interface and adding unit tests for the lifecycle — the `IncidentManager`/`IncidentTransport` split makes this a lot easier to reason about than a Slack-specific implementation would have been. Left some inline comments, mostly around a race between `sendAlert` and the scheduled `resolveIncident`, a couple of edge cases in state handling, and a few small readability items.

## Additional findings

- **monitoring/src/main/java/org/thingsboard/monitoring/service/BaseMonitoringService.java:165** — With `reporter.serviceIsOk(MonitoredServiceKey.GENERAL)` removed from `BaseHealthChecker.check`, the `GENERAL` failure counter in `MonitoringReporter` is no longer reset anywhere, but `MonitoredServiceKey.GENERAL` is still used here as a fallback bucket for the outer `catch (Throwable error)`. The counter now only grows across runs, so after a single transient unexpected error the counter sits at 1+ forever and the usual `repeated_failure_notification` cadence kicks in for any later `GENERAL` error. Also, `GENERAL` can never produce a recovery notification anymore. Might be worth clearing the `GENERAL` counter at the top or bottom of `runChecks`, or dropping `GENERAL` entirely in favour of a per-error-class bucket.
- **monitoring/src/main/java/org/thingsboard/monitoring/data/notification/ServiceFailureNotification.java:57** — `shortName` is a package-private static on a notification data class, called from `ServiceRecoveryNotification`. Feels out of place here — a small `MonitoredServiceNames` utility (or a method on `TransportInfo`/`MonitoredServiceKey`) would read more cleanly.
- **monitoring/src/test/java/org/thingsboard/monitoring/notification/incident/IncidentManagerTest.java:140-142** — mixing fully qualified `java.util.List`/`java.util.ArrayList` while `java.util.List` is already imported at the top is a bit inconsistent. Minor.

---

monitoring/src/main/java/org/thingsboard/monitoring/service/BaseMonitoringService.java
```diff
@@ -117,14 +117,32 @@ public final void runChecks() {
...
-                reporter.reportLatency(Latencies.WS_SUBSCRIBE, stopWatch.getTime());
+                accessToken = tbClient.logIn();
+                reporter.reportLatency(Latencies.LOG_IN, stopWatch.getTime());
+                reporter.serviceIsOk(MonitoredServiceKey.LOGIN);
+            } catch (Exception e) {
+                reporter.serviceFailure(MonitoredServiceKey.LOGIN, e);
+                return;
+            }
+
+            try (WsClient wsClient = createWsClient(accessToken)) {
```
ViacheslavKlimov: Returning `null` from `createWsClient` and then entering a try-with-resources block specifically to `if (wsClient == null) return;` reads awkwardly. Java's try-with-resources is null-safe on close, so this is technically correct, but the control flow is hard to parse — the reader has to confirm that skipping the body on null won't leak a resource. Easier to read if `createWsClient` just throws (wrapping the reporter call inside) and the try-with-resources sits under a normal `try { … } catch (Exception e) { return; }`.

monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackApiClient.java
```diff
@@ -0,0 +1,113 @@
...
+        if (error != null) {
+            switch (error) {
+                case "missing_scope" -> {
+                    String neededScope = response.getNeeded();
+                    error = "bot token scope '" + neededScope + "' is needed";
+                }
+                case "not_in_channel" -> error = "app needs to be added to the channel";
+            }
+        }
+        throw new RuntimeException("Slack API error: " + error);
```
ViacheslavKlimov, line 122: If `response.getError()` is null (possible when `isOk()` is false but the API doesn't populate an error string — e.g. HTTP-layer success with malformed payload) this throws `"Slack API error: null"`. Minor, but a fallback like `error != null ? error : "unknown error (" + response.getWarning() + ")"` would produce something more diagnosable.

monitoring/src/main/java/org/thingsboard/monitoring/notification/channels/impl/SlackApiClient.java
```diff
@@ -0,0 +1,113 @@
...
+        this(botToken, DEFAULT_CALL_TIMEOUT_MS);
+    }
+
+    public SlackApiClient(String botToken, int callTimeoutMs) {
+        this.botToken = botToken;
+        SlackConfig config = new SlackConfig();
+        config.setHttpClientCallTimeoutMillis(callTimeoutMs);
+        config.setHttpClientReadTimeoutMillis(callTimeoutMs);
+        config.setHttpClientWriteTimeoutMillis(callTimeoutMs);
+        this.slack = Slack.getInstance(config);
```
ViacheslavKlimov, line 46: `Slack.getInstance(config)` creates a fresh `Slack` backed by an OkHttp client. There's no `close()` / `shutdown()` hook on this class, so the underlying thread pool and connection pool aren't released on Spring shutdown. It's a monitoring service so probably fine, but consider adding a small `close()` here and calling it from `SlackNotificationChannel.destroy()`.

smatvienko-tb:
## Follow-up suggestion: make the incident affected-service short name pluggable via a small interface

**Context** — this PR introduces `ServiceFailureNotification.shortName(Object serviceKey)` (and `ServiceRecoveryNotification` calls it too), which currently has a single special case:

```java
static String shortName(Object serviceKey) {
    if (serviceKey instanceof TransportInfo info) {
        return info.getShortName();
    }
    return serviceKey.toString();
}
```

The PE fork of this module has an additional service-key type — `IntegrationInfo` (used by `IntegrationsMonitoringService` for PE-only integrations monitoring). Its `toString()` currently produces noisy output like `*MQTT integration* (https://…)`, which looks bad inside the new incident summary line (`🔴 *MQTT integration* (https://…) (3)`).

Rather than accumulating another `instanceof IntegrationInfo` branch in PE (and more for any future service-key types), it would be cleaner to extract a tiny interface at the CE layer. **Zero behavior change for CE**; PE then just has `IntegrationInfo implements ShortNameProvider` and gets the nice output for free.

**To another Claude picking this up:** please implement this as a single commit on top of the current PR branch. Do not broaden scope — only the changes below.

### Changes

1. **Create** `monitoring/src/main/java/org/thingsboard/monitoring/data/notification/ShortNameProvider.java`:
   ```java
   package org.thingsboard.monitoring.data.notification;

   public interface ShortNameProvider {
       String getShortName();
   }
   ```

2. **Edit** `monitoring/src/main/java/org/thingsboard/monitoring/config/transport/TransportInfo.java` — declare it implements the interface. `getShortName()` already exists; no body changes:
   ```java
   public class TransportInfo implements ShortNameProvider {
       // ... existing code unchanged ...
   }
   ```
   Add the import for `org.thingsboard.monitoring.data.notification.ShortNameProvider`.

3. **Edit** `monitoring/src/main/java/org/thingsboard/monitoring/data/notification/ServiceFailureNotification.java` — replace the `shortName` method with a dispatch on the interface:
   ```java
   static String shortName(Object serviceKey) {
       if (serviceKey instanceof ShortNameProvider provider) {
           return provider.getShortName();
       }
       return serviceKey.toString();
   }
   ```
   Remove the now-unused `TransportInfo` import if present.

### Tests

Add a test to `monitoring/src/test/java/org/thingsboard/monitoring/data/notification/ServiceFailureNotificationTest.java` covering the interface dispatch and the fallback. Do not touch the existing `stripResponseBody*` tests.

```java
@Test
void shortNameUsesShortNameProviderWhenAvailable() {
    ShortNameProvider provider = () -> "MQTT";
    assertThat(ServiceFailureNotification.shortName(provider)).isEqualTo("MQTT");
}

@Test
void shortNameFallsBackToToStringForOtherKeys() {
    Object key = new Object() {
        @Override public String toString() { return "LOGIN"; }
    };
    assertThat(ServiceFailureNotification.shortName(key)).isEqualTo("LOGIN");
}
```

### Scope guardrails — please do NOT do any of these

- Do not change `TransportInfo.getShortName()` behavior, signature, or `toString()`.
- Do not touch `MonitoredServiceKey` (enum) — it should keep falling through to `toString()`.
- Do not rename `shortName()` or widen its visibility beyond what this PR already chose.
- Do not refactor `ServiceRecoveryNotification` — it calls `ServiceFailureNotification.shortName(...)` and will pick up the change automatically.
- Do not introduce a separate default method or abstract base class — a bare interface is sufficient.

### Verification

- `mvn -pl monitoring test` passes.
- All existing assertions in `ServiceFailureNotificationTest` still pass unchanged.
- Grep confirms only one `instanceof TransportInfo` / `instanceof ShortNameProvider` site in the `data/notification` package after the change.

PR #15477, opened 2026-04-21
Fixed SMPP SMS encoding for non-ASCII characters

zzzeebra:
## Pull Request description

`SmppSmsSender.sendSms(...)` was calling the single-argument `setShortMessage(String)` overload, so the SMPP library encoded the message bytes with a JVM default ASCII-compatible charset instead of the Coding Scheme selected in UI. All non-ASCII characters (Cyrillic, CJK, Hebrew, etc.) were silently replaced with `?` before the PDU reached the SMSC — the `data_coding` field was set correctly, but the payload bytes did not match it.

The fix switches to `setShortMessage(String value, String encoding)` and resolves the encoding from the configured `coding_scheme`, covering all 13 DCS values exposed in the UI. The `data_coding` field in the `SubmitSM` PDU and the encoding used for the byte conversion are now driven by the same value.

### Tests

Added 13 unit tests to `SmppSmsSenderTest` — one per DCS value supported by the UI — verifying that the bytes written into the `SubmitSM` PDU match `String.getBytes(<expected charset>)` and that the `data_coding` field matches the selected DCS.

### Manual testing                                                                                                                                  
                                                     
Tested locally against hjaf23/smppsim Docker container (SMPP on :2775). For each case: set Coding Scheme in UI, sent a test SMS, verified data_coding and short_message bytes in the captured SUBMIT_SM PDU.  
Case 4 is the scenario that reproduced the original bug — before the fix, non-ASCII characters under UCS-2 were silently replaced with ? bytes.
Screenshot 2026-04-21 at 19 27 40

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
+++ b/application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
@@ -85,8 +85,9 @@ public class SmppSmsSender extends AbstractSmsSender {
                 request.setSourceAddr(new Address(config.getSourceTon(), config.getSourceNpi(), config.getSourceAddress()));
             }
             request.setDestAddr(new Address(config.getDestinationTon(), config.getDestinationNpi(), prepareNumber(numberTo)));
-            request.setShortMessage(message);
-            request.setDataCoding(Optional.ofNullable(config.getCodingScheme()).orElse((byte) 0));
+            byte dataCoding = Optional.ofNullable(config.getCodingScheme()).orElse((byte) 0);
+            request.setShortMessage(message, charsetFor(dataCoding));
+            request.setDataCoding(dataCoding);
             request.setReplaceIfPresentFlag((byte) 0);
             request.setEsmClass((byte) 0);
             request.setProtocolId((byte) 0);
@@ -102,6 +103,27 @@ public class SmppSmsSender extends AbstractSmsSender {
         }
 
         return countMessageSegments(message);
+    }
+
+    private static String charsetFor(byte dataCoding) {
+        return switch (dataCoding) {
+            case 0 -> "US-ASCII"; // SMSC Default (GSM 7-bit subset)
+            case 1 -> "US-ASCII"; // IA5
+            // DCS 2, 4 are "Octet unspecified" (binary) — ISO-8859-1 is bijective for bytes 0-255, so it passes bytes through unchanged.
+            case 2 -> "ISO-8859-1"; // Octet unspecified (binary pass-through)
+            case 3 -> "ISO-8859-1"; // Latin 1
+            case 4 -> "ISO-8859-1"; // Octet unspecified (binary pass-through)
+            case 5 -> "Shift_JIS"; // JIS
+            case 6 -> "ISO-8859-5"; // Cyrillic
+            case 7 -> "ISO-8859-8"; // Latin/Hebrew
+            case 8 -> "UTF-16BE"; // UCS-2
+            // FIXME: Pictogram encoding not natively supported, falling back to ISO-8859-1.
+            case 9 -> "ISO-8859-1"; // Pictogram
+            case 10 -> "ISO-2022-JP"; // Music Codes
+            case 13 -> "JIS_X0212-1990"; // Extended Kanji JIS
+            case 14 -> "EUC-KR"; // KS C 5601
+            default -> throw new UnsupportedOperationException("Unsupported data_coding: " + dataCoding);
+        };
     }
 
     private synchronized void checkSmppSession() {
--- a/application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
+++ b/application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
@@ -33,6 +33,7 @@ import org.smpp.pdu.PDUException;
 import org.smpp.pdu.SubmitSM;
 import org.smpp.pdu.SubmitSMResp;
 import org.thingsboard.rule.engine.api.sms.exception.SmsException;
+import org.thingsboard.rule.engine.api.sms.exception.SmsSendException;
 import org.thingsboard.server.common.data.StringUtils;
 import org.thingsboard.server.common.data.sms.config.SmppSmsProviderConfiguration;
 import org.thingsboard.server.service.sms.AbstractSmsSender;
@@ -99,7 +100,7 @@ public class SmppSmsSender extends AbstractSmsSender {
 
             log.debug("SMPP submit command status: {}", response.getCommandStatus());
         } catch (Exception e) {
-            throw new RuntimeException(e);
+            throw new SmsSendException("Failed to send SMS message - " + ExceptionUtils.getRootCauseMessage(e), e);
         }
 
         return countMessageSegments(message);
@@ -107,8 +108,8 @@ public class SmppSmsSender extends AbstractSmsSender {
 
     private static String charsetFor(byte dataCoding) {
         return switch (dataCoding) {
-            case 0 -> "US-ASCII"; // SMSC Default (GSM 7-bit subset)
-            case 1 -> "US-ASCII"; // IA5
+            case 0 -> Data.ENC_GSM7BIT; // SMSC Default Alphabet (GSM 7-bit) — opensmpp-charset SPI charset "X-Gsm7Bit"
+            case 1 -> "US-ASCII"; // IA5 (CCITT T.50) / ASCII
             // DCS 2, 4 are "Octet unspecified" (binary) — ISO-8859-1 is bijective for bytes 0-255, so it passes bytes through unchanged.
             case 2 -> "ISO-8859-1"; // Octet unspecified (binary pass-through)
             case 3 -> "ISO-8859-1"; // Latin 1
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
```diff
@@ -104,6 +105,27 @@ public int sendSms(String numberTo, String message) throws SmsException {
         return countMessageSegments(message);
     }
 
+    private static String charsetFor(byte dataCoding) {
+        return switch (dataCoding) {
+            case 0 -> "US-ASCII"; // SMSC Default (GSM 7-bit subset)
```
ViacheslavKlimov: DCS `0` is the SMSC default alphabet (GSM 7-bit) and it's also the configuration default, so this is the most common path. Mapping it to `US-ASCII` is a behavioral change from the old code: the single-arg `setShortMessage(message)` defaulted to `Data.ENC_GSM7BIT` (`X-Gsm7Bit`), and `opensmpp-charset` registers that charset on the classpath (transitive dep of `opensmpp-core`, SPI-registered), so GSM-alphabet characters were previously encoded correctly.

With `US-ASCII`, characters that GSM 7-bit supports — `é à ñ è ü £ € §` etc. — are now silently replaced with `?`. Worse, several plain-ASCII bytes don't share code points with the GSM 7-bit alphabet (`$`→`¤`, `@`→`¡`, `_`→`§`, `[ \ ] ^ { | } ~`), so an SMSC that reads `data_coding=0` as GSM 7-bit code points will render those differently than intended.

Was downgrading DCS 0 (and 1) to ASCII intentional? Using `Data.ENC_GSM7BIT` for `case 0` would preserve the prior, more correct behavior for the default scheme. A plain-ASCII manual test wouldn't surface this since A–Z/a–z/digits coincide in both.

application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
```diff
@@ -104,6 +105,27 @@ public int sendSms(String numberTo, String message) throws SmsException {
...
+            case 5 -> "Shift_JIS"; // JIS
+            case 6 -> "ISO-8859-5"; // Cyrillic
+            case 7 -> "ISO-8859-8"; // Latin/Hebrew
+            case 8 -> "UTF-16BE"; // UCS-2
+            // FIXME: Pictogram encoding not natively supported, falling back to ISO-8859-1.
+            case 9 -> "ISO-8859-1"; // Pictogram
+            case 10 -> "ISO-2022-JP"; // Music Codes
+            case 13 -> "JIS_X0212-1990"; // Extended Kanji JIS
+            case 14 -> "EUC-KR"; // KS C 5601
+            default -> throw new UnsupportedOperationException("Unsupported data_coding: " + dataCoding);
```
ViacheslavKlimov: This is a behavioral regression for out-of-range values. Previously an unrecognized coding scheme just set `data_coding` and sent the message; now `sendSms` throws `UnsupportedOperationException`, which the `catch` below wraps into a `RuntimeException` and fails the send entirely.

The `@Schema(allowableValues = {"0-10","13-14"})` on `codingScheme` is documentation only — it isn't enforced — so a value like `11`, `12`, or `15` can still be persisted through the REST API (or already exist in a saved config). Since SMS backs 2FA and notifications, hard-failing every send is a blunt fallback. Consider defaulting unknown values to a safe charset (e.g. `ISO-8859-1`) or validating `codingScheme` at config-save time so the failure surfaces when the admin sets it, not when an SMS goes out.

application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
```diff
@@ -85,8 +85,9 @@ public int sendSms(String numberTo, String message) throws SmsException {
                 request.setSourceAddr(new Address(config.getSourceTon(), config.getSourceNpi(), config.getSourceAddress()));
             }
             request.setDestAddr(new Address(config.getDestinationTon(), config.getDestinationNpi(), prepareNumber(numberTo)));
-            request.setShortMessage(message);
-            request.setDataCoding(Optional.ofNullable(config.getCodingScheme()).orElse((byte) 0));
+            byte dataCoding = Optional.ofNullable(config.getCodingScheme()).orElse((byte) 0);
+            request.setShortMessage(message, charsetFor(dataCoding));
```
ViacheslavKlimov, line 90: Now that real multibyte encodings are in play, note that `setShortMessage` validates length against `Data.SM_MSG_LEN` (254) measured in *encoded octets*. For UCS-2 (`UTF-16BE`) that effectively halves the limit to ~127 characters, and an over-length message throws `WrongLengthOfStringException`, which the `catch (Exception e) { throw new RuntimeException(e); }` below turns into an opaque runtime error rather than an `SmsException`. The single-PDU sender never split long messages, but it's worth confirming long non-ASCII messages produce a clear error (or are segmented) instead of a generic `RuntimeException`.

application/src/test/java/org/thingsboard/server/service/sms/smpp/SmppSmsSenderTest.java
```diff
@@ -91,6 +95,88 @@ public void testSendSms() throws Exception {
...
+        setDefaultSmppConfig();
+        smppConfig.setCodingScheme(dataCoding);
+
+        smppSmsSender.sendSms("123545", message);
+
+        ArgumentCaptor<SubmitSM> captor = ArgumentCaptor.forClass(SubmitSM.class);
+        verify(smppSession).submit(captor.capture());
+        SubmitSM sent = captor.getValue();
+
+        assertArrayEquals(message.getBytes(expectedCharset), sent.getShortMessageData().getBuffer());
```
ViacheslavKlimov: This encodes the message with `expectedCharset` and compares it to what production emitted — but production also just calls `message.getBytes(charsetFor(dcs))`, so both sides compute the same thing. It does verify the DCS→charset *mapping* (valuable), yet it can't catch a charset that silently substitutes. For DCS 13, `"Hi".getBytes("JIS_X0212-1990")` produces `?` substitute bytes on both sides, so the test passes without exercising any actual Kanji round-trip. For at least the key cases (e.g. UCS-2) consider asserting against a hardcoded expected byte sequence to lock in the real wire bytes.

ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [Fixed SMPP SMS encoding for non-ASCII characters](https://github.com/thingsboard/thingsboard/pull/15477). The core fix — pairing `setShortMessage(value, encoding)` with the charset implied by `data_coding` so the payload bytes match the DCS field — is correct and well covered for the multibyte/8-bit schemes that were the subject of the bug. Left 4 comments inline.

The main concern is the `DCS 0` (default coding scheme) case, which is changed from the library's `X-Gsm7Bit` charset to `US-ASCII` and can regress characters that previously encoded correctly. The rest are about the hard-failure fallback for unknown coding schemes, the encoded-length limit now hitting multibyte messages sooner, and a test assertion that mirrors the production code.

---

application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
zzzeebra: Fixed. `case 0` now maps to `Data.ENC_GSM7BIT` ("X-Gsm7Bit"), restoring the prior GSM 7-bit encoding for the default scheme. Left `case 1` on US-ASCII: DCS 1 is IA5 (CCITT T.50)/ASCII per 3GPP 23.038, so ASCII is the correct charset there, not GSM 7-bit.

application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
zzzeebra: Fixed. Unknown `data_coding` now falls back to ISO-8859-1 (byte-transparent, no substitution) with a warn log instead of throwing, so SMS for 2FA/notifications no longer fails on out-of-range values that the unenforced `@Schema` lets through.

application/src/main/java/org/thingsboard/server/service/sms/smpp/SmppSmsSender.java
zzzeebra: Confirmed octet-based: `setShortMessage(value, encoding)` runs `checkString` against the encoded byte length vs `SM_MSG_LEN` (254), so UCS-2 effectively halves to ~127 chars. The opaque `RuntimeException` is fixed — the catch now throws `SmsSendException("Failed to send SMS message - " + rootCause, e)` like the Twilio/AWS senders, so the `WrongLengthOfStringException` (min/max/actual) surfaces clearly. Segmentation is out of scope for the single-PDU sender.

application/src/test/java/org/thingsboard/server/service/sms/smpp/SmppSmsSenderTest.java
zzzeebra: Fixed. Added `assertEncodedBytes` asserting exact wire bytes for UCS-2 ("Привіт") and Cyrillic ISO-8859-5, independent of the production charset. Strengthened the DCS 0 test to "café" → `63 61 66 05`, which fails under a US-ASCII fallback (é → `3F`), locking the GSM 7-bit behavior. Mapping-only cases (incl. substituting charsets like JIS) stay on the original helper.

PR #15498, opened 2026-04-23
Fixed JavaScript modules with special characters corrupted when uploading files

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist


vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Fixed JavaScript modules with special characters corrupted when uploading files](https://github.com/thingsboard/thingsboard/pull/15498). Left 2 comments inline.

Context from the reported bug: existing JS module data on the server contains raw Latin-1 bytes (German-locale Windows uploads with `ü`/`ö`/`ä`); the old `base64toString` calls `decodeURIComponent` which throws `URIError: URI malformed` on those bytes, the unhandled exception aborts form initialization, and the editor shows empty content.

The `base64toString` rewrite directly addresses that — TextDecoder + UTF-8/Latin-1 fallback decodes existing Latin-1 server data without throwing, so the editor renders the content again. That part is solid.

The `file-input.component.ts` change is more concerning. Switching the text-read path from default UTF-8 to `'ISO-8859-1'` is a global behavior change for every `tb-file-input` consumer (CSV/JSON imports, JS module editor, AI/SMS service-account JSON, rule-node credentials, etc.) and corrupts UTF-8 file uploads — the dominant case. Details inline.

## Additional findings

- **ui-ngx/src/app/core/utils.ts:233** — `base64toObj` still uses the old `decodeURIComponent` pattern that throws on invalid UTF-8. With `base64toString` now silently falling back to ISO-8859-1, the two functions behave inconsistently for the same encoded byte sequences. Worth rewriting `base64toObj` as `JSON.parse(base64toString(b64Encoded))` for symmetry, so any other place that reads a JSON payload from base64 gets the same robustness.
- **ui-ngx/src/app/modules/home/pages/admin/resource/js-resource.component.ts:114** — `formValue.data = stringToBase64((formValue as any).content)` is the JS module save path. Combined with the new ISO-8859-1 reading in `file-input.component.ts`, a UTF-8 module containing non-ASCII characters is read as a Latin-1-decoded JS string and then UTF-8-encoded a second time by `stringToBase64`, so the bytes stored on the server are doubly encoded. See the inline comment for a full walk-through.

ui-ngx/src/app/shared/components/file-input.component.ts
```diff
@@ -235,7 +235,7 @@ export class FileInputComponent extends PageComponent implements AfterViewInit,
       if (this.readAsBinary) {
         reader.readAsBinaryString(file.file);
       } else {
-        reader.readAsText(file.file);
+        reader.readAsText(file.file, 'ISO-8859-1');
```
vvlladd28: Forcing `'ISO-8859-1'` here is a global behavior change for every `tb-file-input` consumer that doesn't set `readAsBinary` — CSV/JSON imports, the JS module editor, the AI/SMS service-account JSON inputs, the rule-node credential file inputs, etc. — not just the resource upload that motivated this fix. ISO-8859-1 maps each byte 1:1 to a Unicode code point, which is fine for binary-style consumers, but it actively corrupts UTF-8 input files (the dominant encoding for source files).

Concrete walk-through for a UTF-8 module containing `ü` (bytes `0xC3 0xBC`):

1. `readAsText(file, 'ISO-8859-1')` returns the two-character string `"Ã¼"` (`U+00C3`, `U+00BC`).
2. `uploadContentFromFile` patches that into the `content` form control, so the editor immediately shows `"Ã¼"` instead of `"ü"`.
3. On save, `js-resource.component.ts:114` calls `stringToBase64("Ã¼")`. That function `encodeURIComponent`s the input, which UTF-8-encodes each char a second time (`Ã` → `0xC3 0x83`, `¼` → `0xC2 0xBC`), and base64s the resulting four bytes — so the data sent to the server is doubly UTF-8-encoded.
4. On reload, `base64toString` decodes those four bytes as UTF-8 and returns `"Ã¼"`, so the mojibake round-trips.

The same regression hits `import-dialog.component.ts:90` (`JSON.parse(content)` against a Latin-1-decoded UTF-8 file) and `import-dialog-csv.component.html` for any UTF-8 import file with non-ASCII characters. The previous default (browser-native UTF-8) handled these correctly.

It's also worth noting that the originally reported bug — modules already on the server with Latin-1 bytes that crash the editor — is fully fixed by the `base64toString` change alone; the `readAsText` change here is only about NEW uploads, and it trades the Latin-1-upload case for the UTF-8-upload case.

A cleaner approach is to read the file once as an `ArrayBuffer` and let a UTF-8/Latin-1 fallback decoder pick the right encoding (mirroring what the new `base64toString` does):

```ts
file.file.arrayBuffer().then(buf => {
  const bytes = new Uint8Array(buf);
  let text: string;
  try {
    text = new TextDecoder('utf-8', { fatal: true }).decode(bytes);
  } catch {
    text = new TextDecoder('iso-8859-1').decode(bytes);
  }
  // ...feed `text` into the existing code path
});
```

That handles both the Latin-1 German-locale case and the UTF-8 case correctly, and keeps the change scoped to the actual problem instead of forcing every text consumer onto Latin-1.

ui-ngx/src/app/core/utils.ts
```diff
@@ -209,8 +209,20 @@ export function objToBase64(obj: any): string {
     }));
 }
 
+const textDecoderUtf8 = new TextDecoder('utf-8', { fatal: true });
+const textDecoderLatin1 = new TextDecoder('iso-8859-1');
```
vvlladd28: Minor: the labels accepted by `TextDecoder` are case-insensitive, but the rest of the codebase consistently uses uppercase encoding labels (`'ISO-8859-1'` in the file-input change in this same PR). Worth aligning these for grep-ability — e.g., `new TextDecoder('iso-8859-1')` here vs. `reader.readAsText(file.file, 'ISO-8859-1')` in `file-input.component.ts:238`.

ui-ngx/src/app/shared/components/file-input.component.ts
```diff
@@ -239,7 +239,18 @@ export class FileInputComponent extends PageComponent implements AfterViewInit,
       if (this.readAsBinary) {
         reader.readAsBinaryString(file.file);
       } else {
-        reader.readAsText(file.file);
+        file.file.arrayBuffer().then(buf => {
+          const bytes = new Uint8Array(buf);
+          let text: string;
+          try {
+            text = new TextDecoder('utf-8', {fatal: true}).decode(bytes);
```
vvlladd28: This UTF-8-with-ISO-8859-1-fallback decode is now a near-verbatim copy of the new logic in `base64toString` (`core/utils.ts`). Since both decode a `Uint8Array` the same way, it'd be worth extracting a small shared helper (e.g. `bytesToString(bytes: Uint8Array): string` in `utils.ts`) and calling it from both places — otherwise any future change to the decoding strategy has to be remembered in two spots.

ui-ngx/src/app/shared/components/file-input.component.ts
```diff
@@ -239,7 +239,18 @@ export class FileInputComponent extends PageComponent implements AfterViewInit,
...
+          } catch {
+            text = new TextDecoder('iso-8859-1').decode(bytes);
+          }
+           Object.defineProperty(reader, 'readyState', {value: FileReader.DONE, configurable: true});
```
vvlladd28: This branch overwrites the reader's read-only `readyState`/`result` via `Object.defineProperty` and then manually invokes `onload` with a synthesized `ProgressEvent` — the `FileReader` isn't actually reading anything here, it's just borrowing its callbacks, which is pretty fragile. Cleaner would be to resolve the promise directly from the `arrayBuffer().then(...)` branch (build the `{fileContent, fileName, mediaType}` result inline, mirroring the `onload` body) and keep the `FileReader` purely for the `readAsBinaryString` path. Extracting the result-building into a shared `buildResult(content)` used by both paths would drop the property-spoofing entirely. Minor: this line also has a stray leading space, and the two `defineProperty` calls use manual column alignment that the formatter/linter will likely undo.

ui-ngx/src/app/shared/components/file-input.component.ts
```diff
@@ -239,7 +239,18 @@ export class FileInputComponent extends PageComponent implements AfterViewInit,
...
+          Object.defineProperty(reader, 'result',    {value: text,            configurable: true});
+          reader.onload(new ProgressEvent('load') as ProgressEvent<FileReader>);
+        }).catch(() => reader.onerror(new ProgressEvent('error') as ProgressEvent<FileReader>));
```
vvlladd28: The single `.catch` here covers both an `arrayBuffer()` read failure and any error thrown synchronously inside the `onload` handler — including a throw from `contentConvertFunction`. So a content-conversion failure gets re-reported as a read error and resolves to a null/empty result rather than surfacing. Was that intended? If you keep the manual dispatch, it'd be safer to scope the `.catch` to the read itself and let decoding/handler errors propagate (or handle them explicitly).

vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Fixed JavaScript modules with special characters corrupted when uploading files](https://github.com/thingsboard/thingsboard/pull/15498). Left 3 comment(s) inline.

The `base64toString` rewrite in `utils.ts` is a clean, correct improvement (decode to bytes, then UTF-8 with an ISO-8859-1 fallback) and reusing the module-level `TextDecoder` instances is safe since non-streaming `decode()` resets per call. The concerns are all in `file-input.component.ts`, where the non-binary branch fabricates `FileReader` internals via `Object.defineProperty` and manually invokes the handlers — it works, but it's fragile, duplicates the decode logic, and conflates two different error paths.

---

vvlladd28 (approved):

Issue #15481, opened 2026-04-21
Build log hygiene: categorize and clean up 1000+ maven warnings on lts-4.2


smatvienko-tb:
## Context

Developers rebuilding `lts-4.2` with the `TEST_FAST.md` command

```bash
MAVEN_OPTS="-Xmx1024m" mvn clean install -T6 -DskipTests -Dpkg.skip=true
```

see **1040 `[WARNING]`** lines and **7 `[ERROR]`** lines in an otherwise green build. This is too much noise to spot a real regression. This issue catalogues what the log contains and proposes tiered cleanup so the LTS build output becomes meaningful again.

Baseline taken on commit `75d2613a73e` (Apply LTS schema update during full-version upgrade).

## Reproduction / analysis commands

```bash
MAVEN_OPTS="-Xmx1024m" mvn clean install -T6 -DskipTests -Dpkg.skip=true 2>&1 | tee /tmp/build.log

# Top unique warning categories
grep -E '^\[WARNING\]' /tmp/build.log \
  | sed -E 's/[0-9]+/N/g; s#/[^ :]*thingsboard[^ :]*#PATH#g' \
  | sort | uniq -c | sort -rn | head -40

# All ERROR lines
grep -E '^\[ERROR\]' /tmp/build.log
```

## Tier 1 — pom-only, zero code change, zero risk

**Status: addressed in #15478** — drops totals to **843 warnings / 0 errors**.

| Count | Anomaly | Fix |
|---:|---|---|
| 55 | `'build.plugins.plugin.version' for org.apache.maven.plugins:maven-clean-plugin is missing` at `pom.xml:969` | Pin `<version>3.3.2</version>` (Maven 3.9 bundled default — no behavior change). Also eliminates the cascading *"Some problems were encountered while building the effective model for …"* message for every child module (~140 extra lines). |
| 8 | license-plugin `Unknown file extension` for `lombok.config`, `eslint.config.mjs`, `*.crt`, `*.key`, `*.pem`, `*.otf`, `config.monitoring` | Add these patterns to the `license-maven-plugin` `<excludes>` block |
| 4 | license-plugin `Unable to find a comment style definition for some files` (companion to above) | Same fix |
[...]

PR #15499, opened 2026-04-23
Updated PEM credentials to the non-deprecated certificate API

zzzeebra:
## Pull Request description

Addresses the `X509Certificate.getSubjectDN() deprecated` row of #15481 (Tier 2). Safe, mechanical edits only — no new `@SuppressWarnings`, no changes to expected test strings, no runtime behavior change. Hygiene, not enforcement.

### Measured impact

| Metric | Before | After |
|---|---:|---:|
| `getSubjectDN()` warnings (rule-engine-components) | 7 | **0** |
| `getIssuerDN()` warnings (rule-engine-components) | 1 | **0** |
| `CertPemCredentialsTest`                            | 7/7 | 7/7 |
| `rule-engine-components` full `mvn test`            | 1464/1464 | 1464/1464 |

### What changed

- `CertPemCredentials` (production): keystore aliases in `createAndInitTrustManagerFactory` and `loadKeyStore` now derive from `getSubjectX500Principal().getName(X500Principal.RFC1779)` — 2 call sites.
- `CertPemCredentialsTest`: 5 assertions migrated to the same non-deprecated API. The adjacent `cert.getIssuerDN().getName()` lookup in `testLoadKeyStore` was migrated to `getIssuerX500Principal().getName(X500Principal.RFC1779)` in lock-step to keep alias symmetry with the production aliasing and to clear the last deprecation in the touched file.

### Why `RFC1779` (not the default `RFC2253`)

Empirically verified on all 6 PEM fixtures in `rule-engine-components/src/test/resources/pem/`: `getSubjectX500Principal().getName(X500Principal.RFC1779)` returns a string byte-identical to `getSubjectDN().getName()` for both subject and issuer, including the `"ThingsBoard, Inc."` quoted-comma RDN value. Default `getName()` (RFC 2253) differs (escaped commas, no spaces) and would require rewriting 5 hardcoded expected strings — outside the hygiene scope.

### What we explicitly did not touch

- Other Tier 2 / Tier 3 items in #15481.
- Expected DN format in test assertions — preserved as-is.
- Files outside the two that contain the target deprecations.
- No `@SuppressWarnings` added anywhere.

### Crosslinks

- Tracking issue: #15481

## General checklist

- [x] Changes are backward compatible. Zero runtime behavior change (alias format byte-identical; verified empirically on repo PEM test fixtures).

## Back-End feature checklist

- [x] `rule-engine-components` full `mvn test` green (1464/1464).

### DIFF
--- a/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/credentials/CertPemCredentials.java
+++ b/rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/credentials/CertPemCredentials.java
@@ -25,6 +25,7 @@ import org.thingsboard.server.common.data.StringUtils;
 
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.TrustManagerFactory;
+import javax.security.auth.x500.X500Principal;
 import java.security.KeyStore;
 import java.security.PrivateKey;
 import java.security.cert.CertPath;
@@ -77,7 +78,7 @@ public class CertPemCredentials implements ClientCredentials {
         KeyStore caKeyStore = KeyStore.getInstance(KeyStore.getDefaultType());
         caKeyStore.load(null, null);
         for (X509Certificate caCert : caCerts) {
-            caKeyStore.setCertificateEntry(CA_CERT_CERT_ALIAS_PREFIX + caCert.getSubjectDN().getName(), caCert);
+            caKeyStore.setCertificateEntry(CA_CERT_CERT_ALIAS_PREFIX + caCert.getSubjectX500Principal().getName(X500Principal.RFC1779), caCert);
         }
 
         TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
@@ -99,7 +100,7 @@ public class CertPemCredentials implements ClientCredentials {
         keyStore.load(null);
         List<X509Certificate> unique = certificates.stream().distinct().collect(Collectors.toList());
         for (X509Certificate cert : unique) {
-            keyStore.setCertificateEntry(CERT_ALIAS_PREFIX + cert.getSubjectDN().getName(), cert);
+            keyStore.setCertificateEntry(CERT_ALIAS_PREFIX + cert.getSubjectX500Principal().getName(X500Principal.RFC1779), cert);
         }
 
         if (privateKey != null) {
--- a/rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/credentials/CertPemCredentialsTest.java
+++ b/rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/credentials/CertPemCredentialsTest.java
@@ -23,6 +23,7 @@ import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.thingsboard.common.util.SslUtil;
 
+import javax.security.auth.x500.X500Principal;
 import java.io.File;
 import java.io.IOException;
 import java.security.Key;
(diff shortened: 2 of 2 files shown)


rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/credentials/CertPemCredentials.java
```diff
@@ -77,7 +78,7 @@ protected TrustManagerFactory createAndInitTrustManagerFactory() throws Exceptio
         KeyStore caKeyStore = KeyStore.getInstance(KeyStore.getDefaultType());
         caKeyStore.load(null, null);
         for (X509Certificate caCert : caCerts) {
-            caKeyStore.setCertificateEntry(CA_CERT_CERT_ALIAS_PREFIX + caCert.getSubjectDN().getName(), caCert);
+            caKeyStore.setCertificateEntry(CA_CERT_CERT_ALIAS_PREFIX + caCert.getSubjectX500Principal().getName(X500Principal.RFC1779), caCert);
```
ViacheslavKlimov, line 81: This `getSubjectX500Principal().getName(X500Principal.RFC1779)` expression is now repeated verbatim here and in `loadKeyStore` below (and reconstructed again in the test). Since `SslUtil` is already imported and is where the codebase keeps its X509 helpers, would it be worth extracting a small `SslUtil.getSubjectName(cert)` so the alias format lives in one place? One thing to watch if you do: `TBRedisCacheConfiguration` (lines 194/219) builds its aliases the same way but with the default no-arg `getName()` (RFC 2253), so a shared helper can't just be reused there without changing those alias strings.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/credentials/CertPemCredentials.java
```diff
@@ -99,7 +100,7 @@ protected KeyStore loadKeyStore() throws Exception {
         keyStore.load(null);
         List<X509Certificate> unique = certificates.stream().distinct().collect(Collectors.toList());
         for (X509Certificate cert : unique) {
-            keyStore.setCertificateEntry(CERT_ALIAS_PREFIX + cert.getSubjectDN().getName(), cert);
+            keyStore.setCertificateEntry(CERT_ALIAS_PREFIX + cert.getSubjectX500Principal().getName(X500Principal.RFC1779), cert);
```
ViacheslavKlimov, line 103: The `RFC1779` argument is doing subtle, load-bearing work here: it was picked because it reproduces the old `getSubjectDN().getName()` output byte-for-byte, whereas the no-arg `getName()` defaults to RFC 2253 and would produce different alias strings. A one-line comment noting that would stop a future reader from "simplifying" this to `getName()` and silently changing the keystore aliases.

rule-engine/rule-engine-components/src/test/java/org/thingsboard/rule/engine/credentials/CertPemCredentialsTest.java
```diff
@@ -103,7 +104,7 @@ public void testLoadKeyStore(String certPath, String keyPath, String password, S
 
         List<X509Certificate> certs = SslUtil.readCertFile(certContent);
         for (X509Certificate cert : certs) {
-            String alias = CERT_ALIAS_PREFIX + cert.getIssuerDN().getName();
+            String alias = CERT_ALIAS_PREFIX + cert.getIssuerX500Principal().getName(X500Principal.RFC1779);
```
ViacheslavKlimov, line 107: This looks up the entry by the **issuer** DN, but `loadKeyStore` stores it keyed by the **subject** DN — so the lookup only succeeds because these fixtures are self-signed (subject == issuer). That predates this PR, so it isn't something you introduced, but while the line is being touched it's worth flagging: the assertion wouldn't catch a regression where production switched to an issuer-based alias, and it would break if a non-self-signed cert were ever added to the fixtures. Mirroring production with `getSubjectX500Principal()` here would make the test actually verify the subject keying it's asserting on.

ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [Updated PEM credentials to the non-deprecated certificate API](https://github.com/thingsboard/thingsboard/pull/15499). Left 3 comment(s) inline.

The core change is safe: in both `createAndInitTrustManagerFactory` and `loadKeyStore` the alias string is only written into the `KeyStore` and then handed to `TrustManagerFactory.init()` / `KeyManagerFactory.init()` — it's never read back — so the exact name format has no runtime effect and the deprecation swap can't regress behavior. The comments below are about keeping the intent and duplication maintainable, plus one test-robustness note.

---

PR #15515, opened 2026-04-27
Technical cleanup of deprecated Curator API in ZooKeeper discovery

zzzeebra:
## Pull Request description

Build-log hygiene. Follows up on the Tier 3 plan in issue #15481. Replaces the deprecated Apache Curator `PathChildrenCache` recipe with the modern `CuratorCache` replacement inside `ZkDiscoveryService`, eliminating 11 deprecation warnings on `lts-4.2`.

### Measured impact

| Metric | Before | After |
|---|---:|---:|
| `[WARNING]` lines (repo-wide) | 843 | **832** |
| `PathChildrenCache deprecation` warnings | 11 | **0** |
| `[ERROR]` lines | 0 | 0 |

(`MAVEN_OPTS="-Xmx1024m" mvn clean install -T6 -DskipTests -Dpkg.skip=true`, repo-wide; aggregator from #15481.)

### What changed (2 files, +40/-38)

1. **`ZkDiscoveryService` (production)**
   - Field type: `PathChildrenCache cache` → `CuratorCache cache`.
   - Construction: `new PathChildrenCache(client, zkNodesDir, true)` → `CuratorCache.builder(client, zkNodesDir).build()`.
   - Listener registration: `cache.getListenable().addListener(this)` → `cache.listenable().addListener(this::onCacheEvent)`. Class no longer implements `PathChildrenCacheListener`.
   - Listener body: `childEvent(CuratorFramework, PathChildrenCacheEvent)` → `onCacheEvent(CuratorCacheListener.Type, ChildData oldData, ChildData newData)`. `NODE_CREATED`/`NODE_DELETED` map 1:1 to the original `CHILD_ADDED`/`CHILD_REMOVED` branches; `NODE_CHANGED` keeps the previous `default: break;` no-op.
   - `getOtherServers()`: `cache.getCurrentData().stream()` → `cache.stream().filter(parentPathFilter(zkNodesDir))` (skips the parent node that `CuratorCache` also caches by default).
   - Extra `zkNodesDir.equals(data.getPath())` guard inside `onCacheEvent` for the same reason.

2. **`ZkDiscoveryServiceTest`**
   - Mock type: `PathChildrenCache` → `CuratorCache`.
   - Stub: `when(cache.getCurrentData()).thenReturn(...)` → `when(cache.stream()).thenAnswer(inv -> dataList.stream())`, plus `zkNodesDir` reflection set.
   - Helpers `startNode`/`stopNode` now invoke `onCacheEvent(NODE_CREATED, null, data)` / `onCacheEvent(NODE_DELETED, data, null)`.
   - All 3 existing tests kept verbatim, still pass.

### Behavioral compatibility

- `CuratorCache` is recursive by default, but the cached path `/thingsboard/nodes/<seq>` only contains EPHEMERAL_SEQUENTIAL leaf nodes, so recursive == flat for this use case.
- The parent-node entry that `CuratorCache` adds (and `PathChildrenCache` did not) is filtered out in both `getOtherServers()` (`parentPathFilter`) and `onCacheEvent` (path-equals guard).
- Listener event types map directly: `NODE_CREATED` ↔ `CHILD_ADDED`, `NODE_DELETED` ↔ `CHILD_REMOVED`, `NODE_CHANGED` ↔ ignored (matches the previous `default: break;`).

### Crosslinks

- Tracking issue: #15481

## General checklist

- [x] Changes are backward compatible (Curator 5.x ships both APIs; same observable behavior).

## Back-End feature checklist

- [x] No new Java production class, no new REST endpoint, no new yml property. `mvn clean install -T6 -DskipTests -Dpkg.skip=true` still green.

### DIFF
--- a/common/queue/src/main/java/org/thingsboard/server/queue/discovery/ZkDiscoveryService.java
+++ b/common/queue/src/main/java/org/thingsboard/server/queue/discovery/ZkDiscoveryService.java
@@ -26,9 +26,8 @@ import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.imps.CuratorFrameworkState;
 import org.apache.curator.framework.recipes.cache.ChildData;
-import org.apache.curator.framework.recipes.cache.PathChildrenCache;
-import org.apache.curator.framework.recipes.cache.PathChildrenCacheEvent;
-import org.apache.curator.framework.recipes.cache.PathChildrenCacheListener;
+import org.apache.curator.framework.recipes.cache.CuratorCache;
+import org.apache.curator.framework.recipes.cache.CuratorCacheListener;
 import org.apache.curator.framework.state.ConnectionState;
 import org.apache.curator.framework.state.ConnectionStateListener;
 import org.apache.curator.retry.RetryForever;
@@ -54,12 +53,12 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
-import static org.apache.curator.framework.recipes.cache.PathChildrenCacheEvent.Type.CHILD_REMOVED;
+import static org.apache.curator.framework.recipes.cache.CuratorCacheAccessor.parentPathFilter;
 
 @Service
 @ConditionalOnProperty(prefix = "zk", value = "enabled", havingValue = "true", matchIfMissing = false)
 @Slf4j
-public class ZkDiscoveryService implements DiscoveryService, PathChildrenCacheListener {
+public class ZkDiscoveryService implements DiscoveryService {
 
     @Value("${zk.url}")
     private String zkUrl;
@@ -84,7 +83,7 @@ public class ZkDiscoveryService implements DiscoveryService, PathChildrenCacheLi
     private ScheduledExecutorService zkExecutorService;
     @Getter
     private CuratorFramework client;
-    private PathChildrenCache cache;
+    private CuratorCache cache;
     private String nodePath;
     private String zkNodesDir;
 
@@ -117,7 +116,8 @@ public class ZkDiscoveryService implements DiscoveryService, PathChildrenCacheLi
 
     @Override
     public List<TransportProtos.ServiceInfo> getOtherServers() {
-        return cache.getCurrentData().stream()
+        return cache.stream()
+                .filter(parentPathFilter(zkNodesDir))
                 .filter(cd -> !cd.getPath().equals(nodePath))
                 .map(cd -> {
                     try {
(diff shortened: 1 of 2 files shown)


ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [Technical cleanup of deprecated Curator API in ZooKeeper discovery](https://github.com/thingsboard/thingsboard/pull/15515). Left 2 comment(s) inline.

The `PathChildrenCache` → `CuratorCache` migration reads as behaviorally faithful: the `NODE_CREATED`/`NODE_DELETED` ↔ `CHILD_ADDED`/`CHILD_REMOVED` mapping, the `oldData`/`newData` selection, the parent-node filtering, and the unchanged `getOtherServers()` interface contract all check out. The two comments below are maintainability / test-coverage suggestions rather than correctness problems.

---

common/queue/src/main/java/org/thingsboard/server/queue/discovery/ZkDiscoveryService.java
```diff
@@ -117,7 +116,8 @@ public void init() {
 
     @Override
     public List<TransportProtos.ServiceInfo> getOtherServers() {
-        return cache.getCurrentData().stream()
+        return cache.stream()
+                .filter(parentPathFilter(zkNodesDir))
```
ViacheslavKlimov, line 120: The "skip the recursively-cached parent node" intent is now expressed two different ways: here `getOtherServers()` uses `parentPathFilter(zkNodesDir)`, while `onCacheEvent` uses exact equality `zkNodesDir.equals(data.getPath())`. They aren't quite equivalent — `parentPathFilter` keeps only *direct children* of `zkNodesDir` (it would also drop any deeper descendant), whereas the listener guard only excludes the exact parent path. For the current flat layout (ephemeral-sequential leaves directly under `zkNodesDir`) both are correct, but it's a small trap for whoever changes the ZK node layout later, since the two rules have to stay in sync. Worth aligning them on a single shared predicate so the parent-exclusion rule lives in one place?

common/queue/src/test/java/org/thingsboard/server/queue/discovery/ZkDiscoveryServiceTest.java
```diff
@@ -178,14 +177,14 @@ public void startAnotherNodeDuringRestartTest() throws Exception {
         verify(partitionService, times(1)).recalculatePartitions(eq(currentInfo), eq(List.of(anotherInfo, childInfo)));
     }
 
-    private void startNode(ChildData data) throws Exception {
-        cache.getCurrentData().add(data);
-        zkDiscoveryService.childEvent(curatorFramework, new PathChildrenCacheEvent(CHILD_ADDED, data));
+    private void startNode(ChildData data) {
+        dataList.add(data);
+        zkDiscoveryService.onCacheEvent(NODE_CREATED, null, data);
```
ViacheslavKlimov, line 182: The tests only drive `onCacheEvent` with `NODE_CREATED`/`NODE_DELETED` for child paths, so a few behaviors this PR specifically introduces never get exercised. The parent-node guard `zkNodesDir.equals(data.getPath())` in the listener is never hit (no test fires an event whose path is `/thingsboard/nodes`), and the new `.filter(parentPathFilter(zkNodesDir))` in `getOtherServers()` is effectively a no-op here since `dataList` only ever holds child paths — the parent is never added, so nothing is filtered out. Since dropping the recursively-cached parent is the whole reason `CuratorCache` needs handling that `PathChildrenCache` didn't, could a test add a parent-node `ChildData` (path `/thingsboard/nodes`) and assert it's ignored by the listener and excluded from `getOtherServers()`? Likewise, the `NODE_CHANGED` → `default: break` no-op is intentional but untested — firing `onCacheEvent(NODE_CHANGED, ...)` and asserting no partition recalculation happens would document that intent and guard against someone wiring recalculation into the default branch later.

PR #15522, opened 2026-04-27
Replaced 5 deprecated APIs with non-deprecated variants

zzzeebra:
## Pull Request description

Build-log hygiene. Follows up on the Tier 3 plan in issue #15481. Replaces five small deprecated API call sites with their non-deprecated equivalents. No production behavior change — verified empirically per migration.

### Measured impact

`mvn clean install -T6 -DskipTests -Dpkg.skip=true` on `lts-4.2 @ 4fdc92287d`:

| Metric | Before | After |
|---|---:|---:|
| `[WARNING]` lines | 843 | **831** |
| `[ERROR]` lines | 0 | 0 |
| Target deprecation matches | 17 | **0** |

The 17 → 0 covers all 7 unique warning shapes targeted in this PR. The +5 between target delta (−17) and total delta (−12) is unrelated multi-threaded javac emission noise (3× `MockBean`, 1× `ProtoUtils.getTbMsgProto`, 1× `TelemetryCmdsWrapper`) in files this PR does not touch.

### What changed

5 small, behavior-preserving migrations across 9 files (16 +, 26 −):

1. **`JsonNode.fields()` → `properties()`** (4 + 2 warnings, 4 files)
   `common/util/JacksonUtil.java`, `tools/.../TranslationPruner.java`, `rule-engine-components/.../TbCopyKeysNode.java`, `rule-engine-components/.../Langchain4jJsonSchemaAdapter.java`. Iterator-style loops rewritten as enhanced for-each over the `Set<Map.Entry<String, JsonNode>>` returned by `properties()`. `forEachRemaining(lambda)` on the iterator → `forEach(lambda)` on the set.

2. **`HttpMessageNotReadableException(String)` → `(String, HttpInputMessage)`** (3 warnings, 1 test file)
   `common/transport/http/.../DeviceApiControllerTest.java`. Added explicit `(HttpInputMessage) null` cast (avoids overload ambiguity with `(String, Throwable)`); test only triggers `callback.onError(...)` and does not read `httpInputMessage`.

3. **`org.junit.Assert.assertThat(T, Matcher)` → `org.hamcrest.MatcherAssert.assertThat`** (3 warnings, 1 test file)
   `application/src/test/.../AbstractLwM2MIntegrationTest.java`. Static-import swap only; same Hamcrest matcher, same call sites.

4. **`io.jsonwebtoken.SignatureException` → `io.jsonwebtoken.security.SignatureException`** (2 warnings, 2 files)
   `application/.../JwtTokenFactory.java`, `application/.../OAuth2AppTokenFactory.java`. Import-rename only.

5. **`StopWatch.getLastTaskTimeNanos()` / `getLastTaskTimeMillis()` → `lastTaskInfo().getTimeNanos()` / `getTimeMillis()`** (2 + 1 warnings, 1 file)
   `common/util/.../TbStopWatch.java`. Includes the sister `getLastTaskTimeMillis()` call (same class, same file) for symmetry.

### Behavioral compatibility

Each migration verified empirically against the actual library version on the classpath:

- **Jackson 2.20+**: `ObjectNode.properties()` returns `LinkedHashMap.entrySet()` — same backing map, same insertion order, same `Map.Entry<String, JsonNode>` element type as `fields()`. No call site relied on `Iterator.remove()` or iterator reuse.
- **Spring 6.2.9 `HttpMessageNotReadableException`**: the 2-arg `(String, HttpInputMessage)` ctor exists; `null` is accepted; the test never inspects `getHttpInputMessage()`.
- **JUnit 4 / Hamcrest**: `MatcherAssert.assertThat` is the documented replacement. Same matcher type, same failure messages.
- **JJWT 0.12.5**: inspected `jjwt-impl-0.12.5.jar` bytecode — every `new …SignatureException` references `io.jsonwebtoken.security.SignatureException`. The legacy `io.jsonwebtoken.SignatureException` is a parent of the new one (`security.SignatureException extends io.jsonwebtoken.SignatureException`), so a `catch` on the new type catches the same throw set the library actually emits.
- **Spring 6.2.9 `StopWatch`**: `lastTaskInfo()` is non-deprecated; `TaskInfo.getTimeNanos()` / `getTimeMillis()` return the same `long` as the deprecated accessors and throw the same `IllegalStateException` when no task has run.

### Tests

Targeted module tests pass on the change:

- `common/util`, `common/transport/http`, `rule-engine/rule-engine-components`: 1464 tests, 0 failures, 0 errors.
- `application` — `JwtTokenFactoryTest` (4) + `Oauth2AuthenticationSuccessHandlerTest` (1): 5 tests, 0 failures, 0 errors.

### Out of scope (intentionally not touched)

- `application/src/test/.../VersionControlTest.java:427` uses the same `JsonNode.fields()` API but does not surface as a `[WARNING]` in the baseline build. Different module, no warning, left for a future sweep.
- All other Tier 3 deprecation groups from #15481.

### Crosslinks

- Tracking issue: #15481

## General checklist

- [x] Changes are backward compatible (verified per migration above).

## Back-End feature checklist

- [x] No new Java production code, no new REST endpoint, no new yml property. `mvn clean install -T6 -DskipTests -Dpkg.skip=true` still green.


ViacheslavKlimov (commented):
## Review summary

Reviewed 9 changed files in [Replaced 5 deprecated APIs with non-deprecated variants](https://github.com/thingsboard/thingsboard/pull/15522). Left 4 comment(s) inline.

This is a clean, well-documented, behavior-preserving deprecation sweep. I independently verified the two migrations with real semantic risk: the `JsonNode.fields()` → `properties()` rewrites are drop-in equivalent (same backing `LinkedHashMap`, same insertion order, and none of the loops mutate the node they iterate), and by disassembling `jjwt-impl-0.12.5` I confirmed the `io.jsonwebtoken.security.SignatureException` catch narrowing drops no exception the library actually throws in the pinned version. The comments are all minor.

---

application/src/main/java/org/thingsboard/server/service/security/model/token/JwtTokenFactory.java
```diff
@@ -23,9 +23,9 @@
 import io.jsonwebtoken.JwtParser;
 import io.jsonwebtoken.Jwts;
 import io.jsonwebtoken.MalformedJwtException;
-import io.jsonwebtoken.SignatureException;
 import io.jsonwebtoken.UnsupportedJwtException;
 import io.jsonwebtoken.security.Keys;
+import io.jsonwebtoken.security.SignatureException;
```
ViacheslavKlimov, line 28: Worth noting: the new `io.jsonwebtoken.security.SignatureException` is a *subclass* of the old `io.jsonwebtoken.SignatureException` (the `security.*` one `extends` the legacy one), so the `catch` at line 228 is now strictly narrower than before — it no longer catches a bare `io.jsonwebtoken.SignatureException`. I disassembled `jjwt-impl-0.12.5` and confirmed the impl only ever instantiates the `security.*` type (0 references to the legacy parent, 27 to the child), so this is behavior-preserving for the pinned version. The only residual risk is a future JJWT bump that throws the legacy parent directly — it would then propagate uncaught instead of becoming a `JwtExpiredTokenException`. Fine to keep as-is; just flagging the narrowing.

application/src/main/java/org/thingsboard/server/service/security/model/token/OAuth2AppTokenFactory.java
```diff
@@ -20,9 +20,9 @@
 import io.jsonwebtoken.Jws;
 import io.jsonwebtoken.Jwts;
 import io.jsonwebtoken.MalformedJwtException;
-import io.jsonwebtoken.SignatureException;
 import io.jsonwebtoken.UnsupportedJwtException;
 import io.jsonwebtoken.security.Keys;
+import io.jsonwebtoken.security.SignatureException;
```
ViacheslavKlimov, line 25: Same catch-narrowing as in `JwtTokenFactory` — the new `security.SignatureException` is a subclass of the legacy `io.jsonwebtoken.SignatureException`, so the multi-catch at line 47 now excludes a bare legacy `SignatureException`. Verified safe for jjwt 0.12.5 (the impl only throws the `security.*` type); just noting it would silently narrow should a future JJWT ever throw the legacy parent directly.

rule-engine/rule-engine-components/src/main/java/org/thingsboard/rule/engine/ai/Langchain4jJsonSchemaAdapter.java
```diff
@@ -94,7 +94,7 @@ private static JsonObjectSchema.Builder parseObject(JsonNode objectSchema) {
 
         JsonNode propertiesNode = objectSchema.get("properties");
         if (propertiesNode != null) {
-            propertiesNode.fields().forEachRemaining(entry -> {
+            propertiesNode.properties().forEach(entry -> {
```
ViacheslavKlimov, line 97: Every other site in this PR moved to an enhanced-for over `properties()` (which now returns an `Iterable`), but this one keeps `.forEach(lambda)`. Not wrong, but an enhanced-for here would read consistently with the rest of the change and let the body use plain locals instead of a closure — was there a reason to keep the lambda?

common/util/src/main/java/org/thingsboard/common/util/TbStopWatch.java
```diff
@@ -52,17 +52,17 @@ public long stopAndGetTotalTimeMillis(){
 
     public long stopAndGetTotalTimeNanos(){
         stop();
-        return getLastTaskTimeNanos();
+        return lastTaskInfo().getTimeNanos();
```
ViacheslavKlimov, line 55: Touching this line surfaces a latent naming mismatch: `stopAndGetTotalTimeNanos()` returns the *last-task* time (`lastTaskInfo().getTimeNanos()`), while its sibling `stopAndGetTotalTimeMillis()` just above returns the actual total (`getTotalTimeMillis()`). Two methods differing only in unit now behave differently, which is easy to misuse. It's pre-existing and harmless at the current single-task caller, so no need to fix it in this cleanup — but if you want to, either rename this to `stopAndGetLastTaskTimeNanos` or switch to `getTotalTimeNanos()` to match the name.

PR #15527, opened 2026-04-28
Added inactivity timeout setting to Device Profile

zzzeebra:
## Summary
- New `inactivityTimeoutMs` field on `DeviceProfileData`. Devices that belong to a profile and do not have a per-device `inactivityTimeout` server attribute inherit the profile-level value.
- Resolution order at runtime: per-device attribute → profile field → server-level default from `thingsboard.yml` (`state.defaultInactivityTimeoutInSec`). Existing per-device override attributes keep priority and are not migrated.
- Profile changes propagate to every cluster node via `@EventListener(ComponentLifecycleMsg.class)`. A per-profile resolved-timeout cache short-circuits iteration when the value did not change, so saving an unrelated profile field (description, alarms, ...) is `O(1)`.
- Device reassignment between profiles is handled in `onQueueMsg`: cached `deviceProfileId` updates and timeout re-resolves from the new profile (unless the device has its own override attribute).
- UI: new **Device inactivity timeout** field on the Device Profile Details page using `<tb-time-unit-input>` with seconds/minutes/hours/days unit selector. Empty input maps to `null` in storage; backend validator allows `null` or `> 0`.
- `DeviceProfileData` now carries `@JsonIgnoreProperties(ignoreUnknown = true)` for consistency with sibling DTOs and defense against non-`IGNORE_UNKNOWN` ObjectMapper instances.
- Reworded the `state.defaultInactivityTimeoutInSec` comment block in `thingsboard.yml` to describe the new three-tier resolution order.

## Automated testing
- `DefaultDeviceStateServiceTest` (108 tests) covers: three-tier resolution on cold/hot paths, per-device override priority, profile-update propagation across overridden vs non-overridden devices, profile cleared to null falling back to YAML default, profile-update cache hit on unchanged timeout, device reassignment with and without override, non-`DEVICE_PROFILE` and `DELETED` event filtering.
- `DeviceProfileDataValidatorTest` (12 tests) covers `inactivityTimeoutMs` null / positive / zero / negative validation.

## Manual testing
Verified locally on a single-node deployment with PostgreSQL.

- [x] **Test 1 — Form save/load.** Created Device Profile, set inactivity timeout to 5 minutes and saved; reopened and confirmed the value persists. Cleared the field and saved; confirmed `device_profile.profile_data->'inactivityTimeoutMs'` is JSON `null` (not `0`).
- [x] **Test 2 — Profile timeout marks device inactive.** Set profile timeout to 30 seconds, sent telemetry, waited past 30 seconds. Device transitioned to `active=false` and `inactivityAlarmTime` server attribute appeared.
- [x] **Test 3 — Per-device override beats profile.** With profile timeout at 30 seconds, added a per-device `inactivityTimeout` server attribute set to 5 minutes, sent telemetry, waited past 30 seconds — device stayed `active=true` (override takes priority over the profile value).
- [x] **Test 4 — Delete override falls back to profile.** Removed the per-device `inactivityTimeout` attribute, sent fresh telemetry, waited past 30 seconds — device became `active=false` (profile timeout applied).
- [x] **Test 5 — Profile update propagates to existing devices.** Updated the profile inactivity timeout while devices were running and confirmed non-overridden devices applied the new value without restarting the node.

Screenshot 2026-04-28 at 21 20 11

Screenshot 2026-04-28 at 21 02 12

## Documentation
Companion docs PR: [thingsboard/thingsboard.io#426](https://github.com/thingsboard/thingsboard.io/pull/426) — documents the new **Device inactivity timeout** profile field and the three-tier resolution order (per-device attribute → profile field → server default). This PR also updates the `state.defaultInactivityTimeoutInSec` comment block in `thingsboard.yml`, which is the source for the auto-generated `reference/configuration/core-rule-engine-config.mdx` page.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
+++ b/application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
@@ -34,6 +34,7 @@ import org.checkerframework.checker.nullness.qual.NonNull;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.context.event.EventListener;
 import org.springframework.stereotype.Service;
 import org.thingsboard.common.util.JacksonUtil;
 import org.thingsboard.common.util.ThingsBoardExecutors;
@@ -44,10 +45,12 @@ import org.thingsboard.server.common.data.ApiUsageRecordKey;
 import org.thingsboard.server.common.data.AttributeScope;
 import org.thingsboard.server.common.data.Device;
 import org.thingsboard.server.common.data.DeviceIdInfo;
+import org.thingsboard.server.common.data.DeviceProfile;
 import org.thingsboard.server.common.data.EntityType;
 import org.thingsboard.server.common.data.StringUtils;
 import org.thingsboard.server.common.data.exception.TenantNotFoundException;
 import org.thingsboard.server.common.data.id.DeviceId;
+import org.thingsboard.server.common.data.id.DeviceProfileId;
 import org.thingsboard.server.common.data.id.EntityId;
 import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.common.data.id.UUIDBased;
@@ -62,6 +65,7 @@ import org.thingsboard.server.common.data.msg.TbMsgType;
 import org.thingsboard.server.common.data.notification.rule.trigger.DeviceActivityTrigger;
 import org.thingsboard.server.common.data.page.PageData;
 import org.thingsboard.server.common.data.page.PageDataIterable;
+import org.thingsboard.server.common.data.plugin.ComponentLifecycleEvent;
 import org.thingsboard.server.common.data.query.EntityData;
 import org.thingsboard.server.common.data.query.EntityDataPageLink;
 import org.thingsboard.server.common.data.query.EntityDataQuery;
@@ -72,6 +76,7 @@ import org.thingsboard.server.common.msg.TbMsg;
 import org.thingsboard.server.common.msg.TbMsgDataType;
 import org.thingsboard.server.common.msg.TbMsgMetaData;
 import org.thingsboard.server.common.msg.notification.NotificationRuleProcessor;
+import org.thingsboard.server.common.msg.plugin.ComponentLifecycleMsg;
 import org.thingsboard.server.common.msg.queue.ServiceType;
 import org.thingsboard.server.common.msg.queue.TbCallback;
 import org.thingsboard.server.common.msg.queue.TopicPartitionInfo;
@@ -85,6 +90,7 @@ import org.thingsboard.server.gen.transport.TransportProtos;
 import org.thingsboard.server.queue.discovery.PartitionService;
 import org.thingsboard.server.queue.util.TbCoreComponent;
 import org.thingsboard.server.service.partition.AbstractPartitionBasedService;
+import org.thingsboard.server.service.profile.TbDeviceProfileCache;
 import org.thingsboard.server.service.telemetry.TelemetrySubscriptionService;
 
 import java.util.ArrayList;
@@ -168,6 +174,9 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
     @Autowired
     @Lazy
     private TelemetrySubscriptionService tsSubService;
+    @Autowired
+    @Lazy
+    private TbDeviceProfileCache deviceProfileCache;
 
     @Value("#{${state.defaultInactivityTimeoutInSec} * 1000}")
     private long defaultInactivityTimeoutMs;
@@ -191,6 +200,7 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
     private ListeningExecutorService deviceStateCallbackExecutor;
 
     final ConcurrentMap<DeviceId, DeviceStateData> deviceStates = new ConcurrentHashMap<>();
+    final ConcurrentMap<DeviceProfileId, Long> profileResolvedInactivityTimeoutMs = new ConcurrentHashMap<>();
 
     @PostConstruct
     public void init() {
@@ -307,13 +317,78 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
         if (cleanDeviceStateIfBelongsToExternalPartition(tenantId, deviceId)) {
             return;
         }
-        if (inactivityTimeout <= 0L) {
-            inactivityTimeout = defaultInactivityTimeoutMs;
-        }
-        log.trace("[{}] on Device Activity Timeout Update device id {} inactivityTimeout {}", tenantId.getId(), deviceId.getId(), inactivityTimeout);
         DeviceStateData stateData = getOrFetchDeviceStateData(deviceId);
-        stateData.getState().setInactivityTimeout(inactivityTimeout);
+        boolean overridden = inactivityTimeout > 0L;
+        long resolved = resolveInactivityTimeout(inactivityTimeout, tenantId, stateData.getDeviceProfileId());
+        log.trace("[{}] on Device Activity Timeout Update device id {} inactivityTimeout {} (resolved {}, overridden {})",
+                tenantId.getId(), deviceId.getId(), inactivityTimeout, resolved, overridden);
+        stateData.setInactivityTimeoutOverridden(overridden);
+        stateData.getState().setInactivityTimeout(resolved);
         checkAndUpdateState(deviceId, stateData);
+    }
+
+    /**
+     * Three-tier resolution for the device inactivity timeout:
+     * <ol>
+     *   <li>per-device server attribute (positive value supplied here),</li>
+     *   <li>profile-level {@code inactivityTimeoutMs} (when set and positive),</li>
+     *   <li>global YAML default ({@code state.defaultInactivityTimeoutInSec}).</li>
+     * </ol>
+     */
+    private long resolveInactivityTimeout(long fromAttributeOrZero, TenantId tenantId, DeviceProfileId profileId) {
+        if (fromAttributeOrZero > 0L) {
+            return fromAttributeOrZero;
+        }
+        Long fromProfile = profileTimeoutMs(tenantId, profileId);
+        if (fromProfile != null && fromProfile > 0L) {
+            return fromProfile;
+        }
+        return defaultInactivityTimeoutMs;
+    }
+
+    private Long profileTimeoutMs(TenantId tenantId, DeviceProfileId profileId) {
+        if (tenantId == null || profileId == null) {
+            return null;
+        }
+        DeviceProfile profile = deviceProfileCache.get(tenantId, profileId);
+        if (profile == null || profile.getProfileData() == null) {
+            return null;
+        }
+        return profile.getProfileData().getInactivityTimeoutMs();
+    }
+
+    @EventListener(ComponentLifecycleMsg.class)
+    public void onComponentLifecycleEvent(ComponentLifecycleMsg event) {
+        if (!EntityType.DEVICE_PROFILE.equals(event.getEntityId().getEntityType())) {
+            return;
+        }
+        DeviceProfileId profileId = new DeviceProfileId(event.getEntityId().getId());
+        if (ComponentLifecycleEvent.DELETED.equals(event.getEvent())) {
+            profileResolvedInactivityTimeoutMs.remove(profileId);
+            return;
+        }
+        if (!ComponentLifecycleEvent.UPDATED.equals(event.getEvent())) {
+            return;
+        }
+        Long timeoutFromProfile = profileTimeoutMs(event.getTenantId(), profileId);
+        long resolvedNew = (timeoutFromProfile != null && timeoutFromProfile > 0L) ? timeoutFromProfile : defaultInactivityTimeoutMs;
+        Long previousResolved = profileResolvedInactivityTimeoutMs.put(profileId, resolvedNew);
+        if (previousResolved != null && previousResolved == resolvedNew) {
+            return;
+        }
+        deviceStateCallbackExecutor.submit(() -> deviceStates.forEach((deviceId, stateData) -> {
+            if (!profileId.equals(stateData.getDeviceProfileId())) {
+                return;
+            }
+            if (stateData.isInactivityTimeoutOverridden()) {
+                return;
+            }
+            if (stateData.getState().getInactivityTimeout() == resolvedNew) {
+                return;
+            }
+            stateData.getState().setInactivityTimeout(resolvedNew);
+            checkAndUpdateState(deviceId, stateData);
+        }));
     }
 
     @Override
@@ -384,6 +459,15 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
                         md.putValue("deviceLabel", device.getLabel());
                         md.putValue("deviceType", device.getType());
                         stateData.setMetaData(md);
+                        DeviceProfileId newProfileId = device.getDeviceProfileId();
+                        if (!Objects.equals(newProfileId, stateData.getDeviceProfileId())) {
+                            stateData.setDeviceProfileId(newProfileId);
+                            if (!stateData.isInactivityTimeoutOverridden()) {
+                                long resolved = resolveInactivityTimeout(0L, device.getTenantId(), newProfileId);
+                                stateData.getState().setInactivityTimeout(resolved);
+                                checkAndUpdateState(device.getId(), stateData);
+                            }
+                        }
                         callback.onSuccess();
                     }
                 } else {
@@ -698,7 +782,9 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
                 try {
                     long lastActivityTime = getEntryValue(data, LAST_ACTIVITY_TIME, 0L);
                     long inactivityAlarmTime = getEntryValue(data, INACTIVITY_ALARM_TIME, 0L);
-                    long inactivityTimeout = getEntryValue(data, INACTIVITY_TIMEOUT, defaultInactivityTimeoutMs);
+                    long timeoutFromAttribute = getEntryValue(data, INACTIVITY_TIMEOUT, 0L);
+                    boolean overridden = timeoutFromAttribute > 0L;
+                    long resolvedTimeout = resolveInactivityTimeout(timeoutFromAttribute, device.getTenantId(), device.getDeviceProfileId());
                     // Actual active state by wall-clock will be updated outside this method. This method is only for fetching persistent state
                     final boolean active = getEntryValue(data, ACTIVITY_STATE, false);
                     DeviceState deviceState = DeviceState.builder()
@@ -707,7 +793,7 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
                             .lastDisconnectTime(getEntryValue(data, LAST_DISCONNECT_TIME, 0L))
                             .lastActivityTime(lastActivityTime)
                             .lastInactivityAlarmTime(inactivityAlarmTime)
-                            .inactivityTimeout(inactivityTimeout > 0 ? inactivityTimeout : defaultInactivityTimeoutMs)
+                            .inactivityTimeout(resolvedTimeout)
                             .build();
                     TbMsgMetaData md = new TbMsgMetaData();
                     md.putValue("deviceName", device.getName());
@@ -717,9 +803,12 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
                             .customerId(device.getCustomerId())
                             .tenantId(device.getTenantId())
                             .deviceId(device.getId())
+                            .deviceProfileId(device.getDeviceProfileId())
                             .deviceCreationTime(device.getCreatedTime())
                             .metaData(md)
-                            .state(deviceState).build();
+                            .state(deviceState)
+                            .inactivityTimeoutOverridden(overridden)
+                            .build();
                     log.debug("[{}] Fetched device state from the DB {}", device.getId(), deviceStateData);
                     return deviceStateData;
                 } catch (Exception e) {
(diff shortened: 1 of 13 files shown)


application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -307,15 +317,80 @@ public void onDeviceInactivityTimeoutUpdate(TenantId tenantId, DeviceId deviceId
...
+        if (!ComponentLifecycleEvent.UPDATED.equals(event.getEvent())) {
+            return;
+        }
+        Long timeoutFromProfile = profileTimeoutMs(event.getTenantId(), profileId);
+        long resolvedNew = (timeoutFromProfile != null && timeoutFromProfile > 0L) ? timeoutFromProfile : defaultInactivityTimeoutMs;
+        Long previousResolved = profileResolvedInactivityTimeoutMs.put(profileId, resolvedNew);
+        if (previousResolved != null && previousResolved == resolvedNew) {
+            return;
+        }
+        deviceStateCallbackExecutor.submit(() -> deviceStates.forEach((deviceId, stateData) -> {
```
ViacheslavKlimov: `deviceStateCallbackExecutor` is a multi-threaded work-stealing pool (init at line 210), so two profile-UPDATED events for the same profile arriving in quick succession with different resolved timeouts (T1 then T2) can submit two iteration tasks that run concurrently — and the per-device write order between them is non-deterministic, so devices can be left with the stale T1 even though `profileResolvedInactivityTimeoutMs` already holds T2. The `previousResolved == resolvedNew` short-circuit at line 376 only helps when consecutive events resolve to the same value.

In practice this is narrow — `ComponentLifecycleMsg` is typically fanned out from a single consumer thread and a follow-up event would converge state again — but two cheap mitigations would close the gap entirely: (a) re-read `profileResolvedInactivityTimeoutMs.get(profileId)` inside the lambda and bail out if it no longer equals the captured `resolvedNew`, or (b) just run the iteration synchronously here, since the heavy short-circuiting (cache hit, profile mismatch, override flag, equal-timeout) means the common case is already O(1) and the listener already runs off the consumer thread.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -307,15 +317,80 @@ public void onDeviceInactivityTimeoutUpdate(TenantId tenantId, DeviceId deviceId
...
+            profileResolvedInactivityTimeoutMs.remove(profileId);
+            return;
+        }
+        if (!ComponentLifecycleEvent.UPDATED.equals(event.getEvent())) {
+            return;
+        }
+        Long timeoutFromProfile = profileTimeoutMs(event.getTenantId(), profileId);
+        long resolvedNew = (timeoutFromProfile != null && timeoutFromProfile > 0L) ? timeoutFromProfile : defaultInactivityTimeoutMs;
+        Long previousResolved = profileResolvedInactivityTimeoutMs.put(profileId, resolvedNew);
+        if (previousResolved != null && previousResolved == resolvedNew) {
```
ViacheslavKlimov, line 375: Minor maintainability nit: `previousResolved == resolvedNew` works correctly today (auto-unboxing on the `Long`), but it's a known Java footgun — if a later refactor changes `resolvedNew` from `long` to `Long`, this silently becomes a reference comparison. `previousResolved.longValue() == resolvedNew` (or `Objects.equals`) would be safer against that.

ui-ngx/src/app/modules/home/components/profile/device-profile.component.ts
```diff
@@ -229,13 +232,22 @@ export class DeviceProfileComponent extends EntityComponent<DeviceProfile> {
...
+    delete formValue.inactivityTimeoutSec;
     const deviceProvisionConfiguration: DeviceProvisionConfiguration = formValue.profileData.provisionConfiguration;
     formValue.provisionType = deviceProvisionConfiguration.type;
     formValue.provisionDeviceKey = deviceProvisionConfiguration.provisionDeviceKey;
     delete deviceProvisionConfiguration.provisionDeviceKey;
     return super.prepareFormValue(formValue);
   }
 
+  private toInactivityTimeoutSec(ms: number | null | undefined): number {
+    return ms && ms > 0 ? ms / SECOND : 0;
```
ViacheslavKlimov, line 246: Returning `0` for an unset/null timeout means new and existing-but-unconfigured profiles render the field as `0 Seconds`, which mismatches the hint text *"Leave empty to use the server default"* — the field is never actually empty by default. `tb-time-unit-input.writeValue()` falls back to `time: 0, timeUnit: SECONDS` for any null/0 input regardless, so this isn't fully fixable from here, but at minimum the hint copy could be reworded (e.g. *"Leave at 0 to use the server default"*) so users aren't looking for an empty state that the component can't show.

ViacheslavKlimov (commented):
## Review summary

Reviewed 12 changed files in [Added inactivity timeout setting to Device Profile](https://github.com/thingsboard/thingsboard/pull/15527). The three-tier resolution (per-device attribute → profile → YAML default) is implemented cleanly, the cached `profileResolvedInactivityTimeoutMs` short-circuit is a nice optimization, and the test coverage is thorough. Left 3 comments inline on a possible race in the propagation path, a UX rough edge on the new form field, and the rest-client surface.

## Additional findings

- **Manual Test 5 was not executed** — the PR description explicitly lists *"Test 5 — Profile update propagates to existing devices. Not executed."* but that scenario is the headline behavior of the new feature (the whole `onComponentLifecycleEvent` propagation logic, the `profileResolvedInactivityTimeoutMs` cache, and the override-skipping iteration). Unit tests cover it, but please run a manual end-to-end check before merging: set a profile timeout, observe a non-overridden device pick up the new value, then change the profile timeout again and confirm the device transitions accordingly without restarting the node.
- **`DeviceIdInfo` constructor signature changed** (`common/data/.../DeviceIdInfo.java`) — the only previous 3-arg constructor was replaced with a 4-arg one. All in-tree call sites are updated, but `DeviceIdInfo` is part of the published `tb-common-data` artifact, so any external code that constructed it directly will break at compile time. Probably fine (it's an internal-leaning DTO), but worth being aware of for changelog purposes.

---

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
zzzeebra: Fixed in 77d165dea6 — replaced the `submit(...)` with synchronous iteration in the listener. The cache `put(...)` and the `deviceStates` traversal now share the same happens-before, so the T1/T2 interleaving is gone. `deviceStateCallbackExecutor` is still used elsewhere in the class so I left the field as-is.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
zzzeebra: Leaving as-is. `previousResolved` is null-checked one line above, and `Long == long` auto-unboxes correctly here. Hardening against a hypothetical `long → Long` refactor isn't worth a maintenance note in this hot path.

ui-ngx/src/app/modules/home/components/profile/device-profile.component.ts
zzzeebra: Reworded the hint to "Set to 0 to use the server default" in 65275e082e. Matches what `tb-time-unit-input` actually displays for a null/0 value.

zzzeebra:
Re: review summary findings

- **Test 5** — was actually executed; the "Not executed" line was a description error. PR description updated to reflect the manual run.
- **`DeviceIdInfo` constructor signature** — acknowledged. Will note for the changelog.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -168,6 +174,9 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
     @Autowired
     @Lazy
     private TelemetrySubscriptionService tsSubService;
+    @Autowired
+    @Lazy
```
ViacheslavKlimov: Why Lazy?

ViacheslavKlimov (commented):
## Review summary

Reviewed 13 changed files in [Added inactivity timeout setting to Device Profile](https://github.com/thingsboard/thingsboard/pull/15527). The three-tier resolution (per-device attribute → profile field → YAML default) is well structured, the validator/UI/serialization pieces line up, and the test coverage is genuinely thorough. I traced the cluster-propagation path and a few things check out cleanly that are worth recording: `DeviceIdInfo` carries a `serialVersionUID` and is only ever a transient query-result holder (never cross-node cached/serialized), so the constructor signature change is safe; and `handleComponentLifecycleMsg` evicts-and-refreshes the device profile cache *before* `eventPublisher.publishEvent(...)` on the same thread, so the new listener always reads the up-to-date profile (no stale read, no `@Order` dependency).

My main concern is around the synchronous propagation path — it executes potentially unbounded work on the shared TB-Core notifications consumer thread. Details inline. The rest are smaller cleanliness/edge-case notes.

Left 4 comments inline.

## Additional findings

These touch code outside the diff (the lifecycle-event delivery path) that's relevant to the inline finding on `onComponentLifecycleEvent`:

- **application/src/main/java/org/thingsboard/server/service/queue/processing/AbstractConsumerService.java:202** — `eventPublisher.publishEvent(componentLifecycleMsg)` is the call that drives the new `@EventListener`. It runs on the notifications consumer thread (via `DefaultTbCoreConsumerService`), and the default Spring `@EventListener` is synchronous. So every device-profile lifecycle event in the whole cluster is dispatched to `onComponentLifecycleEvent` on that thread. This is the context for the blocking concern noted inline at `DefaultDeviceStateService.java`. All other `ComponentLifecycleMsg` listeners on this path (profile caches, notification-rule cache, etc.) do cheap O(1)/eviction work; this new one is the only one doing per-device fan-out, so it's worth treating differently from the established pattern.
- Decreasing a profile's `inactivityTimeoutMs` immediately re-evaluates every non-overridden device of that profile in one pass, so a single save can fan out into a burst of simultaneous `INACTIVITY_EVENT` rule-engine messages, alarm writes, and notifications. This is inherent to immediate propagation (the periodic checker would do the same on its tick), but the burst here is triggered synchronously by the save — worth being aware of for large profiles.

---

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -307,15 +317,80 @@ public void onDeviceInactivityTimeoutUpdate(TenantId tenantId, DeviceId deviceId
...
+        if (!ComponentLifecycleEvent.UPDATED.equals(event.getEvent())) {
+            return;
+        }
+        Long timeoutFromProfile = profileTimeoutMs(event.getTenantId(), profileId);
+        long resolvedNew = (timeoutFromProfile != null && timeoutFromProfile > 0L) ? timeoutFromProfile : defaultInactivityTimeoutMs;
+        Long previousResolved = profileResolvedInactivityTimeoutMs.put(profileId, resolvedNew);
+        if (previousResolved != null && previousResolved == resolvedNew) {
+            return;
+        }
+        deviceStates.forEach((deviceId, stateData) -> {
```
ViacheslavKlimov: This is the part I'd think hardest about given the performance/architecture focus.

This `forEach` runs synchronously on the TB-Core notifications consumer thread (`handleComponentLifecycleMsg` → `eventPublisher.publishEvent`). A few consequences worth weighing:

1. **Full-map scan, not profile-scoped.** Every device-profile UPDATE across the entire cluster whose *resolved* timeout changed iterates this node's complete `deviceStates` map — including devices of unrelated profiles — and calls `checkAndUpdateState` for matching ones. There's no `DeviceProfileId → Set<DeviceId>` index, so cost scales with total devices on the node, not with the affected profile's device count. The `profileResolvedInactivityTimeoutMs` cache only short-circuits the *no-change* case (and not on the first update of each profile after startup, when `previousResolved` is null).

2. **It blocks the shared consumer thread.** While this iterates, the notifications consumer can't process other `ToCore` notifications (other lifecycle events, RPC, etc.). An earlier commit ("Run profile inactivity timeout propagation synchronously") deliberately changed this from `deviceStateCallbackExecutor.submit(...)` to synchronous — was blocking the consumer thread the intent? Offloading to `deviceStateExecutor`/`deviceStateCallbackExecutor` (as the original did) would keep the consumer responsive; a secondary profile→devices index would cut the scan itself.

3. **Concurrency.** This thread now writes `state.setInactivityTimeout(...)` and runs `checkAndUpdateState` for a device concurrently with the scheduled `checkStates()` thread and the activity-event executor, on a `DeviceState` whose fields are plain (non-volatile, unsynchronized) longs. It's consistent with the existing unsynchronized access pattern in this class, but it adds a third writer — so visibility of the new timeout to the scheduler thread isn't guaranteed to be prompt, and concurrent `updateInactivityStateIfExpired` on the same device is now more likely. Worth a deliberate decision rather than relying on the existing looseness.

ui-ngx/src/app/modules/home/components/profile/device-profile.component.ts
```diff
@@ -229,13 +232,22 @@ export class DeviceProfileComponent extends EntityComponent<DeviceProfile> {
     if (formValue.defaultEdgeRuleChainId) {
       formValue.defaultEdgeRuleChainId = new RuleChainId(formValue.defaultEdgeRuleChainId);
     }
+    if (formValue.profileData) {
```
ViacheslavKlimov: This `if (formValue.profileData)` guard is inconsistent with the line just below it — `const deviceProvisionConfiguration: ... = formValue.profileData.provisionConfiguration` dereferences `profileData` unconditionally. So if `profileData` could ever be falsy here, line 240 would NPE regardless; and if it can't (which is the case — it's always a non-optional `fb.group`), the guard's only effect is a dead branch where the timeout would be silently dropped. I'd drop the `if` and set `inactivityTimeoutMs` directly for clarity.

ui-ngx/src/app/modules/home/components/profile/device-profile.component.ts
ViacheslavKlimov, line 246: `ms / SECOND` can produce a fractional value. The backend validator and REST API accept any positive `Long` (e.g. `1500` or `500` ms), but `tb-time-unit-input.parseTime()` only returns a unit/value pair when the seconds value divides evenly into one of days/hours/minutes/seconds — otherwise it returns `undefined`, which then gets patched into the inner form. So a profile whose `inactivityTimeoutMs` was set to a non-1000-multiple via the API won't render correctly in this field. Either round here, or confirm sub-second / non-1000-multiple profile timeouts are out of scope (and consider validating multiple-of-1000 on the backend if the UI is meant to be the only writer).

common/data/src/main/java/org/thingsboard/server/common/data/device/profile/DeviceProfileData.java
```diff
@@ -24,6 +25,7 @@
 
 @Schema
 @Data
+@JsonIgnoreProperties(ignoreUnknown = true)
```
ViacheslavKlimov, line 28: Worth flagging that this changes deserialization strictness for the *entire* `DeviceProfileData` DTO, not just the new field — any unknown/renamed/mistyped property will now be silently ignored instead of (potentially) surfacing. The PR description frames this as defensive consistency with sibling DTOs, which is reasonable; just calling it out so it's a conscious decision rather than a side effect of adding the timeout field.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
zzzeebra: Removed. `DefaultTbDeviceProfileCache` depends only on `DeviceProfileService` and `DeviceService` — no cycle back to this service, so `@Lazy` was unnecessary.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
zzzeebra: Moved the iteration to the single-threaded `scheduledExecutor` (the one running `checkStates`). That takes it off the notifications consumer thread, serializes it with the periodic checker so it no longer races on the same `DeviceState`, and preserves event order; a supersede guard re-reads the resolved-timeout cache and drops a task whose value was already replaced. Left the full-map scan as-is — a `DeviceProfileId -> devices` index would add per-connect/reassign upkeep for what is a rare profile-update event.

ui-ngx/src/app/modules/home/components/profile/device-profile.component.ts
zzzeebra: Dropped the guard, set `inactivityTimeoutMs` directly.

ui-ngx/src/app/modules/home/components/profile/device-profile.component.ts
zzzeebra: `DeviceProfileDataValidator` now rejects timeouts that are not a whole number of seconds, so the field always receives a multiple of 1000 and `parseTime` resolves. Chose reject over rounding here to avoid silently mutating an API-supplied value.

common/data/src/main/java/org/thingsboard/server/common/data/device/profile/DeviceProfileData.java
zzzeebra: Intentional — defensive consistency with the sibling DTOs and non-`IGNORE_UNKNOWN` mappers. Keeping it.

vvlladd28 (commented):
## Review summary

Reviewed the **ui-ngx** portion of [Added inactivity timeout setting to Device Profile](https://github.com/thingsboard/thingsboard/pull/15527) (4 of 13 changed files) per the requested scope — `device-profile.component.html`, `device-profile.component.ts`, `shared/models/device.models.ts` and `locale.constant-en_US.json`. Backend, DAO and test files were not reviewed in this pass.

The changed front-end code itself is correct: `tb-time-unit-input` is a seconds-based CVA, so the `ms ↔ sec` conversion in `toInactivityTimeoutSec()` / `prepareFormValue()` is right in both directions; `parseTime()` always resolves now that the backend validator rejects non-whole-second values; `inactivityTimeoutSec` lives outside the `profileData` group and is deleted before save, so it never leaks into the persisted DTO; and `labelText`/`hintText` are pre-translated, matching every other `tb-time-unit-input` call site in the repo. All three UI findings from the previous automated review are addressed in the current head.

One finding, on the creation path — it's in unchanged code so it can't be anchored inline. Details below.

## Additional findings

This is about existing code outside the PR's diff — spotted while reading surrounding context.

- **ui-ngx/src/app/modules/home/components/profile/add-device-profile-dialog.component.ts:108** — **The new setting is unreachable when creating a device profile.** The "Add device profile" wizard's details step mirrors the details page field for field and in the same order (`name` → `defaultRuleChainId` → `defaultDashboardId` → `defaultQueueName` → `defaultEdgeRuleChainId`, plus `image` and `description`), but it does not get `inactivityTimeoutSec`. That makes the inactivity timeout the only details-page field a user cannot set at creation time: they have to create the profile, reopen it, switch to edit mode and save a second time — and until they do, every device already assigned to the fresh profile silently runs on the YAML default, which is precisely the case this feature exists to fix.

  The PR description scopes the UI to the details page, so this may be a deliberate deferral — but it should be an explicit one, because this field now behaves differently from every one of its neighbours in the same form. Three small edits close it:
  - `add-device-profile-dialog.component.ts:116` — add `inactivityTimeoutSec: [0, []]` to `deviceProfileDetailsFormGroup`, reusing the details page's `toInactivityTimeoutSec()`/`SECOND` conversion.
  - `add-device-profile-dialog.component.html:54` — add the same `<tb-time-unit-input>` block right after the `defaultRuleChainId` autocomplete, so the wizard keeps matching the details page's field order.
  - `add-device-profile-dialog.component.ts:190` — add `inactivityTimeoutMs` to the `profileData` literal in `add()`, mapping `0`/empty to `null` exactly as `prepareFormValue()` does.

---

zzzeebra:
Addressed the review finding: the inactivity timeout was not reachable from the device profile creation wizard, so a profile created there always fell back to the YAML default until it was reopened and saved a second time.

`add-device-profile-dialog.component` now carries the field, mirroring the details page: same `tb-time-unit-input` block between **Default rule chain** and **Mobile dashboard**, `inactivityTimeoutSec` control placed in the form group where the details page puts it, and the same `0`/empty → `null` mapping in `add()` as `prepareFormValue()` does. No module changes were needed — `TimeUnitInputComponent` is already exported by `SharedModule`.

### Manual testing

Continuing the numbering from the description (Tests 1–5 covered the details page; these cover the creation wizard). Verified locally on a single-node deployment with PostgreSQL.

- [x] **Test 6 — Wizard saves the value immediately.** Created a profile via the Device profiles table with a 2 second timeout. `profile_data->'inactivityTimeoutMs'` is `2000` without reopening the profile.
- [x] **Test 7 — Empty field stores `null`, not `0`.** Created a profile without touching the field. The key is present with `jsonb_typeof` = `null`, so the runtime falls through to the server default and the validator (which rejects `0`) is not tripped.
- [x] **Test 8 — Unit selector math.** Created a profile with a 1 hour timeout. Stored value is `3600000`, confirming the component's hours→seconds conversion and the `* SECOND` step chain correctly.
- [x] **Test 9 — Round-trip to the details page.** The profile from Test 8 opens on the details page showing **1 hours**, not 3600 seconds.
- [x] **Test 10 — Second entry point.** Created a profile inline from `tb-device-profile-autocomplete` in the device wizard (a separate call site from the profiles table). Field is present and the value persists (`10000`).
- [x] **Test 11 — Empty field does not block the step.** Leaving the field empty, and clearing an entered value, both keep **Next** enabled. `required` is not set on this call site, so `Validators.required` is never added to the inner control.
- [x] **Test 12 — End-to-end, wizard-created profile.** Created a profile via the wizard with a 30 second timeout, assigned a device, posted telemetry. The device went `active=true` with a matching `lastActivityTime`; 69 seconds after the last activity it flipped to `active=false` with a fresh `inactivityAlarmTime`, without the profile ever being saved a second time.

On Test 12, the 69 seconds is the expected shape rather than an exact 30: the timeout elapsed on schedule at +30s, and the transition was recorded by the next `state.defaultStateCheckIntervalInSec` (60s) sweep. The server default of 600s would have delayed the transition by ten minutes, so the profile value set in the wizard is demonstrably the one being applied.

Device reassignment between profiles was exercised incidentally — the test device was moved from a 10s profile to the 30s one and picked up the new timeout.

`ng build` and ESLint are clean on the changed files.

Issue #15541, opened 2026-04-29
RPC monitoring: integrate as per-target companion check on existing transports


smatvienko-tb:
## 1. Background

`tb-monitoring` is an external service that probes a running ThingsBoard
deployment from outside the cluster. On startup it provisions every entity
it needs under a dedicated monitoring tenant — devices, profiles, calculated
fields, a rule chain, the latencies asset, and a public dashboard — and then
runs a continuous loop that, for every configured transport target, sends a
telemetry uplink and validates that the value arrives back over a WebSocket
subscription.

What the loop does **not** cover today is the **server-to-device direction**:
a working telemetry uplink does not prove that a two-way RPC issued by the
REST API will actually reach the device through the rule engine plus the
target transport. There are production failure modes where the uplink
remains green but RPC delivery is broken — for example, a stuck rule engine
partition, or a transport microservice that can publish telemetry inwards
but cannot deliver an RPC outwards.

Adding an opt-in RPC companion check to each transport target closes that
gap with a single new round-trip per cycle, reusing every entity and every
transport client the monitoring service already owns.

### 2.1 Bootstrap (`ThingsboardMonitoringApplication.startMonitoring`)

1. `MonitoringEntityService.checkEntities()` upserts the rule chain for the
   monitoring tenant, finds-or-creates the `[Monitoring] Latencies` asset,
   finds-or-creates the `[Monitoring] Cloud monitoring` dashboard, assigns
   both to the public customer, and remembers the dashboard public link.
2. For each `BaseMonitoringService` bean, `init()` walks its config's
   targets and calls `entityService.checkEntities(config, target)` to
   lazily create the device, device profile and (when
   `monitoring.calculated_fields.enabled`) the calculated field. The
   resulting `DeviceConfig` (id + access token) is written back into the
   target.
3. The scheduler stagger-schedules each `BaseMonitoringService` with
   `(monitoring_rate_ms / count) * i` initial delay.
4. A `:rocket: Monitoring started` info notification is emitted with a
   clickable public-dashboard link.

### 2.2 Per-cycle loop (`BaseMonitoringService.runChecks`)

Each tick is a single REST login → single WebSocket connect → single WS
subscription that covers every device this service owns → for each
`BaseHealthChecker` call `check(WsClient)`. Each phase has its own service
key:

```
LOGIN              "Login"
WS_CONNECT         "WS Connect"
WS_SUBSCRIBE       "WS Subscribe"
[...]

PR #15542, opened 2026-04-29
Monitoring feature: RPC monitoring per transport check (#15541)

smatvienko-tb:
Issue #15541.

## Summary

Adds an opt-in RPC companion check to every transport target in tb-monitoring. Today the loop only proves the device-to-cloud direction (telemetry uplink + WS update); a stuck rule-engine partition or a transport microservice that can publish telemetry inwards but cannot deliver an RPC outwards goes undetected. This PR closes that gap with a single extra round-trip per cycle on the same auto-provisioned device, reusing every transport client tb-monitoring already owns.

Scope (single PR, see ``monitoring/docs/2026-04-29-rpc-monitoring-design.md``):

- **Hook scaffolding** — ``RpcCheckConfig``, ``RpcInfo`` (``ShortNameProvider`` wrapper with stable equals/hashCode), ``TransportMonitoringTarget.rpc``, ``BaseHealthChecker.doRpcCheck()`` no-op invoked between ``checkWsUpdates`` and ``serviceIsOk``, ``Latencies.rpcRoundTrip(key)``.
- **MQTT** — ``initClient`` also subscribes ``v1/devices/me/rpc/request/+`` on the same Paho client and echoes ``params`` back to ``v1/devices/me/rpc/response/{id}``.
- **HTTP** — daemon long-poll thread per target on ``GET /api/v1/{token}/rpc?timeout=1000``; on 200, POST the params back to ``/api/v1/{token}/rpc/{id}``. Idempotent, lifecycle-bound to ``initClient``/``destroyClient``.
- **CoAP** — second ``CoapClient`` opens a CoAP OBSERVE on ``coap://host/api/v1/{token}/rpc``; each notification echoes ``params`` to ``coap://host/api/v1/{token}/rpc/{id}``.
- **LwM2M** — ``doRpcCheck`` only (no ``initClient`` change): server-initiated Read on ``/3/0/0`` via ``POST /api/rpc/twoway/{deviceId}`` with ``method:""Read""``; asserts non-blank instead of uuid-equal — see design doc §4.6 for the rationale.

For MQTT/HTTP/CoAP the call shape is identical (``method:""monitoringCheck""``, ``params:{value:<uuid>}``) and the assertion is uuid-equal, reported under ``<key>RpcRoundTrip``. Failures route through ``MonitoringReporter.serviceFailure(rpcInfo, …)`` so the IncidentManager header gains a distinct ``:red_circle: <transport> RPC (n)`` row alongside the existing ``:red_circle: <transport> (n)``. A telemetry-uplink failure short-circuits the RPC sub-check for that cycle.

## New YAML keys

Each ``monitoring.transports.{mqtt,http,coap,lwm2m}.targets[N]`` accepts an optional ``rpc:`` sub-block:

| Key | Type | Default | Notes |
|-----|------|---------|-------|
| ``rpc.enabled`` | bool | ``false`` | Opt in. |
| ``rpc.request_timeout_ms`` | int | transport ``request_timeout_ms`` (LwM2M: ``6000``) | Sent to ThingsBoard as the RPC ``timeout``. Must be ``< monitoring.rest.request_timeout_ms``. |

Secure transports (``ssl://``, client certs) inherit credentials from the parent target, so ``rpc.enabled: true`` on an ``ssl://`` target gets RPC for free. Full key matrix and per-transport behaviour live in ``monitoring/src/main/resources/README.md``.

## LwM2M Read-vs-echo asymmetry (deliberate)

The other three transports prove the cloud-to-device path by validating that a uuid round-trips. LwM2M can't echo because the existing ``Lwm2mClient`` already serves Object 3 / resource 0 via the standard LwM2M Read; adding a parallel echo handler would either duplicate or destabilise that pathway. So the LwM2M check is a server-initiated Read with a non-blank assertion, and the conservative ``6000 ms`` default reflects the extra hop through Leshan registration. Documented in design §4.6 and the resources README.

## Test plan

- [x] ``mvn test -pl monitoring`` — 75 tests pass (50 new for RPC + 25 pre-existing).
- [x] ``mvn license:check -pl monitoring`` — clean.
- [x] Build the module: ``mvn install -pl monitoring -DskipTests -Dpkg.skip=true``.
- [ ] Smoke-test against a live cluster with ``rpc.enabled: true`` on each transport and confirm ``mqttRpcRoundTripLatency`` etc. show up on the ``[Monitoring] Latencies`` asset.
- [ ] Trigger an RPC outage (e.g., scale rule-engine partition for ``Main`` to 0) with telemetry left healthy and confirm the IncidentManager header gains a ``:red_circle: MQTT RPC (n)`` row independently of the green ``MQTT`` row.
- [ ] Confirm secure-MQTT (``ssl://...:8883``) target with ``rpc.enabled: true`` works without extra credential wiring.

### DIFF
--- a/monitoring/docs/2026-04-29-rpc-monitoring-design.md
+++ b/monitoring/docs/2026-04-29-rpc-monitoring-design.md
@@ -1,6 +1,6 @@
 # RPC Monitoring — Design
 
-| Status | Draft for review |
+| Status | Implemented (PR for #15541) |
 |--------|------------------|
 | Author | Sergii Matviienko (`smatvienko@thingsboard.io`) |
 | Date | 2026-04-29 |
@@ -313,6 +313,42 @@ just another `ServiceFailureNotification` whose `AffectedService` carries
 | Both broken | Two rows: `:red_circle: MQTT (n), :red_circle: MQTT RPC (m)`. |
 | RPC recovers first | MQTT RPC row turns `:large_green_circle:`, MQTT stays red until uplink recovers. |
 
+### 4.6 LwM2M Read-vs-echo asymmetry
+
+Three of the four transports use the same value-echo round-trip:
+
+```
+POST /api/rpc/twoway/{deviceId}  {method:"monitoringCheck", params:{value:<uuid>}}
+                                          ↓ rule engine + transport
+                              device echoes params back, value == uuid
+```
+
+LwM2M is the exception. There is no echo handler to add on the monitoring
+side because the existing `Lwm2mClient` already implements the standard
+Object 3 (Device) Read pathway: `Lwm2mClient.read(server, /3/0/0)` returns
+the value that was last set via `send()` (which is what the telemetry uplink
+phase does immediately before). So the LwM2M companion check uses a
+server-initiated Read instead of an echo:
+
+```
+POST /api/rpc/twoway/{deviceId}  {method:"Read", params:{key:"/3/0/0"}, timeout:6000}
+                                          ↓ rule engine + LwM2M Read
+                              device returns the resource value
+```
+
+The assertion is therefore weaker — non-blank instead of value-equal —
+because the resource value is not under the monitoring service's control on
+the wire (the rule-engine + Leshan registration path round-trips first).
+This is sufficient as a liveness signal for the cloud-to-device LwM2M Read
+pathway and matches what the existing fixtures (`device_profile.json`,
+`resource.json`) already expose. If a stronger assertion is wanted later,
+a Write+Read pair against a custom resource on the test model would close
+the gap, but that is out of scope for this PR.
+
+The `request_timeout_ms` default is therefore `6000` ms for LwM2M (vs `4000`
+ms for the other three transports) because the LwM2M Read traverses one
+extra hop (Leshan registration → device coap loop → response).
+
 ### 4.5 Proposed flow
 
 ```text
@@ -439,55 +475,73 @@ sequenceDiagram
 
 ## 5. Implementation phases
 
-Each phase is an independently reviewable commit:
+Each phase is an independently reviewable commit landing in the same PR:
 
 1. **Land this design paper.** No code changes.
 2. **Add the per-target hook.** Introduce `RpcCheckConfig`,
    `TransportMonitoringTarget.rpc`, `RpcInfo`, and the no-op `doRpcCheck`
    on `BaseHealthChecker`. Wire `BaseHealthChecker.check` to invoke
-   `doRpcCheck` after WS validation when enabled. Tests at base level.
+   `doRpcCheck` after WS validation. Tests at base level.
 3. **MQTT implementation.** Extend `MqttTransportHealthChecker.initClient`
-   to subscribe `v1/devices/me/rpc/request/+` on the existing client when
-   `rpc.enabled`, and implement `doRpcCheck` against
-   `TbClient.handleTwoWayDeviceRPCRequest`. Includes the device-side echo
-   callback (Paho async thread). Unit tests cover happy path, value
-   mismatch, REST exception, response timeout, and RPC-disabled (no
-   subscription).
-4. **Documentation polish.** Update
-   `monitoring/src/main/resources/README.md` with the new YAML keys, a
+   to subscribe `v1/devices/me/rpc/request/+` on the existing Paho client
+   when `rpc.enabled`. Implement the value-echo `doRpcCheck` once on
+   `TransportHealthChecker` so HTTP and CoAP can inherit it. Unit tests
+   cover happy path, value mismatch, REST exception, response timeout,
+   RPC-disabled (no subscription), and idempotent subscribe.
+4. **HTTP implementation.** Extend `HttpTransportHealthChecker.initClient`
+   to spawn an idempotent daemon long-poll thread on
+   `GET /api/v1/{token}/rpc?timeout=1000`; on each 200 response, POST the
+   echoed params back to `/api/v1/{token}/rpc/{id}`. The thread is
+   lifecycle-bound to `initClient` / `destroyClient`. Tests cover the
+   inherited `doRpcCheck` shape plus poll-thread start/stop/idempotency
+   and `pollOnce` behaviour for 200/408/204.
+5. **CoAP implementation.** Extend `CoapTransportHealthChecker.initClient`
+   to open a CoAP OBSERVE on `coap://host/api/v1/{token}/rpc` via a
+   second `CoapClient`; on each notification echo the params back to
+   `coap://host/api/v1/{token}/rpc/{id}`. `destroyClient` cancels the
+   observe and shuts down the rpc client. Tests cover the inherited
+   `doRpcCheck` shape, observe lifecycle, and notification handler
+   parsing.
+6. **LwM2M implementation.** Override `doRpcCheck` only — no
+   `initClient` changes — to issue
+   `POST /api/rpc/twoway/{deviceId} {method:"Read", params:{key:"/3/0/0"}}`
+   and assert non-blank, per §4.6.
+7. **Documentation polish.** Add `monitoring/src/main/resources/README.md`
+   with the per-transport behaviour table, the new YAML keys, the
    secure-MQTT example, and a small troubleshooting note on
-   `request_timeout_ms` ordering vs `monitoring.rest.request_timeout_ms`.
-5. **Follow-ups (separate issues).** HTTP RPC (long-poll
-   `GET /api/v1/{token}/rpc?timeout=…`) and CoAP RPC plug into the same
-   `doRpcCheck` hook with no further base-class changes; tracking issue
-   per transport.
+   `rpc.request_timeout_ms` ordering vs `monitoring.rest.request_timeout_ms`.
 
 ## 6. Open decisions
 
-Listed for reviewer push-back before implementation begins:
+Listed for reviewer push-back before implementation began. All four
+transports ship in this PR; the rest of the table below records the calls
+that survived review.
 
-1. **MQTT only in scope here.** HTTP/CoAP follow as separate issues. The
-   `doRpcCheck` hook is generic so they slot in without further base-class
-   churn.
+1. **All four transports in scope.** Hook + MQTT + HTTP + CoAP + LwM2M ship
+   together so the IncidentManager header gains a coherent set of
+   `:red_circle: <transport> RPC` rows in one go. The hook itself is
+   generic enough that adding more transports later (SNMP, custom) is
+   just another override.
 2. **Two-way RPC only.** Round-trip value validation is the strongest
    signal; one-way RPC would only confirm the request reached the device,
    which the telemetry uplink check already covers from the other
    direction.
 3. **RPC failure key shape.** Use `RpcInfo(TransportInfo)` implementing
    `ShortNameProvider` returning `"<transport.shortName> RPC"`. Alternative:
-   a string constant like `"MQTT RPC"`. Recommendation: the wrapper, so
+   a string constant like `"MQTT RPC"`. Decision: the wrapper, so
    non-default queues (`"MQTT Foo RPC"`) are handled with no extra code.
 4. **Latency suffix.** Reuse the transport target's `getKey()` derivation
    (`mqtt`, `mqttFoo`) and append `RpcRoundTrip` — no separate per-target
    label. This lines up with `<key>Request` / `<key>WsUpdate` already used
    for the telemetry side and means the existing latencies dashboard
    widget needs no schema changes.
-5. **Echo handler placement.** The MQTT echo callback (subscribe handler
-   that publishes `params` back as the response) lives inside
-   `MqttTransportHealthChecker.initClient` — same place where the telemetry
-   client is created, on the same Paho thread. Alternative: a separate
-   helper class. Recommendation: keep it inline. Echo logic is six lines
-   and is only meaningful in the context of one connected client.
+5. **Echo handler placement (MQTT/HTTP/CoAP).** Each transport's echo
+   path lives inline in its `initClient` rather than in a separate helper
+   class. Echo logic is small (a handful of lines per transport) and is
+   only meaningful in the context of one connected client.
+6. **LwM2M Read instead of echo.** See §4.6. The asymmetry is documented
+   in `monitoring/src/main/resources/README.md` so operators understand
+   why the LwM2M assertion is non-blank rather than uuid-equal.
 
 ## 7. Code references (`rc`)
 
--- a/monitoring/docs/2026-04-29-rpc-monitoring-design.md
+++ b/monitoring/docs/2026-04-29-rpc-monitoring-design.md
@@ -1,0 +1,502 @@
+# RPC Monitoring — Design
+
+| Status | Draft for review |
+|--------|------------------|
+| Author | Sergii Matviienko (`smatvienko@thingsboard.io`) |
+| Date | 2026-04-29 |
+| Module | `monitoring/` |
+| Base | `upstream/rc` (CE) |
+
+## 1. Background
+
+`tb-monitoring` is an external service that probes a running ThingsBoard
+deployment from outside the cluster. On startup it provisions every entity
+it needs under a dedicated monitoring tenant — devices, profiles, calculated
+fields, a rule chain, the latencies asset, and a public dashboard — and then
+runs a continuous loop that, for every configured transport target, sends a
+telemetry uplink and validates that the value arrives back over a WebSocket
+subscription.
+
+What the loop does **not** cover today is the **server-to-device direction**:
+a working telemetry uplink does not prove that a two-way RPC issued by the
+REST API will actually reach the device through the rule engine plus the
+target transport. There are production failure modes where the uplink
+remains green but RPC delivery is broken — for example, a stuck rule engine
+partition, or a transport microservice that can publish telemetry inwards
+but cannot deliver an RPC outwards.
+
+Adding an opt-in RPC companion check to each transport target closes that
+gap with a single new round-trip per cycle, reusing every entity and every
+transport client the monitoring service already owns.
+
+## 2. Current architecture (`rc`)
+
+### 2.1 Bootstrap
+
+`ThingsboardMonitoringApplication.startMonitoring` runs once on
+`ApplicationReadyEvent`:
+
+1. `MonitoringEntityService.checkEntities()` upserts the rule chain for the
+   monitoring tenant, finds-or-creates the `[Monitoring] Latencies` asset,
+   finds-or-creates the `[Monitoring] Cloud monitoring` dashboard, assigns
+   both to the public customer, and remembers the dashboard public link.
+2. For each `BaseMonitoringService` bean (transports today; RPC will fold in
+   here too), `init()` walks its config's targets and calls
+   `entityService.checkEntities(config, target)` to lazily create the
+   device, device profile and (when `monitoring.calculated_fields.enabled`)
+   the calculated field. The resulting `DeviceConfig` (id + access token) is
+   written back into the target.
+3. The scheduler stagger-schedules each `BaseMonitoringService` with
+   `(monitoring_rate_ms / count) * i` initial delay — so two services on a
+   10 s rate fire at 0 s and 5 s, not on top of each other.
+4. A `:rocket: Monitoring started` info notification is emitted with a
+   clickable public-dashboard link.
+
+### 2.2 Per-cycle loop (`BaseMonitoringService.runChecks`)
+
+Each tick is a single REST login → single WebSocket connect → single WS
+subscription that covers every device this service owns → for each
+`BaseHealthChecker` call `check(WsClient)`. Each phase has its own service
+key so failures land precisely:
+
+```
+LOGIN              (MonitoredServiceKey.LOGIN)            "Login"
+WS_CONNECT         (MonitoredServiceKey.WS_CONNECT)       "WS Connect"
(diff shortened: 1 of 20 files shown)


smatvienko-tb:
## Code Review

Solid scope, design doc, and test coverage. Main concerns are around the new HTTP long-poll thread and a few smaller correctness/idiom items.

### Overview

Opt-in RPC companion check per transport target reusing the existing per-target client. Echo-on-cloud-side proves the server→device direction that the current uplink+WS loop misses. LwM2M asymmetry (Read with non-blank assertion) is justified in the design doc and acceptable.

### The new daemon Thread (`HttpTransportHealthChecker.rpcPollThread`) — main concern

`new Thread(...).setDaemon(true).start()` is the part you flagged, and yes, it's worth replacing. Specific issues:

1. **`Thread#interrupt()` does not unblock blocking I/O on `RestTemplate`.** Apache HttpClient / `URLConnection` reads are not interruption-aware. Once `pollOnce()` is inside `restTemplate.getForEntity(...)`, `destroyClient()` calling `interrupt()` will set the flag but the read keeps running until the underlying socket times out. Consequence: the thread can outlive `destroyClient()` by `POLL_TIMEOUT_MS` plus the connect/read timeout, and `rpcPollThread = null` (`HttpTransportHealthChecker.java:1011`) drops the only handle while the thread is still alive. Mitigation: configure the `RestTemplate` `ClientHttpRequestFactory` with explicit connect/read timeouts ≤ `POLL_TIMEOUT_MS + small slack`, and call `join(timeoutMs)` after `interrupt()` so destroy actually waits.

2. **No `join()` on shutdown.** Even with the issue above fixed, the current code doesn't wait. Add a bounded `join(...)` so shutdown is deterministic — important if anything ever wraps lifecycle in tests or cycle-restart logic.

3. **Manual `new Thread()` is the wrong primitive here.** Prefer `Executors.newSingleThreadScheduledExecutor(threadFactory)` (with `ThreadFactoryBuilder().setNameFormat(...).setDaemon(true)`) and `scheduleWithFixedDelay`, or — since the service already has many checkers — a *single shared* `ScheduledExecutorService` at the component/service level with one task per target. Per-target platform threads cost ~1 MB stack each and N targets = N threads forever. With a shared pool, lifecycle becomes `Future#cancel(true)` + `awaitTermination`, which is the standard idiom and removes the need for the `AtomicBoolean rpcPolling` guard. (If you can rely on Java 21 + virtual threads, `Thread.ofVirtual().name(...).start(...)` is also fine — but lifecycle still needs `join`.)

4. **`AtomicBoolean rpcPolling` + plain `Thread rpcPollThread` is a partial guard.** `compareAndSet(true, false)` is atomic for the flag but `rpcPollThread = null` and the read of it are not part of that CAS. Concurrent `initClient`/`destroyClient` (or two `destroyClient` calls) can race. The `AtomicBoolean` reads as defensive but doesn't actually protect the field. With a `ScheduledFuture<?>` from a shared executor this whole concern disappears.

5. **`catch (Throwable e)` swallowed at `log.debug`** (`HttpTransportHealthChecker.java:959-960`). Catching `Throwable` and logging at debug will hide `OutOfMemoryError`, `NoClassDefFoundError`, etc. — and the loop happily continues with `Thread.sleep(500)`. Tighten to `catch (Exception e)` and log at `warn` for non-poll-timeout cases. Also: if you do leave a `Throwable` catch, the thread should still set `rpcPolling=false` on `Error` so the next `initClient` can restart it. Right now an `Error` propagates up, the thread dies, `rpcPolling` stays `true`, and the idempotent `compareAndSet(false, true)` guard prevents recovery — silently broken RPC check until the process restarts.

6. **Hard-coded `Thread.sleep(500)` backoff** with no jitter, no exponential growth. If TB is unreachable the loop hammers it 2× per second per target. Suggest exponential backoff with cap (e.g., 500ms → 5s) and jitter.

7. **Long-poll url uses raw `?timeout=` query param** that's also the read-side timeout. Today they happen to align via `POLL_TIMEOUT_MS = 1000`. Worth a one-liner comment that they must stay related, or compute the read timeout from the constant.

### CoAP (`CoapTransportHealthChecker`)

- `postRpcResponse` constructs a fresh `CoapClient` per notification just to POST and immediately calls `shutdown()`. That works but allocates. Consider reusing `rpcCoapClient` with `setURI(...)` per call, or keep it but document why a new client is required.
- `handleRpcNotification` does `idNode.asLong()` — same fragility as HTTP path; defensive `isNumber()` check or `asText()` would be safer if a transport ever returns string ids.
- `onError()` is silent except for a debug log; consider whether a sustained observe failure should surface anywhere (today it's invisible because the only signal is `doRpcCheck` failing on the *next* server-side send, which is the right behavior — worth a comment).

### `BaseHealthChecker` / `TransportHealthChecker` design

- The hook (`doRpcCheck()` between `checkWsUpdates` and `serviceIsOk`) is clean. Good that telemetry-uplink failure short-circuits.
- The `HttpTransportHealthChecker.doRpcCheck()` and `CoapTransportHealthChecker.doRpcCheck()` overrides that just call `super.doRpcCheck()` are dead code — remove them. Same effect, less noise.
- `protected TbClient tbClient` injected on `TransportHealthChecker` (`@Scope(SCOPE_PROTOTYPE)` parent) — works, but consider constructor injection for testability and immutability.

### Smaller correctness items

- `HttpTransportHealthChecker.pollOnce()` (`:993-994`): `restTemplate.postForLocation(url, params == null ? null : params)` — posting `null` body via `RestTemplate` typically yields no `Content-Type`/`Content-Length`, which TB's HTTP transport may reject as 400. Fallback to `JacksonUtil.newObjectNode()` is safer.
- `restTemplate` field visibility was changed from `private` to package-private to allow Mockito injection in tests. Cleaner: keep `private`, expose a constructor or `@Setter` test seam, or use `ReflectionTestUtils.setField` (you already use `ReflectionTestUtils.getField` in this PR's tests, so consistency is fine either way — just don't widen production visibility purely for tests).
- `RpcInfo` reuses `TransportInfo` for stable `equals/hashCode`. LGTM, but make sure `IncidentManager` actually de-dups by `RpcInfo` and not by `TransportInfo` — a single `RpcInfo` per target should produce a *separate* row from the telemetry row. Worth one assertion test on the reporter side if not already covered.
- `getRpcTimeoutMs()` doesn't enforce the documented `< monitoring.rest.request_timeout_ms` constraint. If a misconfigured YAML sets it equal/greater, the `tbClient.handleTwoWayDeviceRPCRequest` call will hit the REST timeout *before* TB times out the RPC, producing a false negative. Consider validating at startup (fail-fast) or clamping with a warn log.

### Tests

50 new tests is good coverage. Notes:

- `initClientStartsPollingThreadWhenRpcEnabled` asserts `isDaemon()` on a freshly-started thread. There's an implicit race — if you switch to an executor, this test asserts an internal field that no longer exists, but the *behavior* (poll happens) is what matters. Suggest asserting through `verify(restTemplate, timeout(500).atLeastOnce()).getForEntity(...)` instead of inspecting field shape.
- No test for the failure mode where the poll thread dies (e.g., `Error` thrown). With the current implementation, that's a permanent breakage and there's no test to catch it. Worth adding once the lifecycle is reworked.
- No test that `destroyClient()` actually stops polling (as opposed to just nulling the field). Same point as above — assert via `verifyNoMoreInteractions(restTemplate)` after a quiescent period.

### Suggested minimum changes (priority order)

1. Replace `new Thread` + `AtomicBoolean` with a `ScheduledExecutorService` (per-checker single-thread, or a shared one on the service); manage with `Future#cancel(true)` and `awaitTermination`.
2. Set `RestTemplate` connect/read timeout explicitly so `cancel()`/`interrupt()` actually unblocks within bounded time; verify by running shutdown under load.
3. Tighten `catch (Throwable)` → `catch (Exception)`; add `warn`-level log for non-timeout errors; ensure `rpcPolling`/lifecycle state recovers if the loop dies.
4. Drop the no-op `doRpcCheck()` overrides in HTTP/CoAP checkers.
5. Fall back to empty `ObjectNode` instead of posting `null` body in `pollOnce`.
6. Validate `rpc.request_timeout_ms < monitoring.rest.request_timeout_ms` at startup.

Nothing blocking on the *feature* — the design is right. The thread management is the rough edge worth polishing before merge.

smatvienko-tb:
Pushed [`20828ca2980`](https://github.com/thingsboard/thingsboard/pull/15542/commits/20828ca2980) addressing the review.

### Daemon thread (main concern)

1. **`Thread#interrupt()` does not unblock blocking I/O** — fixed. `RestTemplate` read timeout is now bounded to `POLL_TIMEOUT_MS + 1s` so a blocked socket terminates within ~2 s of `destroyClient()`.
2. **No `join()` on shutdown** — replaced with `ScheduledExecutorService.shutdownNow()` + `awaitTermination(5s)` (logged WARN if it doesn't terminate).
3. **Manual `new Thread()`** — replaced with `Executors.newSingleThreadScheduledExecutor` + daemon `ThreadFactory` + `scheduleWithFixedDelay(this::pollTask, 0, 1, MS)`. Lifecycle is now `Future#cancel(true)` + `awaitTermination`. The `AtomicBoolean` is gone.
4. **Partial guard race** — gone with the `AtomicBoolean`. Lifecycle is just the `Future<?>` reference.
5. **`catch (Throwable e)` at debug** — tightened to `catch (Exception e)` at WARN. `pollTask` also resets state if the scheduled task ever stops; `initClient` re-schedules when `rpcPollFuture.isDone()` (covers an `Error` killing the task — verified by a new `initClientReschedulesAfterPollFutureDies` test).
6. **Hard-coded 500 ms backoff** — replaced with exponential 500 ms → 5 s + 0–50 % jitter via `ThreadLocalRandom`. Reset to 0 on a successful poll. New test: `pollTaskBackoffDoesNotKillScheduler`.
7. **Long-poll `?timeout=` vs read timeout** — explicit comment in `pollOnce` and the read timeout is now derived from `POLL_TIMEOUT_MS` directly so they cannot drift.

### CoAP

- `postRpcResponse` allocating a fresh client — kept and commented why: the observe relation owns `rpcCoapClient`, and reusing it via `setURI(...)` would race with incoming notifications.
- `idNode.asLong()` — added `isNumber()` guard before parsing.
- `onError()` — comment added explaining that sustained observe failure surfaces via the next server-side `doRpcCheck()` timing out (which is the right behavior).
- `catch (Throwable)` in `handleRpcNotification` tightened to `catch (Exception)` and bumped to WARN.

### `BaseHealthChecker` / `TransportHealthChecker` design

- **No-op `doRpcCheck()` overrides in HTTP/CoAP** — pushing back: these are not pure noise. `TransportHealthChecker.doRpcCheck` is `protected` and lives in `o.t.m.service.transport`; the `*RpcTest` classes live in `o.t.m.service.transport.impl`. Removing the override breaks the test build because `protected` cross-package access requires a subclass relationship the test doesn't have. Kept the override and added a comment explaining the role. Open to swapping to `ReflectionTestUtils.invokeMethod` if you'd rather, but that loses checked-exception type fidelity in `assertThatThrownBy`.
- **Constructor injection for `tbClient`** — agreed in principle but out of scope for this PR; would be a parent-class refactor touching every transport checker.

### Smaller items

- **`postForLocation(url, null)`** — fixed. Falls back to `JacksonUtil.newObjectNode()` when `params` is absent. New test: `pollOncePostsEmptyBodyWhenParamsMissing`.
- **`restTemplate` field visibility** — restored to `private`. Tests use `ReflectionTestUtils.setField` (consistent with the existing `tbClient`/`reporter` injection).
- **`RpcInfo` vs `TransportInfo` dedup** — added `RpcInfoTest#rpcInfoIncidentKeyDiffersFromTelemetryInfoForSameTarget` asserting `toString()` and `equals()` differ for the same target, so `IncidentManager` (string-keyed) treats them as separate rows.
- **`getRpcTimeoutMs() < monitoring.rest.request_timeout_ms`** — fail-fast added in `TransportHealthChecker.initialize()`. New `TransportHealthCheckerInitializeTest` covers pass / fail-on-equal / fail-on-greater / fallback-to-transport-timeout.

### Tests

- `initClientStartsPollingThreadWhenRpcEnabled` rewritten to verify behavior (`verify(restTemplate, timeout(2000).atLeastOnce()).getForEntity(...)`) instead of asserting on `Thread.isDaemon()`.
- `destroyClientStopsPollingThread` rewritten to assert quiescence (`clearInvocations` + `verifyNoMoreInteractions` after a quiet window).
- New `initClientReschedulesAfterPollFutureDies` covers the loop-died recovery path.
- 85 tests in the monitoring module now (was 79); all green.

smatvienko-tb:
## Review — RPC monitoring per transport check

Solid, scoped, opt-in feature with proportionate test coverage. The design doc + module README make the asymmetries (LwM2M Read vs uuid-echo) and the timeout ordering invariant easy to reason about. Comments below are mostly suggestions, with two items I'd treat as blocking-ish.

### Correctness / safety

- **`MqttTransportHealthChecker.echoRpcRequest` should not propagate exceptions to the Paho callback.** It is registered as `IMqttMessageListener` (`mqttClient.subscribe(..., this::echoRpcRequest)`) and currently `throws Exception`. A malformed payload, broken JSON, or a transient publish failure will bubble into Paho's callback dispatcher; depending on Paho version this can disconnect the session or leave the listener in a bad state — masking the very telemetry uplink that the rest of this checker depends on. Wrap the body in `try/catch` and log + skip, the same way `CoapTransportHealthChecker.handleRpcNotification` already does.

- **LwM2M default timeout vs `monitoring.rest.request_timeout_ms`.** The fail-fast in `TransportHealthChecker.initialize` is the right idea, but the LwM2M default of `6000 ms` will trip it on any deployment where `monitoring.rest.request_timeout_ms` is at its existing (smaller) default. This will surface as a startup `IllegalStateException` rather than a config warning. Worth either (a) verifying the default REST timeout in `tb-monitoring.yml` is `> 6000`, or (b) softening the message to point at the exact YAML keys to bump (`monitoring.rest.request_timeout_ms` and the `LWM2M_RPC_REQUEST_TIMEOUT_MS` env var). The current message is good but doesn't name the env var users will reach for.

### Concurrency / lifecycle

- **`HttpTransportHealthChecker` poll loop delay of `1 ms`.** `scheduleWithFixedDelay(..., 0L, 1L, MILLISECONDS)` is essentially "as fast as possible" — the actual rate is gated by the long-poll's read timeout (`POLL_TIMEOUT_MS + slack`), which is fine when the server long-polls. But for any path that returns immediately (`204`, bodyless `200`, `408`, server short-circuits), the loop becomes a tight reschedule. Consider `0L, POLL_TIMEOUT_MS / 4, MILLISECONDS` (or just `0L, 0L`) so the intent is clear; `1 ms` looks like a placeholder.

- **`destroyClient` interruption is best-effort.** `restTemplate.getForEntity` (default JDK `HttpURLConnection`) does not honor `Thread.interrupt()` on a blocking read, so `cancel(true)` won't unblock the poller until the read timeout fires. The 5 s `SHUTDOWN_TIMEOUT_MS` is comfortably above `POLL_TIMEOUT_MS + slack = 2000 ms`, so this is fine in practice — but the warn-log path will fire if anyone bumps `POLL_READ_TIMEOUT_SLACK_MS`. Worth a comment near `SHUTDOWN_TIMEOUT_MS` so it isn't silently broken later.

- **CoAP `postRpcResponse` allocates a fresh `CoapClient` per echo.** The inline comment explains why (observe relation owns `rpcCoapClient`, `setURI` would race). Agreed, but please add a `try-with-resources`-ish hint that `client.shutdown()` runs in `finally` even if construction throws — the current code is fine because `new CoapClient(uri)` does no I/O, but a future change to that ctor could leak a socket.

### Testability / API surface

- **Package-private mutable fields for test access** — `MqttTransportHealthChecker.mqttClient`, `CoapTransportHealthChecker.rpcCoapClient`, `rpcObserveRelation` were dropped from `private` to package-private so tests can `checker.mqttClient = mock(...)`. This works but it's inconsistent with the rest of the same tests, which use `ReflectionTestUtils.setField`. Pick one: either go all-`ReflectionTestUtils` (preserves `private`), or annotate with `@VisibleForTesting`. Right now there's nothing in the source that signals "do not touch from non-test code."

- **The "package-access bridge" delegating overrides** in MQTT/HTTP/CoAP impls (`@Override protected void doRpcCheck() throws Exception { super.doRpcCheck(); }`) exist only to give the same-package test class visibility into a `protected` method declared in a parent package. The doc comment is helpful, but this is a smell — the public API now has three identical methods whose only purpose is to be callable from one test. Either invoke via `ReflectionTestUtils.invokeMethod`, or make the call site go through a public test-only helper, or accept it as-is and de-duplicate the comment to a single source.

- **`Thread.sleep(100)` / `Thread.sleep(200)` in `HttpTransportHealthCheckerRpcTest.destroyClientStopsPolling` and `initClientDoesNotStartPollingWhenRpcDisabled`** will be flaky on a loaded CI runner. Consider replacing with `Awaitility` or `Mockito.verify(..., timeout(...).times(0))` so the test fails deterministically on a regression rather than intermittently on slow runners.

### Minor / style

- `Lwm2mTransportHealthChecker.READ_RESOURCE_PATH` is declared between methods rather than at the top of the class with the other fields. Move it up.
- `RpcInfo.toString()` ends with the literal `" RPC"`; the test asserts `endsWith("RPC")`. That's correct but means any future change of the suffix breaks all four impl tests at once. Consider extracting `RPC_SUFFIX` so the test asserts against the constant, not a duplicated string.
- The 556-line design doc inside `monitoring/docs/` is great context but is a binary commit cost on every future module checkout. If the team has a docs repo, consider linking from a short stub here.
- `tb-monitoring.yml` examples for secure-MQTT are commented-out but documented in the README and the PR body — fine, just confirm a smoke test of `ssl://...:8883` with `rpc.enabled: true` actually inherits credentials (your unchecked test-plan box).

### What's good

- Failure attribution via `RpcInfo` (distinct `getShortName` + `equals`/`hashCode`) so the IncidentManager row for telemetry doesn't collapse with the RPC row — and the test in `RpcInfoTest.rpcInfoIncidentKeyDiffersFromTelemetryInfoForSameTarget` makes that contract explicit. Nice.
- Telemetry-uplink failure short-circuiting the RPC sub-check (`BaseHealthChecker.check` flow) is the right call — avoids double-counting a single broken transport.
- The fail-fast invariant `rpcTimeoutMs < monitoring.rest.request_timeout_ms` at `initialize()` is exactly the kind of foot-gun you want caught at boot instead of as a confusing false-negative under load.
- Per-transport asymmetry (LwM2M Read vs uuid echo) is documented in the design doc, the README, and the test (`Lwm2mTransportHealthCheckerRpcTest.doRpcCheckHappyPathReadsManufacturerResource`) — three places, all consistent.

### Summary

Two things I'd treat as must-fix before merge:
1. `MqttTransportHealthChecker.echoRpcRequest` should not propagate exceptions back into Paho's callback.
2. Sanity-check the LwM2M `6000 ms` default against the deployed `monitoring.rest.request_timeout_ms` so this PR doesn't break LwM2M targets that opt in with defaults.

Everything else is suggestions / nits.

smatvienko-tb:
Thanks for the thorough review. Addressed in 19e66f25beb:

**Both blocking-ish items**
1. `MqttTransportHealthChecker.echoRpcRequest` is now `try/catch`-wrapped and no longer propagates exceptions into Paho's `IMqttMessageListener` dispatcher — matches `CoapTransportHealthChecker.handleRpcNotification`.
2. Bumped `REST_REQUEST_TIMEOUT_MS` default `5000 → 8000` so the LwM2M `6000 ms` default no longer trips the fail-fast invariant out of the box. Also expanded the `IllegalStateException` message to name `REST_REQUEST_TIMEOUT_MS` and the per-transport `<TYPE>_RPC_REQUEST_TIMEOUT_MS` env var so operators know exactly what to bump.

**Concurrency / lifecycle nits**
- HTTP `1 ms` poll delay replaced with `POLL_TIMEOUT_MS / 4` (250 ms) and documented — bounds tight-spin on immediate returns (204 / bodyless 200 / 408).
- Added a comment on `SHUTDOWN_TIMEOUT_MS` explaining the bound vs `POLL_TIMEOUT_MS + slack` and the JDK `HttpURLConnection` interrupt behaviour.
- `CoapTransportHealthChecker.postRpcResponse` now constructs the `CoapClient` inside the `try` block with a null-guarded `shutdown()` in `finally`, so a future Californium ctor that opens a socket can't leak on init failure.

**Testability / API surface**
- `@VisibleForTesting` (Guava) added to `mqttClient`, `rpcCoapClient`, `rpcObserveRelation`, `pollTask`, `pollOnce`, `echoRpcRequest`, `handleRpcNotification`, `postRpcResponse`, and the three `doRpcCheck` bridge overrides.
- The bridge-rationale comment is now consolidated to a single Javadoc on `TransportHealthChecker.doRpcCheck` (kept the bridges so tests don't need `ReflectionTestUtils.invokeMethod`).
- `Thread.sleep(...)` in `HttpTransportHealthCheckerRpcTest` (`destroyClientStopsPolling`, `initClientDoesNotStartPollingWhenRpcDisabled`) replaced with `Mockito.after(...).never()` — deterministic on slow CI runners.

**Style**
- `Lwm2mTransportHealthChecker.READ_RESOURCE_PATH` moved to the field block.
- `RpcInfo.RPC_SUFFIX` extracted; all four impl tests + `RpcInfoTest` now assert against the constant.

**Not addressed**
- Design doc relocation: keeping `monitoring/docs/2026-04-29-rpc-monitoring-design.md` in-tree for now since there's no separate docs repo.
- Secure-MQTT `ssl://...:8883` smoke test — that's a test-plan checkbox, not a code change. Will run before merge.

All 60 RPC-related tests pass locally (`mvn test -pl monitoring -Dtest='*HealthChecker*RpcTest,RpcInfoTest,TransportHealthCheckerInitializeTest,BaseHealthCheckerRpcHookTest'`).

Issue #1213, opened 2018-11-02
MQTT over SSL & Invalid message received


Dfghhj:
I configured it according to the documentation (https://thingsboard.io/docs/user-guide/mqtt-over-ssl/) and tried to connect using a python script (one-way-ssl-mqtt-client.py).
The console prints "Invalid message received" and msg = [MqttMessage[fixedHeader=, variableHeader=, payload=]]。
![image](https://user-images.githubusercontent.com/20330850/47910319-71943380-decd-11e8-8b58-10511ca5a0d5.png)
![image](https://user-images.githubusercontent.com/20330850/47910812-e9af2900-dece-11e8-998f-eb9d88511d17.png)

anujdeshpande:
@Dfghhj how did you fix this? I am running into the same issue.

ashvayka:
Hi, we have recently fixed main issue introduced in v2.2 but according to your stack, it looks like you have not configured SSL on the server side (in thingsboard.yml). Please retest on the latest master version.

PR #15545, opened 2026-04-30
Fix tb-cassandra docker build: dpkg conffile prompt blocks openjdk-17-jre-headless upgrade

smatvienko-tb:
## Summary

The `tb-cassandra` docker image build fails on `lts-4.2` (and the same Dockerfile on `rc`/`master`) when apt-get pulls in a newer `openjdk-17-jre-headless` to satisfy cassandra's `java11-runtime` dependency. The base image `thingsboard/openjdk17:bookworm-slim` ships a customized `/etc/java-17-openjdk/security/java.security`; on upgrade dpkg detects the modified conffile and prompts:

```
Configuration file '/etc/java-17-openjdk/security/java.security'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
     Y or I  : install the package maintainer's version
     N or O  : keep your currently-installed version
       D     : show the differences between the versions
       Z     : start a shell to examine the situation
*** java.security (Y/I/N/O/D/Z) [default=N] ? dpkg: error processing package
  openjdk-17-jre-headless:amd64 (--configure):
  end of file on stdin at conffile prompt
```

Because there is no stdin in the docker build, dpkg fails. `openjdk-17-jre-headless` ends up unconfigured, which cascades into the misleading downstream error `cassandra depends on openjdk-11-jre-headless | openjdk-8-jre-headless | java11-runtime | java8-runtime` and the build aborts.

Setting `DEBIAN_FRONTEND=noninteractive` alone does **not** fix this — that env var only suppresses *debconf* prompts, not *dpkg conffile* prompts, which are a separate mechanism.

## Fix

Pass `-o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"` to both `apt-get install` invocations so dpkg silently keeps the base image's customized conffile. Also set `DEBIAN_FRONTEND=noninteractive` for completeness.

The choice of `--force-confold` (keep installed/customized version) over `--force-confnew` is deliberate — the modified `java.security` is part of the ThingsBoard base image's hardening and we want to preserve it.

Reference failing build: [TestBlackBoxMicroservices #1213 / build 106476](https://builds.thingsboard.io/buildConfiguration/ThingsBoard_TestBlackBoxMicroservices/106476).

## Test plan

- [x] **Reproduce on bare base image (no fix)**: minimal Dockerfile with the same apt-get sequence on `thingsboard/openjdk17:bookworm-slim` (linux/amd64) fails at the exact same `*** java.security (Y/I/N/O/D/Z) [default=N] ?` line, exit code 100. Matches CI build 106476 line-for-line.
- [x] **Verify fix works**: same minimal Dockerfile with the proposed change builds clean — dpkg now logs `==> Keeping old config file as default.` and proceeds to `Setting up openjdk-17-jre-headless` → `Setting up postgresql-16` → `Setting up cassandra (4.0.20)` → `Setting up cassandra-tools (4.0.20)` → image exported.
- [ ] CI: TestBlackBoxMicroservices on this branch is the real test.

## Notes / follow-ups

The same one-line Dockerfile is identical on `rc` and `master`. Once this lands on `lts-4.2`, the same fix should be ported there (or applied centrally in the `thingsboard/openjdk*` base image via `/etc/apt/apt.conf.d/90-noninteractive` so every consumer inherits it).

### DIFF
--- a/msa/tb/docker-cassandra/Dockerfile
+++ b/msa/tb/docker-cassandra/Dockerfile
@@ -42,6 +42,10 @@ ENV CASSANDRA_LOG=/var/log/cassandra
 
 COPY logback.xml ${pkg.name}.conf start-db.sh stop-db.sh start-tb.sh upgrade-tb.sh install-tb.sh ${pkg.name}.deb /tmp/
 
+# Keep base image's customized conffiles (e.g. /etc/java-17-openjdk/security/java.security)
+# when apt upgrades openjdk-17-jre-headless transitively as cassandra's java11-runtime provider;
+# without this dpkg blocks on a non-interactive conffile prompt and the build fails.
+ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update \
     && apt-get install -y --no-install-recommends wget nmap procps gnupg2 \
     && echo "deb http://apt.postgresql.org/pub/repos/apt/ $(. /etc/os-release && echo -n $VERSION_CODENAME)-pgdg main" | tee --append /etc/apt/sources.list.d/pgdg.list > /dev/null \
@@ -49,7 +53,9 @@ RUN apt-get update \
     && echo "deb https://debian.cassandra.apache.org 40x main" | tee -a /etc/apt/sources.list.d/cassandra.sources.list > /dev/null \
     && wget -q https://downloads.apache.org/cassandra/KEYS -O- | apt-key add - \
     && apt-get update \
-    && apt-get install -y --no-install-recommends cassandra cassandra-tools postgresql-${PG_MAJOR} \
+    && apt-get install -y --no-install-recommends \
+         -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" \
+         cassandra cassandra-tools postgresql-${PG_MAJOR} \
     && rm -rf /var/lib/apt/lists/* \
     && update-rc.d cassandra disable \
     && update-rc.d postgresql disable \


smatvienko-tb:
## Code Review

### Overview
Targeted fix for a real, reproducible CI failure on `tb-cassandra` docker build. Root cause analysis is solid: dpkg conffile prompt (separate mechanism from debconf) blocks the build when apt transitively upgrades `openjdk-17-jre-headless` and detects the base image's modified `/etc/java-17-openjdk/security/java.security`. The fix — `--force-confdef --force-confold` — is the standard, correct remedy, and keeping the hardened conffile is the right policy choice.

### Strengths
- Excellent PR description with reproduction, exact failing line, and verification of the fix.
- Inline comment in the Dockerfile explains the *why*, not just the *what* — future-proof against someone deleting the line as "redundant".
- `--force-confold` (vs. `confnew`) is correctly justified: the customized `java.security` is part of the base image hardening.
- Reference to the failing TeamCity build (#106476) is helpful for traceability.

### Issues / Suggestions

**1. PR description doesn't match the diff (minor).**
The summary says *"Pass ... to **both** apt-get install invocations"*, but the diff only adds the dpkg options to the second invocation (`cassandra cassandra-tools postgresql-*`). The first one (`wget nmap procps gnupg2`) is unchanged. In practice the first install pulls only fresh packages and won't trigger conffile prompts, so this is fine — but please either update the description or apply the options to both for consistency/defense-in-depth.

**2. `ENV DEBIAN_FRONTEND=noninteractive` leaks into the runtime image.**
Using `ENV` persists this into the final image's environment, which is generally considered a Docker anti-pattern (it can affect any apt operation a downstream user runs in the container). The conventional fix is `ARG DEBIAN_FRONTEND=noninteractive` for build-time only. That said, the comment in the PR correctly notes this var alone doesn't fix the issue (`--force-conf*` is what does), so removing the ENV line entirely would also be defensible — the dpkg options are sufficient.

**3. Consider centralizing.**
The follow-up note already calls this out: a permanent fix in `thingsboard/openjdk*` base images via `/etc/apt/apt.conf.d/90-noninteractive` would prevent every downstream consumer from hitting this. Worth filing as an issue/ticket so it doesn't get lost.

**4. Port to `rc` and `master`.**
Description acknowledges this — please make sure the port is tracked (separate PR or cherry-pick) so the same regression doesn't surface there once cassandra/openjdk versions tick again.

### Risk Assessment
- **Correctness**: ✅ Fix matches root cause. `--force-confold` keeps existing files, which is what we want for the hardened `java.security`.
- **Performance**: No impact (build-time only).
- **Security**: ✅ Preserves the hardened `java.security` rather than overwriting with maintainer defaults — actually a *security-positive* choice.
- **Tests**: Dockerfile change; manual reproducer + CI rerun is appropriate. No unit-test gap.

### Verdict
LGTM with minor cleanups (description ↔ diff alignment, optional `ARG` over `ENV`). Safe to merge once CI on this branch goes green.

PR #15549, opened 2026-04-30
Fixed reset of Entities table widget pagination on dashboard state update

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
@@ -42,7 +42,7 @@ import {
 import { IWidgetSubscription } from '@core/api/widget-api.models';
 import { UtilsService } from '@core/services/utils.service';
 import { TranslateService } from '@ngx-translate/core';
-import { deepClone, hashCode, isDefined, isDefinedAndNotNull, isEqual, isObject, isUndefined } from '@core/utils';
+import { deepClone, hashCode, isDefined, isDefinedAndNotNull, isObject, isUndefined } from '@core/utils';
 import cssjs from '@core/css/css';
 import { CollectionViewer, DataSource } from '@angular/cdk/collections';
 import { DataKeyType } from '@shared/models/telemetry/telemetry.models';
@@ -79,6 +79,7 @@ import {
   isValidPageStepIncrement,
   noDataMessage,
   prepareTableCellButtonActions,
+  resetPaginationOnStateChange,
   RowStyleInfo,
   TableCellButtonActionDescriptor,
   TableWidgetDataKeySettings,
@@ -267,16 +268,7 @@ export class EntitiesTableWidgetComponent extends PageComponent implements OnIni
     if (this.displayPagination) {
       this.sort.sortChange.pipe(takeUntil(this.destroy$)).subscribe(() => this.paginator.pageIndex = 0);
 
-      let currentStateParams = deepClone(this.ctx.stateController?.getStateParams());
-      this.ctx.stateController?.stateChanged().pipe(
-        takeUntil(this.destroy$)
-      ).subscribe(() => {
-        const newStateParams = this.ctx.stateController.getStateParams();
-        if (!isEqual(currentStateParams, newStateParams)) {
-          currentStateParams = deepClone(newStateParams);
-          this.paginator.firstPage();
-        }
-      });
+      resetPaginationOnStateChange(this.ctx, this.paginator, this.destroy$);
 
       this.ctx.aliasController?.filtersChanged.pipe(
         takeUntil(this.destroy$)
--- a/ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
@@ -46,7 +46,7 @@ import { deepClone, hashCode, isDefined, isDefinedAndNotNull, isObject, isUndefi
 import cssjs from '@core/css/css';
 import { CollectionViewer, DataSource } from '@angular/cdk/collections';
 import { DataKeyType } from '@shared/models/telemetry/telemetry.models';
-import { BehaviorSubject, fromEvent, merge, Observable, of, Subject } from 'rxjs';
+import { BehaviorSubject, fromEvent, merge, Observable, of, Subject, Subscription } from 'rxjs';
 import { emptyPageData, PageData } from '@shared/models/page/page-data';
 import { EntityId } from '@shared/models/id/entity-id';
 import { EntityType, entityTypeTranslations } from '@shared/models/entity-type.models';
@@ -165,6 +165,7 @@ export class EntitiesTableWidgetComponent extends PageComponent implements OnIni
   private subscription: IWidgetSubscription;
   private widgetResize$: ResizeObserver;
   private destroy$ = new Subject<void>();
+  private paginationResetsSubscription: Subscription;
 
   private defaultPageSize;
   private defaultSortOrder = 'entityName';
@@ -248,6 +249,7 @@ export class EntitiesTableWidgetComponent extends PageComponent implements OnIni
     if (this.widgetResize$) {
       this.widgetResize$.disconnect();
     }
+    this.paginationResetsSubscription?.unsubscribe();
     this.destroy$.next();
     this.destroy$.complete();
   }
@@ -266,7 +268,7 @@ export class EntitiesTableWidgetComponent extends PageComponent implements OnIni
     });
 
     if (this.displayPagination) {
-      setupPaginationResets(this.ctx, this.paginator, this.sort, this.destroy$);
+      this.paginationResetsSubscription = setupPaginationResets(this.ctx, this.paginator, this.sort);
 
       this.ctx.aliasController?.filtersChanged.pipe(
         takeUntil(this.destroy$)
--- a/ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
@@ -42,7 +42,7 @@ import {
 import { IWidgetSubscription } from '@core/api/widget-api.models';
 import { UtilsService } from '@core/services/utils.service';
 import { TranslateService } from '@ngx-translate/core';
-import { deepClone, hashCode, isDefined, isDefinedAndNotNull, isObject, isUndefined } from '@core/utils';
+import { deepClone, hashCode, isDefined, isDefinedAndNotNull, isEqual, isObject, isUndefined } from '@core/utils';
 import cssjs from '@core/css/css';
 import { CollectionViewer, DataSource } from '@angular/cdk/collections';
 import { DataKeyType } from '@shared/models/telemetry/telemetry.models';
@@ -266,6 +266,17 @@ export class EntitiesTableWidgetComponent extends PageComponent implements OnIni
 
     if (this.displayPagination) {
       this.sort.sortChange.pipe(takeUntil(this.destroy$)).subscribe(() => this.paginator.pageIndex = 0);
+
+      let currentStateParams = deepClone(this.ctx.stateController?.getStateParams());
+      this.ctx.stateController?.stateChanged().pipe(
+        takeUntil(this.destroy$)
+      ).subscribe(() => {
+        const newStateParams = this.ctx.stateController.getStateParams();
+        if (!isEqual(currentStateParams, newStateParams)) {
+          currentStateParams = deepClone(newStateParams);
+          this.paginator.firstPage();
+        }
+      });
 
       this.ctx.aliasController?.filtersChanged.pipe(
         takeUntil(this.destroy$)
--- a/ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
@@ -79,8 +79,8 @@ import {
   isValidPageStepIncrement,
   noDataMessage,
   prepareTableCellButtonActions,
-  resetPaginationOnStateChange,
   RowStyleInfo,
+  setupPaginationResets,
   TableCellButtonActionDescriptor,
   TableWidgetDataKeySettings,
   TableWidgetSettings,
@@ -266,9 +266,7 @@ export class EntitiesTableWidgetComponent extends PageComponent implements OnIni
     });
 
     if (this.displayPagination) {
-      this.sort.sortChange.pipe(takeUntil(this.destroy$)).subscribe(() => this.paginator.pageIndex = 0);
-
-      resetPaginationOnStateChange(this.ctx, this.paginator, this.destroy$);
+      setupPaginationResets(this.ctx, this.paginator, this.sort, this.destroy$);
 
       this.ctx.aliasController?.filtersChanged.pipe(
         takeUntil(this.destroy$)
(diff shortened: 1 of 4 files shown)


ui-ngx/src/app/modules/home/components/widget/lib/entity/entities-table-widget.component.ts
```diff
@@ -267,6 +267,15 @@ export class EntitiesTableWidgetComponent extends PageComponent implements OnIni
...
       this.sort.sortChange.pipe(takeUntil(this.destroy$)).subscribe(() => this.paginator.pageIndex = 0);
 
+      this.ctx.aliasController?.entityAliasesChanged.pipe(
+        takeUntil(this.destroy$)
+      ).subscribe((aliasIds) => {
+        let currentEntityAliasId = this.ctx.defaultSubscription.options.datasources?.[0]?.entityAliasId;
+        if (currentEntityAliasId && aliasIds.includes(currentEntityAliasId)) {
+          this.paginator.firstPage();
+        }
+      });
```
vvlladd28: Need check in alarm table and timeseries table widgets

ui-ngx/src/app/modules/home/components/widget/lib/table-widget.models.ts
```diff
@@ -516,3 +517,15 @@ export function isValidPageStepIncrement(value: number): boolean {
 export function isValidPageStepCount(value: number): boolean {
   return Number.isInteger(value) && value > 0 && value <= 100;
 }
+
+export function resetPaginationOnStateChange(ctx: WidgetContext,
```
vvlladd28: Two design thoughts on this helper, both minor:

1. The sibling reset right above each call site — `this.sort.sortChange.pipe(...).subscribe(() => this.paginator.pageIndex = 0)` — is duplicated across all the table widgets and is just as extractable. Since you're already creating a shared pagination-helper home in this file, it might be worth pulling that one out too (or a combined `setupPaginationResets(ctx, paginator, sort, destroy$)`) so the two related concerns don't drift apart over time.
2. Taking `destroy$` and returning `void` is a slightly leaky contract — the caller must own a destroy `Subject` and the subscription can't be torn down any other way. Returning the `Subscription` (or accepting an `Observable<void>` notifier instead of the `Subject`) would be a cleaner, more composable signature.

ui-ngx/src/app/modules/home/components/widget/lib/table-widget.models.ts
```diff
@@ -516,3 +517,15 @@ export function isValidPageStepIncrement(value: number): boolean {
...
+                                             paginator: MatPaginator,
+                                             destroy$: Subject<void>): void {
+  ctx.stateController?.stateChanged().pipe(
```
vvlladd28: Null-handling here is inconsistent and the `?.` doesn't actually protect: if `stateController` were null, `stateController?.stateChanged()` evaluates to `undefined` and the following `.pipe(...)` throws anyway. Meanwhile `map` below calls `ctx.stateController.getStateParams()` without a guard while `startWith` uses `ctx.stateController?.getStateParams()`. Elsewhere in these widgets `stateController` is dereferenced without `?.` (e.g. `alarms-table-widget.component.ts` ~line 983), which suggests it's reliably non-null here — in which case the `?.` could just be dropped for consistency. If it genuinely can be null, a real guard is needed instead.

ui-ngx/src/app/modules/home/components/widget/lib/table-widget.models.ts
```diff
@@ -516,3 +517,15 @@ export function isValidPageStepIncrement(value: number): boolean {
...
+    map(() => deepClone(ctx.stateController.getStateParams())),
```
vvlladd28: A few notes on this operator chain (which is effectively "fire only when the state params actually change"):

- **Timing / staleness:** `stateChanged()` is fed by `stateChangedSubject.next()`. On the URL / query-param-driven path (`StateControllerComponent.ngOnInit`), `next()` is emitted *before* `onStateChanged()` rebuilds `stateObject`, and the follow-up `gotoState(id, false)` doesn't re-emit — so `getStateParams()` here still returns the *previous* params at emission time, `distinctUntilChanged` sees no change, and pagination isn't reset. The programmatic `openState`/`updateState` path updates `stateObject` before emitting, so that (common drill-down) case works. Is the URL / browser back-forward case in scope? If so, this would miss it.
- **`deepClone` is unnecessary:** `isEqual` does a structural comparison and never mutates its inputs, so cloning (twice — here and in `startWith`) is dead work. Params are tiny so it's not a perf concern, but it's slightly misleading since it implies mutation protection matters.
- **Readability:** the `startWith(...) + distinctUntilChanged + skip(1)` stack is a bit opaque. A one-line comment noting that `startWith`/`skip(1)` seed the initial params so the first real change is compared against them would save the next reader from reverse-engineering the chain.

vvlladd28 (commented):
## Review summary

Reviewed 3 changed files in [Fixed reset of Entities table widget pagination on dashboard state update](https://github.com/thingsboard/thingsboard/pull/15549). Left 3 comment(s) inline.

The core change — a shared `resetPaginationOnStateChange` helper wired into the entities and alarms table widgets — is clean and correctly handles the common programmatic drill-down case. The main things worth a look: a timing gap for URL/back-forward state changes, an unnecessary `deepClone`, and a third table widget that has the same concern but wasn't wired.

## Additional findings

- **ui-ngx/src/app/modules/home/components/widget/lib/timeseries-table-widget.component.ts** — The timeseries table widget has the same pagination setup (a `displayPagination` block that resets `pageIndex` on `sortChange`) but wasn't wired to `resetPaginationOnStateChange`. If the stale-pagination-on-state-change bug is real for the entities/alarms tables, the timeseries table has it too. Note it uses multiple paginators (`@ViewChildren(MatPaginator) paginators`), so the single-paginator helper signature doesn't drop in cleanly — worth either handling that case or leaving a note on why it's intentionally excluded.

---

ui-ngx/src/app/modules/home/components/widget/lib/timeseries-table-widget.component.ts
```diff
@@ -653,7 +654,7 @@ export class TimeseriesTableWidgetComponent extends PageComponent implements OnI
       if (this.displayPagination) {
         paginator = this.paginators.toArray()[index];
         this.subscriptions.push(
-          sort.sortChange.subscribe(() => paginator.pageIndex = 0)
+          setupPaginationResets(this.ctx, paginator, sort, this.destroy$)
```
vvlladd28: Unlike the alarms and entities widgets, this component's `ngOnDestroy` (lines 282-290) never calls `this.destroy$.next()/complete()`, and `this.subscriptions` is only unsubscribed inside `initSubscriptionsToSortAndPaginator()` — not on destroy. The previous `sort.sortChange` subscription here was harmless because it was tied to this component's own `MatSort`, which dies with the view. But `setupPaginationResets` now also subscribes to `ctx.stateController.stateChanged()`, which is owned by the dashboard state controller and outlives the widget. So when this widget is destroyed, that subscription is never torn down — it leaks and keeps calling `paginator.firstPage()` on a detached paginator on every subsequent dashboard state change. Emitting `destroy$` in `ngOnDestroy` here (as the sibling widgets already do) would fix it.

Separately, note the teardown is doubled up at this call site: the returned `Subscription` is pushed into `this.subscriptions` *and* `destroy$` is passed in, whereas alarms/entities lean on `destroy$` only. Worth settling on a single mechanism so the three siblings manage the helper's lifecycle the same way.

ui-ngx/src/app/modules/home/components/widget/lib/table-widget.models.ts
```diff
@@ -516,3 +518,27 @@ export function isValidPageStepIncrement(value: number): boolean {
 export function isValidPageStepCount(value: number): boolean {
   return Number.isInteger(value) && value > 0 && value <= 100;
 }
+
+export function setupPaginationResets(ctx: WidgetContext,
```
vvlladd28, line 522: The helper takes a `destroyNotifier$` and wires `takeUntil` internally, but it also returns a `Subscription` — so callers end up with two independent teardown handles for the same work, and the three call sites already disagree on which to use (alarms/entities discard the return value and rely on the internal `takeUntil`; timeseries pushes the returned `Subscription` into its array *and* passes `destroy$`). Might be cleaner to pick one model: if the helper owns teardown via `destroyNotifier$`, return `void`; if callers should manage it, drop the internal `takeUntil` and the notifier param. As-is the dual-mode shape invites each new caller to wire it up differently.

ui-ngx/src/app/modules/home/components/widget/lib/table-widget.models.ts
```diff
@@ -516,3 +518,27 @@ export function isValidPageStepIncrement(value: number): boolean {
...
+
+  subscription.add(
+    sort.sortChange.pipe(
+      takeUntil(destroyNotifier$)
+    ).subscribe(() => paginator.pageIndex = 0)
+  );
+
+  subscription.add(
+    ctx.stateController.stateChanged().pipe(
+      map(() => ctx.stateController.getStateParams()),
```
vvlladd28, line 533: This operator chain is fairly subtle — it discards the state id emitted by `stateChanged()`, remaps to params, seeds with `startWith`, dedups with a deep `isEqual`, then `skip(1)` drops the seed so only genuine param changes after subscription reset the page. The intent isn't obvious at a glance. A one-line comment explaining why `startWith` + `skip(1)` are paired (and why we compare `getStateParams()` rather than the emitted state string) would save the next maintainer a real head-scratch.

vvlladd28 (commented):
## Review summary

Reviewed 4 changed files in [Fixed reset of Entities table widget pagination on dashboard state update](https://github.com/thingsboard/thingsboard/pull/15549). Left 3 comment(s) inline.

The extraction of `setupPaginationResets` and the `startWith` + `distinctUntilChanged(isEqual)` + `skip(1)` approach for resetting pagination only on real state-param changes look sound. The main concern is in the timeseries widget: unlike the alarms/entities widgets, it never drives its `destroy$`, so the new subscription to the long-lived `stateController.stateChanged()` leaks on widget destroy. The rest are design/maintainability notes about the helper's teardown contract.

---

ui-ngx/src/app/modules/home/components/widget/lib/table-widget.models.ts
```diff
@@ -516,3 +518,23 @@ export function isValidPageStepIncrement(value: number): boolean {
...
+                                      paginator: MatPaginator,
+                                      sort: MatSort): Subscription {
+  const subscription = new Subscription();
+
+  subscription.add(
+    sort.sortChange.subscribe(() => paginator.pageIndex = 0)
+  );
+
+  subscription.add(
+    ctx.stateController.stateChanged().pipe(
```
vvlladd28, line 532: `ctx.stateController` is dereferenced without a guard here, and it isn't always defined. `WidgetContext.stateController` resolves to `this.dashboard.stateController` (`widget-component.models.ts:170`), and `stateController` is a plain `@Input()` on `DashboardComponent` (`dashboard.component.ts:88`) with no default — so any host that renders `<tb-dashboard>` without binding `[stateController]` leaves it `undefined`.

`attribute-table.component.html:252` is one such host: the "show on widget" preview carousel on the Attributes / Latest telemetry tab binds `[aliasController]` but not `[stateController]`, and its widget list is whatever bundle the user picks (`attribute-table.component.ts:596-634` only filters out `static` and `rpc` types). Selecting a bundle containing the Entities table or Timeseries table renders these very widgets there with `displayPagination` defaulting to `true`.

In that case `undefined.stateChanged()` throws out of `ngAfterViewInit`. Since the call is the first statement in the `displayPagination` block, everything after it is skipped — the `filtersChanged` subscription, the `merge(sortChange, paginator.page)` → `updateData()` wiring, and the trailing `this.updateData()` — so the table renders permanently empty rather than just losing the pagination reset.

For the record, my earlier comment on this line pushed in the opposite direction (suggesting the `?.` could be dropped since `stateController` looked reliably non-null). That was wrong — though the `?.` wouldn't have saved it either, since `undefined.pipe(...)` throws the same way. A real guard is what's needed: something like bailing out of the state-change wiring when `ctx.stateController` is absent and returning just the `sortChange` subscription.

vvlladd28 (commented):
## Review summary

Reviewed 4 changed files in [Fixed reset of Entities table widget pagination on dashboard state update](https://github.com/thingsboard/thingsboard/pull/15549). Left 1 comment(s) inline.

The consolidation into `setupPaginationResets` reads well, the `startWith` + `distinctUntilChanged(isEqual)` + `skip(1)` seeding is correct against a plain `Subject`-backed `stateChanged()`, the dual teardown contract from the earlier iteration is gone, and the timeseries widget's previously missing `destroy$.next()/complete()` is now wired. The one thing worth a look before merge is in `table-widget.models.ts:532`: `ctx.stateController` is dereferenced without a guard and is `undefined` for any widget hosted by a `<tb-dashboard>` that doesn't bind `[stateController]`, which throws out of `ngAfterViewInit` and leaves the table permanently empty.

## Additional findings

- **ui-ngx/src/app/modules/home/components/widget/lib/timeseries-table-widget.component.ts:653** — `initSubscriptionsToSortAndPaginator()` unsubscribes everything in `this.subscriptions` but never resets the array, so it grows by two entries on every `sorts.changes` emission and re-unsubscribes already-dead subscriptions on each subsequent call. A `this.subscriptions.length = 0;` after the `forEach` would bound it. Pre-existing, but it's adjacent to this PR's subscription work and the new `ngOnDestroy` teardown at line 290 now iterates that same accumulating list.

---

ui-ngx/src/app/modules/home/components/dashboard-page/states/state-controller.component.ts
```diff
@@ -110,10 +110,10 @@ export abstract class StateControllerComponent implements IStateControllerCompon
           const newState = this.decodeStateParam(paramMap.get('state'));
           if (this.currentState !== newState) {
             this.currentState = newState;
-            this.stateChangedSubject.next(this.currentState);
             if (this.inited) {
               this.onStateChanged();
             }
+            this.stateChangedSubject.next(this.currentState);
```
vvlladd28, line 116: Moving the emission after `onStateChanged()` is the right fix — subscribers now read `getStateParams()` after `parseState()` has rebuilt `stateObject`, which is exactly what makes the new `distinctUntilChanged(isEqual)` comparison in `setupPaginationResets` see a real change.

`reInit()` (line 173) still has the old ordering, though: it emits `this.stateChangedSubject.next(this.currentState)` and only afterwards calls `this.init()`, which is where `this.stateObject = this.parseState(initialState)` actually happens. A subscriber reading `getStateParams()` on that emission gets the *previous* params, and no corrected emission follows — `init()` calls `gotoState(..., false)`, so `updateLocation()`/`updateStateParam()` never runs and the subject isn't re-emitted. The result is that `distinctUntilChanged` sees no change and swallows the `paginator.firstPage()` call.

In practice both `reInit()` call sites — `states-component.directive.ts:113` on a `dashboardId` change and `dashboard-page.component.ts:1253` after a version-conflict reload — plausibly recreate the widgets anyway, in which case the paginators start at page 0 regardless and the pagination impact is nil. But the asymmetry is on the same subject in the same file, and `getStateParams()` has other readers. Worth swapping those two lines in `reInit()` as well so every emission path carries fresh params?

vvlladd28 (commented):
## Review summary

Reviewed 5 changed files in [Fixed reset of Entities table widget pagination on dashboard state update](https://github.com/thingsboard/thingsboard/pull/15549). Left 1 comment(s) inline.

The consolidation into `setupPaginationResets` now reads cleanly across all three table widgets: the `if (stateController)` guard closes the unguarded-dereference hole from the previous round, teardown settled on a single mechanism (the returned `Subscription`), and the timeseries widget's previously missing `destroy$.next()/complete()` is wired up. Reordering `stateChangedSubject.next()` after `onStateChanged()` is the right fix for the stale-params timing gap. The one thing worth a look before merge is that `reInit()` in the same file still has the old ordering.

---

PR #15550, opened 2026-04-30
Hardened device credentials validation

dashevchenko:
## Pull Request description

Fixed CWE-93, CWE-94:
1. Enhanced device credential validation
2. Added sanitize logic for YAML sink lines.

## General checklist

## Back-End feature checklist

### DIFF
--- a/dao/src/main/java/org/thingsboard/server/dao/service/validator/DeviceCredentialsDataValidator.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/service/validator/DeviceCredentialsDataValidator.java
@@ -30,12 +30,8 @@ import org.thingsboard.server.dao.device.DeviceService;
 import org.thingsboard.server.dao.exception.DeviceCredentialsValidationException;
 import org.thingsboard.server.dao.service.DataValidator;
 
-import java.util.regex.Pattern;
-
 @Component
 public class DeviceCredentialsDataValidator extends DataValidator<DeviceCredentials> {
-
-    private static final Pattern CONTROL_CHARS = Pattern.compile("[\\x00-\\x1F\\x7F]");
 
     @Autowired
     private DeviceCredentialsDao deviceCredentialsDao;
@@ -92,7 +88,7 @@ public class DeviceCredentialsDataValidator extends DataValidator<DeviceCredenti
     }
 
     private static void rejectControlChars(String value, String fieldName) {
-        if (value != null && CONTROL_CHARS.matcher(value).find()) {
+        if (StringUtils.containsControlChars(value)) {
             throw new DeviceCredentialsValidationException(fieldName + " must not contain control characters!");
         }
     }
--- a/dao/src/main/java/org/thingsboard/server/dao/service/validator/DeviceCredentialsDataValidator.java
+++ b/dao/src/main/java/org/thingsboard/server/dao/service/validator/DeviceCredentialsDataValidator.java
@@ -18,17 +18,24 @@ package org.thingsboard.server.dao.service.validator;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Component;
+import org.thingsboard.common.util.JacksonUtil;
 import org.thingsboard.server.common.data.Device;
 import org.thingsboard.server.common.data.StringUtils;
+import org.thingsboard.server.common.data.device.credentials.BasicMqttCredentials;
 import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.common.data.security.DeviceCredentials;
+import org.thingsboard.server.common.data.security.DeviceCredentialsType;
 import org.thingsboard.server.dao.device.DeviceCredentialsDao;
 import org.thingsboard.server.dao.device.DeviceService;
 import org.thingsboard.server.dao.exception.DeviceCredentialsValidationException;
 import org.thingsboard.server.dao.service.DataValidator;
 
+import java.util.regex.Pattern;
+
 @Component
 public class DeviceCredentialsDataValidator extends DataValidator<DeviceCredentials> {
+
+    private static final Pattern CONTROL_CHARS = Pattern.compile("[\\r\\n\\t\\x00-\\x1F\\x7F]");
 
     @Autowired
     private DeviceCredentialsDao deviceCredentialsDao;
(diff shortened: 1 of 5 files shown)


dao/src/main/java/org/thingsboard/server/dao/util/DeviceConnectivityUtil.java
```diff
@@ -50,6 +50,11 @@ public class DeviceConnectivityUtil {
     public static final String MQTT_IMAGE = "thingsboard/mosquitto-clients ";
     public static final String COAP_IMAGE = "thingsboard/coap-clients ";
     private final static Pattern VALID_URL_PATTERN = Pattern.compile("^(https?)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]");
+    private final static Pattern CONTROL_CHARS = Pattern.compile("[\\x00-\\x1F\\x7F]");
+
+    private static String sanitize(String value) {
```
ViacheslavKlimov: `sanitize` is only wired into the docker-compose path. The sibling command builders in this same class — `getHttpPublishCommand` (line 60), `getMqttPublishCommand` (lines 74/81/84/87) and `getCoapPublishCommand` (line 209) — interpolate the same untrusted fields (`credentialsId`, `clientId`, `userName`, `password`) into shell commands that are returned to the UI for the operator to copy-paste and run. That's the same threat model as the compose file (an operator runs attacker-influenced text), but those values are still rendered raw. The new save-time validator mitigates it for newly created credentials, but (a) credentials created before this fix aren't cleaned, and (b) the control-char-only policy doesn't cover characters that matter for shell commands — e.g. a `"` in `userName` breaks out of the `-u "..."` quoting. Worth deciding whether those sinks need equivalent treatment.

Minor: `sanitize` is a fairly generic name for "replace control chars with `_`" — something like `stripControlChars` would make the intent obvious at the call sites.

dao/src/main/java/org/thingsboard/server/dao/service/validator/DeviceCredentialsDataValidator.java
```diff
@@ -18,18 +18,25 @@
...
 import org.thingsboard.server.dao.device.DeviceService;
 import org.thingsboard.server.dao.exception.DeviceCredentialsValidationException;
 import org.thingsboard.server.dao.service.DataValidator;
 
+import java.util.regex.Pattern;
+
 @Component
 public class DeviceCredentialsDataValidator extends DataValidator<DeviceCredentials> {
 
+    private static final Pattern CONTROL_CHARS = Pattern.compile("[\\x00-\\x1F\\x7F]");
```
ViacheslavKlimov: This `CONTROL_CHARS` pattern is duplicated verbatim in `DeviceConnectivityUtil` (line 53). Since both halves are part of the same fix (reject on save, replace on render), consider hoisting it to a single shared constant so the two definitions can't drift apart later.

dao/src/test/java/org/thingsboard/server/dao/service/validator/DeviceCredentialsDataValidatorTest.java
```diff
@@ -0,0 +1,129 @@
...
+    void rejectsNewlineInMqttPassword() {
+        DeviceCredentials creds = mqttBasic("cid", "user", "pwd\nentrypoint: x");
+
+        assertThatThrownBy(() -> validator.validateDataImpl(tenantId, creds))
+                .isInstanceOf(DeviceCredentialsValidationException.class)
+                .hasMessageContaining("password");
+    }
+
+    @Test
+    void acceptsValidCredentials() {
```
ViacheslavKlimov, line 100: The suite asserts rejection for each MQTT_BASIC field, but the only accepting case is an ACCESS_TOKEN. A happy-path test for a clean MQTT_BASIC credential — valid `clientId`/`userName`/`password` passing through the new parse-and-check branch and reaching `findDeviceById` without throwing — would guard against that branch accidentally over-rejecting valid input.

ViacheslavKlimov (commented):
## Review summary

Reviewed 4 changed files in [Fixed gateway docker-compose YAML Injection](https://github.com/thingsboard/thingsboard/pull/15550). The two-layer approach — rejecting control characters at credential save time (`DeviceCredentialsDataValidator`) and sanitizing them when rendering the gateway docker-compose YAML (`DeviceConnectivityUtil`) — is sound. The newline-based sibling-key injection is the critical vector and it's correctly closed; the tests are meaningful (they fail against the pre-fix code). I traced one suspected bug — the validator's `JacksonUtil.fromString(credentialsValue, ...)` throwing a non-`DataValidationException` — and confirmed it's a non-issue, since `formatSimpleMqttCredentials` already parses and re-serializes the value (raising a clean `DeviceCredentialsValidationException` on malformed JSON) before the validator runs. Left 3 comments inline, mostly about how completely the sanitization covers the sibling sinks and a small test gap.

## Additional findings

- **dao/src/main/java/org/thingsboard/server/dao/util/DeviceConnectivityUtil.java:125** — The `TB_GW_HOST` line interpolates `host` into the compose file without `sanitize`. Risk is lower than the credential fields because `host` is derived server-side via `getHost(...)`/base URL rather than from device data, but by the same principle that motivated this PR it's worth confirming `baseUrl`/`host` can never carry a newline — note `getHost` returns the raw input unchanged when `InetAddress.getByName` throws `UnknownHostException` (line 256).

---

dashevchenko (commented):
## Re-review summary

Re-reviewed [Fixed gateway docker-compose YAML Injection](https://github.com/thingsboard/thingsboard/pull/15550) — verified 4 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/15550#pullrequestreview-4426479409).

| Status | Count |
|--------|-------|
| ✅ Resolved | 4 |
| 💬 Acknowledged | 0 |

All previous findings have been addressed. The render-time sanitization now covers every sink the reviewer called out, the control-char regex lives in one place, and the test gap is closed.

### Finding details

- ✅ **DeviceConnectivityUtil.java** — `sanitize` only wired into the docker-compose path; sibling shell-command builders rendered the same untrusted fields raw, and the control-char-only policy didn't stop a `"` breaking out of `-u "..."` — *Fixed: renamed to `stripControlChars` and applied to `getHttpPublishCommand`, `getCoapPublishCommand` and `TB_GW_HOST`; the double-quoted MQTT args (access token, clientId, userName, password) now go through a new `escapeShellArg` that strips control chars and backslash-escapes `\ " $ ` ` so they can't break out of the quotes or trigger command/variable substitution.*
- ✅ **DeviceCredentialsDataValidator.java** — `CONTROL_CHARS` regex duplicated across the validator and the util — *Fixed: hoisted to a single `StringUtils.CONTROL_CHARS` (with a `containsControlChars` helper); both the validator and `DeviceConnectivityUtil` reference it, so they can't drift.*
- ✅ **DeviceCredentialsDataValidatorTest.java** — only ACCESS_TOKEN had a happy-path accept case — *Fixed: added `acceptsValidMqttBasicCredentials`, which drives a clean MQTT_BASIC credential through the parse-and-check branch to `findDeviceById` without throwing. The new escaping is also pinned by three tests in `DeviceConnectivityUtilTest`.*
- ✅ **DeviceConnectivityUtil.java:125 (additional finding)** — `TB_GW_HOST` interpolated `host` without sanitization — *Fixed: now `stripControlChars(isLocalhost(host) ? HOST_DOCKER_INTERNAL : host)`.*

### Fix-commit notes

Reviewed the new code (`escapeShellArg`, the shared `StringUtils.CONTROL_CHARS`, and the three new escaping tests) for regressions — none found. The escape order is correct (backslash first), the two `StringUtils` references both resolve to `common.data.StringUtils` where the helpers were added, and the validator's now-unused `Pattern` import was removed.

One minor, out-of-scope consistency note (not a regression — pre-existing, and `host` is server-derived rather than device-controlled): `host` is still interpolated raw into `-h <host>` in `getMqttPublishCommand` and into the URL authority in the HTTP/CoAP commands. The flagged `TB_GW_HOST` sink is fixed; these siblings carry the same server-derived value and weren't part of the original finding, so they're noted only for completeness.

---

dao/src/main/java/org/thingsboard/server/dao/util/DeviceConnectivityUtil.java
dashevchenko: Fixed shell commands for getHttpPublishCommand, getMqttPublishCommand, getCoapPublishCommand. Renamed method name.

dao/src/main/java/org/thingsboard/server/dao/service/validator/DeviceCredentialsDataValidator.java
dashevchenko: Fixed. Moved to StringUtils.

dao/src/test/java/org/thingsboard/server/dao/service/validator/DeviceCredentialsDataValidatorTest.java
dashevchenko: Fixed. Added test

PR #15560, opened 2026-05-04
Fixed MAX aggregation for mixed double and long telemetry values

dashevchenko:
## Pull Request description

For devices with negative telemetry keys that store telemetry in both long and double format, the max value was ~4.9E-324 instead of real.
Three locations were fixed: SQL aggregation, Cassandra aggregation, and the TBEL rollingArg.max() used in calculated fields. 

## General checklist

## Back-End feature checklist

### DIFF
--- a/dao/src/test/java/org/thingsboard/server/dao/service/timeseries/BaseTimeseriesServiceTest.java
+++ b/dao/src/test/java/org/thingsboard/server/dao/service/timeseries/BaseTimeseriesServiceTest.java
@@ -723,6 +723,19 @@ public abstract class BaseTimeseriesServiceTest extends AbstractServiceTest {
     }
 
     @Test
+    public void testFindDeviceMaxAggregationOverAllNegativeDoubleTsData() throws Exception {
+        save(deviceId, 5000, -50.0);
+        save(deviceId, 15000, -100.0);
+        save(deviceId, 25000, -75.0);
+
+        List<TsKvEntry> list = tsService.findAll(tenantId, deviceId, Collections.singletonList(new BaseReadTsKvQuery(LONG_KEY, 0,
+                60000, 60000, 1, Aggregation.MAX))).get(MAX_TIMEOUT, TimeUnit.SECONDS);
+
+        assertEquals(1, list.size());
+        assertEquals(java.util.Optional.of(-50.0), list.get(0).getDoubleValue());
+    }
+
+    @Test
     public void testSaveTs_RemoveTs_AndSaveTsAgain() throws Exception {
         save(deviceId, 2000000L, 95);
         save(deviceId, 4000000L, 100);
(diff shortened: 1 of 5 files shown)


dao/src/test/java/org/thingsboard/server/dao/service/timeseries/BaseTimeseriesServiceTest.java
```diff
@@ -710,6 +710,31 @@ public void testFindDeviceLongAndDoubleTsData() throws Exception {
...
+
+        List<TsKvEntry> list = tsService.findAll(tenantId, deviceId, Collections.singletonList(new BaseReadTsKvQuery(LONG_KEY, 0,
+                60000, 60000, 1, Aggregation.MAX))).get(MAX_TIMEOUT, TimeUnit.SECONDS);
+
+        assertEquals(1, list.size());
+        assertEquals(java.util.Optional.of(-50.0), list.get(0).getDoubleValue());
+    }
+
+    @Test
+    public void testFindDeviceMaxAggregationOverAllNegativeDoubleTsData() throws Exception {
```
ViacheslavKlimov, line 726: I don't think this test actually reproduces the bug it's named after. Every row stored here has a non-null `doubleValue` and a null `longValue`, so on the SQL path `COALESCE(tskv.doubleValue, -1.7976931348623157E308)` never returns the sentinel — MAX picks `-50.0` from the real values either way. With the pre-fix code (`Double.MIN_VALUE`) the sentinel still never entered the MAX, so the assertion would have passed before the fix too. The Cassandra path behaves the same: when `hasDouble` is true and only doubles are present, `aggResult.dValue` is set by `processMaxAggregation` and the `orElse(...)` fallback isn't triggered.

The mixed-types test above is the one that actually fails without the fix (it's the case that triggers the SQL COALESCE on rows whose `doubleValue IS NULL`). To make this test load-bearing, consider mixing in at least one long-valued row, or otherwise ensuring some `doubleValue` cells are NULL so the COALESCE default is exercised.

ViacheslavKlimov (commented):
## Review summary

Reviewed 5 changed files in [Fixed MAX aggregation for mixed double and long telemetry values](https://github.com/thingsboard/thingsboard/pull/15560). The core fix is correct: `Double.MIN_VALUE` is `+4.9E-324` (smallest *positive* double), so using it as a "less than any value" sentinel for MAX silently overrode any all-negative result. Replacing it with `-Double.MAX_VALUE` (and the equivalent literal `-1.7976931348623157E308` in JPQL) is the right call. Left 1 comment inline.

## Additional notes

- `dao/src/main/java/org/thingsboard/server/dao/sqlts/timescale/AggregationRepository.java:55` already uses `-1.79769E+308` as the COALESCE default for `MAX(dbl_v)`, so the Timescale path was not affected by the same bug. Worth knowing for parity if anyone later wonders why it wasn't touched. (Strictly speaking the literal there is slightly less negative than `-Double.MAX_VALUE` because of truncation, but it's negative enough for any realistic telemetry.)
- `AggregatePartitionsFunction.java:280` — the `orElse(-Double.MAX_VALUE)` fallback is reached only when `aggResult.dValue` is null. Inside `if (aggResult.hasDouble)` that should not happen in practice, since `processMaxAggregation` always populates `dValue` whenever a partition contributes `doubleCount > 0`. Not a problem with your change — just a heads-up that this site is defensive rather than a code path that produced a wrong answer for the reported bug.

---

PR #15562, opened 2026-05-04
Fixed stale duration check future warning on alarm rule REINIT

zzzeebra:
## Summary
- After PR #15439 made REINIT during an active DURATION countdown actually succeed, the existing `durationCheckFuture` from the previous configuration was preserved through `initRuleState`. The next matching event (or scheduled re-eval) called `setDurationCheckFuture`, which logs a WARN with synthetic stacktrace and silently replaces the future.
- Added `AlarmRuleState#cancelDurationCheckFuture()` that cancels and nulls only the future (without touching `firstEventTs`/`lastCheckTs`/`duration`) and call it from `AlarmCalculatedFieldState#initRuleState` before signaling `reevalNeeded`.

## Automated testing
- `AlarmRulesTest#testChangeDurationConditionFromStaticToDynamic` (added in #15439) — passes, and the WARN `Setting new duration check future while previous is not null` no longer appears in the test output.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/cf/ctx/state/alarm/AlarmRuleState.java
+++ b/application/src/main/java/org/thingsboard/server/service/cf/ctx/state/alarm/AlarmRuleState.java
@@ -256,10 +256,7 @@ public class AlarmRuleState {
         firstEventTs = 0L;
         lastCheckTs = 0L;
         duration = 0L;
-        if (durationCheckFuture != null) {
-            durationCheckFuture.cancel(true);
-            durationCheckFuture = null;
-        }
+        cancelDurationCheckFuture();
     }
 
     public void setDurationCheckFuture(ScheduledFuture<?> durationCheckFuture) {
--- a/application/src/main/java/org/thingsboard/server/service/cf/ctx/state/alarm/AlarmRuleState.java
+++ b/application/src/main/java/org/thingsboard/server/service/cf/ctx/state/alarm/AlarmRuleState.java
@@ -270,6 +270,13 @@ public class AlarmRuleState {
         this.durationCheckFuture = durationCheckFuture;
     }
 
+    public void cancelDurationCheckFuture() {
+        if (durationCheckFuture != null) {
+            durationCheckFuture.cancel(true);
+            durationCheckFuture = null;
+        }
+    }
+
     public boolean isEmpty() {
         return eventCount == 0L && firstEventTs == 0L && lastCheckTs == 0L && durationCheckFuture == null;
     }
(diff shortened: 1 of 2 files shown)


application/src/main/java/org/thingsboard/server/service/cf/ctx/state/alarm/AlarmRuleState.java
```diff
@@ -270,6 +270,13 @@ public void setDurationCheckFuture(ScheduledFuture<?> durationCheckFuture) {
         this.durationCheckFuture = durationCheckFuture;
     }
 
+    public void cancelDurationCheckFuture() {
```
ViacheslavKlimov, line 270: Reuse the new method

Image

application/src/main/java/org/thingsboard/server/service/cf/ctx/state/alarm/AlarmRuleState.java
```diff
@@ -270,6 +270,13 @@ public void setDurationCheckFuture(ScheduledFuture<?> durationCheckFuture) {
...
+        if (durationCheckFuture != null) {
+            durationCheckFuture.cancel(true);
+            durationCheckFuture = null;
+        }
+    }
```
ViacheslavKlimov, line 275: Minor: this duplicates the cancel block already inside `clearDurationConditionState` (lines 259-262). Could `clearDurationConditionState` be simplified to call `cancelDurationCheckFuture()` for the future-handling part? Single source of truth for the cancel/null logic, and any future tweak (e.g. swapping `cancel(true)` for `cancel(false)`) only needs to happen in one place. Purely a cleanup — feel free to skip if you'd rather keep the change minimal.

ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [Fixed stale duration check future warning on alarm rule REINIT](https://github.com/thingsboard/thingsboard/pull/15562). The fix is small, well-targeted, and the placement is correct: `cancelDurationCheckFuture()` is called inside `initRuleState` before `reevalNeeded.set(true)`, so the subsequent `setDurationCheckFuture` call from `init()`'s reeval block sees a null future and skips the warn-and-replace branch. Importantly, the new method nulls only the future and preserves `firstEventTs`/`lastCheckTs`/`duration`, which is required for the follow-up `reeval(System.currentTimeMillis(), ctx)` to compute the correct `leftDuration`. Left one minor maintainability comment inline.

---

application/src/main/java/org/thingsboard/server/service/cf/ctx/state/alarm/AlarmRuleState.java
zzzeebra: Done

PR #15618, opened 2026-05-12
Added /api/rule-engine/v2 endpoint with per-entity ACL enrichment

### LINKED ISSUE #15496: Rule Engine ACL enrichment: /api/ruleEngine/v2 with server-authoritative tb_acl and tb_user_id metadata
**Is your feature request related to a problem? Please describe.**

When the REST API pushes a message to the Rule Engine (`POST /api/ruleEngine/...`), the engine runs under effectively-root privileges. Rule chains can read, write, or delete any entity without knowing whether the *calling user* was actually authorized to do so. The only permission check happens in the controller against a single originator entity (`Operation.WRITE`); once the `TbMsg` enters the queue, user identity and authorization context are lost.

This prevents rule-chain authors from building flows whose behavior depends on the caller's permissions — e.g., "allow this action only if the caller can WRITE to asset X", or "log who initiated this action".

**Describe the solution you'd like**

Add an **optional, additive, non-breaking** enrichment path to the Rule Engine REST API — a single new endpoint `POST /api/ruleEngine/v2/` — that:

1. Accepts a list of entities from the caller alongside the payload.
2. Computes, server-side, the set of operations the caller's **role** is allowed to perform on each entity's resource type.
3. Writes that ACL snapshot into **protected** `TbMsg` metadata keys the caller cannot override.
4. Also writes the caller's user id, for audit/logging inside rule chains.
5. Forwards the enriched `TbMsg` to the Rule Engine as usual.

Rule chains are free to ignore the metadata entirely — existing flows are unaffected.

**Describe alternatives you've considered**

- Denying the request at the API layer based on the ACL snapshot. Rejected: decisions belong in the rule chain so operators can express richer policies.
- Per-entity instance-level permission checks in the first iteration. Deferred: requires entity loads via `EntityServiceRegistry` / `EntityDaoService.findEntityById` and is left for a follow-up.
- Modifying or replacing the existing `POST /api/ruleEngine/...` endpoints. Rejected: backward compatibility. A new `/v2/` endpoint is added alongside them.
- Multiple `/v2/` endpoints with path-variable routing (mirroring the v1 surface). Rejected during design review: collapsed to a single body-based endpoint to keep the surface small and the routing parameters uniformly optional.

---

# 📐 Design doc — Rule Engine ACL Enrichment

> **Status:** Approved for implementation · **Date:** 2026-04-14 · **Last updated:** 2026-05-07 · **Scope:** ThingsBoard CE · **Implementation:** PR #15497

## Scope

ThingsBoard CE. A new `POST /api/ruleEngine/v2/` endpoint. v1 endpoints untouched.

## API Contract

A single new endpoint with all routing parameters in the request body:

- `POST /api/ruleEngine/v2/`

Same authorization as v1: `hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN', 'CUSTOMER_USER')` and existing `AccessValidator` WRITE check on the originator.

### Request body

```json
{
  "originator":   { "entityType": "DEVICE", "id": "784f394c-42b6-..." },
  "messageType":  "REST_API_REQUEST",
  "queueName":    "Main",
  "timeout":      10000,
  "payload":      { "any": "user json" },
  "enrichEntities": [
    { "entityType": "DEVICE", "id": "784f394c-42b6-..." },
    { "entityType": "ASSET",  "id": "abc-123-..." }
  ]
}
```

- `originator` — optional `EntityId`. If absent, defaults to the calling user's id (matches v1 behavior when path variables are omitted).
- `messageType` — optional. Defaults to `REST_API_REQUEST` (matches v1's hardcoded type). Custom strings are accepted.
- `queueName` — optional. When present, overrides the queue selected by the originator's profile (same semantics as v1).
- `timeout` — optional, milliseconds. Defaults to `server.rest.rule_engine.response_timeout` (10s).
- `payload` — required JSON, becomes `TbMsg.data`.
- `enrichEntities` — optional list of `EntityId` (uses the platform's existing polymorphic `EntityId` Jackson representation: `{entityType, id}`). If absent or empty, `tb_acl` is `"[]"`.
- Size limit: configurable via `rule-engine.acl.max-entities` (default **20**). Exceeding returns `400 Bad Request`.

### Response

Unchanged from v1: `DeferredResult<ResponseEntity>` populated by the Rule Engine's terminal "REST Call Reply" node (or `408`/`504` on timeout).

## Protected Metadata

Two new keys, both **always written last** when building `TbMsgMetaData`, so any user-supplied value (if a caller tried to sneak them in via the payload or elsewhere) is unconditionally overwritten.

### `tb_acl`

- Key constant: `TbMsgMetaData.TB_ACL_KEY = "tb_acl"`
- Value: JSON string (serialized `List<EntityAclEntry>`).
- Order: matches the order of `enrichEntities` in the request (deterministic).
- Shape:

```json
[
  {
    "entityType": "DEVICE",
    "entityId": "784f394c-42b6-...",
    "roleAllowed": ["READ", "WRITE", "READ_ATTRIBUTES", "WRITE_ATTRIBUTES", "READ_TELEMETRY"]
  },
  {
    "entityType": "ASSET",
    "entityId": "abc-123-...",
    "roleAllowed": []
  }
]
```

- **`roleAllowed` — role-level capability, NOT per-entity access.** Names from the `Operation` enum the caller's role can perform on the entity's *resource type*. Empty list if the role has no access (including the case where `EntityType` has no matching `Resource`).
- For Customer Users whose access depends on entity assignment, the list reports what the role can do, not whether the caller can do it on the specific `entityId`. Per-entity (instance-level) checks are out of scope for the PoC — see Future Work.
- Format deliberately an array of flat objects (no compound keys like `"DEVICE:uuid"`) so it iterates naturally in JS and TBEL rule nodes.

### `tb_user_id`

- Key constant: `TbMsgMetaData.TB_USER_ID_KEY = "tb_user_id"`
- Value: UUID string of the calling user (`currentUser.getId().getId().toString()`).
- Stable across rename/email changes.
- `tenantId` and `customerId` are already recoverable from `TbMsg` (originator, customerId), so they are not duplicated here.

## Computation Logic

For each descriptor in `enrichEntities`:

1. Map `EntityType → Resource` via existing `Resource.of(entityType)`. If it throws `IllegalArgumentException`, record the entity with `roleAllowed=[]` and continue.
2. For each value in `Operation.values()`:
   - Call `accessControlService.checkPermission(user, resource, op)` (role-level form).
   - If it returns normally → add `op.name()` to `roleAllowed`.
   - If it throws `ThingsboardException` → skip (operation not applicable to this role/resource).
3. Append `new EntityAclEntry(entityType, entityId, roleAllowed)` to the result list.
4. Serialize the list with `JacksonUtil.toString(...)` and write to `TB_ACL_KEY`.

Size-limit validation (`> maxAclEntities`) happens before any permission check and returns `400 BAD_REQUEST_PARAMS`.

### New files

- `common/data/src/main/java/org/thingsboard/server/common/data/rule/engine/EnrichedRuleEngineRequest.java`
  - Fields: `EntityId originator`, `String messageType`, `String queueName`, `Integer timeout`, `JsonNode payload`, `List<EntityId> enrichEntities`
- `common/data/src/main/java/org/thingsboard/server/common/data/rule/engine/EntityAclEntry.java`
  - Fields: `EntityType entityType`, `UUID entityId`, `List<String> roleAllowed`
  - Flat shape (not `EntityId`) so JS/TBEL nodes can iterate naturally.
  - Field name communicates the limitation: it is the **role's** capability on the resource type, not a per-entity access decision.

### Modified files

- `common/message/.../TbMsgMetaData.java`
  - Add `public static final String TB_ACL_KEY = "tb_acl";`
  - Add `public static final String TB_USER_ID_KEY = "tb_user_id";`
- `application/.../controller/RuleEngineController.java`
  - Inject `AccessControlService accessControlService`.
  - Add `@Value("${rule-engine.acl.max-entities:20}") int maxAclEntities;`
  - Add a single `POST /api/ruleEngine/v2/` endpoint method `handleEnrichedRuleEngineRequest(@RequestBody EnrichedRuleEngineRequest)`.
  - Add private helper `buildAclMetadata(SecurityUser, List<EntityId>)` returning the JSON string.
  - The flow: parse `EnrichedRuleEngineRequest` → resolve `originator` (body or current user), `messageType` (body or `REST_API_REQUEST`), `queueName` (body or null), `timeout` (body or `defaultResponseTimeout`) → existing `AccessValidator` WRITE check on originator → inside `onSuccess()`, compute ACL → populate metadata in the order `serviceId, requestUUID, expirationTime, tb_user_id, tb_acl` → build `TbMsg` with `data = payload` and the resolved `messageType` (via `type(String)`, the API for custom message types) → call `ruleEngineCallService.processRestApiCallToRuleEngine(...)` exactly as today.
- `application/src/main/resources/thingsboard.yml`
  - Add the `rule-engine.acl.max-entities` property under the existing `rule-engine` section.

### Zero impact on

- v1 endpoints (`POST /api/ruleEngine/...`) — untouched.
- `AccessValidator`, `AccessControlService`, `Resource`, `Operation` — consumed only.
- `RuleEngineCallService`, `TbClusterService` — unchanged; they receive a `TbMsg` that simply has extra metadata keys.
- Downstream rule nodes — consume or ignore the new metadata at their discretion.

## Flow Diagram

```
   ┌──────┐  POST /api/ruleEngine/v2/
   │      │  { originator, messageType, queueName, timeout,
   │ User │    payload, enrichEntities:[{entityType,id},...] }
   │      │─────────────────────────┐
   │      │                         │
   └──────┘                         ▼
                        ┌──────────────────────┐
                        │ RuleEngineController │
                        │   /v2 endpoint       │
                        └───────┬──────────────┘
                                │ 1. parse JSON → EnrichedRuleEngineRequest
                                │    (resolve defaults: originator,
                                │     messageType, timeout, queueName)
                                │ 2. AccessValidator WRITE on originator
                                │    (existing check, unchanged)
                                ▼
                        ┌──────────────────────┐
                        │  buildAclMetadata()  │
                        │  (inline helper)     │
                        └───────┬──────────────┘
                                │ validate size ≤ maxAclEntities
                                │ for each entity:
                                │   Resource.of(entityType)
                                │   for each Operation:
                                │     checkPermission → allowed?
                                │ serialize List<EntityAclEntry>
                                ▼
                        ┌──────────────────────┐
                        │  TbMsgMetaData       │
                        │  serviceId           │
                        │  requestUUID         │
                        │  expirationTime      │
                        │  tb_user_id ◀──────── SERVER
                        │  tb_acl     ◀──────── OVERWRITE (always last)
                        └───────┬──────────────┘
                                │ build TbMsg(data=payload, metaData=above)
                                ▼
                        ┌──────────────────────┐
                        │ RuleEngineCallService│
                        │ (unchanged)          │
                        └───────┬──────────────┘
                                │ push to queue
                                ▼
                        ┌──────────────────────┐
                        │   Rule Engine        │
                        │   rule chain         │
                        │   script/switch nodes│
                        │   read metadata.tb_acl│
                        │   read metadata.tb_user_id│
                        │   decide: pass/block │
                        │   → REST Call Reply  │
                        └───────┬──────────────┘
                                │ response TbMsg
                                ▼
                        ┌──────────────────────┐
                        │  Response to User    │
                        │  (200 or whatever    │
                        │   the rule chain set)│
                        └──────────────────────┘
```

## Rule Node Usage Examples

> **Read first.** `roleAllowed` reports what the caller's *role* can do on the *resource type*
> (e.g., DEVICE) — it does **not** decide whether the caller has access to the specific
> `entityId`. For role-uniform tenants/admins this is sufficient. For Customer Users with
> entity-assignment-based ACLs, treat `roleAllowed.contains("WRITE")` as a *necessary*
> precondition only, and add an entity-load check downstream when needed.
>
> Entries use a flat `{entityType, entityId, roleAllowed[]}` shape so JS/TBEL nodes don't
> need to know about the polymorphic `EntityId` form.

### JS script node — gate by role capability on the resource type

```js
var acl = JSON.parse(metadata.tb_acl);

var roleCanWriteDevices = acl.some(function(e) {
    return e.entityType === 'DEVICE' && e.roleAllowed.indexOf('WRITE') >= 0;
});

var userId = metadata.tb_user_id;
metadata.auditActor = userId;

return { msg: msg, metadata: metadata, msgType: roleCanWriteDevices ? 'Allowed' : 'Denied' };
```

### TBEL script node

```
var acl = JSON.parse(metadata.tb_acl);
var roleCanWriteDevices = false;
foreach (e : acl) {
    if (e.entityType == "DEVICE" && e.roleAllowed.contains("WRITE")) {
        roleCanWriteDevices = true;
        break;
    }
}
metadata.auditActor = metadata.tb_user_id;
return {msg: msg, metadata: metadata, msgType: roleCanWriteDevices ? "Allowed" : "Denied"};
```

### Filter pattern (entity types whose role grants WRITE)

```js
var writableTypes = JSON.parse(metadata.tb_acl)
    .filter(function(e) { return e.roleAllowed.indexOf('WRITE') >= 0; })
    .map(function(e) { return e.entityType + ':' + e.entityId; });
```

## Error Handling

| Condition | Response |
|---|---|
| Malformed JSON body | `400 Bad Request` (existing behavior) |
| `enrichEntities.size() > maxAclEntities` | `400 Bad Request` with message |
| Unknown `entityType` (`EntityIdFactory` fails) | `400 Bad Request` (existing behavior) |
| No WRITE on originator | `401 Unauthorized` (existing `AccessValidator`) |
| `EntityType` has no matching `Resource` | Entry included with `roleAllowed=[]` — no error |
| Operation inapplicable to role (throws) | Silently skipped from `roleAllowed` — no error |
| Rule Engine processing timeout | `408 Request Timeout` (existing) |

## Security Considerations

- **Immutable server fields.** `tb_acl` and `tb_user_id` are written by the controller after all other metadata population, so any caller-supplied value of the same key is overwritten. `TbMsgMetaData.putValue(...)` is replace semantics — no array/append path.
- **No privilege escalation.** The ACL snapshot reports *only* what the user already has. It can never grant more than the user can do directly.
- **Information disclosure.** The snapshot tells the rule chain (which the user cannot read directly) what the caller is allowed to do on each listed entity. Since the caller supplied the list and the data is about themselves, this is not a new information leak.
- **DoS vector.** Bounded by `max-entities` × `Operation.values().length` permission checks per request — O(20 × 18) = 360 checks worst case, all in-memory. No DB load in role-level form.
- **Audit.** `tb_user_id` inside rule chains lets authors emit richer audit events that today's `REST_API_RULE_ENGINE_CALL` audit log does not propagate.

### Unit

- `buildAclMetadata`:
  - Tenant admin + DEVICE → `roleAllowed` contains at least READ/WRITE/DELETE/CREATE.
  - Customer user with limited role → only the operations their role grants.
  - Empty list → `"[]"`.
  - Null list → `"[]"`.
  - `enrichEntities.size() > maxAclEntities` → `ThingsboardException(BAD_REQUEST_PARAMS)`.
  - Unmapped `EntityType` → entry with `roleAllowed=[]`, no exception.
  - Deterministic order preserved.

### Integration (controller)

- `POST /api/ruleEngine/v2/` with valid request → verify captured `TbMsg` contains `tb_acl` (parseable JSON) and `tb_user_id` (equals caller's UUID).
- Caller places `{"tb_acl": "attack"}` anywhere in their payload → captured metadata still has the server-computed `tb_acl`.
- No enrichment (omit `enrichEntities`) → `tb_acl = "[]"`, `tb_user_id` still present.
- Size over limit → HTTP 400.
- Body `messageType=POST_TELEMETRY_REQUEST`, `timeout=2000` → forwarded `TbMsg.getType()` and `expirationTime` reflect the body values.
- Explicit `originator` from body honored; default falls back to current user.
- JSON wire-format uses `roleAllowed` key (not `allowed`) — guards the rule-chain JS contract.
- Existing v1 endpoints regression suite continues to pass.

### Manual smoke

- Create a simple rule chain: script node reads `metadata.tb_acl`, switches `Allowed`/`Denied` → REST Call Reply.
- Call with Tenant Admin and Customer User credentials → observe different branches.

## Out of Scope / Future Work

- **Per-entity (instance-level) permission checks** using the `checkPermission(user, resource, op, entityId, entity)` form. Requires entity loads via `EntityServiceRegistry` / `EntityDaoService.findEntityById`. Main benefit: correct results for Customer Users whose permissions depend on entity assignment.
- **Caching** of permission checks per request if the same entity is repeated (not expected usage).
- **Extending enrichment** to other controllers that push to the Rule Engine (e.g., RPC v2, Telemetry). Would motivate extracting logic into a shared `AclEnrichmentService` (originally Variant B).
- **Additional enriched fields** (e.g., user email, role name) if rule-chain authors ask for it.

## Manual Smoke-Test Recipe

1. Start ThingsBoard locally (Postgres + core + rule-engine node).
2. Log in as Tenant Admin. Create a simple rule chain:
   - **Script** node reading `metadata.tb_acl` and setting `msgType` to `Allowed` or `Denied`.
   - **REST Call Reply** node.
3. Set this rule chain as root (or route via `queueName`).
4. Call:

   ```bash
   curl -X POST http://localhost:8080/api/rule-engine/v2/ \
     -H "Content-Type: application/json" \
     -H "X-Authorization: Bearer $JWT" \
     -d '{
       "payload": {"k":"v"},
       "enrichEntities": [
         {"entityType":"DEVICE","id":"<some-device-uuid>"}
       ]
     }'
   ```

5. Expected: HTTP 200 with body `{"msgType":"Allowed"}` (tenant admin) or `{"msgType":"Denied"}` (customer user without access).
6. In Rule Chain debug mode, inspect the incoming `TbMsg` — `metadata.tb_acl` and `metadata.tb_user_id` are present and authoritative.

zzzeebra:
Closes #15496

## Summary

Adds a new `POST /api/rule-engine/v2` endpoint that pushes a message to the Rule Engine enriched with two server-authoritative metadata keys:

- `tb_aclSnapshot` — a JSON array describing, per entity in `aclEntities`, the operations the calling user is allowed to perform on **that specific entity instance**. Per-`EntityId`, per-`Operation`.
- `tb_userId` — UUID of the calling user, stable across rename/email changes.

Both keys are written last when building `TbMsgMetaData`, so any value the caller tries to inject via the request payload is unconditionally overwritten by the server-computed value.

v1 endpoints (`POST /api/rule-engine/...`) are untouched (no extra metadata keys emitted) and marked `@Deprecated(since = "4.3.1.2")` in favor of v2.

## What was implemented

All routing parameters live inside the request body — no path variables:

```json
{
  "originator":   { "entityType": "DEVICE", "id": "..." },
  "queueName":    "Main",
  "timeout":      10000,
  "payload":      { "any": "user json" },
  "aclEntities":  [ { "entityType": "DEVICE", "id": "..." } ]
}
```

- Single endpoint with all parameters in the body (vs. four v1 endpoints with path variables).
- `aclEntities` typed as `List<EntityId>` using the platform's existing polymorphic Jackson representation `{entityType, id}` — no custom `EnrichEntityDescriptor` DTO.
- `payload` is required (HTTP 400 if missing). Message type is hardcoded to `REST_API_REQUEST`, matching the documented input type for the terminal `rest call reply` rule node (see [Rule Engine — REST Call Reply node](https://thingsboard.io/docs/user-guide/rule-engine-2-0/nodes/action/rest-call-reply/)).
- Authorization unchanged from v1: \`hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN', 'CUSTOMER_USER')\` + \`AccessValidator\` WRITE on the originator.
- Size limit: \`rule_engine.acl.max_entities\` (default 20, set to 0 to disable). Exceeding returns 400.

## Test plan

13 integration tests in \`RuleEngineControllerV2EnrichmentTest\`, all passing locally:

- [x] Tenant admin sees full per-entity ACL on own DEVICE.
- [x] Customer user sees per-entity ACL on a DEVICE assigned to their customer.
- [x] Customer user cannot WRITE a DEVICE assigned to a different customer.
- [x] Two-customer cross-scenario: deviceA on customer1, deviceB on customer2 — each user sees only their own entity as writable.
- [x] Null \`aclEntities\` → \`tb_aclSnapshot="[]"\`, \`tb_userId\` still present.
- [x] Empty \`aclEntities\` list → \`tb_aclSnapshot="[]"\`, \`tb_userId\` still present.
- [x] Duplicate ids in \`aclEntities\` → output preserves order and multiplicity; \`fetchEntity\` invoked once per unique id (Mockito-verified).
- [x] \`aclEntities.size() > maxAclEntities\` → HTTP 400 (bound configured via \`@TestPropertySource\`, independent of production default).
- [x] \`RULE_NODE\` (no Resource mapping) → entry with \`allowed=[]\`.
- [x] Nonexistent UUID → entry with \`allowed=[]\`.
- [x] Caller embeds \`tb_aclSnapshot\`/\`tb_userId\` in payload → server overwrites.
- [x] Body \`timeout\` honored, message forwarded as \`REST_API_REQUEST\`.
- [x] Missing payload → HTTP 400.
- [x] v1 regression suite (\`RuleEngineControllerTest\`) still passes.

## Differences from the original issue #15496

The original issue described a **role-level** ACL snapshot, with per-entity checks and dedup deferred to future work. This PR is per-entity from day one and includes dedup.

**Pulled in from issue's "Future Work":**
- Per-entity (instance-level) permission checks via the boolean \`hasPermission(user, resource, op, entityId, entity)\` form on \`AccessControlService\`. Requires loading the entity from DB — done via injected \`EntityService.fetchEntity\`. Reason: role-level form gives misleading results for Customer Users (it reports operations the role can perform on the *resource type* in general, not on the specific entity instance).
- Per-request deduplication of repeated \`EntityId\` values via a \`Map<EntityId, EntityAclEntry>\` cache. Output preserves duplicates in input order; DB loads and permission probes happen once per unique id.

**Still deferred:**
- Extending enrichment to other controllers that push to the Rule Engine (\`TelemetryController\`, \`RpcController\`, …).
- Additional enriched fields (\`tb_user_email\`, \`tb_user_authority\`, …) — added on demand from rule-chain authors.

**Field-name change driven by the per-entity shift:**
- \`EntityAclEntry.roleAllowed\` → \`EntityAclEntry.allowed\`. The original name communicated "this is role-level, NOT per-entity" — misleading now.

**Config-path:**
- Top-level \`rule_engine.acl.max_entities\` (snake-case to match \`thingsboard.yml\` convention). Default 20; set to 0 to disable the bound entirely.

**Payload handling:**
- \`payload\` is treated as optional. Null or missing payload becomes the empty JSON object \`{}\` on \`TbMsg.data\` — supports probe-only requests where the caller only wants the ACL snapshot.

**Operations logging:**
- WARN logs added at three early-return paths inside \`computeEntry\` (no Resource mapping, no EntityDaoService, entity not found / cross-tenant / non-\`HasTenantId\`). Not in the issue; surfaces operations-debug signal for stale-id / wiring problems without changing the JSON contract.


application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
```diff
@@ -64,46 +75,91 @@
...
+            notes = "Deprecated since 4.3. Prefer `POST /api/rule-engine/v2` which accepts all routing parameters in the request body and supports optional ACL enrichment.\n\n" +
+                    MSG_DESCRIPTION_PREFIX +
                     "Uses current User Id ( the one which credentials is used to perform the request) as the Rule Engine message originator. " +
                     MSG_DESCRIPTION +
                     "The default timeout of the request processing is 10 seconds."
                     + "\n\n" + ControllerConstants.SECURITY_WRITE_CHECK)
     @PreAuthorize("hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN', 'CUSTOMER_USER')")
     @RequestMapping(value = "/", method = RequestMethod.POST)
     @ResponseBody
+    @Deprecated(since = "4.3")
```
ViacheslavKlimov: Not 4.3, but 4.3.1.2

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
```diff
@@ -64,46 +75,91 @@
...
     private AccessValidator accessValidator;
+    @Autowired
+    private EntityService entityService;
+
+    @ApiOperation(value = "Push user message with ACL enrichment to the rule engine (handleRuleEngineRequestV2)",
+            notes = V2_DESCRIPTION + "\n\nUses the originator from the request body, or the calling User id when omitted."
+                    + "\n\n" + ControllerConstants.SECURITY_WRITE_CHECK)
+    @PreAuthorize("hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN', 'CUSTOMER_USER')")
+    @RequestMapping(value = "/v2", method = RequestMethod.POST)
+    @ResponseBody
```
ViacheslavKlimov: PostMapping

ViacheslavKlimov: ResponseBody not needed

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
zzzeebra: Fixed — bumped `@Deprecated(since = "4.3.1.2")` on all four v1 methods and aligned the matching prefix in their `@ApiOperation.notes`.

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
zzzeebra: Fixed — switched to `@PostMapping("/v2")`.

zzzeebra: Fixed — dropped `@ResponseBody`; redundant under `@RestController`.

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
```diff
@@ -179,6 +253,10 @@ public void onSuccess(@Nullable DeferredResult<ResponseEntity> result) {
                     metaData.put("serviceId", serviceInfoProvider.getServiceId());
                     metaData.put("requestUUID", requestId.toString());
                     metaData.put("expirationTime", Long.toString(expTime));
+                    // Server-authoritative keys — written last so any caller value is overwritten.
+                    metaData.put(TbMsgMetaData.TB_USER_ID_KEY, currentUser.getId().getId().toString());
+                    metaData.put(TbMsgMetaData.TB_ACL_KEY, buildAclSnapshot(currentUser, aclEntities));
```
ViacheslavKlimov: The v1 wrappers all funnel through this shared handler, so v1 endpoints now also emit `tb_userId` and `tb_aclSnapshot="[]"` in metadata — keys they didn't write before. The PR description claims "v1 endpoints (`POST /api/rule-engine/...`) are untouched," which isn't accurate.

This is a contract change for existing v1 consumers: any rule chain that iterates metadata or assumes a fixed key set will now see two extra keys. If the intent is to keep v1 strictly identical, gate these two `metaData.put(...)` calls on `aclEntities != null` (i.e., "this is a v2 call"), or pass an explicit `enrichServerKeys` flag through the private helper. Otherwise, update the PR description and the v1 endpoints' Javadoc to mention the new metadata fields.

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
```diff
@@ -209,6 +287,57 @@ public void onFailure(Throwable e) {
...
+        for (EntityId entityId : entities) {
+            Set<String> allowed = new LinkedHashSet<>();
+            Resource resource;
+            try {
+                resource = Resource.of(entityId.getEntityType());
+            } catch (IllegalArgumentException e) {
+                result.add(new EntityAclEntry(entityId, allowed));
+                continue;
+            }
+            HasId<?> entity = entityService.fetchEntity(user.getTenantId(), entityId).orElse(null);
```
ViacheslavKlimov: `entityService.fetchEntity` is only safe to call here if `entityServiceRegistry.getServiceByEntityType(...)` has a mapping — `DefaultEntityServiceRegistry` throws `IllegalArgumentException("Unsupported entity type ...")` for unregistered types. The try/catch above only protects against missing `Resource` mappings, not missing entity-service mappings, so a `Resource`-mapped type that lacks a registered `EntityDaoService` would surface the exception through the async `onSuccess` callback and produce a 500.

It's reasonable to broaden the catch around `fetchEntity` (treat unsupported types the same as missing entities — `allowed=[]`), or assert in the early-return path that all `Resource` values have a registered `EntityDaoService`.

ViacheslavKlimov (commented):
## Review summary

Reviewed 6 changed files in [Added /api/rule-engine/v2 endpoint with per-entity ACL enrichment](https://github.com/thingsboard/thingsboard/pull/15618). Left 8 comment(s) inline.

The core enrichment logic is sound and well-tested. Main concerns: (1) the shared private handler unconditionally writes `tb_userId` and `tb_aclSnapshot` for v1 endpoints too, which contradicts the PR description's "v1 untouched" claim; (2) several PR-description statements (dedup, config path) don't match the implementation; (3) a missing `IllegalArgumentException` catch around `entityService.fetchEntity` can propagate through the async callback for unsupported entity types; (4) the new YAML key is at the top level with hyphen-style naming, inconsistent with surrounding conventions.

## Additional findings

- **rest-client/src/main/java/org/thingsboard/rest/client/RestClient.java** — the project's `pull_request_template.md` checklist explicitly requires "New REST API endpoints must update `RestClient.java`". The four v1 methods are exposed (lines 4321–4355) but there is no corresponding method for `POST /api/rule-engine/v2`. Missing client coverage for the new endpoint.
- **application/src/test/java/org/thingsboard/server/controller/AbstractWebTest.java:534-549** — `loginDifferentCustomer()` is broken on its second invocation: it uses `savedDifferentCustomer.getEmail()` (the Customer entity's email — typically null, since `createDifferentCustomer` only sets the title) and `CUSTOMER_USER_PASSWORD` instead of the user's `DIFFERENT_CUSTOMER_USER_EMAIL` and `DIFFERENT_CUSTOMER_USER_PASSWORD`. The new test added in this PR (`RuleEngineControllerV2EnrichmentTest.testV2TwoCustomersSeeOnlyTheirOwnDevices`) calls `login(...)` directly with a comment acknowledging the bug — fixing the helper would be a better fix and unblock other suites that may hit the same issue.

---

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
```diff
@@ -209,6 +287,57 @@ public void onFailure(Throwable e) {
...
+     * entry resolves to {@code allowed=[]} — ACL enrichment is effectively a no-op for SYS_ADMIN.
+     *
+     * @return serialized JSON array suitable for writing into {@link TbMsgMetaData#TB_ACL_KEY}.
+     */
+    String buildAclSnapshot(SecurityUser user, List<EntityId> entities) {
+        if (entities == null || entities.isEmpty()) {
+            return "[]";
+        }
+        List<EntityAclEntry> result = new ArrayList<>(entities.size());
+        for (EntityId entityId : entities) {
```
ViacheslavKlimov, line 328: The PR description claims "Per-request deduplication of repeated `EntityId` values via a `Map<EntityId, EntityAclEntry>` cache. Output preserves duplicates in input order; DB loads and permission probes happen once per unique id." — but this loop has no dedup cache. Each duplicate id triggers a fresh `fetchEntity` plus 18 `hasPermission` probes. `testV2DuplicateEntitiesPreservedInOutput` only asserts output ordering, not that work was deduplicated, so the gap isn't visible from tests.

Either add the dedup (a `LinkedHashMap<EntityId, EntityAclEntry>` keyed by `EntityId` would preserve order and let you reuse the computed entry for repeated ids), or drop the dedup mention from the description so the docs match the code.

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
```diff
@@ -209,6 +287,57 @@ public void onFailure(Throwable e) {
...
+     * into the entry. Entries with unmapped EntityTypes, missing entities, or entities
+     * that are not {@link HasTenantId} produce {@code allowed=[]}.
+     *
+     * <p>Note: a {@code SYS_ADMIN} caller operates against the system tenant, so tenant-scoped
+     * entities (DEVICE, ASSET, ...) won't be resolved by the tenant-filtered lookup and the
+     * entry resolves to {@code allowed=[]} — ACL enrichment is effectively a no-op for SYS_ADMIN.
+     *
+     * @return serialized JSON array suitable for writing into {@link TbMsgMetaData#TB_ACL_KEY}.
+     */
+    String buildAclSnapshot(SecurityUser user, List<EntityId> entities) {
```
ViacheslavKlimov: Why package-private? `buildAclSnapshot` is only invoked from within this class. There are no tests reaching it directly (everything goes through HTTP), and `Resource.of`/the access service it uses are already accessible without exposing this method. Tightening to `private` avoids inviting accidental coupling from sibling controllers.

application/src/main/resources/thingsboard.yml
```diff
@@ -116,6 +116,13 @@ server:
       # Default timeout for waiting response of REST API request to Rule Engine in milliseconds
       response_timeout: "${DEFAULT_RULE_ENGINE_RESPONSE_TIMEOUT:10000}"
 
+# Rule Engine REST API v2 settings (see /api/rule-engine/v2)
+rule-engine:
```
ViacheslavKlimov: Two inconsistencies here vs. the surrounding file conventions:

1. **Naming style.** Every other key in `thingsboard.yml` uses underscore-separated keys (`rule_engine`, `response_timeout`, `max_entities`). This new block uses hyphens (`rule-engine`, `max-entities`). Spring's relaxed binding makes both work, but the file should pick one style.
2. **Placement.** The PR description claims this is "Placed alongside existing `server.rest.rule_engine.response_timeout` to match local convention," but in practice it sits at the top level. Moving it under `server.rest.rule_engine.acl.max_entities` would actually achieve what the description says and would group all rule-engine-REST tunables in one section.

common/message/src/main/java/org/thingsboard/server/common/msg/TbMsgMetaData.java
```diff
@@ -29,6 +29,21 @@ public final class TbMsgMetaData implements Serializable {
 
     public static final TbMsgMetaData EMPTY = new TbMsgMetaData(0);
 
+    /**
+     * Reserved metadata key. Populated by the platform (e.g. the Rule Engine REST API
+     * v2 enrichment) with a JSON array of {@code EntityAclEntry}. Rule chains MUST
+     * treat this value as authoritative; the platform overwrites any caller-supplied
+     * value of this key before the message is forwarded to the Rule Engine.
+     */
+    public static final String TB_ACL_KEY = "tb_aclSnapshot";
```
ViacheslavKlimov: The constant name (`TB_ACL_KEY`) and its value (`tb_aclSnapshot`) drift apart — "ACL" vs. "aclSnapshot". `TB_ACL_SNAPSHOT_KEY` would match the value, or alternatively the value could be shortened to `tb_acl` to match the constant. As-is, callers reading the key string will see a different name than the Java identifier and have to do a mental translation.

application/src/test/java/org/thingsboard/server/controller/RuleEngineControllerV2EnrichmentTest.java
```diff
@@ -0,0 +1,315 @@
...
+        assertThat(acl.get(0).getEntityId()).isEqualTo(device.getId());
+        assertThat(acl.get(1).getEntityId()).isEqualTo(device.getId());
+        assertThat(acl.get(2).getEntityId()).isEqualTo(other.getId());
+    }
+
+    @Test
+    public void testV2RejectsRequestExceedingMaxEntities() throws Exception {
+        loginTenantAdmin();
+        List<EntityId> tooMany = new ArrayList<>();
+        for (int i = 0; i < 21; i++) {
```
ViacheslavKlimov: The literal `21` couples this test to the YAML default of 20. If someone changes `RULE_ENGINE_ACL_MAX_ENTITIES` (or its default in `thingsboard.yml`) without touching this test, the test silently stops exercising the bound — it might still pass against the new default (if reduced) without validating anything meaningful, or fail in a misleading way (if increased).

Options: inject `@Value("${rule-engine.acl.max-entities:20}") int maxAclEntities` into the test and use `maxAclEntities + 1`, or override the property via `@TestPropertySource(properties = "rule-engine.acl.max-entities=5")` and use `6` here so the test is self-consistent regardless of production defaults.

application/src/test/java/org/thingsboard/server/controller/RuleEngineControllerV2EnrichmentTest.java
```diff
@@ -0,0 +1,315 @@
...
+        assertThat(acl2.get(1).getEntityId()).isEqualTo(deviceB.getId());
+        assertThat(acl2.get(1).getAllowed()).contains("WRITE", "READ_TELEMETRY");
+    }
+
+    @Test
+    public void testV2EmptyAclEntitiesProducesEmptyAcl() throws Exception {
+        loginTenantAdmin();
+
+        RuleEngineV2Request request = baseRequest();
+        // aclEntities left null
```
ViacheslavKlimov, line 166: This only exercises the `entities == null` branch in `buildAclSnapshot`. The `entities.isEmpty()` branch (caller passes `List.of()`) is also a valid v2 input and is part of the same short-circuit, but isn't covered. A second case calling `request.setAclEntities(List.of())` (or parameterizing this test over both inputs) would close that small gap.

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
zzzeebra: Fixed — v1 wrappers no longer emit the two keys. The shared `handleRuleEngineRequest` helper writes `tb_userId` and `tb_aclSnapshot` only when `aclEntities != null`; v1 wrappers pass `null` and skip the writes. The v2 endpoint always passes a non-null list (defaults to `List.of()` when the caller omits `aclEntities`), so v2 always emits the keys. PR description updated.

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
zzzeebra: Fixed — `computeAclEntry` now also catches `IllegalArgumentException` from `entityService.fetchEntity`, so an EntityType without a registered `EntityDaoService` resolves to `allowed=[]` instead of bubbling a 500 through the async callback.

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
zzzeebra: Fixed — restored the `Map<EntityId, EntityAclEntry>` cache in `buildAclSnapshot`. `testV2DuplicateEntitiesPreservedInOutputAndDedupedInWork` now asserts via Mockito `verify(times(1))` that `entityService.fetchEntity` runs once per unique id.

application/src/main/java/org/thingsboard/server/controller/RuleEngineController.java
zzzeebra: Fixed — visibility tightened to `private`.

application/src/main/resources/thingsboard.yml
zzzeebra: Fixed — yml keys and `@Value` renamed to snake (`rule_engine.acl.max_entities`) to match the surrounding file convention. Placement stays top-level (matches mirror PR thingsboard-pe#4731). PR description updated.

common/message/src/main/java/org/thingsboard/server/common/msg/TbMsgMetaData.java
zzzeebra: Fixed (constant rename only) — `TB_ACL_KEY` → `TB_ACL_SNAPSHOT_KEY`, identifier now matches the value. Wire format kept as `tb_aclSnapshot` (camelCase) to align with existing metadata-key conventions in this codebase: `serviceId`, `requestUUID`, `expirationTime`, `userId`, `assignedCustomerId`, … — all camelCase. Snake values would stand out as foreign.

application/src/test/java/org/thingsboard/server/controller/RuleEngineControllerV2EnrichmentTest.java
zzzeebra: Fixed — added `@TestPropertySource(properties = "rule_engine.acl.max_entities=5")` on the test class, introduced `MAX_ACL_ENTITIES = 5` constant, and the exceeds-max test now uses `MAX_ACL_ENTITIES + 1`. Self-contained, independent of the production default.

application/src/test/java/org/thingsboard/server/controller/RuleEngineControllerV2EnrichmentTest.java
zzzeebra: Fixed — added `testV2EmptyAclEntitiesListProducesEmptyAcl` covering the `entities.isEmpty()` branch with `request.setAclEntities(List.of())`. The existing test was renamed to `testV2NullAclEntitiesProducesEmptyAcl` to keep the two branches separately addressable.

zzzeebra:
Re: auto-review "Additional findings" — **RestClient.java**: fixed in 3d77dd1 (added `handleRuleEngineRequestV2(RuleEngineV2Request request)` mirroring the existing v1 variants).  **`loginDifferentCustomer()` bug**: fixed in 8c49119 (second-invocation login now uses `DIFFERENT_CUSTOMER_USER_EMAIL` + `DIFFERENT_CUSTOMER_USER_PASSWORD` instead of the Customer entity email and `CUSTOMER_USER_PASSWORD`).

PR #15627, opened 2026-05-13
Fixed SSRF bypass in Version Control Git integration

zzzeebra:
## Summary

Closes the Git-specific SSRF protection bypass in Version Control, where a tenant-controlled `repositoryUri` was passed to JGit without applying the configured SSRF validator. This allowed `TENANT_ADMIN` to make the backend reach internal Git / HTTP services through `git://`, `http(s)://`, `ssh://`, and `file://` schemes, even when those targets were configured as blocked.

## What changed

- New `RepositoryUriValidator` invoked on save, access check, and every JGit network operation (`clone`, `test`, `fetch`, `push`)
- Scheme whitelist (`http`, `https`, `ssh`) — rejects `file://`, `git://`, etc. regardless of the `SSRF_PROTECTION_ENABLED` flag
- Host validation uses `URIish.getHost()` — the same parser JGit transport uses — to avoid parser-disagreement bypasses such as `http://evil.com@127.0.0.1/`
- `HostScopedCredentialsProvider` refuses credentials when JGit follows a cross-host redirect
- JGit HTTP redirects disabled globally via a `SystemReader` override at VC service startup (`http.followRedirects=false`)
- `SsrfProtectionValidator` refactored to expose `validateHost(String)` as public API; existing `validateUri(URI)` behavior is unchanged

## Coverage

| Disclosure item | Closed by |
|---|---|
| Tier 1 #1–5: SSRF bypass, branch/commit metadata, HTTP probe, entity import/restore | `RepositoryUriValidator` at save + checkAccess + JGit sinks |
| Tier 1 #6, Tier 2 #7: `file://` cache, cross-tenant cache | Scheme whitelist (rejects `file://`) |
| Tier 2 #8: Credential cross-host redirect forwarding | `HostScopedCredentialsProvider` + redirect disable |
| Tier 2 #9: Internal Git write | URL does not pass save → push impossible |
| Mitigations M1–M7 | Covered |

## Known limitations

- **M8 — DNS rebinding**: closed under default Java DNS cache (`networkaddress.cache.ttl > 0`). The validator runs immediately before each JGit network operation, so the TOCTOU window between our resolve and JGit's resolve is microseconds and Java's positive DNS cache returns the same address. Degraded if the administrator sets `-Dnetworkaddress.cache.ttl=0` for DNS failover.
- **`localOnly` path traversal**: pre-existing pattern in `DefaultGitRepositoryService.openOrCloneRepository` where `repositoryUri` is concatenated into a filesystem path when `localOnly=true`. Not reachable via tenant API — `AdminController.saveRepositorySettings` and `checkRepositoryAccess` both force `settings.setLocalOnly(false)` server-side. Out of scope for this disclosure-driven fix; can be tightened in a follow-up.
- **Submodule / Git-LFS secondary requests**: not exercised by ThingsBoard's JGit usage (no `setCloneSubmodules(true)` anywhere; no `jgit-lfs` dependency). Reporter does not claim impact.

## Automated tests

- `SsrfProtectionValidatorTest` — 79/79 pass (existing 63 + 16 new direct tests for `validateHost(String)`)
- `RepositoryUriValidatorTest` — 25/25 pass (skip-localOnly, scheme whitelist, host checks, SSRF integration for HTTP and SSH, allow-list bypass, parser-confusion edge cases)
- `HostScopedCredentialsProviderTest` — 6/6 pass (same-host, cross-host, case-insensitive, null handling)
- `mvn -pl common/util,common/version-control,application -am compile` — BUILD SUCCESS

## Manual verification

Performed against a local TB build (lts-4.2 + this branch) via UI → `Advanced features → Version control` → "Check access". URIs grounded in the PoC where applicable.

### Scheme whitelist (SSRF protection disabled)

**1. `file://` scheme rejected**
- URI: `file:///tmp/repositories/2f05d570-490d-11f1-92e5-27670d9a01fc`
- Result: `Repository URI scheme 'file' is not allowed. Allowed schemes: http, https, ssh`

**2. `git://` scheme rejected**
- URI: `git://172.24.0.2/internal-secrets.git`
- Result: `Repository URI scheme 'git' is not allowed. Allowed schemes: http, https, ssh`

**3. Public Git URL passes the validator**
- URI: `https://github.com/thingsboard/thingsboard.git`
- Result: `not authorized` — GitHub's response to unauthenticated push attempt. Validator passed the URL through; rejection comes from GitHub, not SSRF protection.

### SSRF protection enabled (`-Dactors.rule.external.ssrf_protection_enabled=true`)

**4. Loopback rejected**
- URI: `https://127.0.0.1/repo.git`
- Result: `Repository URI is not allowed: URI is invalid: host '127.0.0.1' is not allowed. If this host should be accessible, contact the platform administrator about SSRF allow-list.`

**5. RFC1918 private IP rejected (from PoC)**
- URI: `http://172.24.0.2/internal-secrets.git`
- Result: `host '172.24.0.2' is not allowed...`

**6. AWS instance metadata endpoint rejected**
- URI: `http://169.254.169.254/latest/meta-data/`
- Result: `host '169.254.169.254' is not allowed...`

**7. Parser-confusion bypass attempt rejected**
- URI: `http://github.com@127.0.0.1/foo.git`
- Result: `host '127.0.0.1' is not allowed...`
- `URIish` parses `github.com` as userinfo and `127.0.0.1` as host. Validator and JGit transport see the same target — no parser-disagreement bypass.

**8. Public Git URL still passes with SSRF enabled**
- URI: `https://github.com/thingsboard/thingsboard.git`
- Result: `not authorized` — same GitHub auth response. SSRF protection does not affect legitimate public targets.

### DIFF
--- a/common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/DefaultGitRepositoryService.java
+++ b/common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/DefaultGitRepositoryService.java
@@ -95,6 +95,8 @@ public class DefaultGitRepositoryService implements GitRepositoryService {
         @Override
         public StoredConfig getUserConfig() throws ConfigInvalidException, IOException {
             StoredConfig config = delegate.getUserConfig();
+            // SSRF hardening: keep http.followRedirects off — a followed
+            // redirect on the info/refs GET is itself the probe being closed.
             config.setString("http", null, "followRedirects", "false");
             return config;
         }
--- a/common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/DefaultGitRepositoryService.java
+++ b/common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/DefaultGitRepositoryService.java
@@ -20,6 +20,12 @@ import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
 import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.errors.ConfigInvalidException;
+import org.eclipse.jgit.lib.Config;
+import org.eclipse.jgit.lib.StoredConfig;
+import org.eclipse.jgit.storage.file.FileBasedConfig;
+import org.eclipse.jgit.util.FS;
+import org.eclipse.jgit.util.SystemReader;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Service;
@@ -67,6 +73,70 @@ public class DefaultGitRepositoryService implements GitRepositoryService {
     public void init() {
         if (StringUtils.isEmpty(repositoriesFolder)) {
             repositoriesFolder = defaultFolder;
+        }
+        installRedirectHardeningSystemReader();
+    }
+
+    private static synchronized void installRedirectHardeningSystemReader() {
+        SystemReader current = SystemReader.getInstance();
+        if (current instanceof RedirectHardeningSystemReader) {
+            return;
+        }
+        SystemReader.setInstance(new RedirectHardeningSystemReader(current));
+    }
+
+    private static final class RedirectHardeningSystemReader extends SystemReader {
+        private final SystemReader delegate;
+
+        RedirectHardeningSystemReader(SystemReader delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public StoredConfig getUserConfig() throws ConfigInvalidException, IOException {
+            StoredConfig config = delegate.getUserConfig();
+            config.setString("http", null, "followRedirects", "false");
+            return config;
+        }
+
+        @Override
+        public String getHostname() {
+            return delegate.getHostname();
+        }
+
+        @Override
+        public String getenv(String variable) {
+            return delegate.getenv(variable);
+        }
+
+        @Override
+        public String getProperty(String key) {
+            return delegate.getProperty(key);
+        }
+
+        @Override
+        public FileBasedConfig openUserConfig(Config parent, FS fs) {
+            return delegate.openUserConfig(parent, fs);
+        }
+
+        @Override
+        public FileBasedConfig openSystemConfig(Config parent, FS fs) {
+            return delegate.openSystemConfig(parent, fs);
+        }
+
+        @Override
+        public FileBasedConfig openJGitConfig(Config parent, FS fs) {
+            return delegate.openJGitConfig(parent, fs);
+        }
+
+        @Override
+        public long getCurrentTime() {
+            return delegate.getCurrentTime();
+        }
+
+        @Override
+        public int getTimezone(long when) {
+            return delegate.getTimezone(when);
         }
     }
 
--- a/common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/GitRepository.java
+++ b/common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/GitRepository.java
@@ -57,7 +57,6 @@ import org.eclipse.jgit.transport.RefSpec;
 import org.eclipse.jgit.transport.RemoteRefUpdate;
 import org.eclipse.jgit.transport.SshTransport;
 import org.eclipse.jgit.transport.URIish;
-import org.eclipse.jgit.transport.UsernamePasswordCredentialsProvider;
 import org.eclipse.jgit.transport.sshd.JGitKeyCache;
 import org.eclipse.jgit.transport.sshd.ServerKeyDatabase;
 import org.eclipse.jgit.transport.sshd.SshdSessionFactory;
@@ -129,6 +128,7 @@ public class GitRepository {
     }
 
     public static GitRepository clone(RepositorySettings settings, File directory) throws GitAPIException {
+        RepositoryUriValidator.validate(settings);
         log.debug("Executing clone [{}]", settings.getRepositoryUri());
         CloneCommand cloneCommand = Git.cloneRepository()
                 .setURI(settings.getRepositoryUri())
@@ -179,6 +179,7 @@ public class GitRepository {
         if (settings.isLocalOnly()) {
             return;
         }
+        RepositoryUriValidator.validate(settings);
         log.debug("Executing test [{}]", settings.getRepositoryUri());
         AuthHandler authHandler = AuthHandler.createFor(settings, directory);
         if (settings.isReadOnly()) {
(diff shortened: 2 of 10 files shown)


common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/DefaultGitRepositoryService.java
```diff
@@ -68,6 +74,70 @@ public void init() {
...
+        private final SystemReader delegate;
+
+        RedirectHardeningSystemReader(SystemReader delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public StoredConfig getUserConfig() throws ConfigInvalidException, IOException {
+            StoredConfig config = delegate.getUserConfig();
+            config.setString("http", null, "followRedirects", "false");
```
ViacheslavKlimov, line 100: Disabling `http.followRedirects` here flips it off for *every* JGit HTTP operation in the JVM, not just attacker-controlled URLs. Legitimate flows that rely on redirects will now fail for all VC repos — notably GitHub/GitLab 301s for renamed repositories and `http://`→`https://` upgrades (e.g. a tenant who configured `http://github.com/org/repo.git`).

To be clear, `false` is the correct *security* choice — `INITIAL` would still follow the redirect on the `info/refs` GET, which is exactly the SSRF probe you're closing — so I'm not suggesting weakening it. Since `HostScopedCredentialsProvider` already prevents credentials from being forwarded across a cross-host redirect, the credential-leak vector is closed independently of this flag, and this flag is what closes the bare-probe vector. The thing worth confirming is just that this is an acceptable behavioral change for an LTS patch: legitimate redirect-based clones/fetches stop working platform-wide, including for the internal `DefaultGitSyncService`. A line in the upgrade/release notes would help operators who hit it.

common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/GitRepository.java
```diff
@@ -560,8 +563,16 @@ protected void configureCommand(TransportCommand command) {
...
 
-        private static CredentialsProvider newCredentialsProvider(String username, String password) {
-            return new UsernamePasswordCredentialsProvider(username, password == null ? "" : password);
+        private static CredentialsProvider newCredentialsProvider(String username, String password, String allowedHost) {
+            return new HostScopedCredentialsProvider(username, password, allowedHost);
+        }
+
+        private static String extractHost(String repositoryUri) {
+            try {
+                return new URIish(repositoryUri).getHost();
```
ViacheslavKlimov: `RepositoryUriValidator.validate()` parses `new URIish(uri.trim())`, but here the same URI is parsed untrimmed. For a stored URI with surrounding whitespace the two parsers can disagree on the host (or `getHost()` returns null here), and since `HostScopedCredentialsProvider` refuses to supply credentials when `allowedHost` is null or doesn't match the request host, an otherwise-valid private repo would silently fail authentication. Trimming consistently (`new URIish(repositoryUri.trim())`) keeps the validated host and the credential-scoping host in sync.

common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/GitRepository.java
```diff
@@ -560,8 +563,16 @@ protected void configureCommand(TransportCommand command) {
             }
         }
 
-        private static CredentialsProvider newCredentialsProvider(String username, String password) {
-            return new UsernamePasswordCredentialsProvider(username, password == null ? "" : password);
+        private static CredentialsProvider newCredentialsProvider(String username, String password, String allowedHost) {
+            return new HostScopedCredentialsProvider(username, password, allowedHost);
```
ViacheslavKlimov, line 567: Minor: now that this just forwards to `new HostScopedCredentialsProvider(...)`, the `newCredentialsProvider` indirection no longer earns its keep — could be inlined at the call site.

ViacheslavKlimov (commented):
## Review summary

Reviewed 9 changed files in [Fixed SSRF bypass in Version Control Git integration](https://github.com/thingsboard/thingsboard/pull/15627). This is a well-constructed, defense-in-depth security fix: scheme whitelist + host validation at the four JGit network sinks (`clone`/`test`/`fetch`/`push`) and at save/check-access, a host-scoped credentials provider that won't forward credentials across a cross-host redirect, and a global JGit `http.followRedirects=false`. I traced the wiring and confirmed the placement is sound — `restore()` only refills credentials (never the URI), so validating the pre-restore settings is correct; `AdminController` forces `localOnly=false` so the `localOnly` skip can't be abused via the tenant API; and JGit 6.10.1's `HttpConfig(URL)` reads `getUserConfig()`, so the redirect-disable applies to the repo-less `lsRemoteRepository` path used by read-only `test()` too. The test coverage is thorough.

Left 3 comment(s) inline — the main one is a behavioral trade-off around disabling redirects globally; the rest are minor consistency/cleanup items.

## Additional findings

- **application/src/main/java/org/thingsboard/server/controller/AdminController.java:307 and :336** — The new validator deliberately skips all checks when `settings.isLocalOnly()` is true, so the SSRF defense leans on these two `settings.setLocalOnly(false)` lines forcing local-only off for every API-driven save/check-access. The trailing `// only to be used in tests` comment is misleading: it reads like test-only scaffolding, but it's actually a load-bearing part of the security model. Worth rewording so a future reader doesn't innocently remove it.

---

common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/DefaultGitRepositoryService.java
zzzeebra: Confirmed acceptable for the LTS patch. The only default-on internal git flow is the gateways-dashboard sync (`DashboardSyncService`, enabled by default), whose default URL `https://github.com/thingsboard/gateway-management-extensions-dist.git` serves `info/refs` with HTTP 200 / no redirect — unaffected. Redirects only bite `http://` URLs (GitHub 301→https) or renamed repos. Left a comment at the disable site explaining why redirects stay off.

common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/GitRepository.java
zzzeebra: Fixed in 15fdc9c — `new URIish(repositoryUri.trim()).getHost()`, consistent with the validator.

common/version-control/src/main/java/org/thingsboard/server/service/sync/vc/GitRepository.java
zzzeebra: Leaving as-is — keeps it symmetric with the sibling `newSshdSessionFactory(...)` helper in the same auth branch.

zzzeebra:
Additional finding (AdminController localOnly): reworded the `setLocalOnly(false)` comment on both `saveRepositorySettings` and `checkRepositoryAccess` to state it forces local-only off for the tenant API as part of the SSRF defense, not test-only scaffolding — 15fdc9c.

PR #15640, opened 2026-05-14
Added handler for WebGL unavailability in map widget

mtsymbarov-del:
## Pull Request description

Знімок екрана 2026-05-15 о 12 07 36

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
@@ -24,7 +24,7 @@ import {
   TbPolygonCoordinate,
   TbPolygonCoordinates,
   TbPolygonRawCoordinate,
-  TbPolygonRawCoordinates
+  TbPolygonRawCoordinates, WEBGL_ERROR_EVENT
 } from '@shared/models/widget/maps/map.models';
 import { WidgetContext } from '@home/models/widget-component.models';
 import { DeepPartial } from '@shared/models/common';
@@ -92,18 +92,19 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
     return this.loadLayers().pipe(
       tap((layers: L.TB.LayerData[]) => {
         if (layers.length) {
-          const layer = layers[0];
+          const defaultLayer = layers[0];
           layers.forEach(layer => {
-            layer.layer.once('gl-error', () => {
-              const toastPosition = this.settings.controlsPosition === MapControlsPosition.bottomleft || this.settings.controlsPosition === MapControlsPosition.bottomright ? 'top' : 'bottom';
+            layer.layer.once(WEBGL_ERROR_EVENT, () => {
+              const toastPosition = this.settings.controlsPosition === MapControlsPosition.bottomleft
+              || this.settings.controlsPosition === MapControlsPosition.bottomright ? 'top' : 'bottom';
               this.ctx.showErrorToast(
-                this.ctx.translate.instant('widgets.maps.gl.webgl-not-available'),
+                this.ctx.translate.instant('widgets.maps.layer.webgl-not-available'),
                 toastPosition, 'left', this.ctx.toastTargetId, true
               );
             });
-          })
-          layer.layer.addTo(this.map);
-          this.map.attributionControl.setPrefix(layer.attributionPrefix);
+          });
+          defaultLayer.layer.addTo(this.map);
+          this.map.attributionControl.setPrefix(defaultLayer.attributionPrefix);
           if (layers.length > 1) {
             const sidebar = this.getSidebar();
             L.TB.layers({
--- a/ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
@@ -93,6 +93,12 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
       tap((layers: L.TB.LayerData[]) => {
         if (layers.length) {
           const layer = layers[0];
+          layer.layer.once('gl-error', () => {
+            this.ctx.showErrorToast(
+              this.ctx.translate.instant('widgets.maps.gl.webgl-not-available'),
+              'top', 'center', this.ctx.toastTargetId, true
+            );
+          });
           layer.layer.addTo(this.map);
           this.map.attributionControl.setPrefix(layer.attributionPrefix);
           if (layers.length > 1) {
--- a/ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
@@ -96,7 +96,7 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
           layer.layer.once('gl-error', () => {
             this.ctx.showErrorToast(
               this.ctx.translate.instant('widgets.maps.gl.webgl-not-available'),
-              'top', 'center', this.ctx.toastTargetId, true
+              'bottom', 'left', this.ctx.toastTargetId, true
             );
           });
           layer.layer.addTo(this.map);
--- a/ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
@@ -18,7 +18,7 @@ import {
   DEFAULT_ZOOM_LEVEL,
   defaultGeoMapSettings,
   GeoMapSettings,
-  latLngPointToBounds,
+  latLngPointToBounds, MapControlsPosition,
   MapZoomAction,
   TbCircleData,
   TbPolygonCoordinate,
@@ -93,12 +93,15 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
       tap((layers: L.TB.LayerData[]) => {
         if (layers.length) {
           const layer = layers[0];
-          layer.layer.once('gl-error', () => {
-            this.ctx.showErrorToast(
-              this.ctx.translate.instant('widgets.maps.gl.webgl-not-available'),
-              'bottom', 'left', this.ctx.toastTargetId, true
-            );
-          });
+          layers.forEach(layer => {
+            layer.layer.once('gl-error', () => {
+              const toastPosition = this.settings.controlsPosition === MapControlsPosition.bottomleft || this.settings.controlsPosition === MapControlsPosition.bottomright ? 'top' : 'bottom';
+              this.ctx.showErrorToast(
+                this.ctx.translate.instant('widgets.maps.gl.webgl-not-available'),
+                toastPosition, 'left', this.ctx.toastTargetId, true
+              );
+            });
+          })
           layer.layer.addTo(this.map);
           this.map.attributionControl.setPrefix(layer.attributionPrefix);
           if (layers.length > 1) {
--- a/ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet/leaflet-tb.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet/leaflet-tb.ts
@@ -24,6 +24,7 @@ import { MatIconRegistry } from '@angular/material/icon';
 import { isSvgIcon, splitIconName } from '@shared/models/icon.models';
 import { catchError, take } from 'rxjs/operators';
 import { of } from 'rxjs';
+import { WEBGL_ERROR_EVENT } from '@shared/models/widget/maps/map.models';
 
 L.MarkerCluster = L.MarkerCluster.mergeOptions({ pmIgnore: true });
 
@@ -1212,12 +1213,12 @@ class MapLibreGLLayer extends L.Layer implements TB.MapLibreGL.MapLibreGLLayer {
       this._glMap = new MapLibreGLMap(options);
     } catch (e) {
       this._glError = true;
-      this.fire('gl-error', { error: e });
+      this.fire(WEBGL_ERROR_EVENT, { error: e });
       return;
     }
     this._glMap.once('webglcontextlost', (e) => {
       this._glError = true;
-      this.fire('gl-error', {error: e});
+      this.fire(WEBGL_ERROR_EVENT, {error: e});
     });
     this._glMap.once('load', () => {
       this.fire('load');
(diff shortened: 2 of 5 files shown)


ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
```diff
@@ -93,6 +93,12 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
       tap((layers: L.TB.LayerData[]) => {
         if (layers.length) {
           const layer = layers[0];
+          layer.layer.once('gl-error', () => {
```
vvlladd28: The `gl-error` listener is only attached to `layers[0]`, but a few lines down (`L.TB.layers({ layers, ... })` at line 106) you wire up a switcher that lets the user pick any of the configured layers at runtime. When the user switches to a different `MapLibreGLLayer` and it can't initialize WebGL, no toast will appear. Consider iterating over `layers` and attaching the listener to each one — or moving the listener-wiring into `loadLayers()` / `TbMapLayer.loadLayer` so it's set up uniformly for every layer.

ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
```diff
@@ -93,6 +93,12 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
...
+            this.ctx.showErrorToast(
+              this.ctx.translate.instant('widgets.maps.gl.webgl-not-available'),
+              'bottom', 'left', this.ctx.toastTargetId, true
```
vvlladd28: The toast is hardcoded to `'bottom', 'left'`, but the map's controls position is user-configurable (`MapControlsPosition` — `topleft`/`topright`/`bottomleft`/`bottomright`, see `map.models.ts:720`). When `this.settings.controlsPosition === 'bottomleft'` the toast will overlap the zoom/layers controls; on `'bottomright'` it'll at least share the same row and looks cramped. Worth picking a toast corner that depends on `this.settings.controlsPosition` (e.g. opposite vertical), or anchoring the toast somewhere that won't collide regardless of the controls config.

ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet/leaflet-tb.ts
```diff
@@ -1204,7 +1207,17 @@ class MapLibreGLLayer extends L.Layer implements TB.MapLibreGL.MapLibreGLLayer {
       zoom: this._map.getZoom() - 1,
       attributionControl: false
     });
-    this._glMap = new MapLibreGLMap(options);
+    try {
+      this._glMap = new MapLibreGLMap(options);
+    } catch (e) {
+      this._glError = true;
```
vvlladd28, line 1215: `_glError` is set to `true` here (and in the `webglcontextlost` handler below) but never reset anywhere. If the layer is removed and re-added — e.g. via the layer switcher, or a widget reconfiguration — `_initGL()` runs again, but every guard in this file still sees `_glError === true` and the layer stays inert even when the retry would have succeeded. Resetting `this._glError = false` at the top of `_initGL()` would let retries recover.

ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet/leaflet-tb.ts
```diff
@@ -1204,7 +1207,17 @@ class MapLibreGLLayer extends L.Layer implements TB.MapLibreGL.MapLibreGLLayer {
...
+      this.fire('gl-error', { error: e });
+      return;
+    }
+    this._glMap.once('webglcontextlost', () => {
+      this._glError = true;
+      this.fire('gl-error', {});
```
vvlladd28: Minor inconsistency: the catch branch above fires `gl-error` with `{ error: e }`, but this branch passes `{}`. A consumer that wants to log or surface the cause gets different shapes depending on which path triggered the error. Either pass the `webglcontextlost` event through, or distinguish the two via a `reason` flag — whatever fits the consumer side better.

ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet/leaflet-tb.ts
```diff
@@ -1253,20 +1266,23 @@ class MapLibreGLLayer extends L.Layer implements TB.MapLibreGL.MapLibreGLLayer {
   private _transformGL(gl: MapLibreGLMap) {
     const center = this._map.getCenter();
     const tr = gl._getTransformForUpdate();
+    if (!tr) { return; }
```
vvlladd28, line 1275: This null guard looks unrelated to the WebGL-unavailability story the rest of the PR is about — when does `_getTransformForUpdate()` actually return falsy? If you hit this during testing, a one-line comment explaining the failure case would help future readers; otherwise it reads as a defensive check pulled in from somewhere else and pulls reviewers off track.

ui-ngx/src/app/modules/home/components/widget/lib/maps/map-layer.ts
```diff
@@ -120,6 +120,9 @@ export abstract class TbMapLayer<S extends MapLayerSettings> {
                     let referenceLayerLoaded = false;
                     baseLayer.addTo(layer);
                     referenceLayer.addTo(layer);
+                    baseLayer.once('gl-error', (e: any) => {
```
vvlladd28: Two things here:

1. Only `baseLayer` is forwarding `gl-error`, but `referenceLayer` (loaded via `loadReferenceLayer()` at line 162) is itself a `MapLibreGLLayer` and can hit the same WebGL failures. If only the reference layer fails, the error gets silently swallowed — worth forwarding `gl-error` from `referenceLayer` too.

2. `(e: any)` could be `LeafletEvent` (with the optional `{ error?: unknown }` payload the layer fires). Tightening the type makes the forwarded shape easier to reason about for the consumer in `geo-map.ts`.

ui-ngx/src/assets/locale/locale.constant-en_US.json
```diff
@@ -8249,6 +8249,9 @@
                 "default-map-zoom-level": "Default map zoom level",
                 "entities-limit": "Limit of entities to load"
             },
+            "gl": {
+                "webgl-not-available": "The browser tried to use the GPU to draw the map, but the GPU is turned off or blocked."
```
vvlladd28: Two things:

1. The wording reads vague for a toast — "The browser tried to use the GPU to draw the map, but the GPU is turned off or blocked" doesn't tell the user what to do, and the GPU/WebGL terminology won't mean much to most users. Something action-oriented would land better, e.g.:
   - "The selected map layer requires graphics acceleration to display. Please enable 'Hardware acceleration' in your browser settings, or switch to a different layer."
   - "Unable to load this map layer because your browser's graphics acceleration is disabled or blocked. Try enabling hardware acceleration or choosing another layer."

2. This key feels misplaced under a brand-new `gl` namespace — the message is really about a map layer failing to load, so it would sit more naturally under the existing `widgets.maps.layer.*` namespace (e.g. `widgets.maps.layer.webgl-not-available` or `widgets.maps.layer.acceleration-unavailable`). That avoids introducing a one-off `gl` group and groups related strings together.

vvlladd28 (commented):
## Review summary

Reviewed 4 changed files in [Added handler for WebGL unavailability in map widget](https://github.com/thingsboard/thingsboard/pull/15640). The defensive guards in `MapLibreGLLayer` and the new `gl-error` event plumbing make sense for the failure case, but there are a few gaps worth a look: the toast is only wired up for the first layer (so errors from layers switched in via the layer control are silent), `_glError` is never reset (so a layer that fails once stays inert even if a later retry would have succeeded), the reference layer's WebGL failures aren't propagated, and the toast position can collide with the map controls. Left 7 comment(s) inline.

---
*This is an auto-generated review. Findings may contain errors — please verify before applying changes.*

ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
```diff
@@ -93,6 +93,15 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
       tap((layers: L.TB.LayerData[]) => {
         if (layers.length) {
           const layer = layers[0];
+          layers.forEach(layer => {
+            layer.layer.once('gl-error', () => {
+              const toastPosition = this.settings.controlsPosition === MapControlsPosition.bottomleft || this.settings.controlsPosition === MapControlsPosition.bottomright ? 'top' : 'bottom';
+              this.ctx.showErrorToast(
+                this.ctx.translate.instant('widgets.maps.gl.webgl-not-available'),
```
vvlladd28: This key won't resolve at runtime. The translation was added under `widgets.maps.layer` (see `ui-ngx/src/assets/locale/locale.constant-en_US.json:8264`), but it's looked up here as `widgets.maps.gl.webgl-not-available`. There's no `gl` block under `maps`, so `translate.instant(...)` will return the key itself and the toast will display the literal string `widgets.maps.gl.webgl-not-available` instead of the message. Either change the key here to `widgets.maps.layer.webgl-not-available`, or move the locale entry into a new `gl` block to match.

ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
```diff
@@ -93,6 +93,15 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
       tap((layers: L.TB.LayerData[]) => {
         if (layers.length) {
           const layer = layers[0];
+          layers.forEach(layer => {
```
vvlladd28, line 96: `const layer = layers[0]` (line 95) is immediately shadowed by the `layer` parameter of this `forEach` callback. Two bindings with the same name in adjacent scopes are easy to misread — renaming the callback param (e.g. `layers.forEach(l => l.layer.once('gl-error', ...))`) would avoid the collision.

ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
```diff
@@ -93,6 +93,15 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
...
+            layer.layer.once('gl-error', () => {
```
vvlladd28: `'gl-error'` is a magic string now repeated across this listener, the re-fire in `map-layer.ts`, and both `fire('gl-error', ...)` sites in `leaflet-tb.ts`. A shared constant would prevent a silent break from a typo in any one of them and make the event contract discoverable.

ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
```diff
@@ -93,6 +93,15 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
...
+              const toastPosition = this.settings.controlsPosition === MapControlsPosition.bottomleft || this.settings.controlsPosition === MapControlsPosition.bottomright ? 'top' : 'bottom';
```
vvlladd28: This inline ternary is hard to parse at a glance, and it only handles the two bottom positions explicitly — any other `controlsPosition` value silently falls into `'bottom'`. Extracting a small helper (e.g. `isBottomControls()`) or at least breaking the condition across lines would make the intent ("show the toast away from where the controls are") clearer.

ui-ngx/src/app/modules/home/components/widget/lib/maps/geo-map.ts
```diff
@@ -93,6 +93,15 @@ export class TbGeoMap extends TbMap<GeoMapSettings> {
...
+              this.ctx.showErrorToast(
+                this.ctx.translate.instant('widgets.maps.gl.webgl-not-available'),
+                toastPosition, 'left', this.ctx.toastTargetId, true
+              );
+            });
+          })
```
vvlladd28: Missing semicolon after the `forEach(...)` block — the rest of the file terminates statements, so worth fixing for style/lint consistency.

ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet/leaflet-tb.ts
```diff
@@ -1253,20 +1267,23 @@ class MapLibreGLLayer extends L.Layer implements TB.MapLibreGL.MapLibreGLLayer {
...
   private _pinchZoom() {
+    if (this._glError) { return; }
     this._glMap.jumpTo({
       zoom: this._map.getZoom() - 1,
       center: this._map.getCenter()
     });
   }
 
   private _animateZoom(e: L.ZoomAnimEvent) {
+    if (this._glError || !this._actualCanvas) { return; }
```
vvlladd28: This `_glError`/`_actualCanvas` guard is now copy-pasted across `_update`, `_pinchZoom`, `_animateZoom`, `_zoomEnd`, and `_transitionEnd` (plus its `requestAnimFrame` callback). A single predicate like `private _glReady(): boolean { return !this._glError && !!this._glMap; }` would keep the precondition in one place and make it harder for a future method to forget it; the animation methods could compose the `_actualCanvas` check on top.

ui-ngx/src/app/modules/home/components/widget/lib/maps/map-layer.ts
```diff
@@ -120,6 +120,9 @@ export abstract class TbMapLayer<S extends MapLayerSettings> {
                     let referenceLayerLoaded = false;
                     baseLayer.addTo(layer);
                     referenceLayer.addTo(layer);
+                    baseLayer.once('gl-error', (e: LeafletEvent) => {
```
vvlladd28: This `gl-error` forwarding is wired up only in the `referenceLayer` branch, because here the base layer is wrapped in a `featureGroup` that `geo-map` listens on. In the other branches the `MapLibreGLLayer` is returned directly and fires `gl-error` itself, so no forwarding is needed — but the asymmetry isn't obvious. A short comment explaining why forwarding is only required when the group wraps the base layer would save the next maintainer some head-scratching.

vvlladd28 (commented):
## Review summary

Reviewed 4 changed files in [Added handler for WebGL unavailability in map widget](https://github.com/thingsboard/thingsboard/pull/15640). Left 7 comment(s) inline.

The approach is sound and well-guarded — the `try/catch` around GL map construction plus the `_glError`/`_actualCanvas` checks across the layer lifecycle look correct, and `gl-error` propagation works in both the single-layer and `featureGroup` paths. The one functional issue is the toast's translation key (`widgets.maps.gl.webgl-not-available`) not matching where the string was actually added in the locale file (under `widgets.maps.layer`), which would surface the raw key to users. The remaining comments are readability/maintainability suggestions.

---

vvlladd28 (approved):

vvlladd28 (approved):

PR #15641, opened 2026-05-15
Fixed duplicate alarms created by concurrent rule engine messages

### LINKED ISSUE #15583: Duplicated Alarms
**Describe the bug**
If there are two messages from the same originator triggering the same create alarm rule node just a few ms apart there will be two alarms of the same type created.

**Your Server Environment**
* own setup
  * Deployment: monolith and microservices
  * Deployment type: deb,  k8s
  * ThingsBoard Version 4.2.1.1 PE and 4.3.0.1 PE
  * Professional Edition

**Your Client Environment**
**Desktop (please complete the following information):**

* OS: Windows 11
* Browser MS Edge
* Version Version 147.0.3912.98
* 
**To Reproduce**
1. Create Rule Chain
2. Add two generator nodes (same msg originator) and create alarm node
3. Checkout Alarm Dashboard
4. 
**Expected behavior**
Just one Alarm is being created

**Screenshots**

Image

Image

Image

**Additional context**
Example Rule Chain:

[alarm_test.json](https://github.com/user-attachments/files/27432250/alarm_test.json)

________________________________________________________________
**Disclaimer** 

We appreciate your contribution whether it is a bug report, feature request, or pull request with improvement (hopefully). Please comply with the [Community ethics policy](https://docs.github.com/en/site-policy/acceptable-use-policies/github-acceptable-use-policies), and do not expect us to answer your requests immediately. Also, do not treat *GitHub issues* as a support channel.

zzzeebra:
## Summary

Fixes #15583.

Two messages for the same originator and alarm type arriving simultaneously at the create-alarm rule node produced two identical active alarms instead of one created + one updated.

## Root cause

`create_or_update_active_alarm` (`dao/src/main/resources/sql/schema-functions.sql`) checks for an existing active alarm with `SELECT ... FOR UPDATE` and inserts a new row if none is found. `FOR UPDATE` cannot lock a row that does not yet exist, and the partial index `idx_alarm_originator_alarm_type_active` on `(originator_id, type) WHERE cleared = false` is not `UNIQUE`, so two concurrent callers both saw no existing alarm and both succeeded with `INSERT`.

## Fix

Serialize concurrent callers per `(originator_id, type)` with a transaction-scoped advisory lock (`pg_advisory_xact_lock`) at the top of the function. The lock is automatically released on `COMMIT`/`ROLLBACK`, has microsecond overhead, and does not block calls for different `(originator, type)` pairs. The `UPDATE` branch was already serialized correctly via `FOR UPDATE` on the existing row.

For existing LTS installations, the same `CREATE OR REPLACE FUNCTION` is added to `application/src/main/data/upgrade/lts/schema_update.sql` so `SystemPatchApplier` delivers the fix on patch-level upgrades.

## Upgrade note

The fix is **not retroactive**: it prevents new duplicate active alarms from being created, but does not clean up duplicates that already exist in a database affected by this bug. Operators upgrading from an affected version should run a one-off cleanup of pre-existing duplicate active alarms per `(originator_id, type)` separately.

## Test plan

- [x] New DAO test `AlarmServiceTest#testShouldNotCreateDuplicateActiveAlarmsOnConcurrentCreate` fires 10 parallel `createAlarm` calls behind a `CountDownLatch` barrier and asserts exactly one active alarm. Reproduces 8 duplicates on baseline; passes with the fix.
- [x] `AlarmServiceTest`, `JpaAlarmDaoTest`, `AlarmCommentServiceTest` — regression check, all pass.
- [x] Rule engine: `TbCreateAlarmNodeTest`, `TbClearAlarmNodeTest`, `TbCheckAlarmStatusNodeTest`, `AlarmStateTest`, `AlarmRuleStateTest`, `DeviceStateTest`, `TbDeviceProfileNodeTest` — all pass.
- [x] Application: `AlarmControllerTest`, `AlarmCommentControllerTest`, `AlarmEdgeTest`, `DefaultTbAlarmServiceTest`, `DefaultTbAlarmCommentServiceTest`, `AlarmsCleanUpServiceTest`, `HousekeeperServiceTest`, `RuleChainControllerTest`, `EntityServiceTest`, `NotificationRuleApiTest` — all pass.
- [x] Manual: imported a rule chain with two `generator` nodes (`msgCount: 2`, `periodInSeconds: 1`, same `originatorId`) feeding one `create alarm` node. On `lts-4.2` baseline this consistently produced 2-4 duplicate active `General Alarm` rows for the same device and type. Applied the fix to the running PostgreSQL by reloading `schema-functions.sql` (`CREATE OR REPLACE FUNCTION` is hot, no TB restart needed), verified the new function body contains `pg_advisory_xact_lock`, deleted the pre-existing duplicates, and re-triggered the rule chain — exactly one active alarm was created.

### DIFF
--- a/dao/src/main/resources/sql/schema-functions.sql
+++ b/dao/src/main/resources/sql/schema-functions.sql
@@ -34,6 +34,10 @@ DECLARE
 BEGIN
     -- Serialize concurrent callers for the same (originator, type) so that the SELECT/INSERT
     -- pair below cannot interleave and produce duplicate active alarms.
+    -- Correctness relies on READ COMMITTED (ThingsBoard's default isolation level): after the
+    -- lock is granted, the SELECT below takes a fresh per-statement snapshot and sees the row a
+    -- prior caller just committed. Under REPEATABLE READ / SERIALIZABLE the transaction's older
+    -- snapshot would miss that row and INSERT a duplicate, so do not call this under those levels.
     PERFORM pg_advisory_xact_lock(hashtext(a_o_id::text), hashtext(a_type));
     SELECT * INTO existing FROM alarm a WHERE a.originator_id = a_o_id AND a.type = a_type AND a.cleared = false ORDER BY a.start_ts DESC FOR UPDATE;
     IF existing.id IS NULL THEN
--- a/dao/src/main/resources/sql/schema-functions.sql
+++ b/dao/src/main/resources/sql/schema-functions.sql
@@ -32,6 +32,9 @@ DECLARE
     result    alarm_info;
     row_count integer;
 BEGIN
+    -- Serialize concurrent callers for the same (originator, type) so that the SELECT/INSERT
+    -- pair below cannot interleave and produce duplicate active alarms.
+    PERFORM pg_advisory_xact_lock(hashtext(a_o_id::text), hashtext(a_type));
     SELECT * INTO existing FROM alarm a WHERE a.originator_id = a_o_id AND a.type = a_type AND a.cleared = false ORDER BY a.start_ts DESC FOR UPDATE;
     IF existing.id IS NULL THEN
         IF a_creation_enabled = FALSE THEN
--- a/dao/src/test/java/org/thingsboard/server/dao/service/AlarmServiceTest.java
+++ b/dao/src/test/java/org/thingsboard/server/dao/service/AlarmServiceTest.java
@@ -64,6 +64,7 @@ import org.thingsboard.server.dao.user.UserService;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -1025,6 +1026,7 @@ public class AlarmServiceTest extends AbstractServiceTest {
         ExecutorService executor = Executors.newFixedThreadPool(parallelism);
         CountDownLatch start = new CountDownLatch(1);
         CountDownLatch done = new CountDownLatch(parallelism);
+        List<Throwable> failures = new CopyOnWriteArrayList<>();
         try {
             for (int i = 0; i < parallelism; i++) {
                 executor.submit(() -> {
@@ -1036,7 +1038,8 @@ public class AlarmServiceTest extends AbstractServiceTest {
                                 .type(TEST_ALARM)
                                 .severity(AlarmSeverity.CRITICAL)
                                 .startTs(ts).build());
-                    } catch (Throwable ignored) {
+                    } catch (Throwable t) {
+                        failures.add(t);
                     } finally {
                         done.countDown();
                     }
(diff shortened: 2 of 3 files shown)


dao/src/main/resources/sql/schema-functions.sql
```diff
@@ -32,6 +32,9 @@ DECLARE
     result    alarm_info;
     row_count integer;
 BEGIN
+    -- Serialize concurrent callers for the same (originator, type) so that the SELECT/INSERT
+    -- pair below cannot interleave and produce duplicate active alarms.
+    PERFORM pg_advisory_xact_lock(hashtext(a_o_id::text), hashtext(a_type));
```
ViacheslavKlimov, line 41: Good use of a transaction-scoped advisory lock here. One thing worth being explicit about: this only closes the race under READ COMMITTED. The lock makes the second caller wait, but what actually lets the subsequent `SELECT ... FOR UPDATE` see the row the first caller just committed is the fresh per-statement snapshot of READ COMMITTED (the function defaults to VOLATILE). If this function were ever invoked inside a REPEATABLE READ / SERIALIZABLE transaction, the SELECT would keep the transaction's older snapshot, miss the freshly-committed alarm, and INSERT a duplicate again — the lock alone wouldn't save it. ThingsBoard runs READ COMMITTED by default so this is correct as written; might be worth a one-line note in the comment so that dependency isn't accidentally lost later.

dao/src/test/java/org/thingsboard/server/dao/service/AlarmServiceTest.java
```diff
@@ -1012,4 +1016,47 @@ public void testShouldFailToCreateAlarmWithBadType() {
...
+                executor.submit(() -> {
+                    try {
+                        start.await();
+                        alarmService.createAlarm(AlarmCreateOrUpdateActiveRequest.builder()
+                                .tenantId(tenantId)
+                                .originator(originatorId)
+                                .type(TEST_ALARM)
+                                .severity(AlarmSeverity.CRITICAL)
+                                .startTs(ts).build());
+                    } catch (Throwable ignored) {
```
ViacheslavKlimov: Swallowing every `Throwable` silently makes an unexpected failure under contention (a deadlock, a SQL error, etc.) indistinguishable from a clean run — the test would still pass as long as the final count happened to be 1. Consider collecting the exceptions (e.g. into a `CopyOnWriteArrayList`) and asserting none were thrown, so a regression that manifests as an exception rather than a duplicate row is still caught rather than masked.

ViacheslavKlimov (commented):
## Review summary

Reviewed 3 changed files in [Fixed duplicate alarms created by concurrent rule engine messages](https://github.com/thingsboard/thingsboard/pull/15641). The fix is clean and well-scoped: every active-alarm creation path (rule engine and the REST `save` path in `DefaultTbAlarmService`) funnels through `create_or_update_active_alarm`, so serializing inside that function covers all callers. The two copies of the function (`schema-functions.sql` for fresh installs, `schema_update.sql` for LTS upgrades) are byte-for-byte identical, `CREATE OR REPLACE FUNCTION` is idempotent as the upgrade file requires, and the two-`int4` advisory-lock key space is distinct from the single-`bigint` lock `SystemPatchApplier` uses, so there's no collision there. Left 2 minor comments inline.

---

dao/src/main/resources/sql/schema-functions.sql
zzzeebra: Added a note documenting the READ COMMITTED dependency in the function comment (both schema-functions.sql and the LTS schema_update.sql copy).

dao/src/test/java/org/thingsboard/server/dao/service/AlarmServiceTest.java
zzzeebra: Done. Failures are now collected into a CopyOnWriteArrayList and the test asserts it is empty, so an exception under contention fails the test instead of being masked by a final count of 1.

PR #15648, opened 2026-05-15
Added password reset link for activated users

zzzeebra:
## Summary
- Added admin REST endpoints `GET /api/user/{userId}/passwordResetLink` (text/plain) and `/passwordResetLinkInfo` (JSON) — generate a fresh password reset link for an already-activated user. Available to `SYS_ADMIN` and `TENANT_ADMIN` with tenant-scoped access. Refuses on not-yet-activated and disabled accounts.
- Added "Display password reset link" button on the User Details panel, visible only when the user is activated and the account is enabled. Mirrors the existing "Display activation link" UX with a separate `PasswordResetLinkDialogComponent`.
- Reuses the existing `/api/noauth/resetPassword` redemption flow — the user opens the link and sets a new password.

Each call regenerates `UserCredentials.resetToken`. If the user themselves had requested a reset via "Forgot password" email shortly before, that earlier link becomes invalid — last-issued token wins. Same field is shared between both flows.

## Automated tests
Added to `AuthControllerTest`:
- `testGetPasswordResetLink` — happy path: tenant admin issues a link, the user redeems it via `/api/noauth/resetPassword`, and logs in with the new password.
- `testGetPasswordResetLinkRegeneratesToken` — verifies that every call issues a fresh token and invalidates the previous one.
- `testGetPasswordResetLinkForNotYetActivatedUser` — expects `400 "User is not yet activated!"`.
- `testGetPasswordResetLinkForDisabledUser` — expects `400 "User account is disabled!"`.
- `testGetPasswordResetLinkForbiddenForCustomerUser` — verifies `CUSTOMER_USER` is rejected by `@PreAuthorize`.

## Manual tests
- Activated user — tenant admin opens user details, clicks `Display password reset link`, copies the URL, opens it in an incognito window, sets a new password, and logs in successfully with the new password.
- Newly created (not yet activated) user — only `Display activation link` is visible on the user details panel; `Display password reset link` is not shown.
- User with disabled credentials — `Display password reset link` is hidden.

Screenshot 2026-05-18 at 09 48 58
Screenshot 2026-05-18 at 09 49 12


application/src/main/java/org/thingsboard/server/controller/UserController.java
```diff
@@ -251,6 +252,37 @@ public UserActivationLink getActivationLinkInfo(@Parameter(description = USER_ID
...
+        return getPasswordResetLinkInfo(strUserId, request).value();
+    }
+
+    @ApiOperation(value = "Get password reset link info (getPasswordResetLinkInfo)",
+            notes = "Generate and return a password reset link info for the specified user. " +
+                    "Issues a fresh reset token, invalidating any previously issued reset token for this user. " +
+                    "Available only for users that are already activated and whose account is enabled. " +
+                    "The base url for the link is configurable in the general settings of system administrator. " + SYSTEM_OR_TENANT_AUTHORITY_PARAGRAPH)
+    @PreAuthorize("hasAnyAuthority('SYS_ADMIN', 'TENANT_ADMIN')")
+    @GetMapping(value = "/user/{userId}/passwordResetLinkInfo")
```
ViacheslavKlimov, line 275: Both `passwordResetLink` endpoints regenerate and persist a brand-new reset token on every call (`generatePasswordResetToken` + `saveUserCredentials` in `DefaultUserService`), unconditionally invalidating any token issued before — including one the user just requested via "Forgot password". The PR description calls this out as intended ("last-issued token wins"), so it's a deliberate choice, but two things are worth weighing:

1. This is a `GET` that always mutates state. GET is expected to be safe, so anything that re-issues the request — browser prefetch, a double-click, an HTTP retry on a flaky connection, a link-preview bot — silently kills the previously shared link. An admin who opens this dialog, copies the link, sends it, then reopens the dialog to double-check has just broken the link they sent.

2. `getActivationLink` takes the softer approach: `checkUserActivationToken` only regenerates when the token is within 15 min of expiry and otherwise returns the existing one, so repeated calls are idempotent. Reusing a still-valid reset token (or moving these to `POST`) would avoid the footgun while keeping explicit regeneration where you actually want it.

application/src/main/java/org/thingsboard/server/service/entitiy/user/DefaultUserService.java
```diff
@@ -97,4 +98,20 @@ public UserActivationLink getActivationLink(TenantId tenantId, CustomerId custom
...
+    public UserPasswordResetLink getPasswordResetLink(TenantId tenantId, CustomerId customerId, UserId userId, HttpServletRequest request) throws ThingsboardException {
+        UserCredentials userCredentials = userService.findUserCredentialsByUserId(tenantId, userId);
+        if (!userCredentials.isEnabled()) {
+            if (userCredentials.getActivateToken() != null) {
+                throw new ThingsboardException("User is not yet activated!", ThingsboardErrorCode.BAD_REQUEST_PARAMS);
+            }
+            throw new ThingsboardException("User account is disabled!", ThingsboardErrorCode.BAD_REQUEST_PARAMS);
+        }
+        userCredentials = userService.generatePasswordResetToken(userCredentials);
+        userCredentials = userService.saveUserCredentials(tenantId, userCredentials);
```
ViacheslavKlimov: Generating a reset link for an already-active account is a sensitive admin action — it effectively lets a tenant admin obtain a working link to set any of their users' passwords — yet nothing is recorded here. `save`/`delete` in this same class log via `logEntityActionService`; consider logging an entity/audit action for reset-link generation too, so there's a trail of who minted a link for whom. (`getActivationLink` doesn't log either, but its risk profile is lower since it only applies to not-yet-activated users.)

ViacheslavKlimov (commented):
## Review summary

Reviewed 39 changed files in [Added password reset link for activated users](https://github.com/thingsboard/thingsboard/pull/15648). The change is well-scoped and closely mirrors the existing activation-link feature: permission checks (`WRITE_CREDENTIALS`, slightly stronger than activation's `READ`), tenant scoping via `checkUserId`, the frontend button visibility (`isUserActivated()` / `isUserCredentialsEnabled()` backed by `checkUserInfo`), and the dialog's pipe injection all line up correctly, and the tests cover the happy path, regeneration, the not-activated/disabled cases, and the customer-user 403. Left 2 comments inline plus one note about a missing client method.

## Additional findings

- **rest-client/src/main/java/org/thingsboard/rest/client/RestClient.java:2830** — The PR adds two new REST endpoints (`GET /api/user/{userId}/passwordResetLink` and `/passwordResetLinkInfo`) but `RestClient` wasn't updated to expose them. The sibling `getActivationLink` already has a client method here (line 2830), and the PR checklist requires new REST endpoints to be mirrored in `RestClient`. Worth adding `getPasswordResetLink`/`getPasswordResetLinkInfo` equivalents.

---

application/src/main/java/org/thingsboard/server/controller/UserController.java
zzzeebra: Fixed. `getPasswordResetLink` now calls `checkUserPasswordResetToken`, which mirrors `checkUserActivationToken` — reuses the current token and regenerates only within 15 min of expiry. Repeated calls are idempotent, so prefetch/double-click/reopen no longer invalidates a shared link. Updated the API docs and reworked the regeneration test into `testGetPasswordResetLinkReusesValidToken`.

application/src/main/java/org/thingsboard/server/service/entitiy/user/DefaultUserService.java
zzzeebra: Added. Logs a `CREDENTIALS_READ` audit action with the responsible admin and target user, matching `getDeviceCredentialsByDeviceId`.

zzzeebra:
Added `getPasswordResetLink` and `getPasswordResetLinkInfo` to `RestClient`.

PR #15651, opened 2026-05-18
Fixed app hanging on transport port conflict at startup

zzzeebra:
This PR fixes two cases where ThingsBoard's JVM hangs after a transport bind failure at startup because non-daemon Californium threads survive a failed `init()`. Both transports are addressed with the same pattern: release Californium resources in a `try/catch` so the bean fails cleanly and the JVM exits with code 1.

### Symptom

If the CoAP transport fails to start (for example, because `coap.bind_port` is already in use), the JVM hangs instead of exiting. `DefaultCoapServerService.createCoapServer()` creates a `CoapServer` and adds endpoints before calling `server.start()`. When the only endpoint fails to bind, Californium throws `IllegalStateException("None of the server endpoints could be started")`, but the underlying partial state plus protocol-stage executor leave non-daemon Californium threads alive. Spring treats the bean as failed and does not invoke the `@PreDestroy` `shutdown()` method, so the process stays up after the failure.

### Fix

Wrap the endpoint/build/`server.start()` logic in `createCoapServer()` in a `try/catch`. On any `RuntimeException` or `UnknownHostException`: shut down `dtlsSessionsExecutor` if it was created, call `server.destroy()` to close any partially-bound endpoint and release Californium executors, null out instance fields, and rethrow the original exception so Spring marks the bean as failed.

### Automated tests

New `DefaultCoapServerServiceTest` with one unit test that occupies a UDP port via `DatagramSocket` and invokes `init()`:

- `whenPlainBindFails_thenInitThrowsAndReleasesCoapServer` — plain CoAP port is occupied. Verifies `IllegalStateException` is rethrown and the `server`, `dtlsSessionsExecutor`, `dtlsConnector`, `dtlsCoapEndpoint`, `tbDtlsCertificateVerifier` instance fields are all nulled.

### Manual tests

- CoAP port 5683 occupied (`nc -6 -u -l 5683`). **Before the fix**: TB hangs after `BindException` / `Application run failed`; non-daemon Californium threads keep JVM alive, requires SIGKILL. **After the fix**: process exits on its own with `Process finished with exit code 1`.

### Symptom

The LwM2M bootstrap transport (`transport.lwm2m.bootstrap.bind_port` / `secure.bind_port`, default UDP 5687/5688) has the same hang on init failure. `LwM2MTransportBootstrapService.init()` runs in `@PostConstruct`, so a failure during Spring `refresh()` does not invoke `@PreDestroy shutdown()`. The Californium `CoapServer` created inside `LeshanBootstrapServer` spawns non-daemon protocol-stage executor threads (`CoapServer(main)#N`, `CoapServer(secondary)#N`) in its constructor — before `start()`. When both UDP endpoints fail to bind, Leshan rethrows `IllegalStateException("None of the server endpoints could be started")` but the executors stay alive and keep the JVM up.

The main LwM2M transport (`DefaultLwM2mTransportService`, UDP 5685/5686) does not need a fix: it runs in `@AfterStartUp` (an `ApplicationReadyEvent` listener), so its failure propagates back through `EventPublishingRunListener.ready()` to `SpringApplication.run()`, which calls `handleRunFailure(context)` → `context.close()` → the bean's existing `@PreDestroy` runs and releases Californium resources cleanly.

### Fix

Wrap `getLhBootstrapServer()` + `bootstrapServer.start()` in `LwM2MTransportBootstrapService.init()` in a `try/catch`. On any `RuntimeException`: call `bootstrapServer.destroy()` to release the Californium `CoapServer` executors, null the `server` field, suppress any cleanup exception via `addSuppressed`, and rethrow the original. Add a `server != null` null-check to `@PreDestroy shutdown()` in case it ever runs after a failed init.

### Automated tests

New `LwM2MTransportBootstrapServiceTest` with one unit test that occupies both UDP ports via `DatagramSocket` and invokes `init()`:

- `whenEndpointsFailToStart_thenInitThrowsAndReleasesBootstrapServer` — both plain and secure ports are occupied. Verifies `IllegalStateException` with `"None of the server endpoints could be started"` is rethrown and the `server` instance field is nulled.

### Manual tests

- LwM2M bootstrap UDP ports 5687 + 5688 occupied (`nc -6 -u -l 5687` and `nc -6 -u -l 5688` in two terminals). **Before the fix**: TB prints `IllegalStateException` / `Application run failed`, JVM stays up; `jcmd <pid> Thread.print | grep -v daemon` shows 3 × `CoapServer(main)#N` + 2 × `CoapServer(secondary)#N` non-daemon threads holding `DestroyJavaVM`. **After the fix**: process exits on its own with `Process finished with exit code 1`, no leaked threads.
- Sanity: occupy only main LwM2M ports 5685 + 5686 → unchanged (TB already exits cleanly on its own, since the main service runs in `@AfterStartUp`).
- Sanity: normal start with no occupied ports → unchanged behaviour.

### DIFF
--- a/common/coap-server/src/main/java/org/thingsboard/server/coapserver/DefaultCoapServerService.java
+++ b/common/coap-server/src/main/java/org/thingsboard/server/coapserver/DefaultCoapServerService.java
@@ -106,26 +106,46 @@ public class DefaultCoapServerService implements CoapServerService, SmartInitial
     private CoapServer createCoapServer() throws UnknownHostException {
         Configuration networkConfig = createNetworkConfiguration();
         server = new CoapServer(networkConfig);
+        try {
+            CoapEndpoint.Builder noSecCoapEndpointBuilder = new CoapEndpoint.Builder();
+            InetAddress addr = InetAddress.getByName(coapServerContext.getHost());
+            InetSocketAddress sockAddr = new InetSocketAddress(addr, coapServerContext.getPort());
+            noSecCoapEndpointBuilder.setInetSocketAddress(sockAddr);
 
-        CoapEndpoint.Builder noSecCoapEndpointBuilder = new CoapEndpoint.Builder();
-        InetAddress addr = InetAddress.getByName(coapServerContext.getHost());
-        InetSocketAddress sockAddr = new InetSocketAddress(addr, coapServerContext.getPort());
-        noSecCoapEndpointBuilder.setInetSocketAddress(sockAddr);
+            noSecCoapEndpointBuilder.setConfiguration(networkConfig);
+            CoapEndpoint noSecCoapEndpoint = noSecCoapEndpointBuilder.build();
+            server.addEndpoint(noSecCoapEndpoint);
+            if (isDtlsEnabled()) {
+                createDtlsEndpoint(networkConfig);
+                dtlsSessionsExecutor = ThingsBoardExecutors.newSingleThreadScheduledExecutor(getClass().getSimpleName());
+                dtlsSessionsExecutor.scheduleAtFixedRate(this::evictTimeoutSessions, new Random().nextInt((int) getDtlsSessionReportTimeout()), getDtlsSessionReportTimeout(), TimeUnit.MILLISECONDS);
+            }
+            Resource root = server.getRoot();
+            TbCoapServerMessageDeliverer messageDeliverer = new TbCoapServerMessageDeliverer(root);
+            server.setMessageDeliverer(messageDeliverer);
 
-        noSecCoapEndpointBuilder.setConfiguration(networkConfig);
-        CoapEndpoint noSecCoapEndpoint = noSecCoapEndpointBuilder.build();
-        server.addEndpoint(noSecCoapEndpoint);
-        if (isDtlsEnabled()) {
-            createDtlsEndpoint(networkConfig);
-            dtlsSessionsExecutor = ThingsBoardExecutors.newSingleThreadScheduledExecutor(getClass().getSimpleName());
-            dtlsSessionsExecutor.scheduleAtFixedRate(this::evictTimeoutSessions, new Random().nextInt((int) getDtlsSessionReportTimeout()), getDtlsSessionReportTimeout(), TimeUnit.MILLISECONDS);
+            server.start();
+            return server;
+        } catch (RuntimeException | UnknownHostException e) {
+            log.error("Failed to start CoAP server, releasing resources", e);
+            try {
+                if (dtlsSessionsExecutor != null) {
+                    dtlsSessionsExecutor.shutdownNow();
+                }
+                if (server != null) {
+                    server.destroy();
+                }
+            } catch (Exception suppressed) {
+                e.addSuppressed(suppressed);
+            } finally {
+                server = null;
+                dtlsSessionsExecutor = null;
+                dtlsConnector = null;
+                dtlsCoapEndpoint = null;
+                tbDtlsCertificateVerifier = null;
+            }
+            throw e;
         }
-        Resource root = server.getRoot();
-        TbCoapServerMessageDeliverer messageDeliverer = new TbCoapServerMessageDeliverer(root);
-        server.setMessageDeliverer(messageDeliverer);
-
-        server.start();
-        return server;
     }
 
     private boolean isDtlsEnabled() {
--- a/common/coap-server/src/main/java/org/thingsboard/server/coapserver/DefaultCoapServerService.java
+++ b/common/coap-server/src/main/java/org/thingsboard/server/coapserver/DefaultCoapServerService.java
@@ -85,7 +85,9 @@ public class DefaultCoapServerService implements CoapServerService, SmartInitial
             dtlsSessionsExecutor.shutdownNow();
         }
         log.info("Stopping CoAP server!");
-        server.destroy();
+        if (server != null) {
+            server.destroy();
+        }
         log.info("CoAP server stopped!");
     }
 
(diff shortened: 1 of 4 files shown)


common/coap-server/src/main/java/org/thingsboard/server/coapserver/DefaultCoapServerService.java
```diff
@@ -106,26 +106,46 @@ public ConcurrentMap<TbCoapDtlsSessionKey, TbCoapDtlsSessionInfo> getDtlsSession
     private CoapServer createCoapServer() throws UnknownHostException {
         Configuration networkConfig = createNetworkConfiguration();
         server = new CoapServer(networkConfig);
```
ViacheslavKlimov: `server = new CoapServer(networkConfig)` runs outside the `try`. If the Californium `CoapServer` constructor spawns the non-daemon protocol-stage threads (as the PR description notes for the LwM2M `CoapServer`), a failure thrown by the constructor itself would leak them, since the catch's `server.destroy()` only covers failures inside the `try`. In practice the bind failure you're targeting surfaces at `server.start()` (inside the `try`), so this is a defensive nit — but moving the construction inside the `try` would close the gap. Same pattern applies to `LwM2MTransportBootstrapService.init()`.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/bootstrap/LwM2MTransportBootstrapService.java
```diff
@@ -82,13 +82,29 @@ public void afterSingletonsInstantiated() {
     @PostConstruct
     public void init() {
         log.info("Starting LwM2M transport bootstrap server...");
-        this.server = getLhBootstrapServer();
-        this.server.start();
-        log.info("Started LwM2M transport bootstrap server.");
+        LeshanBootstrapServer bootstrapServer = getLhBootstrapServer();
```
ViacheslavKlimov: `getLhBootstrapServer()` (which calls `builder.build()`, constructing the Californium `CoapServer` that — per your own description — spawns the non-daemon protocol-stage threads in its constructor) runs outside the `try`. If `build()` throws after constructing that server, the threads it already spawned would leak because `bootstrapServer.destroy()` only runs for exceptions inside the `try`. The targeted port-conflict failure happens at `start()` (inside the `try`), so the common case is covered; consider moving `getLhBootstrapServer()` inside the `try` for full robustness against the very leak this PR is fixing.

common/coap-server/src/test/java/org/thingsboard/server/coapserver/DefaultCoapServerServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Test
+    public void whenPlainBindFails_thenInitThrowsAndReleasesCoapServer() {
+        assertThatThrownBy(() -> service.init())
+                .isInstanceOf(IllegalStateException.class)
+                .hasMessageContaining("None of the server endpoints could be started");
+
+        assertThat(ReflectionTestUtils.getField(service, "server")).isNull();
+        assertThat(ReflectionTestUtils.getField(service, "dtlsSessionsExecutor")).isNull();
```
ViacheslavKlimov, line 89: With `getDtlsSettings()` stubbed to `null`, `isDtlsEnabled()` is false, so `dtlsSessionsExecutor`, `dtlsConnector`, `dtlsCoapEndpoint`, and `tbDtlsCertificateVerifier` are never assigned during `init()`. These four assertions therefore pass trivially — they aren't exercising the DTLS cleanup branch of the catch (the `dtlsSessionsExecutor.shutdownNow()` path) at all. A DTLS-enabled variant (where the executor actually gets created and then nulled) would give real coverage of that branch, which is the more interesting part of the cleanup.

ViacheslavKlimov (commented):
## Review summary

Reviewed 4 changed files in [Fixed app hanging on transport port conflict at startup](https://github.com/thingsboard/thingsboard/pull/15651). The fix is well-targeted and the reasoning is sound — confirmed that `ThingsBoardThreadFactory` produces non-daemon threads, so shutting down `dtlsSessionsExecutor` and calling `destroy()` in the catch is exactly what's needed to let the JVM exit. Left 3 comments inline, mostly minor/defensive. One additional consistency note about code outside the diff.

## Additional findings

This observation is about existing code outside the PR's diff — spotted while reading surrounding context.

- **common/coap-server/src/main/java/org/thingsboard/server/coapserver/DefaultCoapServerService.java:88** — The LwM2M `shutdown()` gained a `server == null` guard in this PR, but the CoAP `shutdown()` still calls `server.destroy()` unguarded. Since `createCoapServer()` now nulls `server` on failure, the two services are now inconsistent. Spring won't invoke `@PreDestroy` after a failed `@PostConstruct`, so this is defensive rather than a live bug — but for parity with the LwM2M change and the same "in case it ever runs after a failed init" reasoning, consider adding the same null guard here.

---

common/coap-server/src/main/java/org/thingsboard/server/coapserver/DefaultCoapServerService.java
zzzeebra: Fixed — moved `new CoapServer(networkConfig)` inside the try.

common/transport/lwm2m/src/main/java/org/thingsboard/server/transport/lwm2m/bootstrap/LwM2MTransportBootstrapService.java
zzzeebra: Fixed — `getLhBootstrapServer()` now runs inside the try, with a null-guarded `destroy()` in the catch.

common/coap-server/src/test/java/org/thingsboard/server/coapserver/DefaultCoapServerServiceTest.java
zzzeebra: Fixed — added `whenDtlsEnabledAndStartFails_thenInitShutsDownDtlsExecutorAndReleasesCoapServer`, which makes the DTLS endpoint build succeed (so `dtlsSessionsExecutor` is created) and fails `server.start()`, then verifies `shutdownNow()` runs and the fields are nulled.

zzzeebra:
Additional finding (CoAP `shutdown()` calling `server.destroy()` unguarded): Fixed — added the same `server != null` guard for parity with the LwM2M change.

PR #15658, opened 2026-05-19
Added configurable iteration limit for rule chain input nodes

zzzeebra:
## Summary

Loops that revisit the same `TbRuleChainInputNode` within a single message have been blocked since #15102. This broke legacy rule-chain designs that rely on bounded multi-iteration patterns — e.g. fetching data for 200 devices from an external API in 10 paginated batches of 20, looped across several rule chains until the last page.

This PR adds an opt-in escape hatch: `actors.rule.chain.input_loop_max_visits` (env `TB_RULE_CHAIN_INPUT_LOOP_MAX_VISITS`), default `1`.

* `1` (default) — strict cycle detection introduced in #15102: any rule chain input node revisited within one message is rejected immediately. **Existing installations see no behavior change.**
* `> 1` — the input node is allowed to fire up to N times per message. Exceeding the limit fails the message with a clear error that names the property.
* `< 1` — clamped to `1`. A one-time WARN is logged at startup so the misconfiguration is discoverable, and the effective limit (`1`) is what appears in the failure message, not the misconfigured input.

`DefaultTbContext.input()` performs a single comparison `countOccurrences(...) >= Math.max(1, cap)` for both default and raised caps — no separate strict/permissive code paths. `TbMsgProcessingCtx.isAlreadyInStack` is removed in favor of `countOccurrences >= 1`.

Direct self-references (configured target chain == own chain) remain unconditionally blocked by `TbRuleChainInputNode.onMsg()` regardless of this setting — there is no legitimate use case and it pins rule-engine at 100% CPU.

The failure message names the property and env var, so an admin upgrading from 4.2.1 who hits the regression has a discoverable opt-out.

### Unit tests
- [x] `TbMsgProcessingCtxTest` — 7/7. `countOccurrences` edge cases: empty / no-match / single match / multiple matches / popped / same-chain-different-node / same-node-different-chain.
- [x] `DefaultTbContextTest` — 62/62. Six tests cover the single visit-count path:
  - default cap=1 + node already in stack → tellFailure with property/env names
  - cap=5 + fresh message → enqueue
  - cap=2 + count=2 (unrelated stack entries do not contribute) → tellFailure with `visit limit 2 reached`
  - cap=10 + count=3 (customer's pagination case) → enqueue
  - negative cap (`-3`) + node already in stack → clamped to 1, fails with `visit limit 1 reached`
  - default cap=1 + fresh message → enqueue
- [x] `TbRuleChainInputNodeTest` — 17/17 unchanged (Layer 1 direct A→A guard untouched).

### Manual verification
Two rule chains wired as `A→B`, `B→A` via `Rule chain input` nodes; device profile default rule chain = A; single telemetry POST per scenario; observed via the rule-node debug events.

| # | Scenario | Expected | Result |
|---|---|---|---|
| 1 | YAML default `1`, indirect loop A↔B | fail on 2nd visit; message `visit limit 1 reached` and references `actors.rule.chain.input_loop_max_visits` + `TB_RULE_CHAIN_INPUT_LOOP_MAX_VISITS` | ✅ |
| 2 | YAML `5`, indirect loop A↔B | 5 successful visits through node A, fail on 6th, `visit limit 5 reached` | ✅ |
| 3 | Layer 1, direct A→A | fail from `TbRuleChainInputNode.onMsg` without Kafka round-trip; visit-cap not consulted | ✅ |
| 4 | YAML `-5` (clamp) | behaves as cap=1, fails on 2nd visit, error reports effective limit `1`, no startup warning/error | ✅ |

## Notes for reviewers

- `maxRuleNodeExecutionsPerMessage` (tenant profile) remains an independent runaway cap.
- Affects all tenants on this rule-engine instance — meant for self-hosted deployments; documented in the YAML comment.
- Visits are counted per individual input-node fire. For typical linear pagination this equals iteration count; fan-out designs that route the same message through the same input node multiple times per logical iteration consume one unit per visit.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
@@ -587,6 +587,10 @@ public class ActorSystemContext {
     @Getter
     private long ruleChainErrorPersistFrequency;
 
+    @Value("${actors.rule.chain.input_loop_max_iterations:0}")
+    @Getter
+    private int ruleChainInputLoopMaxIterations;
+
     @Value("${actors.rule.node.error_persist_frequency:3000}")
     @Getter
     private long ruleNodeErrorPersistFrequency;
--- a/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
@@ -635,6 +635,12 @@ public class ActorSystemContext {
                     "enable SSRF protection by setting SSRF_PROTECTION_ENABLED=true. If your rule chains need to access " +
                     "devices on local networks, use SSRF_ALLOWED_HOSTS to whitelist specific addresses or ranges.");
         }
+        if (ruleChainInputLoopMaxVisits < 1) {
+            log.warn("actors.rule.chain.input_loop_max_visits is configured to {}, which is below the minimum of 1. " +
+                    "Clamping to 1 (strict cycle detection - each rule chain input node fires at most once per message). " +
+                    "Set ACTORS_RULE_CHAIN_INPUT_LOOP_MAX_VISITS to a value >= 1 to allow intentional loops.",
+                    ruleChainInputLoopMaxVisits);
+        }
     }
 
     @Value("${actors.tenant.create_components_on_init:true}")
--- a/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
@@ -587,9 +587,9 @@ public class ActorSystemContext {
     @Getter
     private long ruleChainErrorPersistFrequency;
 
-    @Value("${actors.rule.chain.input_loop_max_iterations:0}")
+    @Value("${actors.rule.chain.input_loop_max_visits:1}")
     @Getter
-    private int ruleChainInputLoopMaxIterations;
+    private int ruleChainInputLoopMaxVisits;
 
     @Value("${actors.rule.node.error_persist_frequency:3000}")
     @Getter
--- a/application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
@@ -186,15 +186,33 @@ public class DefaultTbContext implements TbContext {
         }
         RuleChainId selfRuleChainId = nodeCtx.getSelf().getRuleChainId();
         RuleNodeId selfId = nodeCtx.getSelf().getId();
-        if (msg.isAlreadyInStack(selfRuleChainId, selfId)) {
-            log.warn("[{}] Detected rule chain processing loop for rule node [{}] in rule chain [{}]. " +
-                    "The message will be failed to prevent infinite loop. " +
-                    "Please check the rule chain configuration for circular references.",
-                    nodeCtx.getTenantId(), selfId, selfRuleChainId);
-            tellFailure(msg, new RuntimeException(
-                    "Detected rule chain processing loop for rule node [" + selfId + "] " +
-                    "in rule chain [" + selfRuleChainId + "]. " +
-                    "Please check the rule chain configuration for circular references."));
+        // Stack-based revisit check (Layer 2). Layer 1 — direct A->A where the input node's
+        // configured target chain equals its own chain — is enforced unconditionally in
+        // TbRuleChainInputNode.onMsg() and is not affected by this setting.
+        int maxIterations = mainCtx.getRuleChainInputLoopMaxIterations();
+        if (maxIterations <= 0) {
+            if (msg.isAlreadyInStack(selfRuleChainId, selfId)) {
+                log.warn("[{}] Detected rule chain processing loop for rule node [{}] in rule chain [{}]. " +
+                        "The message will be failed to prevent infinite loop. " +
+                        "Please check the rule chain configuration for circular references. " +
+                        "If the loop is intentional, configure actors.rule.chain.input_loop_max_iterations " +
+                        "(env TB_RULE_CHAIN_INPUT_LOOP_MAX_ITERATIONS) to a positive value.",
+                        nodeCtx.getTenantId(), selfId, selfRuleChainId);
+                tellFailure(msg, new RuntimeException(
+                        "Detected rule chain processing loop for rule node [" + selfId + "] " +
+                        "in rule chain [" + selfRuleChainId + "]. " +
+                        "Please check the rule chain configuration for circular references. " +
+                        "If the loop is intentional, configure actors.rule.chain.input_loop_max_iterations " +
+                        "(env TB_RULE_CHAIN_INPUT_LOOP_MAX_ITERATIONS) to a positive value."));
+                return;
+            }
+        } else if (msg.countOccurrences(selfRuleChainId, selfId) >= maxIterations) {
+            String reason = "Rule chain input loop iteration limit " + maxIterations + " reached for rule node ["
+                    + selfId + "] in rule chain [" + selfRuleChainId + "]. "
+                    + "Check the rule chain configuration or raise TB_RULE_CHAIN_INPUT_LOOP_MAX_ITERATIONS "
+                    + "(yaml: actors.rule.chain.input_loop_max_iterations) if the loop is intentional.";
+            log.warn("[{}] {}", nodeCtx.getTenantId(), reason);
+            tellFailure(msg, new RuntimeException(reason));
             return;
         }
         TbMsg tbMsg = msg.copy()
--- a/application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
+++ b/application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
@@ -186,31 +186,12 @@ public class DefaultTbContext implements TbContext {
         }
         RuleChainId selfRuleChainId = nodeCtx.getSelf().getRuleChainId();
         RuleNodeId selfId = nodeCtx.getSelf().getId();
-        // Stack-based revisit check (Layer 2). Layer 1 — direct A->A where the input node's
-        // configured target chain equals its own chain — is enforced unconditionally in
-        // TbRuleChainInputNode.onMsg() and is not affected by this setting.
-        int maxIterations = mainCtx.getRuleChainInputLoopMaxIterations();
-        if (maxIterations <= 0) {
-            if (msg.isAlreadyInStack(selfRuleChainId, selfId)) {
-                log.warn("[{}] Detected rule chain processing loop for rule node [{}] in rule chain [{}]. " +
-                        "The message will be failed to prevent infinite loop. " +
-                        "Please check the rule chain configuration for circular references. " +
-                        "If the loop is intentional, configure actors.rule.chain.input_loop_max_iterations " +
-                        "(env TB_RULE_CHAIN_INPUT_LOOP_MAX_ITERATIONS) to a positive value.",
-                        nodeCtx.getTenantId(), selfId, selfRuleChainId);
-                tellFailure(msg, new RuntimeException(
-                        "Detected rule chain processing loop for rule node [" + selfId + "] " +
-                        "in rule chain [" + selfRuleChainId + "]. " +
-                        "Please check the rule chain configuration for circular references. " +
-                        "If the loop is intentional, configure actors.rule.chain.input_loop_max_iterations " +
-                        "(env TB_RULE_CHAIN_INPUT_LOOP_MAX_ITERATIONS) to a positive value."));
-                return;
-            }
-        } else if (msg.countOccurrences(selfRuleChainId, selfId) >= maxIterations) {
-            String reason = "Rule chain input loop iteration limit " + maxIterations + " reached for rule node ["
+        int maxVisits = Math.max(1, mainCtx.getRuleChainInputLoopMaxVisits());
+        if (msg.countOccurrences(selfRuleChainId, selfId) >= maxVisits) {
+            String reason = "Rule chain input node visit limit " + maxVisits + " reached for rule node ["
                     + selfId + "] in rule chain [" + selfRuleChainId + "]. "
-                    + "Check the rule chain configuration or raise TB_RULE_CHAIN_INPUT_LOOP_MAX_ITERATIONS "
-                    + "(yaml: actors.rule.chain.input_loop_max_iterations) if the loop is intentional.";
+                    + "If the loop is intentional, raise actors.rule.chain.input_loop_max_visits "
+                    + "(env TB_RULE_CHAIN_INPUT_LOOP_MAX_VISITS).";
             log.warn("[{}] {}", nodeCtx.getTenantId(), reason);
             tellFailure(msg, new RuntimeException(reason));
             return;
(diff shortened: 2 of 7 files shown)


ViacheslavKlimov (commented):
## Review summary

Reviewed 7 changed files in [Added configurable iteration limit for rule chain input nodes](https://github.com/thingsboard/thingsboard/pull/15658). The core change is solid: switching from the boolean `isAlreadyInStack` guard to a `countOccurrences(...) >= Math.max(1, cap)` threshold preserves the existing strict behavior at the default of `1` (count starts at 0, so the first visit always passes and the second fails — equivalent to the old guard), and the `>=` boundary correctly allows exactly N fires before failing. The stack semantics hold up: `output()` pops nested frames so legitimate sequential reuse of an input node doesn't accumulate, while a true loop never reaches `output()` and so accumulates and trips the cap. Test coverage is thorough. Left 3 minor comments inline — naming consistency and operability nits, no functional bugs found.

---

application/src/main/resources/thingsboard.yml
```diff
@@ -549,6 +549,14 @@ actors:
...
       # Errors for particular actors are persisted once per specified amount of milliseconds
       error_persist_frequency: "${ACTORS_RULE_CHAIN_ERROR_FREQUENCY:3000}"
+      # Maximum number of times a single rule chain input node may fire for one message.
+      # Default 1 = the node fires at most once per message — any revisit (direct A->A or
+      # indirect A->B->...->A) is failed immediately to prevent infinite loops.
+      # Raise to N > 1 to allow legacy designs that intentionally loop through the same input
+      # node up to N times (e.g. paginated external API calls). Values below 1 are clamped to 1.
+      # Direct self-references (a rule chain input node configured to forward to its own rule
+      # chain) are always rejected by TbRuleChainInputNode regardless of this setting.
+      input_loop_max_visits: "${TB_RULE_CHAIN_INPUT_LOOP_MAX_VISITS:1}"
```
ViacheslavKlimov: The env var name breaks the local convention in this block. Every other property under `actors.rule.chain` (and the wider `actors.rule.*` section) uses the `ACTORS_RULE_*` prefix — `ACTORS_RULE_CHAIN_ERROR_FREQUENCY`, `ACTORS_RULE_CHAIN_DEBUG_MODE_RATE_LIMITS_PER_TENANT_ENABLED`, etc. This one introduces a `TB_`-prefixed name (`TB_RULE_CHAIN_INPUT_LOOP_MAX_VISITS`). For consistency and discoverability, consider `ACTORS_RULE_CHAIN_INPUT_LOOP_MAX_VISITS`. (Worth doing now since the failure message bakes the env var name in too — renaming later means touching both.)

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -186,15 +186,14 @@ public void input(TbMsg msg, RuleChainId ruleChainId) {
...
-        if (msg.isAlreadyInStack(selfRuleChainId, selfId)) {
-            log.warn("[{}] Detected rule chain processing loop for rule node [{}] in rule chain [{}]. " +
-                    "The message will be failed to prevent infinite loop. " +
-                    "Please check the rule chain configuration for circular references.",
-                    nodeCtx.getTenantId(), selfId, selfRuleChainId);
-            tellFailure(msg, new RuntimeException(
-                    "Detected rule chain processing loop for rule node [" + selfId + "] " +
-                    "in rule chain [" + selfRuleChainId + "]. " +
-                    "Please check the rule chain configuration for circular references."));
+        int maxVisits = Math.max(1, mainCtx.getRuleChainInputLoopMaxVisits());
```
ViacheslavKlimov, line 189: A non-positive `input_loop_max_visits` is silently clamped to 1 here on every message, with no feedback to the admin. An operator who sets `0` or `-5` (perhaps expecting "unlimited") gets strict behavior back with no indication why their intentional loop still fails on the second visit. I see this is deliberate (the test asserts no startup warning), but a one-time WARN at startup when the configured value is < 1 would make the misconfiguration discoverable without the per-message cost. Worth considering. (Side note: the PR description likens this to `DefaultActorService.poolSize`'s `Math.max(1, ...)`, but that one clamps a derived `cores/2` value, not raw user input — different situation.)

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
```diff
@@ -186,15 +186,14 @@ public void input(TbMsg msg, RuleChainId ruleChainId) {
...
+        if (msg.countOccurrences(selfRuleChainId, selfId) >= maxVisits) {
+            String reason = "Rule chain input node visit limit " + maxVisits + " reached for rule node ["
+                    + selfId + "] in rule chain [" + selfRuleChainId + "]. "
+                    + "If the loop is intentional, raise actors.rule.chain.input_loop_max_visits "
```
ViacheslavKlimov, line 194: This failure message only points the admin toward *raising* the limit ("If the loop is intentional, raise ..."). For the far more common case — an accidental circular reference — that's the wrong nudge; the previous message's guidance to "check the rule chain configuration for circular references" was the right one. Since the same message fires for both the unintentional-loop and intentional-loop cases, consider keeping a hint that an unexpected loop should be fixed in the rule chain config, not masked by bumping the cap (which, with a genuine infinite loop, just defers the failure to `maxRuleNodeExecutionsPerMessage`).

application/src/main/resources/thingsboard.yml
zzzeebra: Renamed to `ACTORS_RULE_CHAIN_INPUT_LOOP_MAX_VISITS` in the yml and the failure message. d6648890d9

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
zzzeebra: Added a one-time WARN at startup in `ActorSystemContext.init()` when the configured value is < 1. Agreed on the `poolSize` analogy — dropped it from the description.

application/src/main/java/org/thingsboard/server/actors/ruleChain/DefaultTbContext.java
zzzeebra: Message now advises checking the rule chain for circular references first, then mentions raising the cap for intentional loops.

ViacheslavKlimov (commented):
## Review summary

Reviewed 7 changed files in [Added configurable iteration limit for rule chain input nodes](https://github.com/thingsboard/thingsboard/pull/15658). Left 6 comment(s) inline.

The config wiring, the clamp behavior in `input()`, the boundary equivalence with the old `isAlreadyInStack`, and the proto round trip all check out — there is no serialization change, so rolling upgrades are unaffected, and `isAlreadyInStack` has no callers left behind. The one substantive finding is on `thingsboard.yml:552`: because `DefaultTbContext.output()` pops the frame that `input()` pushed, the counter bounds *un-popped nesting depth*, not total fires per message — a chain that returns through an output node and re-enters the same input node sees `countOccurrences == 0` every iteration, so the YAML comment (and the "customer's pagination case" test comment) describes a stronger guarantee than the guard gives. Worth a wording fix at minimum, and worth confirming it matches the intended semantics. Secondary: the startup WARN says "Clamping to 1" but `ActorSystemContext.init()` never reassigns the field — the real clamp lives in `DefaultTbContext`.

One doc-only nit outside the code: the PR description names the env var `TB_RULE_CHAIN_INPUT_LOOP_MAX_VISITS`, but the code consistently uses `ACTORS_RULE_CHAIN_INPUT_LOOP_MAX_VISITS` (`thingsboard.yml:559`, `ActorSystemContext.java:641`, `DefaultTbContext.java:195`). The code is self-consistent; only the description is stale.

---

application/src/main/resources/thingsboard.yml
```diff
@@ -549,6 +549,14 @@ actors:
     chain:
       # Errors for particular actors are persisted once per specified amount of milliseconds
       error_persist_frequency: "${ACTORS_RULE_CHAIN_ERROR_FREQUENCY:3000}"
+      # Maximum number of times a single rule chain input node may fire for one message.
```
ViacheslavKlimov, line 552: The counter is a return-stack depth, not a per-message fire count, so this description promises more than the guard delivers. `DefaultTbContext.output()` calls `msg.popFormStack()`, so the frame pushed by `input()` is removed as soon as the target chain reaches its output node. A topology where chain B ends in an output node and chain A wires that output back into the same input node (arguably the pagination shape this PR is meant to serve) has `countOccurrences == 0` on every iteration, so the node fires unbounded times even with the default of 1 — the only backstop there is the tenant profile's `maxRuleNodeExecsPerMessage`. What the property actually bounds is concurrent nesting, i.e. re-entering the node while an earlier visit is still unwound (A->B->...->A).

Could the comment say something like "maximum number of nested (not yet returned) visits of a single rule chain input node per message"? The same applies to the `givenCurrentNodeRepeatedBelowCap_whenInput_thenEnqueue` comment in `DefaultTbContextTest`, which labels three un-popped pushes as the "customer's pagination case" — those three frames are nesting depth 3, not three completed iterations.

application/src/main/java/org/thingsboard/server/actors/ActorSystemContext.java
```diff
@@ -631,6 +635,12 @@ public void init() {
                     "enable SSRF protection by setting SSRF_PROTECTION_ENABLED=true. If your rule chains need to access " +
                     "devices on local networks, use SSRF_ALLOWED_HOSTS to whitelist specific addresses or ranges.");
         }
+        if (ruleChainInputLoopMaxVisits < 1) {
```
ViacheslavKlimov, line 638: The warning says "Clamping to 1", but nothing here actually clamps — the field keeps the misconfigured value and `getRuleChainInputLoopMaxVisits()` will keep returning e.g. `-3` forever, with the real clamp done per message by `Math.max(1, ...)` in `DefaultTbContext.input()`. That splits one invariant across two files: the next person who adds a second consumer of the getter has to remember to re-apply the clamp, and nothing in the field or getter hints at it.

Could `init()` just do `ruleChainInputLoopMaxVisits = 1;` inside this branch? Then the log message becomes true, the getter always returns a valid value, `DefaultTbContext` drops the `Math.max` from the hot path, and the effective value is observable and testable in one place.

application/src/test/java/org/thingsboard/server/actors/rule/DefaultTbContextTest.java
```diff
@@ -157,24 +158,222 @@ public void givenMsgWithCurrentNodeAlreadyInStack_whenInput_thenTellFailureToPre
...
+        assertThat(failureMsg.getFailureMessage()).contains("visit limit 1 reached");
+        assertThat(failureMsg.getFailureMessage()).contains("ACTORS_RULE_CHAIN_INPUT_LOOP_MAX_VISITS");
+        assertThat(failureMsg.getFailureMessage()).contains("actors.rule.chain.input_loop_max_visits");
+
+        then(mainCtxMock).should().getRuleChainInputLoopMaxVisits();
+        then(mainCtxMock).shouldHaveNoMoreInteractions();
+    }
+
+    @Test
+    public void givenVisitsBelowCap_whenInput_thenEnqueue() {
```
ViacheslavKlimov, line 184: All coverage of the new behavior is against a mocked `ActorSystemContext`, so the tests assert what `input()` does given a cap, but nothing verifies that the cap actually reaches `input()` end to end — that `actors.rule.chain.input_loop_max_visits` binds, and that a message really does traverse an `A -> B -> A` rule-chain-input loop N times and then fails on visit N+1. `AbstractRuleEngineFlowIntegrationTest` already builds two chains linked by a rule chain input node, so a test there with `@TestPropertySource(properties = "actors.rule.chain.input_loop_max_visits=3")` looks achievable, and it would cover the property wiring plus the interaction with `pushToStack`/`popFormStack` across real hops.

Separately, the clamp branch in `ActorSystemContext.init()` (the below-1 WARN) has no test at all — the `-3` unit test only exercises the duplicate clamp in `DefaultTbContext`, not the startup one.

application/src/test/java/org/thingsboard/server/actors/rule/DefaultTbContextTest.java
```diff
@@ -157,24 +158,222 @@ public void givenMsgWithCurrentNodeAlreadyInStack_whenInput_thenTellFailureToPre
...
+
+        then(mainCtxMock).should().getRuleChainInputLoopMaxVisits();
+        then(mainCtxMock).shouldHaveNoMoreInteractions();
+    }
+
+    @Test
+    public void givenCurrentNodeRepeatedBelowCap_whenInput_thenEnqueue() {
+        // GIVEN - customer's pagination case: same RuleChainInputNode visited three times already.
+        // With cap > occurrences it must proceed.
+        given(mainCtxMock.getRuleChainInputLoopMaxVisits()).willReturn(10);
```
ViacheslavKlimov, line 269: The six new/updated input tests are ~35 lines each of near-identical setup (rule node + debug settings + node ctx stubs, callback mock, `TbMsg` builder, and for the pass cases the tpi/clusterService stubbing); they differ only in the cap value, the number of prior `pushToStack` calls, and the expected outcome. That's a lot of surface to keep in sync — a future change to `input()` means touching all six.

Would a couple of small helpers, or a single `@ParameterizedTest` over (cap, priorVisits, expectFailure) with the two assertion helpers factored out, express the same matrix more compactly? The file already uses `@ParameterizedTest`/`@MethodSource`, so the machinery is there. The boundary cases worth keeping readable (count == cap vs count == cap-1) would then stand out instead of being buried in boilerplate.

application/src/main/resources/thingsboard.yml
```diff
@@ -549,6 +549,14 @@ actors:
...
+      # Default 1 = the node fires at most once per message — any revisit (direct A->A or
+      # indirect A->B->...->A) is failed immediately to prevent infinite loops.
+      # Raise to N > 1 to allow legacy designs that intentionally loop through the same input
+      # node up to N times (e.g. paginated external API calls). Values below 1 are clamped to 1.
+      # Direct self-references (a rule chain input node configured to forward to its own rule
+      # chain) are always rejected by TbRuleChainInputNode regardless of this setting.
```
ViacheslavKlimov, line 558: Seven comment lines for one property is well beyond what the rest of this file does (neighbouring entries get one or two), and it leaks an internal Java class name at an operator who is editing YAML — `TbRuleChainInputNode` means nothing outside the codebase. Suggest trimming to what an admin needs in order to pick a value: what the number counts, that 1 means strict, that values below 1 are clamped, and that a rule chain input node pointing at its own chain is always rejected — dropping the "legacy designs" framing and the class name.

common/message/src/main/java/org/thingsboard/server/common/msg/TbMsg.java
```diff
@@ -253,8 +253,8 @@ public void pushToStack(RuleChainId ruleChainId, RuleNodeId ruleNodeId) {
         ctx.push(ruleChainId, ruleNodeId);
     }
 
-    public boolean isAlreadyInStack(RuleChainId ruleChainId, RuleNodeId ruleNodeId) {
-        return ctx.isAlreadyInStack(ruleChainId, ruleNodeId);
+    public int countOccurrences(RuleChainId ruleChainId, RuleNodeId ruleNodeId) {
```
ViacheslavKlimov, line 256: The old name `isAlreadyInStack` told the caller *what* was being inspected; `countOccurrences` on a class as widely used as `TbMsg` doesn't — occurrences of what, in what? Consider `countStackOccurrences` (or `countOccurrencesInStack`) to keep the connection to the processing stack that `pushToStack`/`popFormStack` already establish on this class.

PR #15662, opened 2026-05-19
Fixed RPC call request rule node returning null body

dashevchenko:
## Pull Request description

Initial issue: TbSendRPCRequestNode produces empty string as data instead of empty JSON in cluster mode. After investigation, the issue was found in handling FromDeviceRpcResponse.response.

- queue.proto: string response → optional string response (wire-compatible; only adds a presence bit to the API)
- 3 receivers (ProtoUtils, DefaultTbCoreConsumerService, DefaultTbRuleEngineConsumerService): use proto.hasResponse() ? proto.getResponse() : null to round-trip null correctly
- 2 receivers (the two consumer services): error check fixed from > 0 to >= 0 so RpcError.NOT_FOUND (ordinal 0) isn't silently dropped
- Regression test added: protoFromDeviceRpcResponseOnewaySerialization round-trips FromDeviceRpcResponse(id, null, null)

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
+++ b/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
@@ -465,8 +465,8 @@ public class DefaultTbCoreConsumerService extends AbstractConsumerService<ToCore
         return firmwareStateService.process(msg.getValue());
     }
 
-    private void forwardToCoreRpcService(FromDeviceRPCResponseProto proto, TbCallback callback) {
-        RpcError error = proto.getError() >= 0 ? RpcError.values()[proto.getError()] : null;
+    void forwardToCoreRpcService(FromDeviceRPCResponseProto proto, TbCallback callback) {
+        RpcError error = RpcError.fromProtoErrorCode(proto.getError());
         FromDeviceRpcResponse response = new FromDeviceRpcResponse(new UUID(proto.getRequestIdMSB(), proto.getRequestIdLSB())
                 , proto.hasResponse() ? proto.getResponse() : null, error);
         tbCoreDeviceRpcService.processRpcResponseFromRuleEngine(response);
--- a/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
+++ b/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
@@ -466,9 +466,9 @@ public class DefaultTbCoreConsumerService extends AbstractConsumerService<ToCore
     }
 
     private void forwardToCoreRpcService(FromDeviceRPCResponseProto proto, TbCallback callback) {
-        RpcError error = proto.getError() > 0 ? RpcError.values()[proto.getError()] : null;
+        RpcError error = proto.getError() >= 0 ? RpcError.values()[proto.getError()] : null;
         FromDeviceRpcResponse response = new FromDeviceRpcResponse(new UUID(proto.getRequestIdMSB(), proto.getRequestIdLSB())
-                , proto.getResponse(), error);
+                , proto.hasResponse() ? proto.getResponse() : null, error);
         tbCoreDeviceRpcService.processRpcResponseFromRuleEngine(response);
         callback.onSuccess();
     }
--- a/common/proto/src/test/java/org/thingsboard/server/common/util/ProtoUtilsTest.java
+++ b/common/proto/src/test/java/org/thingsboard/server/common/util/ProtoUtilsTest.java
@@ -227,6 +227,17 @@ class ProtoUtilsTest {
     }
 
     @Test
+    void protoFromDeviceRpcResponseOnewaySerialization() {
+        // Oneway RPC success: response and error are both null. Relies on the proto
+        // 'optional string response' presence bit so the receiver round-trips null
+        // rather than seeing the proto3 default "".
+        FromDeviceRpcResponseActorMsg msg = new FromDeviceRpcResponseActorMsg(23, tenantId, deviceId, new FromDeviceRpcResponse(id, null, null));
+        TransportProtos.ToDeviceActorNotificationMsgProto serializedMsg = ProtoUtils.toProto(msg);
+        Assertions.assertNotNull(serializedMsg);
+        assertThat(ProtoUtils.fromProto(serializedMsg)).as("deserialized").isEqualTo(msg);
+    }
+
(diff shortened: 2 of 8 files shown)


ViacheslavKlimov (commented):
## Review summary

Reviewed 5 changed files in [[Rpc call node] Fixed output data for cluster mode](https://github.com/thingsboard/thingsboard/pull/15662). Left 2 comment(s) inline.

The fix is correct and well-targeted. Two genuinely separate bugs are addressed for the cluster-mode RPC response path:
- **null response collapsing to `""`** — `string response` → `optional string response` plus `hasResponse() ? getResponse() : null` on all three receivers now round-trips a null response instead of returning the proto3 default empty string.
- **`NOT_FOUND` silently dropped** — `error > 0` → `error >= 0` in the two consumer services. Since every sender (`DefaultTbClusterService` x2, `ProtoUtils.toProto`) uses `-1` as the "no error" sentinel and `RpcError.NOT_FOUND` is ordinal `0`, the old `> 0` check mapped a real `NOT_FOUND` to null. `>= 0` is the right boundary here.

Backward compatibility checks out: the proto change is wire-compatible (proto3 `optional` is a synthetic oneof — same field number and wire type), `optional` is already used throughout `queue.proto`, and `TransportProtos` is generated at build time so nothing committed needs regenerating. The `ProtoUtils.fromProto` refactor to a local `rpcResponse` variable is a clean readability improvement. Comments below are minor.

---

common/proto/src/test/java/org/thingsboard/server/common/util/ProtoUtilsTest.java
```diff
@@ -226,6 +226,17 @@ void protoFromDeviceRpcResponseSerialization() {
         assertThat(ProtoUtils.fromProto(serializedMsg)).as("deserialized").isEqualTo(msg);
     }
 
+    @Test
+    void protoFromDeviceRpcResponseOnewaySerialization() {
```
ViacheslavKlimov, line 230: Good regression test for the null-response round-trip. Worth noting it only exercises the `ProtoUtils.toProto`/`fromProto` actor-message path. The change this PR is actually about — the `error > 0` → `error >= 0` fix (recovering `NOT_FOUND`) and the `hasResponse()` guard in `DefaultTbCoreConsumerService.forwardToCoreRpcService` and `DefaultTbRuleEngineConsumerService.handleNotification` — is the cluster-mode notification path, and it has no test at all. The `0 → NOT_FOUND` recovery in particular (the core of the bug) is only covered for the ProtoUtils path via the pre-existing `protoFromDeviceRpcResponseSerialization`, never for the consumer services. Since the logic is now duplicated across three sites, a small test that drives a `FromDeviceRPCResponseProto` with `error = 0` / unset response through the consumer-service handling would guard the paths that were actually broken.

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
```diff
@@ -466,9 +466,9 @@ private boolean handleOtaPackageUpdates(TbProtoQueueMsg<ToOtaPackageStateService
     }
 
     private void forwardToCoreRpcService(FromDeviceRPCResponseProto proto, TbCallback callback) {
-        RpcError error = proto.getError() > 0 ? RpcError.values()[proto.getError()] : null;
+        RpcError error = proto.getError() >= 0 ? RpcError.values()[proto.getError()] : null;
```
ViacheslavKlimov: `RpcError.values()[proto.getError()]` now guards the lower bound (`>= 0`) but still has no upper bound. With today's senders only emitting `-1..4` this can't trip, but in a mixed-version cluster a newer node that adds an `RpcError` constant would send an ordinal this older node can't index, throwing `ArrayIndexOutOfBoundsException` right here. Since this line is being touched anyway, an explicit bounds check (or a lookup that returns null for unknown ordinals) would make it robust. Same indexing pattern is repeated in `DefaultTbRuleEngineConsumerService.handleNotification` and `ProtoUtils.fromProto`. Low priority / pre-existing — just flagging while it's in scope.

common/proto/src/test/java/org/thingsboard/server/common/util/ProtoUtilsTest.java
dashevchenko: It was not the main problem that I fixed, but I have added test.

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbCoreConsumerService.java
dashevchenko: Fixed. Added a single null-safe helper on the enum.

dashevchenko (commented):
## Re-review summary

Re-reviewed [[Rpc call node] Fixed output data for cluster mode](https://github.com/thingsboard/thingsboard/pull/15662) — verified 2 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/15662#pullrequestreview-4426466915).

| Status | Count |
|--------|-------|
| ✅ Resolved | 2 |

### Finding details

- ✅ **ProtoUtilsTest.java** — consumer-service paths (`error >= 0` recovery + `hasResponse()` guard) had no test, only the ProtoUtils actor-message path was covered — *Fixed in code: added `DefaultTbCoreConsumerServiceTest.givenNotFoundErrorAndNoResponse_whenForwardToCoreRpcService...` and a new `DefaultTbRuleEngineConsumerServiceTest`, both driving `error = NOT_FOUND.ordinal()` (0) with an unset response through the consumer-service handling — exactly the previously broken combination.*
- ✅ **DefaultTbCoreConsumerService.java:469** — `RpcError.values()[proto.getError()]` had no upper bound; a newer node in a mixed-version cluster could send an ordinal this node can't index, throwing `ArrayIndexOutOfBoundsException` — *Fixed in code: added `RpcError.fromProtoErrorCode(int)` with an `errorCode >= 0 && errorCode < VALUES.length` bounds check (caching `values()` in a static array), now used at all three call sites (both consumer services and `ProtoUtils.fromProto`). No unguarded `RpcError.values()[...]` indexing remains in the codebase.*

The fix commits are clean: the new helper is correct, the three call sites are consistent, the visibility relaxation of `forwardToCoreRpcService` (private → package-private) is solely to enable the unit test, and senders (`DefaultTbClusterService`, `ProtoUtils.toProto`) still use the `-1` error sentinel and `ifPresent(setResponse)` so a null response round-trips correctly. No new issues found.

---

PR #15098, opened 2026-02-24
Added Kafka OAUTHBEARER (OAuth2) support

jkitn:
## Pull Request description

Add the ability for JS-executors to connect to kafka services, which are using OAUTH and Bearer Tokens as authentication method.

Adding provider to handle the token request and refresh.
Adding custom environment variables
Extend kafkaTemplate to set  kafkaConfig accordingly if sasl mechanism is OAUTHBEARER

Adressing Issue: https://github.com/thingsboard/thingsboard/issues/13170

Tested with GCP managed Kafka and rule node generator with script node

The additional config variables need to be added to the documentation
- client_id: "TB_QUEUE_KAFKA_CONFLUENT_OAUTH_CLIENT_ID"
- client_secret: "TB_QUEUE_KAFKA_CONFLUENT_OAUTH_CLIENT_SECRET"
- endpoint_url: "TB_QUEUE_KAFKA_CONFLUENT_OAUTH_ENDPOINT_URL"
- refresh_threshold: "TB_QUEUE_KAFKA_CONFLUENT_OAUTH_REFRESH_THRESHOLD"

## General checklist

- [X] You have reviewed the guidelines [document](https://docs.google.com/document/d/1wqcOafLx5hth8SAg4dqV_LV3un3m5WYR8RdTJ4MbbUM/edit?usp=sharing).
- [X] [Labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#about-labels) that classify your pull request have been added.
- [X] The [milestone](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) is specified and corresponds to fix version.  
- [X] Description references specific [issue](https://github.com/thingsboard/thingsboard/issues).
- [X] Description contains human-readable scope of changes.
- [X] Description contains brief notes about what needs to be added to the documentation.
- [X] No merge conflicts, commented blocks of code, code formatting issues.
- [X] Changes are backward compatible or upgrade script is provided.
- [X] Similar PR is opened for PE version to simplify merge. Crosslinks between PRs added. Required for internal contributors only.
  
## Front-End feature checklist

- [X] Screenshots with affected component(s) are added. The best option is to provide 2 screens: before and after changes;
- [X] If you change the widget or other API, ensure it is backward-compatible or upgrade script is present.
- [X] Ensure new API is documented [here](https://github.com/thingsboard/thingsboard-ui-help)

## Back-End feature checklist

[...]

PR #15670, opened 2026-05-20
Added Kafka OAUTHBEARER (OAuth2) support

AndriiLandiak:
## Pull Request description

This PR builds on Jonas Koch's work in #15098, which first added Kafka OAUTHBEARER support to the js-executor. I started from that branch (merged into the current master), then finished and extended it

## General checklist

## Back-End feature checklist


ViacheslavKlimov (commented):
## Review summary

Reviewed 11 changed files in [Added Kafka OAUTHBEARER (OAuth2) support](https://github.com/thingsboard/thingsboard/pull/15670). The change adds OAuth2 client-credentials auth for Confluent Kafka on both the Java backend (`TbKafkaSettings`, wiring Kafka's native `OAuthBearerLoginCallbackHandler`) and the js-executor (new `oAuthBearerProvider` built on `simple-oauth2`). The Java side and the JAAS-escaping/missing-config tests look solid. Most comments are about robustness of the js token-refresh loop and a shared gap around OAuth scope/audience.

A couple of things I checked that turned out fine: `simple-oauth2@5.1.0` is CommonJS (not ESM-only), so the `module: commonjs` build of js-executor imports it without trouble; and the JAAS-escaping test passes because Spring's `@TestPropertySource` runs each inlined property through `Properties.load`, which collapses the doubled backslashes before injection.

## Additional findings

Minor observations, not blocking:

- **msa/js-executor/config/custom-environment-variables.yml:59-64** and **application/src/main/resources/thingsboard.yml:18-22** — the `client-id` / `client-secret` / `endpoint-url` settings are labelled "Optional", but they are required whenever `sasl.mechanism` is `OAUTHBEARER` (both sides fail fast if any is blank). The parent comment in `thingsboard.yml` clarifies this; the per-field "Optional" labels are slightly misleading on their own.
- **msa/js-executor/queue/kafkaTemplate.ts:128-130** — when `OAUTHBEARER` is selected but an env var is unset, `config.get('kafka.confluent.oauth.client_id')` throws node-config's generic "Configuration property ... is not defined" before the friendly validation inside `oauthBearerProvider` can run. This matches the existing PLAIN path (username/password), so it's consistent, but guarding with `config.has(...)` would let the clearer error message surface.

---

msa/js-executor/queue/oAuthBearerProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  tokenPromise = refreshToken();
+  // The first fetch is retried internally; swallow here, so a startup failure is not an unhandled rejection.
+  tokenPromise.catch(() => { /* retry already scheduled in refreshToken() */ });
+
+  return async function () {
+    return {
+      value: await tokenPromise
```
ViacheslavKlimov: When a scheduled refresh runs, `tokenPromise` is replaced with the new in-flight promise (line 72), and this getter awaits it. If that refresh fails, every call here rejects until a retry succeeds — even though the previously fetched token is typically still valid (the refresh fires `refreshThresholdMs`, default 60s, before expiry). KafkaJS calls this provider on each (re)authentication, so a transient token-endpoint hiccup landing in that window can fail an otherwise-healthy reauth (and a consumer CRASH on a non-retryable error sends SIGTERM here). Consider keeping the last successfully fetched token and only swapping it in after a successful refresh, so a failed background refresh doesn't drop a still-valid token.

msa/js-executor/queue/oAuthBearerProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  async function refreshToken(): Promise<string> {
+    logger.info('Requesting Kafka OAuth bearer token');
+    try {
+      // Client-credentials grant issues no refresh_token, so always request a fresh token.
+      const accessToken: AccessToken = await client.getToken({});
+
+      const expiresIn = accessToken.token.expires_in;
+      if (typeof expiresIn !== 'number' || expiresIn <= 0) {
```
ViacheslavKlimov: This rejects the response unless `expires_in` is strictly a JS number, but some OIDC providers return it as a numeric string (`"3600"`). `simple-oauth2` already tolerates that — it parses the response into `accessToken.token.expires_at` (a `Date`) and exposes `accessToken.expired(windowSeconds)`. Reusing those instead of the raw `expires_in` would handle string/number and absolute-time formats and avoid reimplementing the expiry math here.

msa/js-executor/queue/oAuthBearerProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+        throw new Error(`OAuth token response has invalid "expires_in": ${expiresIn}`);
+      }
+      const accessTokenValue = accessToken.token.access_token;
+      if (typeof accessTokenValue !== 'string' || accessTokenValue.length === 0) {
+        throw new Error('OAuth token response has no "access_token"');
+      }
+
+      const nextRefresh = expiresIn * 1000 - refreshThresholdMs;
```
ViacheslavKlimov: If the IdP issues short-lived tokens whose lifetime is below `refreshThresholdMs` (default 60s), `nextRefresh` goes negative and `scheduleRefresh` clamps it to `MIN_REFRESH_DELAY_MS` (1s) — so the executor re-requests a token every second indefinitely, hammering the token endpoint. Worth guarding against `refreshThresholdMs >= expiresIn*1000` (e.g. cap the threshold to a fraction of the token lifetime, or at least log a warning).

msa/js-executor/queue/oAuthBearerProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+      });
+    }, delay);
+    refreshTimer.unref();
+  }
+
+  async function refreshToken(): Promise<string> {
+    logger.info('Requesting Kafka OAuth bearer token');
+    try {
+      // Client-credentials grant issues no refresh_token, so always request a fresh token.
+      const accessToken: AccessToken = await client.getToken({});
```
ViacheslavKlimov: `getToken({})` requests a token with no `scope` (and no audience/resource). The comment above lists Azure AD as a target, but Azure AD's client-credentials flow *requires* a `scope` (e.g. `api://<id>/.default`), and several other IdPs need a scope/audience to mint a token for the right resource — so as written those providers can't actually be used. There's no way to configure one here, nor on the Java path (`TbKafkaSettings#applyOauthBearerProps`). Consider adding an optional `scope` setting passed to `getToken({ scope })`.

common/queue/src/main/java/org/thingsboard/server/queue/kafka/TbKafkaSettings.java
```diff
@@ -230,6 +245,28 @@ Properties toProps() {
...
+            log.warn("Kafka OAuth token endpoint URL is not HTTPS ({}); client credentials will be sent unencrypted",
+                    oauthEndpointUrl);
+        }
+        props.put(SaslConfigs.SASL_JAAS_CONFIG,
+                "org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required"
+                        + " clientId=\"" + escapeJaasValue(oauthClientId) + "\""
+                        + " clientSecret=\"" + escapeJaasValue(oauthClientSecret) + "\";");
+        props.put(SaslConfigs.SASL_LOGIN_CALLBACK_HANDLER_CLASS,
+                "org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginCallbackHandler");
+        props.put(SaslConfigs.SASL_OAUTHBEARER_TOKEN_ENDPOINT_URL, oauthEndpointUrl);
```
ViacheslavKlimov, line 272: Same gap as the js-executor path: there's no way to pass an OAuth `scope`/audience to the client-credentials request. Kafka's `OAuthBearerLoginCallbackHandler` accepts a `scope` option in the JAAS module config; without it, IdPs that require a scope for client-credentials (e.g. Azure AD's `.default`) can't be used. Consider exposing an optional `queue.kafka.confluent.oauth.scope` and appending `scope="..."` to the JAAS string built above.

AndriiLandiak (commented):
## Review summary

Reviewed 12 changed files in [Added Kafka OAUTHBEARER (OAuth2) support](https://github.com/thingsboard/thingsboard/pull/15670). The change is well-structured: the tb-core path wires Kafka's native `OAuthBearerLoginCallbackHandler` (with correct JAAS escaping and an optional `scope` per KIP-768), and the js-executor path implements a self-refreshing `simple-oauth2` client-credentials provider. JAAS-injection escaping, HTTPS warnings, fail-fast validation, and the short-lived-token clamp are all handled thoughtfully, and the test coverage for the Java side is solid. Left 2 comments inline — both are robustness/consistency nits rather than blocking bugs.

---

msa/js-executor/queue/oAuthBearerProvider.ts
```diff
@@ -0,0 +1,171 @@
...
+/// limitations under the License.
+///
+
+import { AccessToken, ClientCredentials } from 'simple-oauth2';
+import { _logger } from '../config/logger';
+
+interface OauthBearerProviderOptions {
+  clientId: string;
+  clientSecret: string;
+  host: string;
```
AndriiLandiak: `host` is a misleading name here — this field actually holds the full token endpoint URL (the call site in `kafkaTemplate.ts` passes `host: optionalOauthStr('kafka.confluent.oauth.endpoint_url')`, and below it's fed into `new URL(...)` and split into `tokenHost` + `tokenPath`). Renaming it to `endpointUrl` would make both the interface and the call site read more clearly.

AndriiLandiak (commented):
## Review summary

Reviewed 12 changed files in [Added Kafka OAUTHBEARER (OAuth2) support](https://github.com/thingsboard/thingsboard/pull/15670). This adds OAuth2 client-credentials auth for Confluent Kafka on two paths: the tb-core JVM path (`TbKafkaSettings`, wiring Kafka's native `OAuthBearerLoginCallbackHandler`) and the js-executor path (a hand-rolled `simple-oauth2` provider with proactive, coalesced token refresh).

The implementation is solid — JAAS values are escaped against injection, config is validated fail-fast, non-HTTPS endpoints are warned about, the refresh logic coalesces concurrent fetches and guards short-lived tokens, and the PLAIN path is preserved with regression tests. Findings are minor; left 1 comment inline.

## Additional findings

- **common/queue/src/main/java/org/thingsboard/server/queue/kafka/TbKafkaSettings.java:246** — `toProps()` calls `configureSSL(props)` after the confluent block, and `configureSSL` unconditionally overwrites `security.protocol` to `"SSL"` when `queue.kafka.ssl.enabled=true`. If someone enables both Confluent OAUTHBEARER (`use_confluent_cloud=true`, `security.protocol=SASL_SSL`) and `kafka.ssl.enabled=true`, the SASL_SSL protocol gets clobbered to plain SSL and OAuth auth silently breaks. This is pre-existing (the PLAIN confluent path has the same interaction) and requires an unusual dual-config, but the new OAuth path now flows through the same method — worth a guard or at least a documented note that the two SSL flags are mutually exclusive.

---

PR #15682, opened 2026-05-22
Add related items logic

pinkevmladchy:
## Pull Request description

- Adds dependency resolution to IoT Hub item install: new GET /versions/{id}/installPlan and POST /installPlan endpoints walk the item's relatedItems, mark each as Install / Update / Missing, and cascade-install them in
  topological order with rollback on failure.
- Updates the install dialog to display the resolved plan (entries + status badges + counts) and confirm before installing. For items with no dependencies the dialog skips the plan view and installs directly — no UI flash.
- Adds an InstallPlan / InstallPlanEntry / InstallPlanResult DTO trio + the matching frontend models, locale strings, and styling.

## General checklist

## Back-End feature checklist


ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.html
```diff
@@ -101,31 +154,53 @@ <h2 class="tb-iot-hub-install-title">{{ 'iot-hub.install-error-title' | translat
...
+        @if (resolvingPlan) {
+          <mat-spinner diameter="18" class="mr-2 inline-block align-middle"></mat-spinner>
+        }
         {{ 'iot-hub.rule-chain-overwrite-replace' | translate }}
       </button>
     }
+    @case ('plan') {
+      <button mat-button (click)="cancelPlan()">{{ 'action.cancel' | translate }}</button>
+      <button mat-flat-button color="primary"
+              [disabled]="planSummary.willInstall === 0"
```
pinkevmladchy: When the root item and all of its dependencies are already installed, `install()` still shows the plan (because `hasDependencies` is true once `summary.alreadyInstalled > 0`), but here the confirm button is disabled since `planSummary.willInstall === 0`. That leaves the user in a dead-end dialog where the only option is Cancel. Is re-installing a fully-installed tree reachable (e.g. from the items list)? If so, it might be friendlier to either skip straight to a "nothing to do" / success state, or keep the button enabled as a no-op confirm.

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -654,4 +670,167 @@ public void deleteInstalledItem(SecurityUser user, IotHubInstalledItemId install
...
+                    missingItemIds.add(entry.getItemId());
+                    resultEntries.add(resultEntry);
+                }
+                case ALREADY_INSTALLED -> resultEntries.add(resultEntry);
+                case WILL_INSTALL -> {
+                    try {
+                        // Only the root entry receives the user's install data (target profile entityId);
+                        // transitive deps install with defaults.
+                        JsonNode entryData = entry.isRoot() ? data : null;
+                        IotHubInstalledItem installed = doInstallVersion(user, entry.getVersionId(), entryData, request);
```
pinkevmladchy, line 885: `installPlan` installs exactly the `WILL_INSTALL` entries the client sends back, trusting the per-entry `versionId`/`status`/`root` that `resolveInstallPlan` produced earlier. Because nothing is re-validated server-side, a stale or replayed plan can re-install an item that has since been installed, creating a duplicate `IotHubInstalledItem` row (the already-installed check only runs at resolve time). Would it be safer to re-resolve the plan from the root `versionId` on the server, or at least re-check `alreadyInstalledItemIds` here before each install?

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -654,4 +670,167 @@ public void deleteInstalledItem(SecurityUser user, IotHubInstalledItemId install
...
+                        }
+                        resultEntries.add(resultEntry);
+                    } catch (Exception e) {
+                        log.error("[{}] Cascade install failed at entry {} ({}): {}", tenantId,
+                                entry.getName(), entry.getVersionId(), e.getMessage(), e);
+                        resultEntry.setErrorMessage(e.getMessage());
+                        resultEntries.add(resultEntry);
+                        rollbackInstalledItems(user, rollbackIds);
+                        result.setSuccess(false);
+                        result.setRolledBack(true);
```
pinkevmladchy: `rollbackInstalledItems` catches and logs each failed delete but doesn't report it back, yet `result.setRolledBack(true)` is set unconditionally. So if one of the rollback deletes fails, the client is told the cascade was fully rolled back while some entities are actually left behind. Consider tracking whether every rollback succeeded and reflecting a partial-rollback state in the result.

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -654,4 +670,167 @@ public void deleteInstalledItem(SecurityUser user, IotHubInstalledItemId install
...
+                              LinkedHashMap<String, InstallPlanEntry> entries,
+                              boolean root) {
+        String itemId = versionInfo.get("itemId").asText();
+        if (entries.containsKey(itemId)) {
+            return;
+        }
+
+        InstallPlanEntry entry = new InstallPlanEntry();
+        entry.setItemId(itemId);
+        entry.setVersionId(versionInfo.get("id").asText());
```
pinkevmladchy: `name`/`type`/`version` are read defensively with `hasNonNull(...)`, but `itemId` (`versionInfo.get("itemId").asText()`) and this `id` field are dereferenced directly. If the marketplace payload ever omits either, this NPEs and surfaces as a 500 from the install-plan endpoint. Worth guarding these the same way, or failing with a clear message if the version JSON is missing its identifiers.

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -654,4 +670,167 @@ public void deleteInstalledItem(SecurityUser user, IotHubInstalledItemId install
...
+        if (related != null && related.isArray()) {
+            for (JsonNode relatedNode : related) {
+                String relatedItemId = relatedNode.asText();
+                if (relatedItemId == null || relatedItemId.isEmpty()
+                        || relatedItemId.equals(rootItemId) || entries.containsKey(relatedItemId)) {
+                    continue;
+                }
+                JsonNode relatedVersion;
+                try {
+                    relatedVersion = iotHubRestClient.getPublishedVersionByItemId(relatedItemId);
```
pinkevmladchy, line 761: This makes one synchronous marketplace call per related item inside the loop, on top of the initial `getVersionInfo` — N+1 sequential round-trips before the user sees the plan, with the dialog blocking on it. Fine for a handful of deps, but if `relatedItems` can grow it's worth a bounded parallel fetch or a batch endpoint, and perhaps a sanity cap on how many related items we'll resolve.

application/src/main/java/org/thingsboard/server/controller/IotHubController.java
```diff
@@ -71,6 +73,23 @@ public InstallItemVersionResult installItemVersion(@PathVariable String versionI
...
+
+    @PreAuthorize("hasAuthority('TENANT_ADMIN')")
+    @PostMapping("/installPlan")
+    @ResponseBody
+    public InstallPlanResult installPlan(@RequestBody InstallPlanRequest body,
+                                         HttpServletRequest request) throws ThingsboardException {
+        return iotHubService.installPlan(getCurrentUser(), body.plan(), body.data(), request);
+    }
+
+    public record InstallPlanRequest(InstallPlan plan, JsonNode data) {}
```
pinkevmladchy, line 91: The POST endpoint takes the whole `InstallPlan` back as the request body, including per-entry status the server itself assigned moments earlier. That makes the install endpoint depend on the client faithfully echoing server state, which is easy to get wrong and lets the two endpoints drift. It would be cleaner for the install call to accept just the root `versionId` and re-resolve the plan server-side. (Flagging the API shape; the correctness consequence of trusting it is noted on the service.)

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -654,4 +670,167 @@ public void deleteInstalledItem(SecurityUser user, IotHubInstalledItemId install
         iotHubInstalledItemService.deleteById(tenantId, installedItemId);
         log.info("[{}] Deleted installed IoT Hub item: {}", tenantId, installedItem.getItemName());
     }
+
+    @Override
+    public InstallPlan resolveInstallPlan(SecurityUser user, String versionId) {
```
pinkevmladchy, line 718: I don't see any tests in this PR for `resolveInstallPlan` or `installPlan`, despite a lot of branching worth locking down: `relatedItems` null/empty/array, a related id equal to the root, duplicate ids, fetch-failure and null-version → MISSING, already-installed detection, and the cascade install with reverse-order rollback on failure plus the partial-success / all-missing outcomes. An integration test over a couple of these paths would catch a lot. Was leaving them untested intentional?

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -103,48 +106,61 @@ public class DefaultIotHubService implements IotHubService {
...
-            installedItem.setDescriptor(descriptor);
-            iotHubInstalledItemService.save(tenantId, installedItem);
+    /**
+     * Fetch + apply a single marketplace version, persist the installed-item record, and ping
+     * the marketplace install counter. Throws on failure so callers (cascade install) can roll back.
+     */
+    private IotHubInstalledItem doInstallVersion(SecurityUser user, String versionId, JsonNode data, HttpServletRequest request) throws Exception {
+        TenantId tenantId = user.getTenantId();
+        JsonNode versionInfo = iotHubRestClient.getVersionInfo(versionId);
+        String itemType = versionInfo.get("type").asText();
```
pinkevmladchy: The marketplace version field names (`"type"`, `"name"`, `"itemId"`, `"version"`, `"id"`, `"relatedItems"`) are now parsed as bare string literals in both `doInstallVersion` and `addPlanEntry`. Since both read the same payload shape, pulling these into shared constants (or a small `JsonNode` → typed-record mapper) would keep the contract in one place so the two parsers can't drift.

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -654,4 +670,167 @@ public void deleteInstalledItem(SecurityUser user, IotHubInstalledItemId install
...
+
+        return new InstallPlan(versionId, new ArrayList<>(entries.values()));
+    }
+
+    /**
+     * Build an {@link InstallPlanEntry} for a single marketplace version and append it to the
+     * plan. No traversal of {@code relatedItems} happens here — related items are resolved one
+     * level deep by {@link #resolveInstallPlan}.
+     */
+    private void addPlanEntry(JsonNode versionInfo,
```
pinkevmladchy, line 786: The trailing `boolean root` flag isn't self-documenting at the call sites (`addPlanEntry(..., false)` / `addPlanEntry(..., true)`). A two-state boolean like this reads better as a named enum or two thin wrappers (`addRootEntry` / `addRelatedEntry`) — it's the kind of flag that's easy to flip by accident.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
```diff
@@ -232,44 +247,130 @@ export class TbIotHubInstallDialogComponent extends DialogComponent<TbIotHubInst
...
   install(): void {
+    // Keep the current state visible (confirm / select-entity / confirm-overwrite) while resolving —
+    // a transient 'resolving-plan' state caused a visible "blink" for items without dependencies.
+    this.resolvingPlan = true;
+    const versionId = this.item.id as string;
+    this.iotHubApiService.resolveInstallPlan(versionId, { ignoreLoading: true }).subscribe({
+      next: (plan) => {
+        this.resolvingPlan = false;
+        const summary = this.summarizePlan(plan);
+        const hasDependencies = plan.entries.length > 1
```
pinkevmladchy: `hasDependencies` actually means "is there anything worth showing the user" — it's true for already-installed or missing items, not just dependencies. The name and the `entries.length > 1` check (which leans on the root always being present) make the intent harder to follow. Consider renaming to `shouldShowPlan`, or having the server hint whether a plan UI is warranted so the two sides agree on what counts as a trivial install.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.scss
```diff
@@ -57,3 +57,92 @@ tb-entity-select {
...
+
+.tb-iot-hub-plan-entry-status {
+  font-size: 12px;
+  font-weight: 500;
+  padding: 2px 8px;
+  border-radius: 10px;
+  white-space: nowrap;
+
+  &.tb-status-will-install {
+    color: #1565c0;
```
pinkevmladchy, line 135: The status/badge styles use hardcoded hex colors (`#1565c0`, `#2e7d32`, `#b26a00`, `#fff8e1`, …) rather than the app's Material theme palette/variables. That tends to break dark mode and theme consistency. Could these reuse the existing theme tokens / mat palette variables instead of literal hexes?

ui-ngx/src/assets/locale/locale.constant-en_US.json
```diff
@@ -3843,6 +3843,16 @@
         "goto-main-dashboard": "Go to main dashboard",
         "install-error-title": "Installation Failed",
         "install-error-message": "Failed to install '{{name}}'.",
+        "install-plan-title": "Install '{{name}}' and its dependencies",
+        "install-plan-will-install": "{{count}} item(s) will be installed.",
+        "install-plan-will-update": "{{count}} item(s) will be updated.",
+        "install-plan-missing-warning": "{{count}} required item(s) could not be found on the marketplace. '{{ name }}' may not work fully without them.",
+        "install-plan-status-will": "Install",
+        "install-plan-status-installed": "Update",
```
pinkevmladchy: The `ALREADY_INSTALLED` rows are labeled "Update" (and `install-plan-will-update` reads "… will be updated"), but those entries are skipped — `installPlan` does nothing for `ALREADY_INSTALLED`, it doesn't upgrade them. That likely misleads the user into expecting an upgrade. Would "Already installed" / "Skipped" match the actual behavior better?

pinkevmladchy (commented):
## Review summary

Reviewed 14 changed files in [Add related items logic](https://github.com/thingsboard/thingsboard/pull/15682). The dependency-resolution flow is well structured — the `doInstallVersion` extraction is clean, rollback genuinely deletes the installed entities (not just the tracking rows), and the no-deps fast-path avoids the UI flash as described. Left 12 comments inline spanning correctness (a dead-end plan dialog, server trust of the client-supplied plan, a rollback-status flag, and an NPE-guard inconsistency) and quality/design (N+1 marketplace calls, missing tests, a misleading "Update" label, and a few readability/theming points).

---

PR #15694, opened 2026-05-26
Added ability for creating alarm rules across multiple entities

mtsymbarov-del:
## Pull Request description

Before:
Знімок екрана 2026-05-26 о 17 02 15

After:
Знімок екрана 2026-05-26 о 17 01 45

## General checklist

## Back-End feature checklist


ChantsovaEkaterina (commented):
## Review summary

Reviewed 7 changed files in [Added ability for creating alarm rules across multiple entities](https://github.com/thingsboard/thingsboard/pull/15694). Left 5 comments inline.

The bulk-create flow and the entity filter extension are the right direction, but there are a few correctness issues to address.

---
*This is an auto-generated review. Findings may contain errors — please verify before applying changes.*

ui-ngx/src/app/shared/models/calculated-field.models.ts
```diff
@@ -527,7 +528,29 @@ export const ArgumentEntityTypeParamsMap =new Map<ArgumentEntityType, ArgumentEn
   [ArgumentEntityType.Customer, { title: 'calculated-fields.customer-name', entityType: EntityType.CUSTOMER }],
 ])
 
-export const getCalculatedFieldCurrentEntityFilter = (entityName: string, entityId: EntityId) => {
+export const getCalculatedFieldCurrentEntityFilter = (entityName: string, entityId: EntityId | EntityId[]): EntityFilter => {
+  if (Array.isArray(entityId)) {
+    const entityType = entityId[0]?.entityType as EntityType;
+    const ids = entityId.map(value => value.id);
```
ChantsovaEkaterina: The function has two structural problems worth fixing together:

**1. `ids` is a misleading name for profile types, and profile handling is split across two switch statements.**
For DEVICE_PROFILE/ASSET_PROFILE, the values extracted via `.id` are profile **names** (not UUIDs), because `argsEntityId` in the dialog stores profile names there. Naming them `ids` obscures this. More importantly, the profile logic is duplicated — one `case` for the array path and another for the single-entity path — when both can be unified.

**2. The non-profile array path always produces `entityList`, even for a single entity.**
The single-entity path returns `singleEntity` for a lone device/asset/customer; the array path should match when there's only one element.

Suggested rewrite — check entity type first, unify both single and array under the same profile branch:

```ts
export const getCalculatedFieldCurrentEntityFilter = (entityName: string, entityId: EntityId | EntityId[]): EntityFilter => {
  const entityType = (Array.isArray(entityId) ? entityId[0]?.entityType : entityId?.entityType) as EntityType;

  if (entityType === EntityType.DEVICE_PROFILE || entityType === EntityType.ASSET_PROFILE) {
    // Profile filters use names, not entity IDs.
    // For array input, .id holds the profile name (set by argsEntityId in alarm-rule-dialog).
    const profileNames = Array.isArray(entityId) ? entityId.map(e => e.id) : [entityName];
    return entityType === EntityType.ASSET_PROFILE
      ? { type: AliasFilterType.assetType, assetTypes: profileNames }
      : { type: AliasFilterType.deviceType, deviceTypes: profileNames };
  }

  if (Array.isArray(entityId)) {
    return entityId.length === 1
      ? { type: AliasFilterType.singleEntity, singleEntity: entityId[0] }
      : { type: AliasFilterType.entityList, entityType, entityList: entityId.map(e => e.id) };
  }
  return { type: AliasFilterType.singleEntity, singleEntity: entityId };
};
```

ui-ngx/src/app/modules/home/components/alarm-rules/alarm-rule-dialog.component.html
```diff
@@ -56,23 +56,57 @@ <h2>{{ 'alarm-rule.alarm-rule' | translate}}</h2>
...
+                  }
+                </mat-select>
+              </mat-form-field>
+              @if (isProfileType) {
+                <tb-entity-subtype-list formControlName="entityId"
+                                        class="flex-1"
+                                        appearance="outline"
+                                        required
+                                        subscriptSizing="dynamic"
+                                        [entityType]="profileSubtypeEntityType"
```
ChantsovaEkaterina: `tb-entity-subtype-list` sets its `subtypeListEmptyText` (the required-field validation hint) once in `ngOnInit()` based on the initial `entityType` input. When the user switches between DEVICE_PROFILE and ASSET_PROFILE, `isProfileType` stays `true`, so the component is not recreated — `entityType` changes but the hint stays stale. This causes the mismatch seen in the screenshot: selecting "Asset profile" still shows "At least one **device** profile should be selected."

The simplest fix is to replace the single `@if (isProfileType)` guard with two separate blocks so Angular destroys and recreates the component when the profile type changes:

```html
@if (entityTypeControl.value === EntityType.DEVICE_PROFILE) {
  <tb-entity-subtype-list ... [entityType]="EntityType.DEVICE"></tb-entity-subtype-list>
} @else if (entityTypeControl.value === EntityType.ASSET_PROFILE) {
  <tb-entity-subtype-list ... [entityType]="EntityType.ASSET"></tb-entity-subtype-list>
}
```

ui-ngx/src/app/modules/home/components/alarm-rules/alarm-rule-dialog.component.ts
```diff
@@ -174,21 +199,48 @@ export class AlarmRuleDialogComponent extends DialogComponent<AlarmRuleDialogCom
...
+            return this.alarmRulesService.saveAlarmRule(alarmRule);
+          }))),
+        takeUntilDestroyed(this.destroyRef)
+      ).subscribe({
+        next: (calculatedFields: CalculatedFieldAlarmRule[]) => this.dialogRef.close(calculatedFields[0]),
+        error: () => this.isLoading = false
+      });
     } else {
       this.fieldFormGroup.get('name').markAsTouched();
       this.entitySelect?.entityAutocompleteMarkAsTouched();
```
ChantsovaEkaterina: When `createNew = true`, `entitySelect` is `null` (the `#entitySelect` ref lives inside the `@else` branch), so `entitySelect?.entityAutocompleteMarkAsTouched()` silently does nothing. The entity selector — `tb-entity-subtype-list` or `tb-entity-list` — won't show its required validation error when the user submits without selecting entities.

Adding `this.fieldFormGroup.get('entityId')?.markAsTouched()` here would surface the error.

ui-ngx/src/app/modules/home/components/calculated-fields/components/calculated-field-arguments/calculated-field-argument-panel.component.ts
```diff
@@ -341,4 +341,8 @@ export class CalculatedFieldArgumentPanelComponent implements OnInit, AfterViewI
       this.entityNameSubject.next(null);
     }
   }
+
+  get entityIdType() {
+    return Array.isArray(this.entityId) ? this.entityId[0].entityType : this.entityId.entityType;
```
ChantsovaEkaterina: If `entityId` is an empty array — which can happen momentarily after the entity type dropdown resets the selection — `this.entityId[0]` is `undefined` and accessing `.entityType` throws a TypeError. `getCalculatedFieldCurrentEntityFilter` already guards against this with `entityId[0]?.entityType`; the same optional chaining should be applied here:

```ts
return Array.isArray(this.entityId) ? this.entityId[0]?.entityType : this.entityId?.entityType;
```

ui-ngx/src/app/shared/components/entity/entity-list-select.component.html
```diff
@@ -35,6 +36,7 @@
   @if (modelValue.entityType) {
     <tb-entity-list
       class="flex-1"
+      [labelText]="displayEntityLabel ? (entityTypeTranslations.get(modelValue.entityType).typePlural | translate) : ''"
```
ChantsovaEkaterina: `displayEntityLabel: boolean` makes the component auto-compute the label via `entityTypeTranslations.get(modelValue.entityType).typePlural`, which throws if the entity type isn't in the translations map. The existing `entityTypeLabel: string` input on this component already handles the entity-type selector label via a direct string — the same pattern would be cleaner here.

Consider replacing `displayEntityLabel: boolean` with a string input, e.g. `entityListLabel: string`. Callers pass the label text directly, no map lookup is needed, and the potential NPE is eliminated.

ui-ngx/src/app/modules/home/components/alarm-rules/alarm-rule-dialog.component.ts
```diff
@@ -174,21 +190,52 @@ export class AlarmRuleDialogComponent extends DialogComponent<AlarmRuleDialogCom
...
+            : of([] as CalculatedFieldAlarmRule[])),
+        takeUntilDestroyed(this.destroyRef)
+      ).subscribe({
+        next: (calculatedFields: CalculatedFieldAlarmRule[]) =>
+          calculatedFields.length ? this.dialogRef.close(calculatedFields[0]) : this.isLoading = false,
+        error: () => this.isLoading = false
+      });
     } else {
       this.fieldFormGroup.get('name').markAsTouched();
+      this.fieldFormGroup.get('entityId')?.markAsTouched();
```
ChantsovaEkaterina, line 223: **Still unresolved:** `fieldFormGroup.get('entityId')?.markAsTouched()` marks the Angular reactive-form control touched in the model, but both `tb-entity-subtype-list` and `tb-entity-list` have empty `registerOnTouched` implementations — they completely ignore the external touch signal.

Each component drives its own error visibility through an internal `dirty` flag that is only ever set by user interaction inside the component (`writeValue`, add/remove item, etc.). Calling `markAsTouched()` on the parent form control never sets this flag, so the required-field error (red border and error text) still does not appear when the user clicks Add without selecting any entities.

The correct fix is to implement `registerOnTouched` properly in both components:

```ts
private onTouched = () => {};

registerOnTouched(fn: any): void {
  this.onTouched = fn;
}
```

Then, in the same place where `propagateChange` is called (user has made a change), also call `this.onTouched()`. Additionally, inside the `onTouched` callback itself, set `dirty = true` so the next focus/render cycle shows the error — matching the pattern the `dirty` flag already uses.

ui-ngx/src/app/shared/models/calculated-field.models.ts
```diff
@@ -527,24 +528,22 @@ export const ArgumentEntityTypeParamsMap =new Map<ArgumentEntityType, ArgumentEn
...
+      ? { type: AliasFilterType.assetType, assetTypes: profileNames }
+      : { type: AliasFilterType.deviceType, deviceTypes: profileNames };
+  }
+
+  if (Array.isArray(entityId)) {
+    return entityId.length === 1
+      ? { type: AliasFilterType.singleEntity, singleEntity: entityId[0] }
+      : { type: AliasFilterType.entityList, entityType, entityList: entityId.map(value => value.id) };
   }
+  return { type: AliasFilterType.singleEntity, singleEntity: entityId };
```
ChantsovaEkaterina, line 546: The last two branches handle non-profile cases but express the same intent across two separate paths. Both a bare `EntityId` and a single-element array produce `singleEntity`, so they can be merged into one check while keeping the common single-entity path simple:

```ts
if (Array.isArray(entityId) && entityId.length > 1) {
  return { type: AliasFilterType.entityList, entityType, entityList: entityId.map(e => e.id) };
}
return { type: AliasFilterType.singleEntity, singleEntity: Array.isArray(entityId) ? entityId[0] : entityId };
```

This removes the separate `Array.isArray` guard and the trailing bare-entity branch, making the function a clean two-branch structure: profiles at the top, all other entities at the bottom, with the common single-entity case as the default path.

ChantsovaEkaterina (commented):
## Re-review summary

Re-reviewed [Added ability for creating alarm rules across multiple entities](https://github.com/thingsboard/thingsboard/pull/15694) — verified 5 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/15694#pullrequestreview-4487657704).

| Status | Count |
|--------|-------|
| ✅ Resolved | 4 |
| ❌ Unresolved | 1 |

Also found 1 new observation in the updated code, commented inline.

### Finding details

- ✅ **calculated-field.models.ts** — `getCalculatedFieldCurrentEntityFilter` restructured: profile branch unified, `entityList` vs `singleEntity` made consistent — *Fixed in code.*
- ✅ **alarm-rule-dialog.component.html** — Split single `@if (isProfileType)` into two separate blocks (one per profile type) so Angular destroys and recreates `tb-entity-subtype-list` on type change — *Fixed in code.*
- ❌ **alarm-rule-dialog.component.ts:210** — Calling `markAsTouched()` on the outer reactive-form control does not trigger the required-field error inside `tb-entity-subtype-list` / `tb-entity-list` — *Still present, see inline comment.*
- ✅ **calculated-field-argument-panel.component.ts** — `entityIdType` getter now uses optional chaining (`entityId[0]?.entityType`) to avoid TypeError on empty array — *Fixed in code.*
- ✅ **entity-list-select.component.html/ts** — `displayEntityLabel: boolean` replaced with `entityListLabel: string` input — *Fixed in code.*

---
*This is an auto-generated re-review. Findings may contain errors — please verify before applying changes.*

PR #15718, opened 2026-06-02
Fixed device/asset profile import when default rule chain no longer exists

mtsymbarov-del:
## Pull Request description

Знімок екрана 2026-06-02 о 13 51 42

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/shared/import-export/import-export.service.ts
+++ b/ui-ngx/src/app/shared/import-export/import-export.service.ts
@@ -89,6 +89,7 @@ import {
 import { FormProperty, propertyValid } from '@shared/models/dynamic-form.models';
 import { CalculatedFieldsService } from '@core/http/calculated-fields.service';
 import { CalculatedField } from '@shared/models/calculated-field.models';
+import { RuleChainId } from '@shared/models/id/rule-chain-id';
 
 export type editMissingAliasesFunction = (widgets: Array<Widget>, isSingleWidget: boolean,
                                           customTitle: string, missingEntityAliases: EntityAliases) => Observable<EntityAliases>;
@@ -749,7 +750,9 @@ export class ImportExportService {
               type: 'error'}));
           throw new Error('Invalid device profile file');
         } else {
-            return this.deviceProfileService.saveDeviceProfile(this.prepareImport(deviceProfile));
+          return this.checkIfRuleChainExist(deviceProfile).pipe(
+            mergeMap(profile => this.deviceProfileService.saveDeviceProfile(this.prepareImport(profile)))
+          );
         }
       }),
       catchError(() => of(null))
@@ -776,10 +779,30 @@ export class ImportExportService {
               type: 'error'}));
           throw new Error('Invalid asset profile file');
         } else {
-          return this.assetProfileService.saveAssetProfile(this.prepareImport(assetProfile));
+          return this.checkIfRuleChainExist(assetProfile).pipe(
+            mergeMap(profile => this.assetProfileService.saveAssetProfile(this.prepareImport(profile)))
+          )
         }
       }),
       catchError(() => of(null))
+    );
+  }
+
+  private checkIfRuleChainExist<T extends {defaultRuleChainId?: RuleChainId}>(profile: T) {
+    if (!profile.defaultRuleChainId?.id) {
+      return of(profile);
+    }
+    return this.ruleChainService.getRuleChain(profile.defaultRuleChainId.id,
+      {ignoreErrors: true, ignoreLoading: true}).pipe(
+      map(() => profile),
+      catchError(() => {
+        this.store.dispatch(new ActionNotificationShow(
+          {message: this.translate.instant('rulechain.rulechain-not-found-error'),
+            type: 'warn'}));
+        const cleanedProfile = deepClone(profile);
+        cleanedProfile.defaultRuleChainId = null;
+        return of(cleanedProfile);
+      })
     );
   }
 
--- a/ui-ngx/src/app/shared/import-export/import-export.service.ts
+++ b/ui-ngx/src/app/shared/import-export/import-export.service.ts
@@ -33,7 +33,8 @@ import {
 } from '@shared/models/alias.models';
 import { MatDialog } from '@angular/material/dialog';
 import { ImportDialogComponent, ImportDialogData } from '@shared/import-export/import-dialog.component';
-import { forkJoin, Observable, of, Subject } from 'rxjs';
+import { forkJoin, Observable, of, Subject, throwError } from 'rxjs';
+import { HttpErrorResponse } from '@angular/common/http';
 import { catchError, map, mergeMap, switchMap, take, tap } from 'rxjs/operators';
 import { DashboardUtilsService } from '@core/services/dashboard-utils.service';
 import { EntityService } from '@core/http/entity.service';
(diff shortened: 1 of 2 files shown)


vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Fixed device/asset profile import when default rule chain no longer exists](https://github.com/thingsboard/thingsboard/pull/15718). Left 6 comment(s) inline.

The core fix is sound — checking each referenced rule chain for existence and nulling out the missing ones before save is the right approach. The main suggestion is to replace the per-id `getRuleChain` loop with a single `getRuleChainsByIds([...])` batch call: it's tenant-scoped and returns only the chains that exist, so you derive the missing fields by set difference in one request, and it removes the per-id 404/error-handling that currently introduces a silent-abort failure mode. The rest are quality/DRY observations.

---

ui-ngx/src/app/shared/import-export/import-export.service.ts
```diff
@@ -776,13 +780,48 @@ export class ImportExportService {
...
+    const ruleChainFields: Array<'defaultRuleChainId' | 'defaultEdgeRuleChainId'> = ['defaultRuleChainId', 'defaultEdgeRuleChainId'];
+    const checks = ruleChainFields
+      .filter(field => profile[field]?.id)
+      .map(field => this.ruleChainExists(profile[field].id).pipe(
+        map(exists => ({field, exists}))
+      ));
+    if (!checks.length) {
+      return of(profile);
+    }
+    return forkJoin(checks).pipe(
```
vvlladd28: Consider replacing this `forkJoin` of per-id `getRuleChain` calls with a single `ruleChainService.getRuleChainsByIds([...ids], {ignoreErrors: true, ignoreLoading: true})`. The backend (`findRuleChainsByTenantIdAndIds`) returns only the rule chains that exist for the current tenant, so you can collect the requested ids, fire one request, and mark a field as missing when its id isn't present in the returned set. That turns N requests into one and removes the per-id 404 handling in `ruleChainExists` entirely (which also eliminates the silent-abort behavior noted below). For reference, there's no existence-only/HEAD endpoint for rule chains, so this batch call is the lightest correct option available.

ui-ngx/src/app/shared/import-export/import-export.service.ts
```diff
@@ -776,13 +780,48 @@ export class ImportExportService {
...
+        missingFields.forEach(field => cleanedProfile[field] = null);
+        return cleanedProfile;
+      })
+    );
+  }
+
+  private ruleChainExists(ruleChainId: string): Observable<boolean> {
+    return this.ruleChainService.getRuleChain(ruleChainId, {ignoreErrors: true, ignoreLoading: true}).pipe(
+      map(() => true),
+      catchError((err: HttpErrorResponse) => err?.status === 404 ? of(false) : throwError(() => err))
```
vvlladd28: If the per-id approach is kept: any non-404 error here (network blip, 500, 403, a malformed-id 400) is re-thrown, propagates through `forkJoin` and the `mergeMap`, and is then swallowed by the outer `catchError(() => of(null))` in the importers — so the whole import silently fails with no notification. Note this network call didn't exist on the import path before this PR, so it's a new way for an otherwise-valid import to be blocked. A couple of options: the error body does carry `err.error.errorCode`/`err.error.message` if you want to distinguish causes, but the built-in `{resendRequest: true}` retry only kicks in on HTTP 429 — it won't retry transient 5xx/network, so you'd need an explicit `retry(...)` for that. The simplest fix is probably to treat an unknown/transient error as "assume it exists" and let the backend be the final authority, rather than aborting. Switching to `getRuleChainsByIds` (comment above) sidesteps this entirely.

ui-ngx/src/app/shared/import-export/import-export.service.ts
```diff
@@ -749,7 +751,9 @@ export class ImportExportService {
               type: 'error'}));
           throw new Error('Invalid device profile file');
         } else {
-            return this.deviceProfileService.saveDeviceProfile(this.prepareImport(deviceProfile));
+          return this.clearMissingProfileRuleChains(deviceProfile).pipe(
```
vvlladd28: `importDeviceProfile` and `importAssetProfile` now carry the identical `clearMissingProfileRuleChains(...).pipe(mergeMap(profile => save(prepareImport(profile))))` wrapper, differing only by which `...ProfileService.save` is called. Small as it is, this is the kind of duplication that rots — a future change to the clearing flow has to be made in two places. Consider passing the save callback into `clearMissingProfileRuleChains` (or a shared helper) so the orchestration lives once.

ui-ngx/src/app/shared/import-export/import-export.service.ts
```diff
@@ -776,13 +780,48 @@ export class ImportExportService {
...
+    return forkJoin(checks).pipe(
+      map(results => {
+        const missingFields = results.filter(result => !result.exists).map(result => result.field);
+        if (!missingFields.length) {
+          return profile;
+        }
+        this.store.dispatch(new ActionNotificationShow(
+          {message: this.translate.instant('rulechain.rulechain-not-found-warning'),
+            type: 'warn'}));
+        const cleanedProfile = deepClone(profile);
```
vvlladd28: `prepareImport` (called immediately after this on the returned value) already does `deepClone(data)` internally, so on the missing-fields path the profile gets cloned twice. Minor, but you could drop this `deepClone` and null the fields on a shallow copy, or rely on `prepareImport`'s clone — whichever reads cleaner.

ui-ngx/src/app/shared/import-export/import-export.service.ts
```diff
@@ -776,13 +780,48 @@ export class ImportExportService {
...
+            mergeMap(profile => this.assetProfileService.saveAssetProfile(this.prepareImport(profile)))
+          );
         }
       }),
       catchError(() => of(null))
     );
   }
 
+  private clearMissingProfileRuleChains<T extends { defaultRuleChainId?: RuleChainId; defaultEdgeRuleChainId?: RuleChainId }>(profile: T): Observable<T> {
+    const ruleChainFields: Array<'defaultRuleChainId' | 'defaultEdgeRuleChainId'> = ['defaultRuleChainId', 'defaultEdgeRuleChainId'];
```
vvlladd28: The two field names are spelled out three times — in the generic constraint, the array's type annotation, and the array literal — and all three must be kept in sync if a third rule-chain field is ever added. Deriving the array from a single typed `const` tuple (and feeding the generic from it) would remove the triple-maintenance, though it's minor with only two fields today.

ui-ngx/src/assets/locale/locale.constant-en_US.json
```diff
@@ -5137,6 +5137,7 @@
         "create-new-rulechain": "Create new rule chain",
         "rulechain-file": "Rule chain file",
         "invalid-rulechain-file-error": "Unable to import rule chain: Invalid rule chain data structure.",
+        "rulechain-not-found-warning": "Default rule chain not found. The profile was saved without one — you can assign it in the profile settings.",
```
vvlladd28, line 5140: This is the only em-dash (—) in the entire en_US locale file; every other string uses plain ASCII punctuation. For consistency (and to avoid encoding surprises in downstream translations) consider a period or hyphen, e.g. "...saved without one. You can assign it in the profile settings."

vvlladd28 (approved):

PR #15729, opened 2026-06-04
Skip notification push to unknown service id to prevent orphan notification topics

smatvienko-tb:
## Problem

Per-service notification topics are named `<notifications-topic>.<serviceId>` (`TopicService.buildNotificationTopicName`) and are **auto-created by the Kafka producer on first send** (`TbKafkaProducerTemplate.createTopicIfNotExist`).

`DefaultTbClusterService.pushNotificationToCore(...)` and `pushNotificationToRuleEngine(...)` forwarded the target `serviceId` as-is. That id is not always a local node id — it can be carried in message metadata (`originServiceId` / `serviceId`) and read back by RPC-reply and integration-downlink flows (`TbSendRPCRequestNode`, `TbSendRestApiCallReplyNode`, `DefaultTbContext.pushToIntegration`).

If that metadata holds a value that does not correspond to any cluster node, the producer still creates a topic for it. Such topics are never consumed, and there is no cleanup for `tb_core.notifications.*` / `tb_rule_engine.notifications.*` (only `KafkaEdgeTopicsCleanUpService` cleans edge topics), so the topic count can grow without bound.

## Fix

Validate the target service id against current cluster membership (`PartitionService.getAllServiceIds`) before sending. If the id is not a known `TB_CORE` / `TB_RULE_ENGINE` service, skip the push and complete the callback instead of creating an orphan topic.

The check **fails open** when the membership snapshot is empty/unknown (e.g. right after a node starts, before the first partition assignment), so legitimate notifications are never dropped. Each `pushNotification*` method validates against its own service type.

Pure validation — no API or schema change, backward compatible.

## Tests

`DefaultTbClusterServiceTest`:
- unknown core service id → not sent, callback acked
- known core service id → sent
- empty membership (topology unknown) → fails open, sent
- unknown rule-engine service id → not sent, callback acked

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
+++ b/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
@@ -253,7 +253,7 @@ public class DefaultTbClusterService implements TbClusterService {
     }
 
     private void skipNotificationToUnknownService(ServiceType serviceType, String serviceId, TbQueueCallback callback) {
-        log.warn("Skipping notification push to unknown {} service id [{}]", serviceType, serviceId);
+        log.debug("Skipping notification push to unknown {} service id [{}]", serviceType, serviceId);
         if (callback != null) {
             // The message was not delivered to any node, so signal failure (not success) to the caller.
             callback.onFailure(new RuntimeException("Target " + serviceType + " service id [" + serviceId + "] is not a member of the cluster"));
@@ -377,6 +377,10 @@ public class DefaultTbClusterService implements TbClusterService {
             if (callback != null) {
                 callback.onSuccess(null); //callback that message already sent, no useful payload expected
             }
+            return;
+        }
+        if (isUnknownService(ServiceType.TB_TRANSPORT, serviceId)) {
+            skipNotificationToUnknownService(ServiceType.TB_TRANSPORT, serviceId, callback);
             return;
         }
         TopicPartitionInfo tpi = topicService.getNotificationsTopic(ServiceType.TB_TRANSPORT, serviceId);
--- a/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
+++ b/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
@@ -255,8 +255,9 @@ public class DefaultTbClusterService implements TbClusterService {
     private void skipNotificationToUnknownService(ServiceType serviceType, String serviceId, TbQueueCallback callback) {
         log.debug("Skipping notification push to unknown {} service id [{}]", serviceType, serviceId);
         if (callback != null) {
-            // The message was not delivered to any node, so signal failure (not success) to the caller.
-            callback.onFailure(new RuntimeException("Target " + serviceType + " service id [" + serviceId + "] is not a member of the cluster"));
+            // Deliberate skip - the target is not a cluster member. Complete the callback as a no-op success,
+            // matching pushNotificationToTransport, so callers don't treat an intentional skip as a delivery error.
+            callback.onSuccess(null);
         }
     }
 
--- a/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
+++ b/application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
@@ -15,11 +15,14 @@
  */
 package org.thingsboard.server.service.queue;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.context.event.EventListener;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Service;
 import org.thingsboard.common.util.JacksonUtil;
@@ -96,6 +99,7 @@ import org.thingsboard.server.queue.common.TbProtoQueueMsg;
 import org.thingsboard.server.queue.common.TbRuleEngineProducerService;
 import org.thingsboard.server.queue.discovery.PartitionService;
 import org.thingsboard.server.queue.discovery.TopicService;
+import org.thingsboard.server.queue.discovery.event.ServiceListChangedEvent;
 import org.thingsboard.server.queue.provider.TbQueueProducerProvider;
 import org.thingsboard.server.service.gateway_device.GatewayNotificationsService;
 import org.thingsboard.server.service.ota.OtaPackageStateService;
@@ -129,6 +133,13 @@ public class DefaultTbClusterService implements TbClusterService {
     private final AtomicInteger toTransportNfs = new AtomicInteger(0);
     private final AtomicInteger toEdgeMsgs = new AtomicInteger(0);
     private final AtomicInteger toEdgeNfs = new AtomicInteger(0);
+
+    private static final int RECENTLY_SEEN_SERVICES_MAX_SIZE = 1024;
+    // serviceType:serviceId -> seen. Bounded LRU of cluster members ever observed, used to grace a service id that is
+    // briefly missing from the live snapshot (see isUnknownService). Fed from ServiceListChangedEvent.
+    private final Cache<String, Boolean> recentlySeenServices = Caffeine.newBuilder()
+            .maximumSize(RECENTLY_SEEN_SERVICES_MAX_SIZE)
+            .build();
 
     @Autowired
     private PartitionService partitionService;
@@ -241,15 +252,28 @@ public class DefaultTbClusterService implements TbClusterService {
      * first send. A {@code serviceId} that is not a member of the cluster - e.g. a stale node id or a value injected
      * via message metadata - would therefore create an orphan topic that nobody ever consumes and that is never
      * cleaned up. To prevent that, notifications whose target service id is not a current member of the cluster are
-     * skipped. The check fails open while the cluster topology is not yet known (empty/absent membership, e.g. right
-     * after startup) so legitimate notifications are never dropped.
+     * skipped.
+     * <p>
+     * This is a safety net for never-real ids, not a precise gatekeeper, so it is deliberately lenient. A service id is
+     * treated as known if it is either a current member ({@link PartitionService#getAllServiceIds}) or a recently seen
+     * one ({@link #recentlySeenServices}). The recently-seen grace covers nodes that are briefly absent from the live
+     * snapshot - a discovery-propagation lag, a rolling restart or a transient scale-down - so legitimate replies to
+     * them are not dropped; genuine non-delivery to a node that is actually gone is already handled by the caller-side
+     * RPC/REST timeout. The check also fails open while the cluster topology is not yet known (empty/absent membership,
+     * e.g. right after startup).
      */
     private boolean isUnknownService(ServiceType serviceType, String serviceId) {
         if (serviceId == null || serviceId.isEmpty()) {
             return true;
         }
         Set<String> serviceIds = partitionService.getAllServiceIds(serviceType);
-        return serviceIds != null && !serviceIds.isEmpty() && !serviceIds.contains(serviceId);
+        if (serviceIds == null || serviceIds.isEmpty()) {
+            return false; // topology not known yet - fail open
+        }
+        if (serviceIds.contains(serviceId)) {
+            return false; // currently a live member
+        }
+        return recentlySeenServices.getIfPresent(recentlySeenServiceKey(serviceType, serviceId)) == null;
     }
 
     private void skipNotificationToUnknownService(ServiceType serviceType, String serviceId, TbQueueCallback callback) {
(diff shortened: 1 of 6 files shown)


smatvienko-tb:
### Performance validation

The added guard (`isUnknownService` → `PartitionService.getAllServiceIds`) is a purely in-memory check — no ZK/DB/network, and a lock-free read of the cached cluster-membership list (`currentOtherServices`, reassigned only on topology change). Per call it allocates ~2 short-lived `HashSet`s and iterates the other-nodes list (cluster node count is small), with a tiny `getServiceTypesList().contains(...)` per node.

- **Known service id (normal path):** the check is orders of magnitude cheaper than the protobuf build + `producer.send()` it precedes — well under 1% overhead.
- **Unknown service id:** net *faster* — it skips the send and, more importantly, the first-send `createTopicIfNotExists` AdminClient round-trip to the broker.

These methods sit on request/response paths (RPC responses, REST-API-call replies, integration downlink acks), not the telemetry ingestion hot path, so the per-call cost is immaterial in practice. No new locks or contention are introduced.

A zero-allocation variant (a cached `Set<String>` per service type in `HashPartitionService` plus an `isServiceAvailable(...)` predicate) is possible but intentionally omitted as YAGNI given the above.

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -222,12 +226,40 @@ public void pushNotificationToCore(String serviceId, FromDeviceRpcResponse respo
...
+     * cleaned up. To prevent that, notifications whose target service id is not a current member of the cluster are
+     * skipped. The check fails open while the cluster topology is not yet known (empty/absent membership, e.g. right
+     * after startup) so legitimate notifications are never dropped.
+     */
+    private boolean isUnknownService(ServiceType serviceType, String serviceId) {
+        if (serviceId == null || serviceId.isEmpty()) {
+            return true;
+        }
+        Set<String> serviceIds = partitionService.getAllServiceIds(serviceType);
+        return serviceIds != null && !serviceIds.isEmpty() && !serviceIds.contains(serviceId);
```
smatvienko-tb: The fail-open here is binary — it only triggers when the membership snapshot is completely empty. During a brief discovery-propagation window (a core/rule-engine node has just joined and registered, but this node's `currentOtherServices` snapshot hasn't refreshed yet), the set is non-empty but won't yet contain the new node's id, so a legitimate notification to it would be skipped rather than delivered.

For the RPC/REST-reply flows the target is the node that originated the request, so in practice it's almost always already in the snapshot and this window is narrow — but it's worth consciously accepting the trade-off: a (rare) dropped reply during scale-up vs. the orphan-topic cost. If that matters, treating "recently seen" ids as known, or a short grace period before dropping, would close the gap.

smatvienko-tb (commented):
## Review summary

Reviewed 2 changed files in [Skip notification push to unknown service id to prevent orphan notification topics](https://github.com/thingsboard/thingsboard/pull/15729). The change is well-targeted, the fail-open guard is sensible, and the tests cover the main paths (skip / send / fail-open). Left 3 comments inline — they're about edge cases and consistency rather than correctness.

## Additional findings

- **common/queue/src/main/java/org/thingsboard/server/queue/discovery/HashPartitionService.java:112** — `currentOtherServices` is a plain (non-`volatile`) field, written under `synchronized recalculatePartitions(...)` but read without synchronization in `getOtherServices(...)` (line ~564). This PR adds a per-notification reader of it (`getAllServiceIds(...)` now runs on every RPC/REST-reply push), so the unsynchronized read happens much more often than before. There's no happens-before guarantee for the reference or its contents across threads. Since the write replaces the whole reference rather than mutating in place, marking the field `volatile` would be enough to make the new reader see a consistent snapshot.
- **application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java:374** — `pushNotificationToTransport(...)` builds `tb_transport.notifications.<serviceId>` through the same auto-creating producer and is **not** covered by the new membership check (it only guards null/empty serviceId). If transport service ids can also arrive from stale metadata, transport notification topics have the same orphan-topic growth exposure this PR is fixing for core/rule-engine. Was transport intentionally left out of scope?

---

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -222,12 +226,40 @@ public void pushNotificationToCore(String serviceId, FromDeviceRpcResponse respo
...
+    }
+
+    private void skipNotificationToUnknownService(ServiceType serviceType, String serviceId, TbQueueCallback callback) {
+        log.warn("Skipping notification push to unknown {} service id [{}]", serviceType, serviceId);
```
smatvienko-tb: This logs at WARN on every skip. The PR's own premise is that unknown service ids legitimately show up in message metadata during normal operation (that's exactly what was creating the orphan topics), so depending on how often that occurs this could turn into a recurring WARN in the logs. Consider DEBUG, or throttling, if it turns out to fire regularly in practice.

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -222,12 +226,40 @@ public void pushNotificationToCore(String serviceId, FromDeviceRpcResponse respo
...
+        if (callback != null) {
+            // The message was not delivered to any node, so signal failure (not success) to the caller.
+            callback.onFailure(new RuntimeException("Target " + serviceType + " service id [" + serviceId + "] is not a member of the cluster"));
```
smatvienko-tb: A contract inconsistency worth noting: the other skip path in this class, `pushNotificationToTransport` (line ~378), completes its callback with `onSuccess(null)` when it drops a message for a missing serviceId, whereas here a dropped message reports `onFailure`. Both cases are "we deliberately did not send," so the opposite signals are surprising. All current callers pass a `null` callback, so there's no impact today — but a future caller that branches on success/failure would observe different outcomes for the same kind of deliberate skip. Worth settling on one convention.

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -130,6 +134,13 @@ public class DefaultTbClusterService implements TbClusterService {
     private final AtomicInteger toEdgeMsgs = new AtomicInteger(0);
     private final AtomicInteger toEdgeNfs = new AtomicInteger(0);
 
+    private static final int RECENTLY_SEEN_SERVICES_MAX_SIZE = 1024;
+    // serviceType:serviceId -> seen. Bounded LRU of cluster members ever observed, used to grace a service id that is
+    // briefly missing from the live snapshot (see isUnknownService). Fed from ServiceListChangedEvent.
+    private final Cache<String, Boolean> recentlySeenServices = Caffeine.newBuilder()
```
smatvienko-tb: There's already a small utility for this exact shape — `common/util`'s `SetCache<K>` (Caffeine cache + dummy value, `add`/`contains`/`remove`), currently used by `TaskProcessor` for the same "have I seen this key recently" purpose. This re-implements the same pattern by hand with a raw `Cache<String, Boolean>`. `SetCache` today only supports TTL eviction (`expireAfterWrite`) rather than the `maximumSize`/LRU bound used here — but extending it with a `maximumSize` constructor overload would avoid a second, slightly different copy of the same "bounded seen-set" concept living in application code.

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -222,12 +237,92 @@ public void pushNotificationToCore(String serviceId, FromDeviceRpcResponse respo
...
+     * snapshot - a discovery-propagation lag, a rolling restart or a transient scale-down - so legitimate replies to
+     * them are not dropped; genuine non-delivery to a node that is actually gone is already handled by the caller-side
+     * RPC/REST timeout. The check also fails open while the cluster topology is not yet known (empty/absent membership,
+     * e.g. right after startup).
+     */
+    private boolean isUnknownService(ServiceType serviceType, String serviceId) {
+        if (serviceId == null || serviceId.isEmpty()) {
+            return true;
+        }
+        Set<String> serviceIds = partitionService.getAllServiceIds(serviceType);
```
smatvienko-tb, line 270: `getAllServiceIds(serviceType)` → `getAllServices` → `getOtherServices` rebuilds a `HashSet<ServiceInfo>` from the full `currentOtherServices` list and then streams/collects that into a second `HashSet<String>` — two allocations plus a full scan, on every call. This used to run only on relatively infrequent broadcast paths; now it runs once per RPC reply and once per REST-callback reply. Probably fine for typical cluster sizes, but would a cheaper `boolean isKnownServiceId(serviceType, serviceId)` on `PartitionService` (backed by the existing per-type map without collecting into new sets) be worth adding instead of rebuilding two collections just to test membership of one id?

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -222,12 +237,92 @@ public void pushNotificationToCore(String serviceId, FromDeviceRpcResponse respo
...
+            return false; // topology not known yet - fail open
+        }
+        if (serviceIds.contains(serviceId)) {
+            return false; // currently a live member
+        }
+        return recentlySeenServices.getIfPresent(recentlySeenServiceKey(serviceType, serviceId)) == null;
+    }
+
+    private void skipNotificationToUnknownService(ServiceType serviceType, String serviceId, TbQueueCallback callback) {
+        log.debug("Skipping notification push to unknown {} service id [{}]", serviceType, serviceId);
```
smatvienko-tb, line 278: This class already tracks `toCoreNfs`/`toRuleEngineNfs`/`toTransportNfs` counters that get logged periodically via stats when `cluster.stats.enabled` is on, but notifications skipped here aren't counted anywhere and only surface at DEBUG. Since the whole point of this change is to catch a leak that was previously invisible in production, would a "skipped" counter (or a rate-limited WARN) alongside the existing stats help operators see how often this guard is actually triggering?

application/src/main/java/org/thingsboard/server/service/queue/DefaultTbClusterService.java
```diff
@@ -343,6 +442,10 @@ public void pushNotificationToTransport(String serviceId, ToTransportMsg respons
             }
             return;
         }
+        if (isUnknownService(ServiceType.TB_TRANSPORT, serviceId)) {
```
smatvienko-tb, line 437: Now that `isUnknownService` treats a null/empty `serviceId` as unknown (returns `true` a few lines up in that method), doesn't this overlap with the pre-existing null/empty guard a few lines above in this method (`if (serviceId == null || serviceId.isEmpty()) { ... callback.onSuccess(null); return; }`)? Both paths end up doing the same thing — skip and complete the callback as a no-op success. Could that older block be dropped now that `isUnknownService` covers it, so there's a single place deciding whether a `serviceId` is usable?

application/src/test/java/org/thingsboard/server/service/queue/DefaultTbClusterServiceTest.java
```diff
@@ -288,6 +291,111 @@ public void testPushNotificationToCoreWithRestApiCallResponseMsgProto() {
...
+        doReturn(tpi).when(topicService).getNotificationsTopic(any(ServiceType.class), any(String.class));
+        when(producerProvider.getTbCoreNotificationsMsgProducer()).thenReturn(tbCoreQueueProducer);
+
+        clusterService.pushNotificationToCore("rpc-1778979165457-749", new FromDeviceRpcResponse(UUID.randomUUID(), null, null), callback);
+
+        verify(tbCoreQueueProducer, times(1)).send(eq(tpi), any(TbProtoQueueMsg.class), eq(callback));
+    }
+
+    @Test
+    public void testPushNotificationToRuleEngineToUnknownServiceIdIsSkipped() {
```
smatvienko-tb, line 359: The "recently seen" grace-period and "fails open when topology unknown" branches of `isUnknownService` are only tested for `pushNotificationToCore` (see `testPushNotificationToCoreToRecentlySeenButOfflineServiceIdIsSent` / `testPushNotificationToCoreFailsOpenWhenClusterTopologyIsUnknown` above). `pushNotificationToRuleEngine` and `pushNotificationToTransport` only get an unknown-id and known-id test each — the grace-period and fail-open paths for those two call sites aren't exercised, even though they share the same `isUnknownService`/`recentlySeenServiceKey` logic keyed by `ServiceType`. Worth mirroring the two core-specific tests for rule engine and transport so a regression scoped to those service types (e.g. a key-format mismatch) doesn't slip through.

smatvienko-tb (commented):
## Review summary

Reviewed 3 changed files in [Skip notification push to unknown service id to prevent orphan notification topics](https://github.com/thingsboard/thingsboard/pull/15729). Left 5 comment(s) inline.

The core fix is sound: `isUnknownService` correctly checks live membership first, falls back to a bounded recently-seen cache to grace transient membership-snapshot lag, and fails open when the topology snapshot is empty — I traced this through `HashPartitionService`, the `ServiceListChangedEvent` publication path, and all real callers (`DefaultTbRuleEngineRpcService`, `DefaultTbCoreDeviceRpcService`, where `callback` is always `null` in practice, exercising the null-guard) and didn't find a correctness issue. The `currentOtherServices` → `volatile` change in `HashPartitionService` is a legitimate fix, since this guard now reads that field from far more threads than before. Remaining comments below are quality/design suggestions, not bugs.

---

PR #15732, opened 2026-06-05
Edge: debounce disconnect notifications for unstable connections

AndriiLandiak:
## Pull Request description

 Edges with poor connectivity connect/disconnect frequently. Each transition fired an EdgeConnectionTrigger immediately; the existing dedup keys on the connected flag, so alternating connect/disconnect events each notified and spammed the notification center.

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -324,6 +334,7 @@ public void deleteEdge(TenantId tenantId, EdgeId edgeId) {
             } finally {
                 newEventLock.unlock();
             }
+            cancelPendingDisconnectNotification(edgeId);
```
ViacheslavKlimov: This cancel only runs when there's a live connected session — but a pending disconnect notification can only exist when the edge is *disconnected* (it's scheduled in `onEdgeDisconnect` and cancelled on the next `onEdgeConnect`). So if an edge is deleted during its disconnect-notification window (no active session), this guard is skipped and the pending task still fires `notifyEdgeConnectivity(false)` for an already-deleted edge. Should it move outside the `if (session != null && session.isConnected())` block so it runs unconditionally?

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -684,6 +694,53 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
...
+                    disconnectNotificationDelayMs, TimeUnit.MILLISECONDS);
+        });
+    }
+
+    private void fireDelayedDisconnectNotification(TenantId tenantId, Edge edge) {
+        EdgeId edgeId = edge.getId();
+        pendingDisconnectNotifications.remove(edgeId);
+        // Re-verify the edge is still disconnected. The cache is cluster-wide, so this also covers the case
+        // where the edge dropped on this node and reconnected to a different TB-Core node within the window.
+        if (sessions.containsKey(edgeId) || edgeIdServiceIdCache.get(edgeId) != null) {
```
ViacheslavKlimov: The cross-node suppression here relies on `edgeIdServiceIdCache` still holding the new node's entry — but `onEdgeDisconnect` evicts this key unconditionally (line 583). If the edge reconnects to a different node *before* this node finishes processing the disconnect (e.g. while waiting out the keep-alive timeout), this node's `evict` can wipe the entry the other node just `put`, and the delayed check then sees `null` and fires a false 'disconnected' notification. Is that ordering guaranteed, or is the residual race considered acceptable?

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -242,6 +246,12 @@ public void destroy() {
                 sessionEdgeEventChecks.remove(edgeId);
             }
         }
+        pendingDisconnectNotifications.values().forEach(task -> {
```
ViacheslavKlimov: The 'cancel the future if non-null and not done' idiom now appears in three places — here, in `scheduleDisconnectNotification`'s `compute` block, and in `cancelPendingDisconnectNotification`. Might be worth a small `cancelIfPending(ScheduledFuture<?>)` helper so the semantics (e.g. `mayInterruptIfRunning=false`) live in one spot and can't drift apart.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -684,6 +694,53 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
...
+        if (disconnectNotificationDelayMs <= 0) {
+            notifyEdgeConnectivity(tenantId, edge, false);
+            return;
+        }
+        EdgeId edgeId = edge.getId();
+        pendingDisconnectNotifications.compute(edgeId, (id, existing) -> {
+            if (existing != null && !existing.isDone()) {
+                existing.cancel(false);
+            }
+            return executorService.schedule(() -> fireDelayedDisconnectNotification(tenantId, edge),
```
ViacheslavKlimov: These tasks run on the single-threaded `executorService` that also drives `cleanupZombieSessions` (every 60s) and the sync-request timeouts. The rule-processor call itself is async (`process` just submits to `notificationExecutor`), so that part is cheap — but the `edgeIdServiceIdCache.get()` re-verify in `fireDelayedDisconnectNotification` is a potentially remote cache read, and with many edges firing around the same time those reads serialize on this one thread and could delay zombie cleanup / sync timeouts. Worth considering `edgeEventProcessingExecutorService` (the pooled one) for these, or confirming disconnect volume stays low.

application/src/main/resources/thingsboard.yml
```diff
@@ -1617,6 +1617,11 @@ edges:
   state:
     # Persist state of edge (active, last connect, last disconnect) into timeseries or attributes tables. 'false' means to store edge state into attributes table
     persistToTelemetry: "${EDGES_PERSIST_STATE_TO_TELEMETRY:false}"
+  connectivity:
+    # Delay (ms) before sending an edge "disconnected" notification. Suppressed if the edge reconnects within
+    # this window - debounces flapping edges from spamming the notification center. Only the notification is
+    # delayed; rule-engine events and state attributes update immediately. Set to 0 to notify immediately.
+    disconnect_notification_delay_ms: "${TB_EDGE_DISCONNECT_NOTIFICATION_DELAY_MS:60000}"
```
ViacheslavKlimov: Every other property in the `edges:` section uses the `EDGES_` env-var prefix (`EDGES_PERSIST_STATE_TO_TELEMETRY`, `EDGES_STATS_ENABLED`, etc.). This one uses `TB_EDGE_DISCONNECT_NOTIFICATION_DELAY_MS`, which breaks the convention — something like `EDGES_DISCONNECT_NOTIFICATION_DELAY_MS` (or `EDGES_CONNECTIVITY_...`) would be consistent and easier to discover.

application/src/test/java/org/thingsboard/server/edge/EdgeConnectionNotificationTest.java
```diff
@@ -0,0 +1,96 @@
...
+import static org.awaitility.Awaitility.await;
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.clearInvocations;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+@DaoSqlTest
+@TestPropertySource(properties = {
```
ViacheslavKlimov: Two new branches aren't covered here: the `disconnectNotificationDelayMs <= 0` immediate-notify path in `scheduleDisconnectNotification`, and the `edgeIdServiceIdCache` cluster-wide re-verify branch in `fireDelayedDisconnectNotification` (the flap test exercises the same-node cancel path, but not the cache fallback that is the stated motivation for that branch). The `<= 0` case is easy to add with a second test class setting the property to 0 and asserting the disconnect notification fires immediately; the cluster path is harder in a single-node test, but at least worth acknowledging the gap.

application/src/test/java/org/thingsboard/server/edge/EdgeConnectionNotificationTest.java
```diff
@@ -0,0 +1,96 @@
...
+        "edges.connectivity.disconnect_notification_delay_ms=5000"
+})
+public class EdgeConnectionNotificationTest extends AbstractEdgeTest {
+
+    private static final long DELAY_MS = 5000L;
```
ViacheslavKlimov: `DELAY_MS` here and the `5000` literal in `@TestPropertySource` above must stay in sync, but nothing enforces it (annotation attributes need compile-time constants, so you can't reference `DELAY_MS` there). A short comment noting the coupling would help a future reader avoid silently desyncing the timing assumptions.

application/src/test/java/org/thingsboard/server/edge/EdgeConnectionNotificationTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void givenEdgeReconnectsWithinDelay_whenEdgeFlaps_thenDisconnectNotificationSuppressed() throws Exception {
+        clearInvocations(notificationRuleProcessor);
+
+        // Edge drops...
+        edgeImitator.disconnect();
+        // Ensure the server processed the disconnect (and scheduled the delayed notification) before reconnecting.
+        TimeUnit.SECONDS.sleep(1);
+
+        // ...and reconnects within the delay window, which must cancel the pending "disconnected" notification.
+        EdgeImitator reconnected = new EdgeImitator(EDGE_HOST, EDGE_PORT, edge.getRoutingKey(), edge.getSecret());
```
ViacheslavKlimov: This reconnect setup (construct + the two `ignoreType(OAuth2...)` calls + `connect`) duplicates `AbstractEdgeTest.setupEdgeTest` (lines 157-162). If the base setup later gains another `ignoreType` or changes construction, this copy drifts silently. A small protected helper on the base test (e.g. `newConnectedImitator()`) that both call would keep them aligned.

application/src/test/java/org/thingsboard/server/edge/EdgeConnectionNotificationTest.java
```diff
@@ -0,0 +1,96 @@
...
+        reconnected.ignoreType(OAuth2DomainUpdateMsg.class);
+        reconnected.connect();
+        edgeImitator = reconnected; // let teardown clean up the live session
+
+        // The "connected" notification still fires immediately on reconnect (we suppress the disconnect only).
+        await().atMost(AbstractWebTest.TIMEOUT, TimeUnit.SECONDS).untilAsserted(() ->
+                verify(notificationRuleProcessor, atLeastOnce()).process(argThat(edgeConnectionTrigger(true))));
+
+        // Wait until the original disconnect-notification window has fully elapsed...
+        TimeUnit.MILLISECONDS.sleep(DELAY_MS + 1000);
```
ViacheslavKlimov: This leans on a fixed `sleep(DELAY_MS + 1000)` to wait out the window, which couples the test to the 5000ms property and adds ~6s of pure sleeping per run. Could it instead await on something observable (e.g. the pending-map entry being cleared, or the connect notification having fired) and keep the residual sleep minimal? That would be both faster and less timing-dependent.

ViacheslavKlimov (commented):
## Review summary

Reviewed 3 changed files in [Edge: debounce disconnect notifications for unstable connections](https://github.com/thingsboard/thingsboard/pull/15732). Left 9 comments inline.

The core debounce approach is sound — splitting the immediate rule-engine state event from the delayed user-facing notification, and leaning on the existing 1-minute dedup window (keyed on `connected`) to collapse the repeated connect events that flapping produces. The main correctness concern is that the new `cancelPendingDisconnectNotification` call in `deleteEdge` sits inside the connected-session guard, so it never fires for the one case it appears intended to cover. The remaining comments cover the cross-node re-verify assumption, the shared single-thread executor, an env-var naming inconsistency, and test coverage/cleanliness.

---

ViacheslavKlimov (commented):
## Review summary

Reviewed 5 changed files in [Edge: debounce disconnect notifications for unstable connections](https://github.com/thingsboard/thingsboard/pull/15732). Left 8 comment(s) inline.

The core approach is sound — splitting the immediate rule-engine state event (`pushStateEventToRuleEngine`) from the notification trigger (`notifyEdgeConnectivity`), deferring only the disconnect notification, and adding a cluster-aware cache guard so a cross-node reconnect doesn't raise a false 'disconnected'. The `@PreDestroy` and `deleteEdge` cleanup of pending futures is handled, and the `edgeSessions` cache (no TTL by default) makes the `cache.get != null` guard reliable. The comments are mostly non-blocking: a question on the 60s default's impact on genuine-outage alerting, a couple of robustness/resource notes around the shared scheduler, an untested cross-node path, and a few test-structure observations.

---

application/src/main/resources/thingsboard.yml
```diff
@@ -1617,6 +1617,11 @@ edges:
   state:
     # Persist state of edge (active, last connect, last disconnect) into timeseries or attributes tables. 'false' means to store edge state into attributes table
     persistToTelemetry: "${EDGES_PERSIST_STATE_TO_TELEMETRY:false}"
+  connectivity:
+    # Delay (ms) before sending an edge "disconnected" notification. Suppressed if the edge reconnects within
+    # this window - debounces flapping edges from spamming the notification center. Only the notification is
+    # delayed; rule-engine events and state attributes update immediately. Set to 0 to notify immediately.
+    disconnect_notification_delay_ms: "${EDGES_DISCONNECT_NOTIFICATION_DELAY_MS:60000}"
```
ViacheslavKlimov, line 1624: With the default of `60000`, every genuine edge disconnect — including edges that go down and stay down — now surfaces its "disconnected" notification a full minute late. For outage alerting that delay is significant. Was 60s the intended default? If the goal is purely to absorb brief flaps, a smaller default (a few seconds) would still debounce flapping while keeping real-outage alerts timely; `0` preserves the previous immediate-notify behavior for anyone relying on it.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -684,6 +701,54 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
...
+
+    private void scheduleDisconnectNotification(TenantId tenantId, Edge edge) {
+        if (disconnectNotificationDelayMs <= 0) {
+            notifyEdgeConnectivity(tenantId, edge, false);
+            return;
+        }
+        EdgeId edgeId = edge.getId();
+        pendingDisconnectNotifications.compute(edgeId, (id, existing) -> {
+            cancelIfPending(existing);
+            return edgeEventProcessingExecutorService.schedule(() -> fireDelayedDisconnectNotification(tenantId, edge),
```
ViacheslavKlimov: These tasks run on the shared `edgeEventProcessingExecutorService`, which is a `ScheduledThreadPoolExecutor` without `removeOnCancelPolicy` set. Every reconnect cancels a pending future, but a cancelled `ScheduledFutureTask` isn't removed from the delay queue until its original fire time (~`disconnectNotificationDelayMs`, 60s by default). For exactly the flapping edges this feature targets, those cancelled tasks pile up in the queue for up to a minute each. Worth considering `setRemoveOnCancelPolicy(true)` on that executor — it would help the edge-event-check cancellations too.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -684,6 +701,54 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
...
+                    disconnectNotificationDelayMs, TimeUnit.MILLISECONDS);
+        });
+    }
+
+    private void fireDelayedDisconnectNotification(TenantId tenantId, Edge edge) {
+        EdgeId edgeId = edge.getId();
+        pendingDisconnectNotifications.remove(edgeId);
```
ViacheslavKlimov: This unconditionally removes whatever future is currently mapped for the edge. In a tight disconnect → reconnect → disconnect sequence right at the fire boundary, a newer future scheduled by the second disconnect could be the entry removed here, leaving it untracked so a later `cancelPendingDisconnectNotification()`/`deleteEdge()` wouldn't cancel it. The guard just below keeps the notification itself correct, so the impact is limited — but `remove(edgeId, <this future>)` keyed on identity would be safer. Minor.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -684,6 +701,54 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
...
+        // Re-verify the edge is still disconnected. The cache is cluster-wide, so this also covers the case
+        // where the edge dropped on this node and reconnected to a different TB-Core node within the window.
+        if (sessions.containsKey(edgeId) || edgeIdServiceIdCache.get(edgeId) != null) {
```
ViacheslavKlimov: This cross-node re-verification (together with `evictServiceIdCacheIfOwnedByThisNode` above) is the subtlest logic in the change, but the new tests are all single-node, so the cluster path — edge dropping here while the cache still points at another node — isn't exercised. A full multi-node integration test is admittedly hard in this harness; if one isn't feasible, even a unit-level test of the ownership/guard predicates would protect against a future regression here.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -684,6 +701,54 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
...
+            return;
+        }
+        notifyEdgeConnectivity(tenantId, edge, false);
+    }
+
+    private void cancelPendingDisconnectNotification(EdgeId edgeId) {
+        cancelIfPending(pendingDisconnectNotifications.remove(edgeId));
+    }
+
+    private static void cancelIfPending(ScheduledFuture<?> future) {
```
ViacheslavKlimov, line 803: This is the third place in the class that cancels a `ScheduledFuture` — `cancelScheduleEdgeEventsCheck` and the `destroy()` loop both do the same `isCancelled`/`isDone` check then `cancel(true)`. Now that there's a reusable helper, consider routing those two through it (parameterizing `mayInterruptIfRunning`) so there's one canonical cancel path rather than three subtly different inline variants.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -386,7 +394,11 @@ private void onEdgeConnect(EdgeId edgeId, EdgeGrpcSession edgeGrpcSession) {
         long lastConnectTs = System.currentTimeMillis();
         save(tenantId, edgeId, LAST_CONNECT_TIME, lastConnectTs);
         edgeIdServiceIdCache.put(edgeId, serviceInfoProvider.getServiceId());
-        pushRuleEngineMessage(tenantId, edge, lastConnectTs, TbMsgType.CONNECT_EVENT);
+        // If the edge reconnected within the disconnect-notification delay window, suppress the pending
+        // "disconnected" notification - the drop was transient (debounce for flapping edges).
+        cancelPendingDisconnectNotification(edgeId);
+        pushStateEventToRuleEngine(tenantId, edge, lastConnectTs, TbMsgType.CONNECT_EVENT);
+        notifyEdgeConnectivity(tenantId, edge, true);
```
ViacheslavKlimov: Splitting the old `pushRuleEngineMessage` into `pushStateEventToRuleEngine` + `notifyEdgeConnectivity` reads well, but the asymmetry it creates is easy to miss: connect calls both immediately, while disconnect pushes the state event now and defers the notification. You already have a nice comment above explaining the cancel — one more line here ("connect notifies immediately; only disconnect is debounced") would make that asymmetry self-documenting.

application/src/test/java/org/thingsboard/server/edge/EdgeImmediateDisconnectNotificationTest.java
```diff
@@ -0,0 +1,64 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+// Covers the disconnect_notification_delay_ms <= 0 branch in EdgeGrpcService.scheduleDisconnectNotification,
+// where the "disconnected" notification must be sent immediately instead of being scheduled with a delay.
+@DaoSqlTest
+@TestPropertySource(properties = {
+        "edges.connectivity.disconnect_notification_delay_ms=0"
+})
+public class EdgeImmediateDisconnectNotificationTest extends AbstractEdgeTest {
```
ViacheslavKlimov: This class and `EdgeConnectionNotificationTest` are near-identical — same superclass, same `@MockitoSpyBean`, same `edgeConnectionTrigger` matcher — differing only in the `disconnect_notification_delay_ms` value. Because each carries a distinct `@TestPropertySource`, Spring builds and caches a separate full application context per class, and `AbstractEdgeTest` boots a heavy context (gRPC server included), so this roughly doubles the most expensive part of these tests. Could the delay be overridden at runtime instead (a test-only setter, or reflectively setting the field) so all three cases share one context?

application/src/test/java/org/thingsboard/server/edge/EdgeConnectionNotificationTest.java
```diff
@@ -0,0 +1,92 @@
...
+        EdgeImitator reconnected = createEdgeImitator();
+        reconnected.connect();
+        edgeImitator = reconnected; // let teardown clean up the live session
+
+        // The "connected" notification still fires immediately on reconnect (we suppress the disconnect only).
+        await().atMost(AbstractWebTest.TIMEOUT, TimeUnit.SECONDS).untilAsserted(() ->
+                verify(notificationRuleProcessor, atLeastOnce()).process(argThat(edgeConnectionTrigger(true))));
+
+        // Wait until the original disconnect-notification window has fully elapsed...
+        TimeUnit.MILLISECONDS.sleep(DELAY_MS + 1000);
```
ViacheslavKlimov: This fixed 6s sleep (plus the earlier 1s) makes the suppression test wait ~7s, and a hardcoded sleep is the slow/fragile pattern the codebase generally avoids in favor of Awaitility. The negative assertion does need to wait out the window, but Awaitility's `during()`/`atLeast()` would express "never fired within this window" more clearly and without the magic `+1000` padding.

application/src/main/resources/thingsboard.yml
AndriiLandiak: No!!!

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
AndriiLandiak: no, executorService -> cleanupExecutorService and use the same as for zoombies.. How I missed it? Who told Claude to use edgeEventProcessingExecutorService??

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -684,6 +690,62 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
...
+        });
+    }
+
+    private void fireDelayedDisconnectNotification(TenantId tenantId, Edge edge, ScheduledFuture<?> self) {
+        EdgeId edgeId = edge.getId();
+        // Identity-keyed remove: don't clobber a newer future if a second disconnect races with this task firing.
+        pendingDisconnectNotifications.remove(edgeId, self);
+        // Re-verify the edge is still disconnected. The cache is cluster-wide, so this also covers the case
+        // where the edge dropped on this node and reconnected to a different TB-Core node within the window.
+        if (sessions.containsKey(edgeId) || edgeIdServiceIdCache.get(edgeId) != null) {
```
ViacheslavKlimov: **Still unresolved:** This cross-node re-verification (together with `evictServiceIdCacheIfOwnedByThisNode`) is still the subtlest logic in the change, and the tests remain single-node — the path where the edge drops on this node while the cluster cache still points at another node isn't exercised. A full multi-node integration test is admittedly hard in this harness, but even a unit-level test of the ownership/guard predicates (`evictServiceIdCacheIfOwnedByThisNode` and this `sessions.containsKey(edgeId) || cache.get(edgeId) != null` check) would protect against a future regression here.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -684,6 +690,62 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
...
+
+    private void scheduleDisconnectNotification(TenantId tenantId, Edge edge) {
+        if (disconnectNotificationDelayMs <= 0) {
+            notifyEdgeConnectivity(tenantId, edge, false);
+            return;
+        }
+        EdgeId edgeId = edge.getId();
+        pendingDisconnectNotifications.compute(edgeId, (id, existing) -> {
+            cancelIfPending(existing);
+            ScheduledFuture<?>[] holder = {null};
```
ViacheslavKlimov: Minor readability: the single-element `holder` array is here because the scheduled task needs a reference to its own future (for the identity-keyed `remove(edgeId, self)`), which doesn't exist until `schedule(...)` returns — a genuine self-reference, so the idiom is justified. But it's the least obvious line in the method, and the only explanation lives further down in `fireDelayedDisconnectNotification`. A one-line comment here noting *why* the holder is needed would save the next reader the puzzle.

ViacheslavKlimov (commented):
## Re-review summary

Re-reviewed [Edge: debounce disconnect notifications for unstable connections](https://github.com/thingsboard/thingsboard/pull/15732) — verified 8 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/15732#pullrequestreview-4450512514).

| Status | Count |
|--------|-------|
| ✅ Resolved | 6 |
| 💬 Acknowledged | 1 |
| ❌ Unresolved | 1 |

Also found 2 new (minor) issue(s) in the fix commits, commented inline.

### Finding details

- ✅ **EdgeGrpcService.java** (`scheduleDisconnectNotification`) — delayed-notification task ran on the contended `edgeEventProcessingExecutorService` — *Moved to the single-thread `executorService` (the zombie-cleanup / sync-timeout executor) per your reply. `removeOnCancelPolicy` is still unset, but that's consistent with the existing pattern on that executor — minor.*
- ✅ **EdgeGrpcService.java** (`fireDelayedDisconnectNotification`) — unconditional `remove(edgeId)` could drop a newer future — *Fixed: now identity-keyed `remove(edgeId, self)` via the holder.*
- ✅ **EdgeGrpcService.java** (`cancelIfPending`) — three divergent inline cancel sites — *Fixed: `destroy()` and `cancelScheduleEdgeEventsCheck` now route through the helper.*
- ✅ **EdgeGrpcService.java** (`onEdgeConnect`) — connect-immediate / disconnect-debounced asymmetry was undocumented — *Fixed: explanatory comment added.*
- ✅ **EdgeImmediateDisconnectNotificationTest** — near-duplicate test forced a second heavy Spring context — *Fixed: deleted and folded into `EdgeConnectionNotificationTest`, sharing one context via `ReflectionTestUtils`.*
- ✅ **EdgeConnectionNotificationTest** (suppression test) — fixed `sleep(DELAY_MS + 1000)` — *Fixed: replaced with Awaitility `during().atMost()`.*
- 💬 **thingsboard.yml** (`disconnect_notification_delay_ms`) — 60s default delays a genuine-outage notification by a full minute — *Developer: "No!!!" — keeping 60s as the intended debounce window. Reasonable as a product call.*
- ❌ **EdgeGrpcService.java** (`fireDelayedDisconnectNotification`) — the cross-node reconnect guard still has no test — *Still uncovered; flagged inline.*

---

application/src/test/java/org/thingsboard/server/edge/EdgeConnectionNotificationTest.java
```diff
@@ -0,0 +1,113 @@
...
+    @MockitoSpyBean
+    private NotificationRuleProcessor notificationRuleProcessor;
+
+    @Autowired
+    private EdgeGrpcService edgeGrpcService;
+
+    // The delay is overridden per test (rather than via a per-class @TestPropertySource) so all cases share a
+    // single Spring application context instead of booting a separate heavy context per delay value.
+    private void setDisconnectNotificationDelayMs(long delayMs) {
+        ReflectionTestUtils.setField(edgeGrpcService, "disconnectNotificationDelayMs", delayMs);
```
ViacheslavKlimov, line 73: Each test sets this explicitly, so it's correct today, and the `@MockitoSpyBean` gives this class its own cached application context — so the mutation doesn't leak into the other `AbstractEdgeTest` subclasses. The remaining risk is intra-class: a test method added here later that forgets to call `setDisconnectNotificationDelayMs` would silently inherit the previous method's value (JUnit doesn't guarantee method order). Capturing the original value and restoring it in `@After` would make the class robust against that, at no cost to the shared-context optimization.

ViacheslavKlimov:
Leaving only the 4.3 PRs

PR #15743, opened 2026-06-08
fix: release DirectByteBuffer after MQTT SSL auth to prevent native memory leak

pon0marev:
## Pull Request description

Holding sslHandler as a final field kept SSLEngine and its JNI-allocated DirectByteBuffers alive for the full session lifetime. On rapid reconnect loops this caused linear native memory growth, eventually hitting OOM.

Null the reference at the end of onValidateDeviceResponse() so SslHandler becomes eligible for Young GC immediately after authentication completes. Add boolean ssl flag to preserve isSSL semantics after the field is cleared.

### What happens

On every MQTT SSL connection Netty creates an `SslHandler` wrapping a JDK `SSLEngine`
(JSSE, not OpenSSL). During the TLS handshake, `SSLEngine` allocates `DirectByteBuffer`
memory via native JNI code. This memory is tracked by JMX as
`java.nio:type=BufferPool,name=direct`.

`MqttTransportHandler` held `sslHandler` as a `final` field — the reference was kept
for the entire lifetime of the handler (from connect to disconnect).

### Why buffers were never freed

`DirectByteBuffer` is released through a `Cleaner` — a `PhantomReference` that calls
`Unsafe.freeMemory()` when the object is garbage collected. But two things prevented this:

1. **GC never collected Old Gen.** Heap was only ~20% occupied, and G1GC only starts Mixed GC
   at `InitiatingHeapOccupancyPercent=45%` (default). That threshold was never reached.

2. **Netty does not close JDK `SSLEngine` on pipeline removal.**
   `SslHandler.handlerRemoved0()` releases the engine only if it is `instanceof ReferenceCounted`.
   `JdkSslEngine` is **not** `ReferenceCounted` — so the engine and its DirectByteBuffers
   stay alive as long as `MqttTransportHandler` is in memory.

After channel close, all external references to `MqttTransportHandler` are cleared
(pipeline, closeFuture, sessions map), but the handler sat in Old Gen uncollected.

### Code fix (MqttTransportHandler.java)

`common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java`

`sslHandler` is used **only** during CONNECT packet processing:
- `processConnect()` — extracts the client X509 certificate
- `onValidateDeviceResponse()` — checks the certificate on auth failure

After `onValidateDeviceResponse()` completes, `sslHandler` is **never used again**,
but was previously held as a `final` field until the end of the session.

### Misbehaving device (200ms reconnect loop)

Before fix:
- Each SSL handshake → `SslHandler` → `SSLEngine` → DirectByteBuffer (~150 KB)
- `sslHandler` held by `final` field → object promoted to Old Gen
- Old Gen never collected (heap 20% < IHOP 45%)
- Memory grows linearly: OOM in 2–3 days

After fix:
- `sslHandler = null` immediately after auth → object eligible for **Young GC**
- `SSLEngine` and its DirectByteBuffer collected on the next Minor GC, never reach Old Gen
- Cleaner runs, `Unsafe.freeMemory()` called — native memory returned to OS

## Why It Works

`DirectByteBuffer` lives as long as `SSLEngine` is alive. `SSLEngine` lives as long as
`SslHandler` is alive. `SslHandler` lived as long as `MqttTransportHandler` because of
the `final` field reference. `MqttTransportHandler` is a long-lived Netty `ChannelHandler`
that exists for the full duration of the TCP connection.

Setting `this.sslHandler = null` breaks this chain immediately after auth.
`SslHandler` has no more strong references from the handler → becomes unreachable →
GC triggers the `Cleaner` → `Unsafe.freeMemory()` → native memory returned to OS.

The key insight: `SslHandler` is not needed after authentication. All subsequent TLS
encrypt/decrypt work is done automatically by the Netty pipeline. `MqttTransportHandler`
only handles MQTT application logic — it never touches TLS directly after the CONNECT phase.

## General checklist

## Back-End feature checklist

### DIFF
--- a/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
+++ b/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
@@ -150,7 +150,6 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
     protected final MqttTransportContext context;
     private final TransportService transportService;
     private final SchedulerComponent scheduler;
-    private final boolean ssl;
     private volatile SslHandler sslHandler;
     private final ConcurrentMap<MqttTopicMatcher, Integer> mqttQoSMap;
 
@@ -173,7 +172,6 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
         this.context = context;
         this.transportService = context.getTransportService();
         this.scheduler = context.getScheduler();
-        this.ssl = sslHandler != null;
         this.sslHandler = sslHandler;
         this.mqttQoSMap = new ConcurrentHashMap<>();
         this.deviceSessionCtx = new DeviceSessionCtx(sessionId, mqttQoSMap, context);
--- a/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
+++ b/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
@@ -150,7 +150,8 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
     protected final MqttTransportContext context;
     private final TransportService transportService;
     private final SchedulerComponent scheduler;
-    private final SslHandler sslHandler;
+    private final boolean ssl;
+    private volatile SslHandler sslHandler;
     private final ConcurrentMap<MqttTopicMatcher, Integer> mqttQoSMap;
 
     final DeviceSessionCtx deviceSessionCtx;
@@ -172,6 +173,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
         this.context = context;
         this.transportService = context.getTransportService();
         this.scheduler = context.getScheduler();
+        this.ssl = sslHandler != null;
         this.sslHandler = sslHandler;
         this.mqttQoSMap = new ConcurrentHashMap<>();
         this.deviceSessionCtx = new DeviceSessionCtx(sessionId, mqttQoSMap, context);
@@ -1373,6 +1375,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
                 }
             });
         }
+        this.sslHandler = null;
     }
 
     @Override


common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -150,7 +150,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
     protected final MqttTransportContext context;
     private final TransportService transportService;
     private final SchedulerComponent scheduler;
-    private final SslHandler sslHandler;
+    private volatile SslHandler sslHandler;
```
smatvienko-tb, line 153: Making the field `volatile` is the right call now that it's written after construction. But it's worth stepping back on what nulling it actually buys for the leak. This `SslHandler` is also added to the Netty pipeline (`MqttTransportServerInitializer:50`) and stays there decrypting every packet for the whole connection, so the *pipeline* holds a strong reference to it until the channel closes — independent of this field. That means nulling here doesn't make the handler GC-eligible while the session is open; the PR description's "eligible for Young GC immediately after auth / collected on the next Minor GC" doesn't quite hold, since the pipeline reference only drops at channel close. Where the fix does help: `registerAsyncSession(..., this)` keeps `MqttTransportHandler` reachable in a long-lived registry, so once the channel closes, this null stops the lingering handler from dragging the `SslHandler` (and its off-heap buffers) along with it. Could you confirm the native-memory measurements were taken against this exact change, so we know the win is real and not from something else (e.g. heap pressure finally crossing IHOP)?

Separately, the field now carries an implicit lifecycle contract — valid only during CONNECT/auth, intentionally null afterwards — that isn't documented anywhere. The three readers (1057, 1125, 1329) all happen to run before the null today, but nothing signals that to the next person who adds a fourth reader. A one-line comment on the field describing the construct → use-during-auth → null lifecycle would make the invariant discoverable.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -1373,6 +1373,7 @@ public void onError(Throwable e) {
                 }
             });
         }
+        this.sslHandler = null;
```
smatvienko-tb, line 1376: A few things converge on this line.

**Concurrency / NPE.** `onValidateDeviceResponse` runs on the transport callback executor, not the Netty event loop, so this `null` write races with reads on the event loop. `getX509Certificate()` (line 1125) dereferences `sslHandler.engine()` directly, and the `sslHandler != null` guard at 1057 (and `== null` at 1329) is a *separate* volatile read from that deref. If a client sends a second CONNECT that's processed on the event loop while a prior CONNECT's `onValidateDeviceResponse` runs this line on the callback thread, the guard can pass and then 1125 reads `null` → NPE (only `SSLPeerUnverifiedException` is caught there, so it escapes `channelRead`). With the old `final`, non-null field this couldn't happen. Snapshotting once closes the window and is cleaner regardless:
```java
private X509Certificate getX509Certificate() {
    SslHandler handler = this.sslHandler;
    if (handler == null) {
        return null;
    }
    Certificate[] certChain = handler.engine().getSession().getPeerCertificates();
    ...
}
```
That also lets the callers drop their separate `sslHandler != null` pre-check.

**Behavioral change on re-CONNECT.** After this null, a repeat CONNECT from an X509 client sees `sslHandler == null` at 1057 and silently falls through to `processAuthTokenConnect` instead of certificate auth (`processMqttMsg`:1277 has no guard against a duplicate CONNECT). Previously it would re-validate via the cert. Probably an edge case, but worth confirming it's acceptable.

**Readability.** This bare assignment at the very end of a ~50-line auth method reads like dead code without context. A short comment on *why* it's nulled (releasing the SSL engine's off-heap DirectByteBuffers once auth is done) would save the next reader a trip to the commit log and discourage a "cleanup" that removes it.

smatvienko-tb (commented):
## Review summary

Reviewed 1 changed file in [fix: release DirectByteBuffer after MQTT SSL auth to prevent native memory leak](https://github.com/thingsboard/thingsboard/pull/15743). Left 2 comments inline.

The diff is small and the leak-fix intent is sound, but three things are worth a closer look before merge: (1) a concurrency window introduced by making the field mutable that can NPE on a repeat CONNECT, (2) a question about whether nulling the field actually frees the off-heap buffers before channel close — the Netty pipeline keeps a strong reference to the `SslHandler` for the entire session, and (3) the absence of any regression test for the new behavior. Details inline and below.

## Additional findings

These are about the change as a whole rather than a single diff line.

- **No regression test for the null-after-auth behavior** — `MqttTransportHandlerTest` already mocks an `SslHandler` and constructs the handler with it (lines 88/105), so the scaffolding is nearly free, yet the suite never exercises `processConnect`'s X509 path or `onValidateDeviceResponse` at all. A future change that moves a `sslHandler` read below the null assignment, or drops the null entirely (reintroducing the leak), would pass CI silently. A test that drives `onValidateDeviceResponse` and asserts the field is nulled afterward, plus one confirming the X509 cert path still resolves *before* nulling, would lock in both the GC intent and the ordering invariant the fix depends on.
- **`isSSL` after the field is cleared (verified safe, no change needed)** — confirmed that dropping the intermediate `boolean ssl` flag is safe today: there are no subclasses of `MqttTransportHandler`, no `isSSL`/`isSsl` method anywhere in the transport modules, and all three `sslHandler` reads (`processConnect`:1057, `getX509Certificate`:1125, `onValidateDeviceResponse`:1329) run during the CONNECT/auth phase before line 1376. IP extraction is independent of this field — it comes from `ctx.channel().remoteAddress()` (`getAddress`:257). The only caveat is a latent trap: if a future feature ever needs to know post-auth that the session was over TLS (audit logging, a connectivity event, etc.), that signal is no longer available and re-adding the field would reintroduce the leak — another reason a comment at the null site is worthwhile.

---

smatvienko-tb:
**Follow-up: cross-edition caveat on dropping the `boolean ssl` flag.**

Came back to this after seeing the root-cause doc, which documents an `isSSL()` change (read the boolean instead of `sslHandler`) that was *required because `channelUnregistered()` calls `isSSL()` after `sslHandler` is nulled*. That doesn't match what I see here, so flagging it:

- Verified the flag removal is safe in **CE**: `channelUnregistered()` in this handler only calls `context.channelUnregistered()` → `connectionsCounter.decrementAndGet()`, and there is no `isSSL()` method anywhere in the CE transport handler. So the `boolean ssl` field really is dead code here.
- But the doc describes the **PE** variant, where `isSSL()` is consulted *after* the null. The PR checklist notes a similar PR is opened for PE. If that PE PR mirrors this one and also drops the `ssl` flag, PE's `isSSL()` reverts to `sslHandler != null` and returns `false` for SSL sessions after auth (or NPEs, depending on the impl) — which is the exact bug the flag was added to prevent.

Recommendation: keep the `boolean ssl` flag + `isSSL() { return ssl; }` in the **PE** PR even though it's genuinely unused in CE.

One more note for anyone cherry-picking: the memory stabilization in the metrics depends on the companion JVM/helm change (`InitiatingHeapOccupancyPercent=20`, periodic G1 GC, `MaxDirectMemorySize`) that directly addresses the "Old Gen never collected below IHOP" root cause. This PR is the code half only — the `SslHandler` is still strongly held by the Netty pipeline for the whole connection, so the field-null mainly stops a lingering (Old-Gen) handler from dragging the engine along after channel close, rather than freeing it "immediately after auth."

PR #15744, opened 2026-06-08
Fixed transport tenant-profile lock convoy under cold-cache reconnect storm

ViacheslavKlimov:
## Problem

The defective code lives in `common/transport/transport-api` and is shared by **all transports** (MQTT, HTTP, CoAP, LwM2M, SNMP). The production incident happened to surface on MQTT, but the fix is not MQTT-specific.

On a cold tenant-profile cache (e.g. after a transport cache clear + restart), a device reconnect storm serializes the whole transport instance behind tenant-profile resolution. The bounded callback pool fully saturates on threads blocked waiting for tenant-profile fetches, throughput collapses to "one tenant-profile fetch at a time", and the instance is effectively unavailable for ~15 minutes per restart until the cache warms.

Two compounding anti-patterns turn the storm into an instance-wide stall:

1. **Global lock across remote I/O** — `DefaultTransportTenantProfileCache.getTenantProfile()` held a single process-wide `ReentrantLock` across the synchronous cross-service `getEntityProfile` Kafka round-trip. Every tenant-profile cache miss in the entire process was serialized one-at-a-time.
2. **`computeIfAbsent`-over-I/O** — `DefaultTransportRateLimitService` performed that same blocking fetch *inside* `ConcurrentHashMap.computeIfAbsent`'s mapping function, holding a CHM bin lock across the remote round-trip (a second serialization layer).

## Fix

- **Per-tenant locking** in `DefaultTransportTenantProfileCache`: replace the single global lock with a bounded set of per-tenant locks (Guava `Striped.lock(1024)`). Different tenants resolve their profiles concurrently; concurrent misses for the *same* tenant are still de-duplicated into one fetch.
- **Fetch outside the bin lock** in `DefaultTransportRateLimitService`: pre-fetch the tenant profile before `computeIfAbsent`, so the mapping function is pure and no CHM bin lock is held across remote I/O.

## Cleanups (no behavior change)

- Collapsed the four near-identical `getXRateLimits` methods into one generic `getRateLimits(map, tenantId, entityId, type, onCreate)` helper.
- Moved the per-type rate-limit getters onto the `TransportLimitsType` enum, removing the `switch` in `createRateLimits`.
- `update(TenantId)` no longer fetches the tenant profile four times; extracted the duplicated info/debug logging branch in `mergeLimits`.

## Notes

- `Striped.lock(1024)` is the eager variant on purpose: the lock array is negligible memory, `get()` is a plain O(1) array index, and there is no weak-reference bookkeeping or GC overhead.
- `HashPartitionService` is intentionally left untouched: its direct `getRoutingInfo` call during partition recalculation is load-bearing for routing freshness — its `tenantRoutingInfoMap` is **not** evicted on tenant-profile updates (only on tenant update/delete and queue changes), unlike the tenant-profile caches, so resolving routing through that cache would risk misrouting an isolated tenant after a profile change. The expensive round-trip is also already de-duplicated per tenant by the profile cache, and with the global lock removed the recalculation path is no longer wedged behind the convoy.
- Backward compatible: no config or schema changes.

## Tests

- `DefaultTransportTenantProfileCacheTest` — a blocking fetch for one tenant must not block resolution of a different tenant.
- `DefaultTransportRateLimitServiceTest` — two concurrent rate-limit checks for the same tenant both reach the profile fetch concurrently (no bin lock held across I/O).

Both were written to fail against the previous code and pass after the fix.

### DIFF
--- a/common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
+++ b/common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
@@ -119,11 +119,13 @@ public class DefaultTransportRateLimitService implements TransportRateLimitServi
 
     @Override
     public void update(TenantId tenantId) {
-        EntityTransportRateLimits tenantRateLimitPrototype = createRateLimits(tenantProfileCache.get(tenantId), TENANT_LIMITS);
-        EntityTransportRateLimits deviceRateLimitPrototype = createRateLimits(tenantProfileCache.get(tenantId), DEVICE_LIMITS);
-        EntityTransportRateLimits gatewayRateLimitPrototype = createRateLimits(tenantProfileCache.get(tenantId), GATEWAY_LIMITS);
-        EntityTransportRateLimits gatewayDeviceRateLimitPrototype = createRateLimits(tenantProfileCache.get(tenantId), GATEWAY_DEVICE_LIMITS);
-        update(tenantId, tenantRateLimitPrototype, deviceRateLimitPrototype, gatewayRateLimitPrototype, gatewayDeviceRateLimitPrototype);
+        TenantProfile profile = tenantProfileCache.get(tenantId);
+        update(tenantId,
+                createRateLimits(profile, TENANT_LIMITS),
+                createRateLimits(profile, DEVICE_LIMITS),
+                createRateLimits(profile, GATEWAY_LIMITS),
+                createRateLimits(profile, GATEWAY_DEVICE_LIMITS)
+        );
     }
 
     private void update(TenantId tenantId, EntityTransportRateLimits tenantRateLimitPrototype, EntityTransportRateLimits deviceRateLimitPrototype,
@@ -231,22 +233,23 @@ public class DefaultTransportRateLimitService implements TransportRateLimitServi
                                                   BiConsumer<T, EntityTransportRateLimits> putFunction) {
         EntityTransportRateLimits oldRateLimits = getFunction.apply(entityId);
         if (oldRateLimits == null) {
-            if (EntityType.TENANT.equals(entityId.getEntityType())) {
-                log.info("[{}] New rate limits: {}", entityId, newRateLimits);
-            } else {
-                log.debug("[{}] New rate limits: {}", entityId, newRateLimits);
-            }
+            logLimits(entityId, "New", newRateLimits);
             putFunction.accept(entityId, newRateLimits);
         } else {
             EntityTransportRateLimits updated = merge(oldRateLimits, newRateLimits);
             if (updated != null) {
-                if (EntityType.TENANT.equals(entityId.getEntityType())) {
-                    log.info("[{}] Updated rate limits: {}", entityId, updated);
-                } else {
-                    log.debug("[{}] Updated rate limits: {}", entityId, updated);
-                }
+                logLimits(entityId, "Updated", updated);
                 putFunction.accept(entityId, updated);
             }
+        }
+    }
+
+    private void logLimits(EntityId entityId, String action, EntityTransportRateLimits limits) {
+        // Tenant-level changes are logged at INFO; the much noisier per-device/gateway ones at DEBUG.
+        if (EntityType.TENANT.equals(entityId.getEntityType())) {
+            log.info("[{}] {} rate limits: {}", entityId, action, limits);
+        } else {
+            log.debug("[{}] {} rate limits: {}", entityId, action, limits);
         }
     }
 
@@ -269,36 +272,12 @@ public class DefaultTransportRateLimitService implements TransportRateLimitServi
         DefaultTenantProfileConfiguration profile = (DefaultTenantProfileConfiguration) profileData.getConfiguration();
         if (profile == null) {
             return new EntityTransportRateLimits(ALLOW, ALLOW, ALLOW);
-        } else {
-            TransportRateLimit regularMsgRateLimit;
-            TransportRateLimit telemetryMsgRateLimit;
-            TransportRateLimit telemetryDpRateLimit;
-            switch (limitsType) {
-                case TENANT_LIMITS -> {
-                    regularMsgRateLimit = newLimit(profile.getTransportTenantMsgRateLimit());
-                    telemetryMsgRateLimit = newLimit(profile.getTransportTenantTelemetryMsgRateLimit());
-                    telemetryDpRateLimit = newLimit(profile.getTransportTenantTelemetryDataPointsRateLimit());
-                }
-                case DEVICE_LIMITS -> {
-                    regularMsgRateLimit = newLimit(profile.getTransportDeviceMsgRateLimit());
-                    telemetryMsgRateLimit = newLimit(profile.getTransportDeviceTelemetryMsgRateLimit());
-                    telemetryDpRateLimit = newLimit(profile.getTransportDeviceTelemetryDataPointsRateLimit());
-                }
-                case GATEWAY_LIMITS -> {
-                    regularMsgRateLimit = newLimit(profile.getTransportGatewayMsgRateLimit());
-                    telemetryMsgRateLimit = newLimit(profile.getTransportGatewayTelemetryMsgRateLimit());
-                    telemetryDpRateLimit = newLimit(profile.getTransportGatewayTelemetryDataPointsRateLimit());
-                }
-                case GATEWAY_DEVICE_LIMITS -> {
-                    regularMsgRateLimit = newLimit(profile.getTransportGatewayDeviceMsgRateLimit());
-                    telemetryMsgRateLimit = newLimit(profile.getTransportGatewayDeviceTelemetryMsgRateLimit());
-                    telemetryDpRateLimit = newLimit(profile.getTransportGatewayDeviceTelemetryDataPointsRateLimit());
-                }
-                default -> throw new IllegalStateException("Unknown limits type: " + limitsType);
-            }
-
-            return new EntityTransportRateLimits(regularMsgRateLimit, telemetryMsgRateLimit, telemetryDpRateLimit);
         }
+        return new EntityTransportRateLimits(
+                newLimit(limitsType.getRegularMsgRateLimit().apply(profile)),
+                newLimit(limitsType.getTelemetryMsgRateLimit().apply(profile)),
+                newLimit(limitsType.getTelemetryDataPointsRateLimit().apply(profile))
+        );
     }
 
     private static TransportRateLimit newLimit(String config) {
@@ -306,31 +285,34 @@ public class DefaultTransportRateLimitService implements TransportRateLimitServi
     }
 
     private EntityTransportRateLimits getTenantRateLimits(TenantId tenantId) {
-        return perTenantLimits.computeIfAbsent(tenantId, k -> createRateLimits(tenantProfileCache.get(tenantId), TENANT_LIMITS));
+        return getRateLimits(perTenantLimits, tenantId, tenantId, TENANT_LIMITS, null);
     }
 
     private EntityTransportRateLimits getDeviceRateLimits(TenantId tenantId, DeviceId deviceId) {
-        return perDeviceLimits.computeIfAbsent(deviceId, k -> {
-            EntityTransportRateLimits limits = createRateLimits(tenantProfileCache.get(tenantId), DEVICE_LIMITS);
-            getTenantDevices(tenantId).add(deviceId);
-            return limits;
-        });
+        return getRateLimits(perDeviceLimits, tenantId, deviceId, DEVICE_LIMITS, () -> getTenantDevices(tenantId).add(deviceId));
     }
 
     private EntityTransportRateLimits getGatewayRateLimits(TenantId tenantId, DeviceId gatewayId) {
-        return perGatewayLimits.computeIfAbsent(gatewayId, k -> {
-            EntityTransportRateLimits limits = createRateLimits(tenantProfileCache.get(tenantId), GATEWAY_LIMITS);
-            getTenantGateways(tenantId).add(gatewayId);
-            return limits;
-        });
+        return getRateLimits(perGatewayLimits, tenantId, gatewayId, GATEWAY_LIMITS, () -> getTenantGateways(tenantId).add(gatewayId));
     }
 
     private EntityTransportRateLimits getGatewayDeviceRateLimits(TenantId tenantId, DeviceId gatewayId) {
-        return perGatewayDeviceLimits.computeIfAbsent(gatewayId, k -> {
-            EntityTransportRateLimits limits = createRateLimits(tenantProfileCache.get(tenantId), GATEWAY_DEVICE_LIMITS);
-            getTenantGatewayDevices(tenantId).add(gatewayId);
-            return limits;
-        });
+        return getRateLimits(perGatewayDeviceLimits, tenantId, gatewayId, GATEWAY_DEVICE_LIMITS, () -> getTenantGatewayDevices(tenantId).add(gatewayId));
+    }
+
+    private <T extends EntityId> EntityTransportRateLimits getRateLimits(ConcurrentMap<T, EntityTransportRateLimits> limitsMap, TenantId tenantId,
+                                                                         T entityId, TransportLimitsType limitsType, Runnable onCreate) {
+        EntityTransportRateLimits limits = limitsMap.get(entityId);
+        if (limits == null) {
+            // Resolve the tenant profile WITHOUT holding the ConcurrentHashMap bin lock: the fetch may
+            // block on a cross-service round-trip, so it must run before computeIfAbsent's mapping function.
+            TenantProfile tenantProfile = tenantProfileCache.get(tenantId);
+            limits = limitsMap.computeIfAbsent(entityId, k -> createRateLimits(tenantProfile, limitsType));
+            if (onCreate != null) {
+                onCreate.run();
+            }
+        }
+        return limits;
     }
 
     private Set<DeviceId> getTenantDevices(TenantId tenantId) {
--- a/common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
+++ b/common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
@@ -107,11 +107,12 @@ public class DefaultTransportRateLimitService implements TransportRateLimitServi
 
     @Override
     public void update(TenantProfileUpdateResult update) {
-        log.info("Received tenant profile update: {}", update.getProfile());
-        EntityTransportRateLimits tenantRateLimitPrototype = createRateLimits(update.getProfile(), TENANT_LIMITS);
-        EntityTransportRateLimits deviceRateLimitPrototype = createRateLimits(update.getProfile(), DEVICE_LIMITS);
-        EntityTransportRateLimits gatewayRateLimitPrototype = createRateLimits(update.getProfile(), GATEWAY_LIMITS);
-        EntityTransportRateLimits gatewayDeviceRateLimitPrototype = createRateLimits(update.getProfile(), GATEWAY_DEVICE_LIMITS);
+        TenantProfile profile = update.getProfile();
+        log.info("Received tenant profile update: {}", profile);
+        EntityTransportRateLimits tenantRateLimitPrototype = createRateLimits(profile, TENANT_LIMITS);
+        EntityTransportRateLimits deviceRateLimitPrototype = createRateLimits(profile, DEVICE_LIMITS);
+        EntityTransportRateLimits gatewayRateLimitPrototype = createRateLimits(profile, GATEWAY_LIMITS);
+        EntityTransportRateLimits gatewayDeviceRateLimitPrototype = createRateLimits(profile, GATEWAY_DEVICE_LIMITS);
         for (TenantId tenantId : update.getAffectedTenants()) {
             update(tenantId, tenantRateLimitPrototype, deviceRateLimitPrototype, gatewayRateLimitPrototype, gatewayDeviceRateLimitPrototype);
         }
@@ -301,15 +302,17 @@ public class DefaultTransportRateLimitService implements TransportRateLimitServi
     }
 
     private <T extends EntityId> EntityTransportRateLimits getRateLimits(ConcurrentMap<T, EntityTransportRateLimits> limitsMap, TenantId tenantId,
-                                                                         T entityId, TransportLimitsType limitsType, Runnable onCreate) {
+                                                                         T entityId, TransportLimitsType limitsType, Runnable onMiss) {
         EntityTransportRateLimits limits = limitsMap.get(entityId);
         if (limits == null) {
             // Resolve the tenant profile WITHOUT holding the ConcurrentHashMap bin lock: the fetch may
             // block on a cross-service round-trip, so it must run before computeIfAbsent's mapping function.
             TenantProfile tenantProfile = tenantProfileCache.get(tenantId);
             limits = limitsMap.computeIfAbsent(entityId, k -> createRateLimits(tenantProfile, limitsType));
-            if (onCreate != null) {
-                onCreate.run();
+            // Runs on every observed miss, including callers that lost the computeIfAbsent race and got an
+            // existing value back - NOT only on actual creation, so the callback must be idempotent.
+            if (onMiss != null) {
+                onMiss.run();
             }
         }
         return limits;
--- a/common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/TransportLimitsType.java
+++ b/common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/TransportLimitsType.java
@@ -15,6 +15,39 @@
  */
 package org.thingsboard.server.common.transport.limits;
 
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import org.thingsboard.server.common.data.tenant.profile.DefaultTenantProfileConfiguration;
+
+import java.util.function.Function;
+
+@Getter
+@RequiredArgsConstructor
 public enum TransportLimitsType {
-    TENANT_LIMITS, DEVICE_LIMITS, GATEWAY_LIMITS, GATEWAY_DEVICE_LIMITS
+
+    TENANT_LIMITS(
+            DefaultTenantProfileConfiguration::getTransportTenantMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportTenantTelemetryMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportTenantTelemetryDataPointsRateLimit
+    ),
+    DEVICE_LIMITS(
+            DefaultTenantProfileConfiguration::getTransportDeviceMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportDeviceTelemetryMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportDeviceTelemetryDataPointsRateLimit
+    ),
+    GATEWAY_LIMITS(
+            DefaultTenantProfileConfiguration::getTransportGatewayMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportGatewayTelemetryMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportGatewayTelemetryDataPointsRateLimit
+    ),
+    GATEWAY_DEVICE_LIMITS(
+            DefaultTenantProfileConfiguration::getTransportGatewayDeviceMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportGatewayDeviceTelemetryMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportGatewayDeviceTelemetryDataPointsRateLimit
+    );
+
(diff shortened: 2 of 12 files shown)


common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCache.java
```diff
@@ -37,14 +38,15 @@
...
 
 @Component
 @TbTransportComponent
 @Slf4j
 public class DefaultTransportTenantProfileCache implements TransportTenantProfileCache {
 
-    private final Lock tenantProfileFetchLock = new ReentrantLock();
+    // Bounded set of per-tenant locks: de-duplicates concurrent misses for the same tenant while
+    // letting different tenants fetch concurrently (eager array - no weak-ref overhead at this size).
+    private final Striped<Lock> tenantProfileFetchLocks = Striped.lock(1024);
```
ViacheslavKlimov: Minor: the stripe count `1024` is a magic number that controls both the memory footprint and the per-tenant collision probability of these locks. Once a deployment has well over ~1k active tenants, birthday-paradox collisions start reintroducing exactly the cross-tenant serialization this change removes — two unrelated tenants sharing a stripe block each other. The comment helps, but would you consider pulling it into a named constant or a `@Value`-backed property so it's discoverable and tunable without editing this class? A fixed 1024 is a perfectly defensible default.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCache.java
```diff
@@ -103,43 +105,53 @@ public Set<TenantId> remove(TenantProfileId profileId) {
...
-        TenantProfile profile = null;
-        TenantProfileId tenantProfileId = tenantIds.get(tenantId);
-        if (tenantProfileId != null) {
-            profile = profiles.get(tenantProfileId);
-        }
+        TenantProfile profile = lookupCached(tenantId);
         if (profile == null) {
-            tenantProfileFetchLock.lock();
+            // Per-tenant lock: de-duplicates concurrent misses for the SAME tenant while allowing
+            // different tenants to resolve their profiles concurrently. A single global lock here
```
ViacheslavKlimov: This comment reads as self-contradictory. The first sentence correctly describes the new per-tenant striped lock, but the next one — 'A single global lock here serializes the synchronous cross-service fetch below across the entire process' — describes the old behavior this PR is removing. A future reader won't know which is true. Since the field-level comment (lines 47-48) already explains the rationale, this block could probably just be trimmed to the first sentence.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
```diff
@@ -269,68 +272,47 @@ private EntityTransportRateLimits createRateLimits(TenantProfile tenantProfile,
...
+
+    private <T extends EntityId> EntityTransportRateLimits getRateLimits(ConcurrentMap<T, EntityTransportRateLimits> limitsMap, TenantId tenantId,
+                                                                         T entityId, TransportLimitsType limitsType, Runnable onCreate) {
+        EntityTransportRateLimits limits = limitsMap.get(entityId);
+        if (limits == null) {
+            // Resolve the tenant profile WITHOUT holding the ConcurrentHashMap bin lock: the fetch may
+            // block on a cross-service round-trip, so it must run before computeIfAbsent's mapping function.
+            TenantProfile tenantProfile = tenantProfileCache.get(tenantId);
+            limits = limitsMap.computeIfAbsent(entityId, k -> createRateLimits(tenantProfile, limitsType));
+            if (onCreate != null) {
```
ViacheslavKlimov: The `onCreate` name suggests this runs only when a new entry is actually created, but here it fires on every cache-miss branch — including when `computeIfAbsent` returns a value another thread inserted first. It's harmless because the registration into the tenant device/gateway set is idempotent, but the name oversells the guarantee (and the old code kept the `add` inside the `computeIfAbsent` lambda, so the semantics shifted slightly). Would `onMiss` / `afterMiss` describe it more honestly?

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/TransportLimitsType.java
```diff
@@ -15,6 +15,39 @@
...
+            DefaultTenantProfileConfiguration::getTransportGatewayTelemetryMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportGatewayTelemetryDataPointsRateLimit
+    ),
+    GATEWAY_DEVICE_LIMITS(
+            DefaultTenantProfileConfiguration::getTransportGatewayDeviceMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportGatewayDeviceTelemetryMsgRateLimit,
+            DefaultTenantProfileConfiguration::getTransportGatewayDeviceTelemetryDataPointsRateLimit
+    );
+
+    private final Function<DefaultTenantProfileConfiguration, String> regularMsgRateLimit;
```
ViacheslavKlimov, line 49: Nice cleanup over the switch. One gap: nothing asserts that each constant resolves to the correct profile field — e.g. that `GATEWAY_DEVICE_LIMITS` points at the `getTransportGatewayDevice*` getters and not the plain gateway ones. The four constructor-arg blocks are near-identical, so a transposed method reference would be easy to introduce and invisible in review. A small parameterized test checking `createRateLimits(profile, type)` against the matching profile field for each `TransportLimitsType` would lock the mapping down cheaply.

common/transport/transport-api/src/test/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitServiceTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        when(tenantProfileCache.get(tenant)).thenAnswer(invocation -> {
+            bothCallersReachedFetch.countDown();
+            releaseFetch.await(5, TimeUnit.SECONDS);
+            return tenantProfile();
+        });
+
+        DefaultTransportRateLimitService service = new DefaultTransportRateLimitService(tenantProfileCache);
+
+        Runnable check = () -> service.checkLimits(tenant, null, null, 1, false);
```
ViacheslavKlimov, line 78: This is the only new test for the service, and it exercises just the tenant-limits path (`getTenantRateLimits`, where `onCreate == null`). The three other refactored paths — `getDeviceRateLimits` / `getGatewayRateLimits` / `getGatewayDeviceRateLimits` — and their `onCreate` callbacks that register the entity into the per-tenant set (so `update(tenantId)` can later propagate limit changes) go untested. A copy-paste slip in that wiring (e.g. adding to `tenantGateways` instead of `tenantGatewayDevices`) wouldn't be caught. Worth a case that resolves a device/gateway limit and then asserts a subsequent `update(tenantId)` actually reaches it.

common/transport/transport-api/src/test/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCacheTest.java
```diff
@@ -0,0 +1,136 @@
...
+        // T1 starts fetching tenantA's profile and blocks inside the cross-service round-trip.
+        Future<TenantProfile> tenantAResult = executor.submit(() -> cache.get(tenantA));
+        assertThat(tenantAFetchStarted.await(5, TimeUnit.SECONDS))
+                .as("tenantA fetch should have started").isTrue();
+
+        // T2 resolves a different tenant - it must NOT wait for tenantA's in-flight fetch.
+        // Fails today (single global lock); passes once locking is per-tenant.
+        TenantProfile tenantBProfile = CompletableFuture
+                .supplyAsync(() -> cache.get(tenantB), executor)
+                .get(2, TimeUnit.SECONDS);
```
ViacheslavKlimov, line 123: Heads-up on a small flake window: this test proves non-blocking only if `tenantA` and `tenantB` land on different stripes of `Striped.lock(1024)`. Since both are random UUIDs, there's roughly a 1/1024 (~0.1%) chance they hash to the same stripe — and when they do, `tenantB`'s resolution blocks on `tenantA`'s held lock and this `.get(2, SECONDS)` times out, failing the test spuriously. Over enough CI runs that will surface occasionally. Could you make the two tenant IDs deterministically map to distinct stripes (or otherwise remove the dependence on the random draw) so the test is reliable?

ViacheslavKlimov (commented):
## Review summary

Reviewed 5 changed files in [Fix transport tenant-profile lock convoy under cold-cache reconnect storm](https://github.com/thingsboard/thingsboard/pull/15744). Left 6 comment(s) inline.

The core concurrency fix is sound: replacing the single global `ReentrantLock` with per-tenant `Striped` locks (double-checked against the concurrent caches) correctly de-duplicates same-tenant fetches while letting different tenants resolve concurrently, and pre-fetching the tenant profile before `computeIfAbsent` removes the CHM bin-lock-over-I/O serialization. The behavior-preserving refactors (generic `getRateLimits` helper, `TransportLimitsType` enum accessors, the hoisted single profile fetch in `update(TenantId)`, and the extracted `logLimits`) all map 1:1 to the previous logic — I verified the enum's type→getter wiring matches the old switch exactly, `TransportLimitsType` has no external consumers the new constructor could break, and the `onCreate` callbacks stay idempotent so the relaxed atomicity is safe. No correctness regressions found; the inline comments are quality and test-coverage observations.

---

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCache.java
smatvienko-tb: The established idiom for per-key locking in this codebase is a `ConcurrentReferenceHashMap` (weak/soft) keyed by id with `computeIfAbsent(k -> new ReentrantLock())` — see `AbstractGatewaySessionHandler.createWeakMap()`, `DefaultTbActorSystem.actorCreationLocks`, `DefaultClusterVersionControlService.tenantRepoLocks`. `Striped` here is reasonable and arguably better (bounded, no per-key allocation, no weak-ref bookkeeping), and the eager `Striped.lock(1024)` choice is well justified in the PR description — but it's the only `Striped` use in the repo, so a one-line note on why you diverged from the house pattern would save the next maintainer a double-take.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
```diff
@@ -269,68 +272,47 @@ private EntityTransportRateLimits createRateLimits(TenantProfile tenantProfile,
...
+        );
     }
 
     private static TransportRateLimit newLimit(String config) {
         return StringUtils.isEmpty(config) ? ALLOW : new SimpleTransportRateLimit(config);
     }
 
     private EntityTransportRateLimits getTenantRateLimits(TenantId tenantId) {
-        return perTenantLimits.computeIfAbsent(tenantId, k -> createRateLimits(tenantProfileCache.get(tenantId), TENANT_LIMITS));
+        return getRateLimits(perTenantLimits, tenantId, tenantId, TENANT_LIMITS, null);
```
smatvienko-tb, line 289: Passing `tenantId` twice — `getRateLimits(perTenantLimits, tenantId, tenantId, TENANT_LIMITS, null)` — reads like it could be a copy-paste slip until you confirm the lookup key and the profile key really are the same object for the tenant case. The 5-arg helper with two same-typed ids plus a nullable trailing `Runnable` is a fair amount to hold in your head. Would it read more cleanly to keep `getTenantRateLimits` as its own one-line `computeIfAbsent` (no `onCreate`, key == tenantId) and share the generic helper only across the three device/gateway variants that actually need the tenant/entity split and the tracking-set callback?

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
```diff
@@ -269,68 +272,47 @@ private EntityTransportRateLimits createRateLimits(TenantProfile tenantProfile,
...
+    private <T extends EntityId> EntityTransportRateLimits getRateLimits(ConcurrentMap<T, EntityTransportRateLimits> limitsMap, TenantId tenantId,
+                                                                         T entityId, TransportLimitsType limitsType, Runnable onCreate) {
+        EntityTransportRateLimits limits = limitsMap.get(entityId);
+        if (limits == null) {
+            // Resolve the tenant profile WITHOUT holding the ConcurrentHashMap bin lock: the fetch may
+            // block on a cross-service round-trip, so it must run before computeIfAbsent's mapping function.
+            TenantProfile tenantProfile = tenantProfileCache.get(tenantId);
+            limits = limitsMap.computeIfAbsent(entityId, k -> createRateLimits(tenantProfile, limitsType));
+            if (onCreate != null) {
+                onCreate.run();
```
smatvienko-tb: `onCreate` reads as "run when this call created the entry", but it actually runs for every caller that observed `limits == null` at the top — including a thread that lost the `computeIfAbsent` race and got an existing value back. It's safe today only because all three callbacks are idempotent set-adds (`getTenantDevices(tenantId).add(deviceId)`); a future non-idempotent body would be silently run more than once, and there's also a marginally wider window now between the entry appearing in `limitsMap` and the device showing up in the tracking set. Consider renaming to `onMiss`/`afterLookup`, or moving the tracking add back inside the mapping function (as the original code had it) so it fires exactly once per real creation — at minimum a comment that idempotency is required.

common/transport/transport-api/src/test/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCacheTest.java
```diff
@@ -0,0 +1,136 @@
...
+        executor = Executors.newCachedThreadPool();
+    }
+
+    @AfterEach
+    void tearDown() {
+        executor.shutdownNow();
+    }
+
+    @Test
+    void fetchForOneTenantDoesNotBlockResolutionOfAnotherTenant() throws Exception {
```
smatvienko-tb, line 96: Nice, targeted test for the cross-tenant win. Two gaps worth closing: (1) the other half of the contract — that concurrent misses for the **same** tenant dedupe to a single `getEntityProfile` round-trip — isn't covered, yet that dedup is the entire reason the per-tenant lock still exists (vs. going lock-free); a sibling test firing N concurrent `get(tenantA)` against a gated mock and asserting `verify(transportService, times(1)).getEntityProfile(any())` would lock it down. (2) `tenantA`/`tenantB` are random UUIDs (L56-57), so with `Striped.lock(1024)` there's a ~1/1024 chance they land on the same stripe, in which case T2 would block on T1 and this test fails spuriously — consider deriving two tenants known to fall on different stripes to make it deterministic.

common/transport/transport-api/src/test/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitServiceTest.java
smatvienko-tb, line 78: This drives only the tenant-level path (`gatewayId`/`deviceId` null), which is the single `getRateLimits` variant that passes a `null` `onCreate`. The device/gateway variants — whose `() -> getTenantDevices(...).add(...)` callbacks the refactor newly moved outside `computeIfAbsent` — get no focused assertion that the tracking set is actually populated (e.g. after a device `checkLimits`, the device shows up in `getTenantDevices(tenant)` so a later `update(tenantId)` re-applies its limits). It's exercised end-to-end elsewhere, so this is a gap rather than a hole, but given the `onCreate` indirection is new and subtle, a small direct assertion would be cheap insurance.

smatvienko-tb (commented):
## Review summary

Reviewed 5 changed files in [Fix transport tenant-profile lock convoy under cold-cache reconnect storm](https://github.com/thingsboard/thingsboard/pull/15744). Left 5 comments inline.

This PR correctly implements two of the three documented root causes: per-tenant `Striped.lock(1024)` in `DefaultTransportTenantProfileCache` (removes the global `tenantProfileFetchLock` convoy), and pre-fetching the tenant profile before `computeIfAbsent` in `DefaultTransportRateLimitService` (removes the CHM-bin-lock-across-I/O layer). I traced the enum / `getRateLimits` / `update` refactors — they are behavior-preserving.

Against the production thread dump that motivated this PR: every BLOCKED/WAITING thread in the convoy is waiting, directly or transitively, on either the global lock `<45d2927d>` or a CHM bin lock held across I/O (`<6c2f5e47>` / `<7c4ce172>`) — both eliminated here. The absent transport-notifications consumer is a downstream effect of `main` being wedged in the `@AfterStartUp(order = DISCOVERY_SERVICE = 2)` listener inside `recalculatePartitions`; removing the global lock lets that listener complete so the `order = TRANSPORT_SERVICE = MAX-1000` `start()` runs and launches the consumer. So the PR resolves the captured stall.

Two structural items remain, both in `Additional findings`: the hardcoded callback-pool size (an uncontrollable bottleneck that survives this fix), and the deliberately un-addressed third root cause in `HashPartitionService`.

## Additional findings

These are about code outside this PR's diff, but directly relevant to the incident it fixes.

- **`common/transport/transport-api/.../service/DefaultTransportService.java:201` — make the callback pool size configurable in this PR.** `transportCallbackExecutor = ThingsBoardExecutors.newWorkStealingPool(20, getClass())` is hardcoded. Fix 1+2 remove the *serialization*, but these 20 threads still each block on the real `getEntityProfile().get()` round-trip, so a cold-cache reconnect storm is still capped at ~20 concurrent tenant-profile fetches — an uncontrollable bottleneck that can't be tuned without a rebuild. Suggest a `@Value("${transport.callback_thread_pool_size:20}")` field (default 20 → backward compatible) passed to `newWorkStealingPool(...)`, with the property + a `TB_TRANSPORT_CALLBACK_THREAD_POOL_SIZE` env override and a description comment added to each `transport/*/src/main/resources/tb-*-transport.yml` and `application/src/main/resources/thingsboard.yml` (this trips the config-files CI check, which requires the description). The deeper fix — async profile resolution so the pool isn't blocked at all — can be a follow-up.

- **`common/queue/.../discovery/HashPartitionService.java` (third root cause) is intentionally left untouched — the residual is narrowed, not removed.** The consumer launch (`@AfterStartUp(order = MAX-1000)`) is still serialized on `main` behind `ZkDiscoveryService.onApplicationEvent (order = 2)` → `synchronized recalculatePartitions` → `resolveByPartitionIdx:715`, which still does serial, synchronous, per-tenant routing round-trips while holding the `HashPartitionService` / `ZkDiscoveryService` monitors. After Fix 1 this no longer dead-stalls, but under a severe tb-core outage (each routing fetch timing out at 10s × N isolated-RE tenants) `main` can still spend a long time before the consumer launches. The PR's rationale for not routing `:715` through the cached `getRoutingInfo` helper is valid — `tenantRoutingInfoMap` is invalidated only on tenant update/delete + queue change (`:231`/`:592`), not on profile update, so the cache would risk misrouting an isolated tenant. But pre-resolving the distinct tenant ids *outside* the `synchronized` block (then running pure partition math inside) avoids both the staleness risk and the under-monitor blocking, and decoupling the consumer launch from discovery is independent of that. Worth at least a tracked follow-up.

- **`DefaultTransportService.java:266` — `transportApiRequestTemplate.send(protoMsg).get()` is still unbounded.** It relies on the template's internal `max_requests_timeout` (10s). An explicit `.get(maxRequestTimeoutMs, MILLISECONDS)` would make the failure explicit and non-fatal (defense-in-depth). Optional.

- **`DefaultTransportRateLimitService.java:111-114` — `update(TenantProfileUpdateResult)` still calls `update.getProfile()` four times**, while the sibling `update(TenantId)` was just consolidated to a single `profile` local. Only a getter, but the asymmetry stands out now that the two sit next to each other; `var profile = update.getProfile();` once would make the pair consistent.

- **Minor:** in `DefaultTransportTenantProfileCache.fetchAndCacheTenantProfile`, two tenants sharing a profile id can now race on `profiles.put` (previously serialized by the global lock). The `existingProfile` reuse still converges and the data is equivalent, so this is benign — just no longer fully de-duplicated to a single instance.

- **Process:** PR base is `lts-4.2`; per the incident notes the defective code is also on `master`, so a forward-port / master PR will be needed.

---

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCache.java
ViacheslavKlimov: Fixed in a1bf69cd276 — extracted the stripe count to the `TENANT_PROFILE_FETCH_LOCK_STRIPES` constant.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCache.java
ViacheslavKlimov: Fixed in a1bf69cd276 — trimmed the stale global-lock sentence; only the per-tenant description remains (the field comment on lines 47-48 already covers the rationale).

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
ViacheslavKlimov: Fixed in a1bf69cd276 — renamed `onCreate` to `onMiss` and added a comment that it runs on every observed miss (not only real creation), so the callback must be idempotent.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/TransportLimitsType.java
ViacheslavKlimov: Fixed in a1bf69cd276 — added a parameterized test (`eachLimitsTypeReadsItsOwnProfileFields`) that asserts each `TransportLimitsType` resolves to its own profile getters, so a transposed method reference (e.g. `GATEWAY_DEVICE_LIMITS` pointing at the plain gateway getters) would fail the build.

common/transport/transport-api/src/test/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitServiceTest.java
ViacheslavKlimov: Fixed in a1bf69cd276 — added `profileUpdateReachesEntityTrackedDuringFirstCheck` (parameterized over device / gateway / gateway-device): it resolves a permissive limit, then tightens the profile and pushes `update(tenantId)`, asserting the freshly merged limit reaches the entity — which only holds if the `onMiss` callback registered it into the per-tenant tracking set.

common/transport/transport-api/src/test/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCacheTest.java
ViacheslavKlimov: Fixed in a1bf69cd276 — `tenantB` is now derived to map to a different stripe than `tenantA` (using a same-size probe `Striped`, whose mapping is identical to the cache's), so the test no longer depends on the random draw.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCache.java
ViacheslavKlimov: Decided to keep it as is: `Striped` is a standard Guava concurrency primitive rather than a bespoke pattern, and it's strictly better suited here — bounded, no per-key allocation, no weak-ref bookkeeping. The existing field comment already explains the eager-array sizing, so I'd rather not add a justification for choosing a well-known utility over the older idiom.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
ViacheslavKlimov: Kept the unified helper: its role isn't only dedup — it performs the blocking `tenantProfileCache.get(tenantId)` *before* `computeIfAbsent` for all four paths, which is the bin-lock-over-I/O fix this PR is about. Splitting the tenant case into a one-line `computeIfAbsent` would either move the fetch back inside the mapping function (reintroducing the convoy) or duplicate that pre-fetch logic. The `tenantId, tenantId` is intentional — for the tenant case the lookup entity *is* the tenant.

common/transport/transport-api/src/main/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitService.java
ViacheslavKlimov: Fixed in a1bf69cd276 — renamed `onCreate` to `onMiss` and added a comment that it runs on every observed miss, including callers that lost the `computeIfAbsent` race and got an existing value back, so the callback must be idempotent.

common/transport/transport-api/src/test/java/org/thingsboard/server/common/transport/service/DefaultTransportTenantProfileCacheTest.java
ViacheslavKlimov: Fixed in a1bf69cd276 — added `concurrentMissesForSameTenantDedupeToSingleFetch`, which fires N concurrent `get(tenantA)` against a gated mock and asserts `verify(transportService, times(1)).getEntityProfile(any())`. Also pinned the cross-tenant test's two tenants to distinct stripes (via a same-size probe `Striped`) to remove the ~1/1024 flake.

common/transport/transport-api/src/test/java/org/thingsboard/server/common/transport/limits/DefaultTransportRateLimitServiceTest.java
ViacheslavKlimov: Fixed in a1bf69cd276 — added `profileUpdateReachesEntityTrackedDuringFirstCheck` (parameterized over device / gateway / gateway-device): after a `checkLimits` resolves the entity, a subsequent `update(tenantId)` with a tightened profile must reach it, which only holds if the `onMiss` callback added it to the correct per-tenant tracking set.

ViacheslavKlimov:
> **`DefaultTransportService.java:201` — make the callback pool size configurable in this PR.** `transportCallbackExecutor = ThingsBoardExecutors.newWorkStealingPool(20, getClass())` is hardcoded… ([from review](https://github.com/thingsboard/thingsboard/pull/15744#pullrequestreview-4449952509))

Done in a1bf69cd276 — added `transport.callback_thread_pool_size` (`@Value`, default 20) wired into `newWorkStealingPool(...)`, with a `TB_TRANSPORT_CALLBACK_THREAD_POOL_SIZE` env override and a described property in `thingsboard.yml` and each `tb-*-transport.yml`. The deeper async-resolution fix remains a follow-up.

ViacheslavKlimov:
> **`common/queue/.../discovery/HashPartitionService.java` (third root cause) is intentionally left untouched — the residual is narrowed, not removed.** … Worth at least a tracked follow-up. ([from review](https://github.com/thingsboard/thingsboard/pull/15744#pullrequestreview-4449952509))

Acknowledged — this third root cause is intentionally out of scope for this PR: it narrows the residual rather than removing it, and the rationale for not reusing the (profile-stale) routing cache stands. Tracking the improvement — pre-resolving the distinct tenant ids outside the `synchronized` block and decoupling the consumer launch from discovery — as a separate follow-up.

ViacheslavKlimov:
> **`DefaultTransportService.java:266` — `transportApiRequestTemplate.send(protoMsg).get()` is still unbounded.** … An explicit `.get(maxRequestTimeoutMs, MILLISECONDS)` … Optional. ([from review](https://github.com/thingsboard/thingsboard/pull/15744#pullrequestreview-4449952509))

Leaving as-is: the request template already enforces its internal `max_requests_timeout` (10s), so the call is bounded. The identical `send(...).get()` pattern is used by sibling methods (e.g. `getQueueRoutingInfo` just below), so adding an explicit timeout to only this one would be inconsistent — a consistent sweep can be a separate change.

ViacheslavKlimov:
> **`DefaultTransportRateLimitService.java:111-114` — `update(TenantProfileUpdateResult)` still calls `update.getProfile()` four times**, while the sibling `update(TenantId)` was just consolidated to a single `profile` local. ([from review](https://github.com/thingsboard/thingsboard/pull/15744#pullrequestreview-4449952509))

Done in a1bf69cd276 — consolidated to a single `var profile = update.getProfile();` reused across the log line and the four `createRateLimits(...)` calls, matching the sibling `update(TenantId)`.

ViacheslavKlimov:
> **Minor:** in `DefaultTransportTenantProfileCache.fetchAndCacheTenantProfile`, two tenants sharing a profile id can now race on `profiles.put` … this is benign … just no longer fully de-duplicated to a single instance. ([from review](https://github.com/thingsboard/thingsboard/pull/15744#pullrequestreview-4449952509))

Acknowledged — confirmed benign: the `existingProfile` reuse means racing threads converge on equivalent profile data; only the guarantee of a single shared instance is relaxed, and nothing depends on that. No change needed.

PR #15753, opened 2026-06-10
sparkplug_fix_bug_for_unique-device-names

nickAS21:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
+++ b/application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
@@ -400,7 +400,7 @@ public class DefaultTransportApiService implements TransportApiService {
                     gateway.getTenantId(),
                     gateway.getId(),
                     existingDevice.getId(),
-                    "Created",
+                    EntityRelation.CONTAINS_TYPE,
                     RelationTypeGroup.COMMON
             );
         } catch (Exception e) {
@@ -448,8 +448,8 @@ public class DefaultTransportApiService implements TransportApiService {
         Device device = new Device();
         device.setTenantId(tenantId);
         device.setName(requestMsg.getDeviceName());
-        if (requestMsg.getIsSparkplug()) {
-            if (topicPath.length == 3) device.setLabel(topicPath[2]);
+        if (requestMsg.getIsSparkplug() && topicPath.length == 3) {
+            device.setLabel(topicPath[2]);
         }
         device.setType(requestMsg.getDeviceType());
         device.setCustomerId(gateway.getCustomerId());
@@ -462,7 +462,7 @@ public class DefaultTransportApiService implements TransportApiService {
         device = deviceService.saveDevice(device);
         relationService.saveRelation(
                 tenantId,
-                new EntityRelation(gateway.getId(), device.getId(), "Created")
+                new EntityRelation(gateway.getId(), device.getId(), EntityRelation.CONTAINS_TYPE)
         );
         return device;
     }
--- a/application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
+++ b/application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
@@ -396,7 +396,6 @@ public class DefaultTransportApiService implements TransportApiService {
         // Security check: verify that the device was created by this gateway
         boolean isRelated = false;
         try {
-            // Security check: verify that the device was originally created by this gateway
             isRelated = relationService.checkRelation(
                     gateway.getTenantId(),
                     gateway.getId(),
@@ -405,37 +404,34 @@ public class DefaultTransportApiService implements TransportApiService {
                     RelationTypeGroup.COMMON
             );
         } catch (Exception e) {
-            // Log the error from the relation service but return null to allow potential recovery
-            log.error("[{}] Error checking relation for device {}", gateway.getId(), existingDevice.getId(), e);
-            return null;
+            log.error("[{}] Failed checking relation for device {}",
+                    gateway.getId(),
+                    existingDevice.getId(),
+                    e);
+            throw new RuntimeException(
+                    "Failed checking relation for device " + existingDevice.getId(),
+                    e
+            );
         }
 
-        // If the device is found but not related to this gateway, it's a security breach
+        // If the device is found but not related to this gateway
         if (!isRelated) {
-            log.error("[{}] Security breach attempt! Gateway tried to rename device [{}] without 'Created' relation.",
-                    gateway.getId(), existingDevice.getId());
-            // Throwing exception to halt the entire connection process
-            throw new RuntimeException("Security breach attempt! Unauthorized device rename.");
+            log.debug("[{}] Device [{}] exists but is not related to gateway. " +
+                            "Skipping rename and allowing creation of Sparkplug device [{}].",
+                    gateway.getId(),
+                    existingDevice.getId(),
+                    requestMsg.getDeviceName());
+
+            return null;
         }
 
         // Logic for renaming the device if it's related and no naming conflicts exist
         boolean changed = false;
         String newName = requestMsg.getDeviceName();
-
         if (!newName.equals(existingDevice.getName())) {
-            // Check if the new name is already taken by another device
-            Device conflictDevice = deviceService.findDeviceByTenantIdAndName(gateway.getTenantId(), newName);
-
-            if (conflictDevice != null) {
-                log.warn("[{}] Cannot rename device [{}] to [{}]: name already exists!",
-                        gateway.getId(), existingDevice.getId(), newName);
-                return existingDevice;
-            }
-
             existingDevice.setName(newName);
 
-            // Update label only if it's empty to avoid overwriting user changes
-            if (existingDevice.getLabel() == null || existingDevice.getLabel().isEmpty()) {
+            if (StringUtils.isEmpty(existingDevice.getLabel())) {
                 existingDevice.setLabel(deviceId);
             }
 
--- a/application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
+++ b/application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
@@ -396,24 +396,13 @@ public class DefaultTransportApiService implements TransportApiService {
 
         // Security check: verify that the device was created by this gateway
         boolean isRelated = false;
-        try {
-            isRelated = relationService.checkRelation(
-                    gateway.getTenantId(),
-                    gateway.getId(),
-                    existingDevice.getId(),
-                    GATEWAY_CREATED_RELATION,
-                    RelationTypeGroup.COMMON
-            );
-        } catch (Exception e) {
-            log.error("[{}] Failed checking relation for device {}",
-                    gateway.getId(),
-                    existingDevice.getId(),
-                    e);
-            throw new RuntimeException(
-                    "Failed checking relation for device " + existingDevice.getId(),
-                    e
-            );
-        }
+        isRelated = relationService.checkRelation(
+                gateway.getTenantId(),
+                gateway.getId(),
+                existingDevice.getId(),
+                GATEWAY_CREATED_RELATION,
+                RelationTypeGroup.COMMON
+        );
 
         // If the device is found but not related to this gateway
         if (!isRelated) {
--- a/application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
+++ b/application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
@@ -141,6 +141,7 @@ public class DefaultTransportApiService implements TransportApiService {
     private final OtaPackageService otaPackageService;
     private final OtaPackageDataCache otaPackageDataCache;
     private final QueueService queueService;
+    public static final String GATEWAY_CREATED_RELATION = "Created";
 
     private final ConcurrentMap<String, ReentrantLock> deviceCreationLocks = new ConcurrentReferenceHashMap<>(16, ConcurrentReferenceHashMap.ReferenceType.WEAK);
 
@@ -400,7 +401,7 @@ public class DefaultTransportApiService implements TransportApiService {
                     gateway.getTenantId(),
                     gateway.getId(),
                     existingDevice.getId(),
-                    EntityRelation.CONTAINS_TYPE,
+                    GATEWAY_CREATED_RELATION,
                     RelationTypeGroup.COMMON
             );
         } catch (Exception e) {
@@ -462,7 +463,7 @@ public class DefaultTransportApiService implements TransportApiService {
         device = deviceService.saveDevice(device);
         relationService.saveRelation(
                 tenantId,
-                new EntityRelation(gateway.getId(), device.getId(), EntityRelation.CONTAINS_TYPE)
+                new EntityRelation(gateway.getId(), device.getId(), GATEWAY_CREATED_RELATION)
         );
         return device;
     }
--- a/application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
+++ b/application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
@@ -402,7 +402,7 @@ public abstract class AbstractMqttV5ClientSparkplugTest extends AbstractMqttInte
 
         String expectedFullPath = groupId + DEVICE_NAME_SPLIT_SEPARATOR + edgeNode + DEVICE_NAME_SPLIT_SEPARATOR + strangerName;
         await().atMost(30, TimeUnit.SECONDS).untilAsserted(() ->
-                doGet("/api/tenant/devices?deviceName=" + expectedFullPath, Device.class, status().isNotFound())
+                doGet("/api/tenant/devices?deviceName=" + expectedFullPath, Device.class, status().isOk())
         );
     }
 
--- a/application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
+++ b/application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
@@ -65,6 +65,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 import static org.thingsboard.common.util.JacksonUtil.newArrayNode;
+import static org.thingsboard.server.service.transport.DefaultTransportApiService.GATEWAY_CREATED_RELATION;
 import static org.thingsboard.server.transport.mqtt.util.sparkplug.MetricDataType.Bytes;
 import static org.thingsboard.server.transport.mqtt.util.sparkplug.MetricDataType.Int16;
 import static org.thingsboard.server.transport.mqtt.util.sparkplug.MetricDataType.Int32;
@@ -128,8 +129,7 @@ public abstract class AbstractMqttV5ClientSparkplugTest extends AbstractMqttInte
         Device device1 = createDevice(deviceName1, deviceProfile.getName(), false);
 
         // 2. Establish 'Created' relation so the transport identifies this gateway as the owner
-        String relationType = "Created";
-        EntityRelation relation1 = createFromRelation(savedGateway, device1, relationType);
+        EntityRelation relation1 = createFromRelation(savedGateway, device1, GATEWAY_CREATED_RELATION);
         doPost("/api/relation", relation1).andExpect(status().isOk());
 
         // 3. Create the second device with a full-path name
@@ -137,7 +137,7 @@ public abstract class AbstractMqttV5ClientSparkplugTest extends AbstractMqttInte
         Device device2 = createDevice(deviceName2, deviceProfile.getName(), false);
 
         // 4. Establish 'Created' relation for the second device as well
-        EntityRelation relation2 = createFromRelation(savedGateway, device2, relationType);
+        EntityRelation relation2 = createFromRelation(savedGateway, device2, GATEWAY_CREATED_RELATION);
         doPost("/api/relation", relation2).andExpect(status().isOk());
 
     }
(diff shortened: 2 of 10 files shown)


application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -394,12 +397,12 @@ protected void unauthorizedRenameAttempt_Test() throws Exception {
...
         // Unauthorized gateway attempts to rename the device via Sparkplug topic
-        client.publish(TOPIC_ROOT_SPB_V_1_0 + "/" + groupId + "/DBIRTH/" + edgeNode + "/" + strangerName,
+        client.publish(TOPIC_ROOT_SPB_V_1_0 + TOPIC_SPLIT_SEPARATOR + groupId + "/DBIRTH/" + edgeNode + TOPIC_SPLIT_SEPARATOR + strangerName,
                 payload.build().toByteArray(), 0, false);
 
-        String expectedFullPath = groupId + ":" + edgeNode + ":" + strangerName;
+        String expectedFullPath = groupId + DEVICE_NAME_SPLIT_SEPARATOR + edgeNode + DEVICE_NAME_SPLIT_SEPARATOR + strangerName;
         await().atMost(30, TimeUnit.SECONDS).untilAsserted(() ->
-                doGet("/api/tenant/devices?deviceName=" + expectedFullPath, Device.class, status().isNotFound())
+                doGet("/api/tenant/devices?deviceName=" + expectedFullPath, Device.class, status().isOk())
```
ViacheslavKlimov, line 410: This assertion flip (from `isNotFound` to `isOk`) correctly reflects the new behavior — an unauthorized gateway no longer aborts; it just gets its own namespaced device created. But notice what's now *not* covered: this wired test only checks that the full-path device appears. The actual security property — that the original, unrelated device is left untouched and keeps its ID (i.e. it wasn't hijacked/renamed) — is asserted only in `unauthorizedRenameAttemptBad_Test`, which still isn't wired to any `@Test`. So after this change there's no running test that verifies a gateway can't take over a device it didn't create. Could you either wire up `unauthorizedRenameAttemptBad_Test` or fold its "original device ID unchanged" assertion into this one, so the security guarantee is actually exercised?

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -438,25 +441,25 @@ protected void concurrentFirstMessageRegistration_Test() throws Exception {
         String concurrentDeviceName = "concurrent_device";
         clientWithCorrectNodeAccessTokenWithNDEATH();
 
-        java.util.concurrent.ExecutorService executor = java.util.concurrent.Executors.newFixedThreadPool(threadCount);
+        java.util.concurrent.ExecutorService executor = newFixedThreadPool(threadCount);
```
ViacheslavKlimov: **Still unresolved:** the static import for `newFixedThreadPool` cleaned up the factory call, but the variable type is still `java.util.concurrent.ExecutorService` fully-qualified. A plain `import java.util.concurrent.ExecutorService;` and `ExecutorService executor = ...` would finish this off.

ViacheslavKlimov (commented):
## Re-review summary

Re-reviewed [sparkplug_fix_bug_for_unique-device-names](https://github.com/thingsboard/thingsboard/pull/15753) — this is the follow-up to my [review on #14987](https://github.com/thingsboard/thingsboard/pull/14987#pullrequestreview-4458114884), verified against the 15 findings there.

The substantive correctness findings are all addressed, and a couple were fixed exactly as suggested (the `seedLegacyAndFullPathDevices()` extraction and the `onDeviceConnect` two-arg overload). One new gap was introduced by the fix: the unauthorized-rename security assertion now lives only in an unwired test method, so it isn't actually exercised anymore.

| Status | Count |
|--------|-------|
| ✅ Resolved | 9 |
| ❌ Unresolved | 6 |

Also found 1 new issue in the fix commits, commented inline.

### Finding details

- ✅ **Security-breach throw blocks legitimate creation** — the throw is gone; when the existing device isn't related to the gateway it now logs at debug and returns null, falling through to create the namespaced device. Exactly the suggested fix.
- ✅ **Dead/redundant conflict re-query** — the `conflictDevice` lookup and its branch were removed.
- ✅ **Relation-check exception silently created a duplicate** — the `catch` now rethrows instead of returning null, so a transient failure propagates rather than spawning a parallel device.
- ✅ **Ineffective outer try/catch in `SparkplugNodeSessionHandler`** — the misleading outer `catch` was removed; the inner one around `setDeviceBirthMetrics` remains.
- ✅ **Duplicate "Security check" comment** — removed.
- ✅ **Raw `:` / `/` literals in tests** — now use `DEVICE_NAME_SPLIT_SEPARATOR` / `TOPIC_SPLIT_SEPARATOR` (no shared helper extracted, but the intent is met).
- ✅ **`// Restored to 40s as requested` comment** — removed. (The `// //` double-slash leftover at `AbstractMqttV5ClientSparkplugTest.java:300` is still there.)
- ✅ **Boolean flag on `beforeSparkplugTest`** — replaced with a no-arg `beforeSparkplugTest()` plus a separate `seedLegacyAndFullPathDevices()` helper.
- ✅ **Boolean positional param on `onDeviceConnect`** — a two-arg `onDeviceConnect(name, type)` overload was added and the non-Sparkplug callers reverted to it.
- ❌ **`"Created"` relation type still a magic string** — still hardcoded at `DefaultTransportApiService.java:403` and `:465` (and `AbstractMqttV5ClientSparkplugTest.java`). Not addressed.
- ❌ **Nested single-line `if`s in `createNewDevice`** — `DefaultTransportApiService.java:451` unchanged (the method wasn't touched).
- ❌ **`getNodeDeviceNameAllPath()` returns `""` for host-app topics** — `SparkplugTopic.java` isn't part of this PR, so it remains in master.
- ❌ **Trailing-whitespace-only line in `SparkplugDeviceSessionContext`** — that file isn't part of this PR; the whitespace remains in master.
- ❌ **Near-duplicate + dead `unauthorizedRenameAttemptBad_Test`** — still not wired to any `@Test` (see the new inline comment — this now matters for coverage).
- ❌ **Fully-qualified `ExecutorService`** — partially: a static import was added for `newFixedThreadPool`, but the variable type at `AbstractMqttV5ClientSparkplugTest.java:444` is still `java.util.concurrent.ExecutorService`. Commented inline.

---

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
nickAS21: Ok

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
nickAS21: Ok

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -396,46 +396,42 @@ private Device tryRenameSparkplugDevice(GetOrCreateDeviceFromGatewayRequestMsg r
         // Security check: verify that the device was created by this gateway
         boolean isRelated = false;
         try {
-            // Security check: verify that the device was originally created by this gateway
             isRelated = relationService.checkRelation(
                     gateway.getTenantId(),
                     gateway.getId(),
                     existingDevice.getId(),
-                    "Created",
+                    EntityRelation.CONTAINS_TYPE,
```
ViacheslavKlimov: I think this change breaks the legacy-device migration, which is the whole point of the feature. My earlier comment was just "extract `\"Created\"` into a constant" — but `EntityRelation.CONTAINS_TYPE` is `"Contains"`, so this swaps the actual relation *value*, not just the literal.

The consequence: the rename path only fires for a device that (a) still has the short legacy name and (b) was created by this gateway. Every such device was created by a *prior* version, whose `createNewDevice` wrote a `"Created"` relation (that's what master does today, two lines… well, at line 465 before this PR). This `checkRelation` now looks for `"Contains"`, finds nothing, returns false, and falls through to `createNewDevice` — so the legacy device is **duplicated** under the full-path name instead of being renamed, and its telemetry history splits across two devices. A device created by *this* version always already has the full-path name, so it never needs renaming — meaning the rename branch is effectively dead for every real legacy device.

The test doesn't catch it: `seedLegacyAndFullPathDevices()` still seeds the relation as `"Created"` (line 131), which no longer matches this `"Contains"` check, and `connectClientWithCorrectAccessTokenWithNDEATHDevicesCreatingBefore_Test` only asserts the resulting `label` — which is `"deviceId_1"` whether the device was renamed *or* freshly created — so it passes either way and never verifies that the original short-named device was actually renamed away.

Suggest keeping the value `"Created"` and extracting a local constant for it (e.g. `private static final String GATEWAY_CREATED_RELATION = "Created";`) used by both `checkRelation` here and `saveRelation` in `createNewDevice`, rather than reusing `CONTAINS_TYPE`. And it'd be worth tightening the migration test to assert the legacy short-named device no longer exists after the rename (not just the label), so this kind of swap can't pass silently.

ViacheslavKlimov (commented):
## Re-review summary

Checked the latest commits (`b6d1e25..dfbd49c`) against my [previous re-review](https://github.com/thingsboard/thingsboard/pull/15753#pullrequestreview-4474611195).

Both inline items I raised are resolved:

- ✅ **Security assertion was in dead code** — `unauthorizedRenameAttempt_Test` now captures the stranger device up front and asserts its ID is unchanged after the DBIRTH, so the "can't hijack someone else's device" property is actually exercised by a running test now. (Side note: `unauthorizedRenameAttemptBad_Test` is now fully redundant with that and could be deleted.)
- ✅ **`ExecutorService` fully-qualified** — `import java.util.concurrent.ExecutorService;` added and the type reference cleaned up. Also a nice bonus pass converting `Assert.*` to static imports.
- ✅ **Nested `if`s in `createNewDevice`** — collapsed to a single condition.

But the fix for the `"Created"` magic-string note introduced a regression that I think is more serious than the original nit — flagged inline below. **It needs another look before merge.**

| Status | Count |
|--------|-------|
| ✅ Resolved | 4 |
| ⚠️ New regression | 1 |

---

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
nickAS21: Ok

ViacheslavKlimov (commented):
## Full review of the Sparkplug unique-device-names feature

First, the follow-ups: the `Created` → `Contains` regression I flagged is fixed (commit `-5` introduced `GATEWAY_CREATED_RELATION = "Created"`, used by both the check and the save and shared with the test seed) — good. Both earlier inline items are also resolved.

Since the original feature merged via #14987 and this PR only carries the fixes, I read the whole feature in the worktree, not just the diff. The logic itself now looks correct. **The thing I'd push on hardest is the tests.** Three of the new integration tests assert weaker properties than their names claim and would pass even if the behavior under test were broken or removed — which is precisely how the `Contains`/`Created` swap got through CI last round. Details inline.

Findings on lines this PR touched are inline; findings in feature code that's unchanged here but still ships are in **Additional findings** below.

## Additional findings

These are in feature code outside this PR's diff (mostly from #14987) but still shipping with it.

- **`AbstractGatewaySessionHandler.java:304` (`onDeviceConnectSparkplug`) — the "legacy fallback" cache probe is dead.** The secondary lookup `devices.get(shortName)` can never hit: the per-session in-memory `devices` map is only ever populated under the **full-path** name (the only writer is `devices.putIfAbsent(deviceName, …)` at line 343, and in the Sparkplug flow `deviceName` is always the `fullPath` from `onDeviceConnect(fullPath, …)`). A device "migrated from an older version" lives in the DB, not in this fresh per-session map, so the short-name branch is unreachable. Either it's doing nothing, or the intended legacy-routing isn't actually happening here (the real reconciliation is server-side in `tryRenameSparkplugDevice`). Worth removing or rethinking — and the comment claiming it "supports devices migrated from older versions" is misleading.
- **`SparkplugTopic.java:333` (`getNodeDeviceNameAllPath`) — returns `""` for host-application topics.** Callers feed the result straight into `checkDeviceName(...)` / `devices.get(...)`, so an empty device name is only caught downstream by a bare `RuntimeException("Device name is empty!")`. It's not reached today (only device topics call it), but it's a latent landmine; returning `null`/`Optional` or asserting the precondition would be safer and is less surprising next to `getNodeDeviceName()` which returns a real value.
- **`AbstractMqttV5ClientSparkplugTest.java:331` (`renameCollisionWhenTargetNameAlreadyExists_Test`) — asserts nothing meaningful.** Both devices are created synchronously *before* the DBIRTH is published, so the `await(... oldDevice != null && conflictDevice != null)` condition is already true at `t=0` and can succeed before the message is even processed. It would pass even if the rename/collision branch were a no-op or threw and was swallowed. To actually cover the collision, assert the short-named device kept its identity (id unchanged) and was not merged into the pre-existing full-path device.
- **`DefaultTransportApiService.java:391` — extra DB round-trips per new Sparkplug device.** On the create path, `findOrCreateDevice` now does the full-path lookup (367), then the short-name probe (391), then `checkRelation` — three sequential calls before creation, for every device a Sparkplug gateway ever births (the short-name probe essentially never hits post-migration). Fine as a one-time cost, but worth a comment, and ideally skip the probe when there's nothing to migrate.
- **`DefaultTransportApiService.java:335` (`handle`) — lock key vs. mutated entity mismatch (low confidence).** The lock is taken on `requestMsg.getDeviceName()` (the full path), but `tryRenameSparkplugDevice` reads-modifies-writes the **short-named** device. Two gateways reporting the same `deviceId` under different group/node would take different locks yet touch the same legacy row. Unlikely in practice, but the read-modify-write isn't serialized on the entity it actually mutates.
- **`AbstractGatewaySessionHandler.java:869` — the `msgId <= 0` handling is split across two blocks.** The block at ~860 already writes an `UNSPECIFIED_ERROR` PubAck and logs "Wrong msg id" on the last success when `msgId <= 0`; then a separate `if (msgId <= 0)` re-evaluates the same condition to either log "allowed" (Sparkplug) or close the session. So a Sparkplug `msgId=0` success both emits an error PubAck *and* logs that the id is allowed. Worth unifying the Sparkplug-vs-standard and ack-vs-close decisions in one place.
- **`SparkplugNodeSessionHandler.java` (`setNodeBirthMetrics`) — fully-qualified `java.util.List` / `SparkplugBProto` in the signature** while the file imports and uses both by short name everywhere else.

---

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -290,10 +296,10 @@ protected void connectClientWithCorrectAccessTokenWithNDEATHDevicesCreatingBefor
                     return device2.get() != null;
                 });
         devices.add(device2.get());
-        Assert.assertEquals(cntDevices, devices.size());
+        assertEquals(cntDevices, devices.size());
         state_ONLINE_ALL (devices, calendar.getTimeInMillis());
         // Without full topic: as it was in the old version. When deviceId is updated to full theme, Label is also updated to old deviceId
-        Assert.assertEquals(deviceIdNameLabel1, device1.get().getLabel());
+        assertEquals(deviceIdNameLabel1, device1.get().getLabel());
```
ViacheslavKlimov, line 302: This is the migration test, and it doesn't actually prove a *rename*. It awaits a device under the full-path name `deviceName1` and asserts the label — but a fresh duplicate created under the same full path would have the exact same name and the same label `deviceId_1`, so this passes whether the legacy device was renamed or simply re-created alongside the lingering short-named row. That's the precise gap that let the `Contains`/`Created` swap through CI last round.

To make it real: capture the seeded device's id in `seedLegacyAndFullPathDevices()` (store it on a field), then here assert `device1.get().getId()` equals that seeded id (proving it's the same row, renamed) **and** that `doGet("/api/tenant/devices?deviceName=" + (deviceId + "_1"))` now returns null/404 (proving the old short name is gone). As written the test is weaker than its name implies.

(Minor, same method: the `// //` double-slash on the line below is a leftover review artifact.)

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -438,25 +451,25 @@ protected void concurrentFirstMessageRegistration_Test() throws Exception {
...
                             payload.build().toByteArray(), 0, false);
                 } catch (Exception e) {
                     log.error("Concurrent publish failed", e);
                 }
             });
         }
 
-        String expectedName = groupId + ":" + edgeNode + ":" + concurrentDeviceName;
+        String expectedName = groupId + DEVICE_NAME_SPLIT_SEPARATOR + edgeNode + DEVICE_NAME_SPLIT_SEPARATOR + concurrentDeviceName;
         await("Wait for concurrent registration result")
```
ViacheslavKlimov, line 471: This fires 5 parallel DBIRTHs but only asserts that *a* device with the expected name eventually exists. Uniqueness on `(tenantId, name)` is already enforced by the DB constraint, so this test would pass even if the `deviceCreationLocks` serialization were removed entirely — it never verifies the concurrency-control behavior it's named for. Could it assert exactly one device exists (and a single `Created` relation to the gateway), e.g. via the device list / relation count, so a regression in the locking would actually fail it?

application/src/test/java/org/thingsboard/server/transport/mqtt/sparkplug/AbstractMqttV5ClientSparkplugTest.java
```diff
@@ -364,8 +370,8 @@ protected void unauthorizedRenameAttemptBad_Test() throws Exception {
                 .untilAsserted(() -> {
                     // Check if the original device still exists with its original ID
                     Device currentStranger = doGet("/api/tenant/devices?deviceName=" + strangerName, Device.class);
-                    Assert.assertNotNull("Original device disappeared!", currentStranger);
-                    Assert.assertEquals("Security breach: Original device ID changed!", originalStrangerId, currentStranger.getId());
+                    assertNotNull("Original device disappeared!", currentStranger);
```
ViacheslavKlimov, line 373: `unauthorizedRenameAttemptBad_Test` is dead code — it's defined but not wired to any `@Test` (the running one is the near-duplicate `unauthorizedRenameAttempt_Test`). It reads like an abandoned first draft. Worth deleting; the one assertion it has that the live test doesn't — that a newly-created full-path device must have a *different* id than the stranger — is actually worth keeping, so consider folding just that line into `unauthorizedRenameAttempt_Test` rather than leaving two copies.

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -396,46 +397,42 @@ private Device tryRenameSparkplugDevice(GetOrCreateDeviceFromGatewayRequestMsg r
...
             );
         } catch (Exception e) {
-            // Log the error from the relation service but return null to allow potential recovery
-            log.error("[{}] Error checking relation for device {}", gateway.getId(), existingDevice.getId(), e);
-            return null;
+            log.error("[{}] Failed checking relation for device {}",
+                    gateway.getId(),
+                    existingDevice.getId(),
+                    e);
+            throw new RuntimeException(
```
ViacheslavKlimov: `checkRelation` declares no checked exception, so this `catch (Exception)` only ever catches unchecked runtimes — and it just logs and rethrows as a bare `RuntimeException`, which then reaches `handle()` where the only typed catches are `JsonProcessingException` and `EntitiesLimitExceededException`, so it propagates unhandled exactly as it would have without the wrap. The try/catch adds no behavior over letting it propagate, and a bare `RuntimeException` deviates from the `ThingsboardException`/`ThingsboardErrorCode` convention used elsewhere in this flow. I'd drop the catch (or, if you want a clearer message, at least don't re-wrap into an untyped runtime).

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -141,6 +141,7 @@ public class DefaultTransportApiService implements TransportApiService {
     private final OtaPackageService otaPackageService;
     private final OtaPackageDataCache otaPackageDataCache;
     private final QueueService queueService;
+    public static final String GATEWAY_CREATED_RELATION = "Created";
```
ViacheslavKlimov, line 144: Making this `public static final` on the service exists mainly so the integration test can `import static ...DefaultTransportApiService.GATEWAY_CREATED_RELATION` — exposing internal service state as public API for test convenience is a bit of a smell. The codebase already centralizes relation types in `EntityRelation` (e.g. `CONTAINS_TYPE`, `MANAGES_TYPE`, `EDGE_TYPE`); a `GATEWAY_CREATED`/`CREATED_TYPE` constant there (or in `DataConstants`) referenced from both prod and test would fit the existing pattern better. Also, it's wedged into the middle of the `final`-field injection block rather than grouped with the static constants — easy to miss.

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
```diff
@@ -452,8 +449,8 @@ private Device createNewDevice(GetOrCreateDeviceFromGatewayRequestMsg requestMsg
         Device device = new Device();
         device.setTenantId(tenantId);
         device.setName(requestMsg.getDeviceName());
-        if (requestMsg.getIsSparkplug()) {
-            if (topicPath.length == 3) device.setLabel(topicPath[2]);
+        if (requestMsg.getIsSparkplug() && topicPath.length == 3) {
```
ViacheslavKlimov, line 441: The `{groupId}:{nodeId}:{deviceId}` shape is encoded as a bare magic `3` plus `topicPath[2]` in three spots (the `split` at 371, the `length != 3` guard + `topicPath[2]` in `tryRenameSparkplugDevice`, and here). It's easy to get the index wrong and there's no single place that documents the layout. A small named constant (`SPARKPLUG_PATH_SEGMENTS = 3`) or a tiny parse helper/record with named accessors would make the intent obvious and keep the segment math in one place.

application/src/main/java/org/thingsboard/server/service/transport/DefaultTransportApiService.java
nickAS21: Ok

PR #15754, opened 2026-06-10
Fixed sync highlight overlay scroll on Safari in string-pattern-autocomplete

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
+++ b/ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
@@ -21,6 +21,7 @@ import {
   ElementRef,
   forwardRef,
   Input,
+  NgZone,
   OnChanges,
   OnInit,
   QueryList,
@@ -133,6 +134,7 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
               private translate: TranslateService,
               private viewContainerRef: ViewContainerRef,
               private destroyRef: DestroyRef,
+              private ngZone: NgZone,
               private cd: ChangeDetectorRef) {
   }
 
@@ -216,16 +218,18 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
     const overlay = this.highlightTextRef?.nativeElement as HTMLElement;
     let lastLeft = -1;
     let lastTop = -1;
-    const tick = () => {
-      if (overlay && (input.scrollLeft !== lastLeft || input.scrollTop !== lastTop)) {
-        lastLeft = input.scrollLeft;
-        lastTop = input.scrollTop;
-        overlay.scrollLeft = lastLeft;
-        overlay.scrollTop = lastTop;
-      }
+    this.ngZone.runOutsideAngular(() => {
+      const tick = () => {
+        if (overlay && (input.scrollLeft !== lastLeft || input.scrollTop !== lastTop)) {
+          lastLeft = input.scrollLeft;
+          lastTop = input.scrollTop;
+          overlay.scrollLeft = lastLeft;
+          overlay.scrollTop = lastTop;
+        }
+        this.scrollSyncRafId = requestAnimationFrame(tick);
+      };
       this.scrollSyncRafId = requestAnimationFrame(tick);
-    };
-    this.scrollSyncRafId = requestAnimationFrame(tick);
+    });
   }
 
   private stopScrollSync() {
@@ -234,7 +238,6 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
       this.scrollSyncRafId = null;
     }
   }
-
 
   optionSelected(value: string) {
     const position = this.inputRef.nativeElement.selectionStart;
--- a/ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
+++ b/ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
@@ -21,7 +21,6 @@ import {
   ElementRef,
   forwardRef,
   Input,
-  NgZone,
   OnChanges,
   OnInit,
   QueryList,
@@ -46,6 +45,7 @@ import {
   PositionStrategy
 } from '@angular/cdk/overlay';
 import { TemplatePortal } from '@angular/cdk/portal';
+import { CancelAnimationFrame, RafService } from '@core/services/raf.service';
 
 @Component({
     selector: 'tb-string-pattern-autocomplete',
@@ -124,7 +124,7 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
 
   private predefinedValuesButtonMode = false;
 
-  private scrollSyncRafId: number | null = null;
+  private scrollSyncCancel: CancelAnimationFrame | null = null;
 
   private propagateChange = (_val: any) => {
   };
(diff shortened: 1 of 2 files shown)


ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
```diff
@@ -199,15 +208,34 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
...
-      this.highlightTextRef.nativeElement.scrollTop = scrollTop;
+  private startScrollSync() {
+    if (this.scrollSyncRafId !== null) {
+      return;
+    }
+    const input = this.inputRef.nativeElement as HTMLInputElement;
+    const overlay = this.highlightTextRef?.nativeElement as HTMLElement;
+    let lastLeft = -1;
+    let lastTop = -1;
+    const tick = () => {
```
vvlladd28, line 225: This app runs with zone.js, which patches `requestAnimationFrame` by default, so this self-rescheduling loop runs *inside* the Angular zone — for the entire time the input is focused it triggers a full application change-detection cycle every frame (~60fps), even when nothing is scrolling. Wrapping the loop in `NgZone.runOutsideAngular(() => ...)` would avoid that (there's precedent in `nav-tree.component.ts`).

More broadly, a perpetual per-frame poll for an input that's usually focused-but-idle is a lot of churn just to keep two scroll offsets in sync. Since the overlay only needs to move when the input's scroll position can actually change, could this be driven off actual interaction (`input`/`keyup`/`select`/`click`/`wheel`) instead — or a short bounded rAF burst after such an event — rather than a loop that never stops until blur?

ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
```diff
@@ -199,15 +208,34 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
...
+  }
+
+  private stopScrollSync() {
+    if (this.scrollSyncRafId !== null) {
+      cancelAnimationFrame(this.scrollSyncRafId);
+      this.scrollSyncRafId = null;
     }
   }
 
+
```
vvlladd28: Extra blank line here — there are two blank lines between `stopScrollSync()` and `optionSelected()`, whereas the rest of the file uses a single blank line between methods.

vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Fixed sync highlight overlay scroll on Safari in string-pattern-autocomplete](https://github.com/thingsboard/thingsboard/pull/15754). Left 2 comment(s) inline.

The Safari fix itself is sound — replacing the unreliable `(scroll)` event with a focus-gated `requestAnimationFrame` poll, with proper teardown on blur and on destroy. The main thing worth weighing is that the rAF loop runs inside the Angular zone (zone.js patches rAF), so it drives whole-app change detection every frame while the input is focused; see the inline comment for the suggested `NgZone.runOutsideAngular` fix.

---

ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
```diff
@@ -199,12 +210,32 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
...
-      this.highlightTextRef.nativeElement.scrollTop = scrollTop;
+  private startScrollSync() {
+    if (this.scrollSyncRafId !== null) {
+      return;
+    }
+    const input = this.inputRef.nativeElement as HTMLInputElement;
+    const overlay = this.highlightTextRef?.nativeElement as HTMLElement;
+    let lastLeft = -1;
+    let lastTop = -1;
+    this.ngZone.runOutsideAngular(() => {
```
vvlladd28: The codebase already has a `RafService` (`app/core/services/raf.service.ts`) that encapsulates exactly this pattern — it runs the frame callback outside Angular's zone, returns a `CancelAnimationFrame` handle for teardown, and falls back to `webkitRequestAnimationFrame`/`setTimeout` when unprefixed rAF isn't available. It's the established convention here (js-func, css, markdown-editor components, widget-subscription, etc. all inject it). Since this fix targets a Safari/webkit quirk, reusing `RafService` — which already handles the webkit-prefixed variants — fits especially well: it would replace the hand-rolled `runOutsideAngular` + `requestAnimationFrame`/`cancelAnimationFrame` with a single injected dependency and a stored cancel handle, drop the `NgZone` import, and keep the same per-frame behavior. Worth considering.

ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
```diff
@@ -199,12 +210,32 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
...
-    if (this.highlightTextRef && this.highlightTextRef.nativeElement) {
-      this.highlightTextRef.nativeElement.scrollLeft = scrollLeft;
-      this.highlightTextRef.nativeElement.scrollTop = scrollTop;
+  private startScrollSync() {
+    if (this.scrollSyncRafId !== null) {
+      return;
+    }
+    const input = this.inputRef.nativeElement as HTMLInputElement;
+    const overlay = this.highlightTextRef?.nativeElement as HTMLElement;
+    let lastLeft = -1;
```
vvlladd28, line 223: Could you add a short comment here explaining why a continuous per-frame poll replaced the old `(scroll)` handler? Two reasons are worth recording: Safari doesn't fire `scroll` reliably on this input, and the poll keeps the overlay synced frame-by-frame during drag-selection / auto-scroll (where an event-driven sync would stutter). Without that note, a future reader will likely see an always-on rAF loop and try to revert it to a scroll event, reintroducing the bug. (The cost — a perpetual loop while focused — is bounded to a single focused input at a time, so it's acceptable; the comment is mostly to protect the intent.)

ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts
```diff
@@ -199,12 +210,32 @@ export class StringPatternAutocompleteComponent implements ControlValueAccessor,
...
-    const scrollTop = (event.target as HTMLInputElement).scrollTop;
-    if (this.highlightTextRef && this.highlightTextRef.nativeElement) {
-      this.highlightTextRef.nativeElement.scrollLeft = scrollLeft;
-      this.highlightTextRef.nativeElement.scrollTop = scrollTop;
+  private startScrollSync() {
+    if (this.scrollSyncRafId !== null) {
+      return;
+    }
+    const input = this.inputRef.nativeElement as HTMLInputElement;
+    const overlay = this.highlightTextRef?.nativeElement as HTMLElement;
```
vvlladd28: Minor: `input` and `overlay` are resolved once here and captured by the long-lived `tick` closure for the whole focus session. They're stable `static: true` ViewChildren so this is fine in practice, but reading them inside the tick (or bailing once up front if `overlay`/`input` is absent) would be marginally more robust.

vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Fixed sync highlight overlay scroll on Safari in string-pattern-autocomplete](https://github.com/thingsboard/thingsboard/pull/15754). Left 3 comment(s) inline.

The scroll-sync rewrite (focus-scoped rAF loop, stopped on blur/destroy) is correct — I traced the start guard, the per-frame reschedule, and all three stop paths (`focusout`, `destroyRef.onDestroy`, and the re-focus after `optionSelected`) and found no race, leak, or dangling reference to the removed `onInputScroll`. The continuous poll is also a reasonable choice here: it keeps the overlay glued frame-by-frame during drag-selection / auto-scroll, which an event-driven sync wouldn't do as smoothly. The main suggestion is to route that polling through the existing `RafService` rather than hand-rolling `NgZone` + `requestAnimationFrame`.

---

vvlladd28 (approved):

PR #15756, opened 2026-06-11
Fixed edge event loss for kafka edges during sync

AndriiLandiak:
## Pull Request description

On Kafka-based edges, an edge event can be silently dropped if it gets polled from the edge-event topic right as an edge sync (or high-priority processing) starts. The event is read from Kafka, skipped without being committed, and the consumer position moves past it — so it's never re-delivered to that session. This PR adds a small "only poll when the session is ready" gate so the event stays in the queue until we can actually process it.

### How it shows up

`DashboardClientTest.testSendDashboardToCloud` has been flaky (~50% on the Kafka edge profile). The test unassigns the edge from a customer and deletes that customer on the cloud, then waits for the customer to disappear on the edge. Every so often the `CUSTOMER/DELETED` event never reaches the edge, the customer is never removed there, and the `getCustomerById(...).isEmpty()` wait times out after 30s.

The in-memory/Postgres edge never reproduces it: `PostgresEdgeGrpcSession` re-reads events by `seqId` from the DB on every iteration, so an interrupted read is simply re-fetched next time around. The Kafka consumer, on the other hand, is created once and polls forward continuously — so a skipped batch is just gone.

### Root cause

`KafkaEdgeGrpcSession.processMsgs(...)` polls a batch and then returns without committing whenever the session isn't ready:

```java
if (!isConnected() || isSyncInProgress() || isHighPriorityProcessing) {
    log.debug("... Skipping iteration");
    return;                 // batch already polled; position advanced; nothing committed
}
...
if (!isInterrupted) {
    consumer.commit();
}
```

The catch: the consumer has *already* polled, so its in-memory position is past that batch. Not committing only makes the *committed* offset lag — the next `poll()` still reads forward, so within that session the skipped events are never seen again.

Here's the exact sequence from a failing run (cloud side, edge `edge-kafka`):

| Time | Event |
|------|-------|
| 00:49:55,035 | `CUSTOMER/DELETED` pushed to the edge-event topic |
| 00:49:55,036 | poll #1 → processes `UNASSIGNED_FROM_CUSTOMER` (a config change) |
| 00:49:55,044 | the config change triggers a full re-sync → `syncInProgress = true`, "General processing interrupted!" |
| 00:49:55,050 | poll #2 → `isSyncInProgress() == true` → "Skipping iteration", returns without `commit()` → the batch (incl. `CUSTOMER/DELETED`) is dropped, position advances |
| 00:49:55,124 | sync completes — re-sends entities via fetchers (ADD-only), so it can't remove the now-stale customer |
| after | consumer keeps polling forward; `CUSTOMER/DELETED` is never reprocessed |

### The fix

Don't poll while the session isn't ready. If we never poll during the sync window, the Kafka position never advances past the event, so it stays in the queue and is delivered the moment the session is ready again — the same no-loss behaviour the Postgres edge gets for free.

`TbQueueConsumer` has no seek/pause, so rewinding the offset would be a high-blast-radius change. Instead this adds an opt-in poll gate to the shared `QueueConsumerManager` and wires only the edge consumer to it:

- `QueueConsumerManager` gets an optional `readinessCheck`; while it reports "not ready", the consumer loop skips polling (sleeps one poll interval and re-checks) instead of polling and then dropping.
- `KafkaEdgeGrpcSession` opts in via `isReadyToProcessGeneralEvents()` = `isConnected() && !isSyncInProgress() && !isHighPriorityProcessing`, and reuses that single predicate for the existing in-`processMsgs` guard too.

This is opt-in by design: every other consumer leaves `readinessCheck` unset, so it's always ready and its behaviour is byte-for-byte unchanged.

One caveat worth flagging for reviewers: if a sync ever ran longer than Kafka's `max.poll.interval.ms` (default 5 min), not polling would trigger a consumer-group rebalance; on rejoin the consumer resumes from the committed offset — still no data loss, just a rebalance. Edge syncs are normally seconds, so this is a rare edge case rather than a blocker.

### Validation

New `QueueConsumerManagerTest` in `common/queue`, covering:

- an event queued while the gate is closed is delivered once it opens — not dropped;
- the consumer isn't polled at all while not-ready, and resumes once ready;
- the default (no `readinessCheck`) path still polls and delivers immediately — the guarantee every other consumer relies on;
- the loop exits cleanly when interrupted mid-wait.

The two core tests are written so they fail on the old behaviour (without the gate it polls when it shouldn't, and the queued event is dropped) and pass with the fix. `common/queue` builds and all four tests are green.

No documentation changes needed — this is purely internal edge-consumer behaviour.

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java
+++ b/application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java
@@ -70,7 +70,9 @@ public class KafkaEdgeGrpcSession extends EdgeGrpcSession {
 
     private void processMsgs(List<TbProtoQueueMsg<ToEdgeEventNotificationMsg>> msgs, TbQueueConsumer<TbProtoQueueMsg<ToEdgeEventNotificationMsg>> consumer) {
         log.trace("[{}][{}] starting processing edge events", tenantId, edge.getId());
-        if (!isConnected() || isSyncInProgress() || isHighPriorityProcessing) {
+        // Defensive backstop: the loop already gates polling on readiness; this only fires on the narrow race
+        // where readiness flips during poll(), and that already-polled batch is dropped here (can't rewind).
+        if (!isReadyToProcessGeneralEvents()) {
             log.debug("[{}][{}] edge not connected, edge sync is not completed or high priority processing in progress, " +
                       "connected = {}, sync in progress = {}, high priority in progress = {}. Skipping iteration",
                     tenantId, edge.getId(), isConnected(), isSyncInProgress(), isHighPriorityProcessing);
@@ -96,6 +98,10 @@ public class KafkaEdgeGrpcSession extends EdgeGrpcSession {
         }
     }
 
+    private boolean isReadyToProcessGeneralEvents() {
+        return isConnected() && !isSyncInProgress() && !isHighPriorityProcessing;
+    }
+
     @Override
     public ListenableFuture<Boolean> migrateEdgeEvents() throws Exception {
         return super.processEdgeEvents();
@@ -103,7 +109,7 @@ public class KafkaEdgeGrpcSession extends EdgeGrpcSession {
 
     @Override
     public ListenableFuture<Boolean> processEdgeEvents() {
-        if (!isConnected() || isSyncInProgress() || isHighPriorityProcessing) {
+        if (!isReadyToProcessGeneralEvents()) {
             log.warn("[{}][{}] Session is not ready (connected={}, syncInProgress={}, highPriority={}), skip starting edge event consumer",
                     tenantId, edge != null ? edge.getId() : null, isConnected(), isSyncInProgress(), isHighPriorityProcessing);
             return Futures.immediateFuture(Boolean.FALSE);
@@ -126,6 +132,7 @@ public class KafkaEdgeGrpcSession extends EdgeGrpcSession {
                         .consumerCreator(() -> tbCoreQueueFactory.createEdgeEventMsgConsumer(tenantId, edge.getId()))
                         .consumerExecutor(consumerExecutor)
                         .threadPrefix("edge-events-" + edge.getId())
+                        .readinessCheck(this::isReadyToProcessGeneralEvents)
                         .build();
                 consumer.subscribe();
                 consumer.launch();
--- a/application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java
+++ b/application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java
@@ -74,7 +74,7 @@ public class KafkaEdgeGrpcSession extends EdgeGrpcSession {
         // where readiness flips during poll(), and that already-polled batch is dropped here (can't rewind).
         if (!isReadyToProcessGeneralEvents()) {
             log.debug("[{}][{}] edge not connected, edge sync is not completed or high priority processing in progress, " +
-                      "connected = {}, sync in progress = {}, high priority in progress = {}. Skipping iteration",
+                            "connected = {}, sync in progress = {}, high priority in progress = {}. Skipping iteration",
                     tenantId, edge.getId(), isConnected(), isSyncInProgress(), isHighPriorityProcessing);
             return;
         }
@@ -147,8 +147,11 @@ public class KafkaEdgeGrpcSession extends EdgeGrpcSession {
     @Override
     public void processHighPriorityEvents() {
         isHighPriorityProcessing = true;
-        super.processHighPriorityEvents();
-        isHighPriorityProcessing = false;
+        try {
+            super.processHighPriorityEvents();
+        } finally {
+            isHighPriorityProcessing = false;
+        }
     }
 
     @Override
--- a/common/queue/src/main/java/org/thingsboard/server/queue/common/consumer/QueueConsumerManager.java
+++ b/common/queue/src/main/java/org/thingsboard/server/queue/common/consumer/QueueConsumerManager.java
@@ -30,6 +30,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
 
 @Slf4j
@@ -40,6 +41,8 @@ public class QueueConsumerManager<M extends TbQueueMsg> {
     private final long pollInterval;
     private final ExecutorService consumerExecutor;
     private final String threadPrefix;
+    /** Optional poll gate: while {@code false} the loop skips polling so the position doesn't advance; {@code null} = always ready (default). */
+    private final BooleanSupplier readinessCheck;
 
     @Getter
     private final TbQueueConsumer<M> consumer;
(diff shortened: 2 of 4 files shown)


MazurenkoNick (commented):
## Review summary

Reviewed 3 changed files in [Fixed edge event loss on Kafka edges during sync](https://github.com/thingsboard/thingsboard/pull/15756). Left 2 comment(s) inline.

The root-cause analysis in the description checks out. Verified against `TbKafkaConsumerTemplate`/`AbstractTbQueueConsumerTemplate`: `KafkaConsumer.poll()` advances the in-memory position regardless of commit, and `doCommit()` only does `commitSync()` (which matters on rebalance / a new session, not within the live consumer). So an uncommitted-but-polled batch is genuinely never re-read by that session — the "consumer position moves past it" claim is correct. The edge consumer is created with a `groupId`, so the `max.poll.interval.ms` rebalance caveat is also accurate.

The gate is a clean, opt-in design and is byte-for-byte unchanged for every other consumer (null `readinessCheck` ⇒ always ready). The comments below are about the one residual race the loop gate doesn't fully close, plus a minor API note — nothing blocking.

application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java
```diff
@@ -70,9 +70,9 @@ public KafkaEdgeGrpcSession(EdgeContextComponent ctx, TopicService topicService,
 
     private void processMsgs(List<TbProtoQueueMsg<ToEdgeEventNotificationMsg>> msgs, TbQueueConsumer<TbProtoQueueMsg<ToEdgeEventNotificationMsg>> consumer) {
         log.trace("[{}][{}] starting processing edge events", tenantId, edge.getId());
-        if (!isConnected() || isSyncInProgress() || isHighPriorityProcessing) {
+        if (!isReadyToProcessGeneralEvents()) {
```
MazurenkoNick, line 75: Now that the consumer loop gates on `isReadyToProcessGeneralEvents()` *before* polling, this guard becomes the second line of defense rather than the primary one. Worth keeping — but worth being explicit about what it now protects. The only way execution reaches here while not-ready is if readiness flips between the loop's gate check and this call, i.e. during the up-to-`pollInterval` `consumer.poll(...)`. When that happens, this returns without committing, but the batch has *already* been polled, so the position has advanced and those events are dropped — the same loss mode this PR fixes, just in a much narrower (~1s) window. Since `TbQueueConsumer` can't rewind, you can't fully close it here, but a one-line comment acknowledging the residual window (and that this duplicate condition is intentional defensive double-checking, not dead code) would stop a future reader from assuming the loop gate eliminated the race entirely.

common/queue/src/main/java/org/thingsboard/server/queue/common/consumer/QueueConsumerManager.java
```diff
@@ -49,12 +51,13 @@ public class QueueConsumerManager<M extends TbQueueMsg> {
     @Builder
     public QueueConsumerManager(String name, MsgPackProcessor<M> msgPackProcessor,
                                 long pollInterval, Supplier<TbQueueConsumer<M>> consumerCreator,
-                                ExecutorService consumerExecutor, String threadPrefix) {
+                                ExecutorService consumerExecutor, String threadPrefix, BooleanSupplier readinessCheck) {
```
MazurenkoNick, line 55: The optional `readinessCheck` is the right shape — existing call sites pass nothing, `isReadyToProcess()` defaults to always-ready, so this stays backward compatible and is hard to misuse. Minor: a short Javadoc on the builder param spelling out the contract (e.g. "when supplied, the loop skips polling while this returns `false`, so the consumer position does not advance") would make the truthy-means-ready semantics obvious to the next consumer that wants to opt in, since a bare `BooleanSupplier` at a call site is otherwise opaque.

AndriiLandiak (commented):
## Review summary

Reviewed 3 changed files in [Fixed edge event loss on Kafka edges during sync](https://github.com/thingsboard/thingsboard/pull/15756). Left 3 comment(s) inline.

The core change — an opt-in `readinessCheck` poll gate on `QueueConsumerManager`, wired only into the edge consumer — looks correct. The replacement predicate is boolean-equivalent to the code it supersedes (`!(a && !b && !c)` ⇔ `!a || b || c`), the gate is null-safe and leaves every other consumer byte-for-byte unchanged (verified: no direct constructor callers, all callers use the Lombok builder so the new param defaults to `null`/always-ready), and the documented consumer-rebalance trade-off when a sync exceeds `max.poll.interval.ms` is a reasonable, clearly-flagged edge case. The inline notes are about test coverage of the edge-specific wiring plus two minor design/maintainability points.

## Additional findings

This observation is about existing code outside the PR's diff — spotted while reading surrounding context.

- **application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java:147** — `processHighPriorityEvents()` sets `isHighPriorityProcessing = true` / `false` around `super.processHighPriorityEvents()` without a `try/finally`. Now that the new `readinessCheck` reads this flag to decide whether the consumer polls *at all*, a stuck `true` would park the consumer indefinitely rather than just skipping one in-line iteration as before. In practice the base method swallows every `Exception` internally, so today the flag can only stick on an `Error` (e.g. OOM) — but a `try/finally` would future-proof the toggle against base-method refactors now that its blast radius is larger.

---

application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java
```diff
@@ -126,6 +132,7 @@ public ListenableFuture<Boolean> processEdgeEvents() {
                         .consumerCreator(() -> tbCoreQueueFactory.createEdgeEventMsgConsumer(tenantId, edge.getId()))
                         .consumerExecutor(consumerExecutor)
                         .threadPrefix("edge-events-" + edge.getId())
+                        .readinessCheck(this::isReadyToProcessGeneralEvents)
```
AndriiLandiak, line 135: The new `QueueConsumerManagerTest` deterministically covers the generic gate, but the edge-specific wiring that actually fixes the bug — this `.readinessCheck(this::isReadyToProcessGeneralEvents)` line — has no deterministic test. If it were ever dropped, every unit test in this PR would still pass; the only thing that might catch the regression is the probabilistic `DashboardEdgeTest` BBT that surfaced the original ~50% flake. Worth a focused test (ideally an edge-level integration test) proving that an event queued while `isSyncInProgress()` survives until sync completes, so the actual fix is pinned rather than just the mechanism.

application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java
```diff
@@ -96,14 +98,18 @@ private void processMsgs(List<TbProtoQueueMsg<ToEdgeEventNotificationMsg>> msgs,
         }
     }
 
+    private boolean isReadyToProcessGeneralEvents() {
```
AndriiLandiak, line 101: Nice consolidation — this predicate is now the single source of truth for the poll gate (via `readinessCheck`), the `processMsgs` backstop, and `processEdgeEvents`. Since the gate and the backstop have to stay semantically identical for the "can't rewind" comment above to hold, a one-line note here that this method backs *both* the poll gate and the in-flight backstop would make that coupling discoverable from the definition rather than only from the call sites.

common/queue/src/main/java/org/thingsboard/server/queue/common/consumer/QueueConsumerManager.java
```diff
@@ -102,6 +112,25 @@ private void consumerLoop(TbQueueConsumer<M> consumer) {
         }
     }
 
+    private boolean isReadyToProcess() {
+        return readinessCheck == null || readinessCheck.getAsBoolean();
```
AndriiLandiak, line 116: Minor: the nullable `readinessCheck` is handled with a null check here, which is what keeps all existing builder callers unchanged. An alternative is to default the field to an always-ready supplier (`() -> true`) in the builder, so this method could just call `readinessCheck.getAsBoolean()` and the nullable field disappears entirely. That trades a tiny constant allocation for one fewer null branch and a field that can't be left null — current form is fine, just flagging so the choice is deliberate.

AndriiLandiak: I prefer to leave like it's now, it's a pure-taste

application/src/main/java/org/thingsboard/server/service/edge/rpc/KafkaEdgeGrpcSession.java
AndriiLandiak: No, it's already clear one

PR #15764, opened 2026-06-11
Fixed "Confirm exit" dialog appearing when the form was not changed

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/calculated-fields/components/entity-aggregation-configuration/entity-aggregation-component.component.html
+++ b/ui-ngx/src/app/modules/home/components/calculated-fields/components/entity-aggregation-configuration/entity-aggregation-component.component.html
@@ -58,7 +58,6 @@
           appearance="outline"
           subscriptSizing="dynamic"
           required
-          userTimezoneByDefault
           formControlName="tz">
         </tb-timezone-select>
       </div>
--- a/ui-ngx/src/app/core/services/calculated-field-form.service.ts
+++ b/ui-ngx/src/app/core/services/calculated-field-form.service.ts
@@ -25,11 +25,11 @@ import {
   CalculatedFieldType,
   OutputStrategyType
 } from '@shared/models/calculated-field.models';
-import { oneSpaceInsideRegex } from '@shared/models/regex.constants';
 import { isDefined } from '@core/utils';
 import { CalculatedFieldsService } from '@core/http/calculated-fields.service';
 import { CalculatedFieldsTableEntity } from '@home/components/calculated-fields/calculated-fields-table-config';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { getTimezoneInfo } from '@shared/models/time/time.models';
 
 @Injectable({ providedIn: 'root' })
 export class CalculatedFieldFormService {
@@ -85,6 +85,9 @@ export class CalculatedFieldFormService {
     if (config.type !== CalculatedFieldType.ALARM) {
       if (isDefined(config?.output) && !config?.output?.strategy) {
         config.output.strategy = { type: OutputStrategyType.RULE_CHAIN };
+      }
+      if (config.type === CalculatedFieldType.ENTITY_AGGREGATION && config.interval) {
+        config.interval.tz = getTimezoneInfo(config.interval.tz, undefined, true)?.id;
       }
     }
     return config;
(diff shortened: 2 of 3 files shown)


ui-ngx/src/app/core/guards/confirm-on-exit.guard.ts
```diff
@@ -59,15 +59,17 @@ export class ConfirmOnExitGuard  {
...
         const confirmForm = component.confirmForm();
         if (confirmForm) {
           isDirty = confirmForm.dirty;
+          isDisabled = confirmForm.disabled;
         }
       } else if (isDefined(component.isDirty)) {
         isDirty = component.isDirty;
       }
-      if (isDirty) {
+      if (isDirty && !isDisabled) {
```
vvlladd28: I'd suggest reverting this guard change. Treating a fully-disabled form as "not modified" is fragile — `FormGroup.disabled` is only `true` when *every* control is disabled — but the bigger concern is the regression it introduces: a form that the user edited and that then gets programmatically disabled while still dirty (e.g. entering a read-only/view mode, or a save in progress) would now silently skip the exit confirmation and lose those edits. That's a behavior change well beyond the bug being fixed.

This reads as compensating for the symptom. The spurious dirty state actually originates in the timezone control re-emitting a normalized value on init (see my comment in `timezone-select.component.ts`); fixing it there makes this guard change unnecessary.

ui-ngx/src/app/shared/components/time/timezone-select.component.ts
```diff
@@ -166,7 +170,7 @@ export class TimezoneSelectComponent implements ControlValueAccessor, OnInit, Af
     const foundTimezone = getTimezoneInfo(value, this.defaultTimezoneId, this.userTimezoneByDefaultValue);
     if (foundTimezone !== null) {
       this.selectTimezoneFormGroup.get('timezone').patchValue(foundTimezone, {emitEvent: false});
-      if (foundTimezone.id !== value) {
+      if (foundTimezone.id !== value && this.propagateChangeOnWrite) {
```
vvlladd28: This is the real root cause of the false "Confirm exit". When the stored value normalizes to a different id (an empty/legacy tz resolved via `userTimezoneByDefault`), `writeValue()` schedules `updateView()` → `propagateChange()`. Per the `ControlValueAccessor` contract `writeValue` shouldn't emit, and Angular marks the bound control dirty on any `onChange` coming from the accessor — so a pristine form gets dirtied purely on load.

I'd avoid the `propagateChangeOnWrite` opt-out: it pushes a workaround onto every call site and leaves the contract violation in a shared component used in 6 places. Instead, normalize the timezone where the data is loaded into the form so the control receives an already-valid value and never needs to re-emit. For this PR that's `CalculatedFieldFormService.prepareConfig()` (see the review body) — it already defaults `output.strategy`, so normalizing `interval.tz` there fixes the entity-aggregation case with no change to this component.

Worth noting the type-change path needs no flag: when the user switches CF type, `setupTypeChange` resets the config and the form is legitimately dirty from the type change anyway, so the existing propagate-on-write correctly seeds the default tz into the model there. The load path is the only one that needs fixing.

ui-ngx/src/app/modules/home/components/calculated-fields/components/entity-aggregation-configuration/entity-aggregation-component.component.html
```diff
@@ -59,6 +59,7 @@
           subscriptSizing="dynamic"
           required
           userTimezoneByDefault
+          [propagateChangeOnWrite]="false"
```
vvlladd28: If the timezone is normalized at load (see my comments in `prepareConfig` / `timezone-select.component.ts`), this `[propagateChangeOnWrite]="false"` — and the new input backing it — can be dropped, keeping the shared component untouched.

vvlladd28 (commented):
## Review summary

Thanks for tracking this down. The diagnosis is right — the form is being marked dirty on load without any user input — but I think both changes treat the symptom rather than the cause, and I'd suggest a different approach.

The root cause is in `TimezoneSelectComponent.writeValue()`: it re-emits a normalized timezone id through `propagateChange()` during initialization, which Angular treats as a user edit and marks the bound control dirty. The cleanest fix is to normalize the timezone where the value is loaded into the form (`CalculatedFieldFormService.prepareConfig()`), so the component receives an already-valid value and never needs to re-emit — leaving the shared component and the guard untouched.

Details inline. Happy to pair on it if useful.

## Additional findings

- **ui-ngx/src/app/core/services/calculated-field-form.service.ts:83** (`prepareConfig`) — this is the natural place to normalize the timezone at load, alongside the existing `output.strategy` defaulting. Something like `if (config.type === CalculatedFieldType.ENTITY_AGGREGATION && config.interval) { config.interval.tz = getTimezoneInfo(config.interval.tz, undefined, true)?.id; }` makes the entity-aggregation form pristine on open and keeps the persisted model consistent (the value is seeded into the parent control before the CVA is bound), with no changes to `TimezoneSelectComponent`.

vvlladd28 (approved):

PR #15786, opened 2026-06-12
Fix zombie edge event consumer on session handover

arminfelder:
## Pull Request description

When stopping the Thingsboard server for a longer time (e.g. 1h), and then bringing it up again, 
the connected edge instances are not receiving downstream messages, via GRPC, till the edge instances are getting  restarted.

I created a LLM assisted testsetup (https://github.com/arminfelder/thingsboard-debug) using first TB 4.3 (branch v4.3), and conducted the automated tests for around an hour., after confirmation, another test run has been conducted against the master branch

As a result the following issue has been identified:
consumerTasks can get get abandoned, and therfor block the edge instance, till restart:
currently in case a loop misses the 10s limit there is just a warning log "Failed to await consumer loop stop" 

therefor this contribution adds:
+            log.warn("[{}] Consumer loop did not stop in {} ms, interrupting it", name, stopTimeoutMs);
+            consumerTask.cancel(true);

to force close the consumer

an LLM written analysis of the issue is also available here: https://github.com/arminfelder/thingsboard-debug/blob/master/docs/edge-sync-bugs-and-fixes-master.md

## General checklist

- [x ] You have reviewed the guidelines [document](https://docs.google.com/document/d/1wqcOafLx5hth8SAg4dqV_LV3un3m5WYR8RdTJ4MbbUM/edit?usp=sharing).
- [ x] No merge conflicts, commented blocks of code, code formatting issues.
  
## Back-End feature checklist

### DIFF
--- a/common/queue/src/main/java/org/thingsboard/server/queue/common/consumer/QueueConsumerManager.java
+++ b/common/queue/src/main/java/org/thingsboard/server/queue/common/consumer/QueueConsumerManager.java
@@ -25,6 +25,7 @@ import org.thingsboard.server.queue.TbQueueMsg;
 
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
@@ -36,6 +37,8 @@ import java.util.function.Supplier;
 @Slf4j
 public class QueueConsumerManager<M extends TbQueueMsg> {
 
+    public static final long DEFAULT_STOP_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(10);
+
     private final String name;
     private final MsgPackProcessor<M> msgPackProcessor;
     private final long pollInterval;
@@ -43,6 +46,7 @@ public class QueueConsumerManager<M extends TbQueueMsg> {
     private final String threadPrefix;
     /** Optional poll gate: while {@code false} the loop skips polling so the position doesn't advance; {@code null} = always ready (default). */
     private final BooleanSupplier readinessCheck;
+    private final long stopTimeoutMs;
 
     @Getter
     private final TbQueueConsumer<M> consumer;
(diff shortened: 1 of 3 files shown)


common/queue/src/main/java/org/thingsboard/server/queue/common/consumer/QueueConsumerManager.java
```diff
@@ -135,11 +141,22 @@ public void stop() {
...
-        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            consumerTask.get(stopTimeoutMs, TimeUnit.MILLISECONDS);
+        } catch (TimeoutException e) {
+            // The loop did not finish in time: it is most likely blocked inside the message pack
+            // processor (e.g. waiting for a downlink delivery that will never be acknowledged).
+            // A leftover consumer keeps its group membership alive via heartbeats and thereby
+            // blocks a replacement consumer of the same group from getting partitions assigned,
+            // so we must interrupt it instead of abandoning it.
+            log.warn("[{}] Consumer loop did not stop in {} ms, interrupting it", name, stopTimeoutMs);
+            consumerTask.cancel(true);
```
MazurenkoNick, line 156: `QueueConsumerManager` defined as `consumer` field in `KafkaBasedEdgeGrpcSessionManager` is being stopped on `destroy`: `consumer.stop();`
There's `consumer.unsubscribe()` Inside `stop` - meaning, it's not possible to have leftover consumer as consumer unsubscribes doesn't matter whether the pack is blocked or not. 
If the `consumerTask.get` doesn't execute on time, the log error will be printed without canceling consumerTask. But, we are eventually closing the `consumerExecutor` on KafkaBasedEdgeGrpcSessionManager:112-117 anyway.

So, there's no need for `consumerTask.cancel`

Caveat one: `awaitConsumerTermination()` catches `InterruptedException` at line 316 and returns without
calling `shutdownNow()`. If the gRPC thread running `destroy()` is itself interrupted during that wait,
the backstop is skipped and the thread does leak unbounded. Narrow, but it's the only master path
where cancel(true) changes an outcome rather than a timing.

Caveat two: `stop()` is shared API — nine callers. On lts-4.2 and lts-4.3, `awaitConsumerTermination()`
is awaitTermination(5, SECONDS) with no `shutdownNow()` at all, so there the thread genuinely leaks per
handover. That's an argument for backporting something, not for this change on master.

So, if you've been able to reproduce the issue on LTS-4.2, LTS-4.3, please make PR to LTS-4.2 branch (oldest affected, forward-merges to lts-4.3)

PR #15808, opened 2026-06-19
LTS versioned migrations

ViacheslavKlimov:
## Summary

Replaces the previous "run the whole cumulative LTS SQL on every version bump" behavior with **per-version migrations** selected by the version range `(dbVersion, packageVersion]`.

Each migration is a Spring bean implementing the new `LtsMigration` interface (`@Component @TbCoreComponent`, package `org.thingsboard.server.service.install.lts`), keyed by the version it ships in. They are auto-discovered as a `List<LtsMigration>`, ordered by a shared `LtsVersion` value type, and run by a new `LtsMigrationService` runner with three entry points:

- `applyMigrations` — no-downtime patch path (`SystemPatchApplier`): per in-range migration, runs `data/upgrade/lts/<version>/schema_update.sql` (by convention, if present) → the bean's `apply()` hook → records the version, all in **one transaction** per migration.
- `runSchemaMigrations` — offline major-upgrade schema phase (`SqlDatabaseUpgradeService`): SQL only.
- `runDataMigrations` — offline major-upgrade data phase (`DefaultDataUpdateService`): `apply()` only.

Migrations are required to be **idempotent** (the offline path records no per-version progress, so a resumed offline upgrade replays every in-range migration).

## CE `lts-4.2` content

- Split the cumulative `lts/schema_update.sql` (iot_hub table + indexes) into the per-version file `lts/4.2.2.3/schema_update.sql`; deleted the cumulative file.
- `V4_2_2_3Migration`: deprecates the widget **types** of the four obsolete system bundles (`air_quality`, `indoor_environment`, `industrial_widgets`, `outdoor_environment`) — keeping the type entities — and deletes **only** the bundle entities.

Covered by unit tests (`LtsVersionTest`, `LtsMigrationServiceTest`, `DefaultDatabaseSchemaSettingsServiceTest`, `V4_2_2_3MigrationTest`, updated `SystemPatchApplierTest`) and a real-context Testcontainers integration test (`LtsMigrationIntegrationTest`).

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/install/DefaultDatabaseSchemaSettingsService.java
+++ b/application/src/main/java/org/thingsboard/server/service/install/DefaultDatabaseSchemaSettingsService.java
@@ -19,6 +19,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.stereotype.Service;
+import org.thingsboard.server.service.install.lts.LtsVersion;
 import org.thingsboard.server.service.install.update.DefaultDataUpdateService;
 
 import java.util.Map;
@@ -132,12 +133,8 @@ public class DefaultDatabaseSchemaSettingsService implements DatabaseSchemaSetti
     }
 
     private long toDbVersion(String version) {
-        String[] versionParts = version.split("\\.");
-        long major = Long.parseLong(versionParts[0]);
-        long minor = versionParts.length > 1 ? Long.parseLong(versionParts[1]) : 0;
-        long maintenance = versionParts.length > 2 ? Long.parseLong(versionParts[2]) : 0;
-        long patch = versionParts.length > 3 ? Long.parseLong(versionParts[3]) : 0;
-        return major * 1_000_000_000L + minor * 1_000_000L + maintenance * 1000L + patch;
+        LtsVersion v = LtsVersion.parse(version);
+        return v.major() * 1_000_000_000L + v.minor() * 1_000_000L + v.maintenance() * 1000L + v.patch();
     }
 
     private void onSchemaSettingsError(String message) {
@@ -146,14 +143,7 @@ public class DefaultDatabaseSchemaSettingsService implements DatabaseSchemaSetti
     }
 
     private String normalizeVersion(String version) {
-        String[] parts = version.split("\\.");
-
-        int major = Integer.parseInt(parts[0]);
-        int minor = parts.length > 1 ? Integer.parseInt(parts[1]) : 0;
-        int maintenance = parts.length > 2 ? Integer.parseInt(parts[2]) : 0;
-        int patch = parts.length > 3 ? Integer.parseInt(parts[3]) : 0;
-
-        return major + "." + minor + "." + maintenance + "." + patch;
+        return LtsVersion.parse(version).toString();
     }
 
 }
--- a/application/src/main/java/org/thingsboard/server/service/install/DefaultDatabaseSchemaSettingsService.java
+++ b/application/src/main/java/org/thingsboard/server/service/install/DefaultDatabaseSchemaSettingsService.java
@@ -74,7 +74,12 @@ public class DefaultDatabaseSchemaSettingsService implements DatabaseSchemaSetti
 
     @Override
     public void updateSchemaVersion() {
-        jdbcTemplate.execute("UPDATE tb_schema_settings SET schema_version = " + getPackageSchemaVersionForDb());
+        updateSchemaVersion(getPackageSchemaVersion());
+    }
+
+    @Override
+    public void updateSchemaVersion(String version) {
+        jdbcTemplate.execute("UPDATE tb_schema_settings SET schema_version = " + toDbVersion(version));
     }
 
     @Override
@@ -123,13 +128,15 @@ public class DefaultDatabaseSchemaSettingsService implements DatabaseSchemaSetti
     }
 
     private long getPackageSchemaVersionForDb() {
-        String[] versionParts = getPackageSchemaVersion().split("\\.");
+        return toDbVersion(getPackageSchemaVersion());
+    }
 
-        long major = Integer.parseInt(versionParts[0]);
-        long minor = Integer.parseInt(versionParts[1]);
-        long maintenance = Integer.parseInt(versionParts[2]);
-        long patch = Integer.parseInt(versionParts[3]);
-
+    private long toDbVersion(String version) {
+        String[] versionParts = version.split("\\.");
+        long major = Long.parseLong(versionParts[0]);
+        long minor = versionParts.length > 1 ? Long.parseLong(versionParts[1]) : 0;
+        long maintenance = versionParts.length > 2 ? Long.parseLong(versionParts[2]) : 0;
+        long patch = versionParts.length > 3 ? Long.parseLong(versionParts[3]) : 0;
         return major * 1_000_000_000L + minor * 1_000_000L + maintenance * 1000L + patch;
     }
 
--- a/application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigration.java
+++ b/application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigration.java
@@ -13,20 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.thingsboard.server.service.install;
+package org.thingsboard.server.service.install.lts;
 
-public interface DatabaseSchemaSettingsService {
+/**
+ * One LTS migration, keyed by the version it ships in (e.g. "4.2.2.3").
+ * The runner runs the version's data/upgrade/lts/&lt;version&gt;/schema_update.sql by convention
+ * (if present), then calls {@link #apply()} for any programmatic work beyond SQL.
+ * Most migrations are SQL-only and only override {@link #getVersion()}.
+ */
+public interface LtsMigration {
 
-    void validateSchemaSettings();
+    String getVersion();
 
-    void createSchemaSettings();
-
-    void updateSchemaVersion();
-
-    void updateSchemaVersion(String version);
-
-    String getPackageSchemaVersion();
-
-    String getDbSchemaVersion();
-
+    default void apply() {
+    }
 }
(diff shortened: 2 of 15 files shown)


application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
```diff
@@ -0,0 +1,131 @@
...
+                throw new IllegalStateException("Duplicate LTS migration version: " + m.getVersion());
+            }
+        }
+        return migrations.stream()
+                .sorted(Comparator.comparing(m -> LtsVersion.parse(m.getVersion())))
+                .toList();
+    }
+
+    /** No-downtime path: per migration in (from, to] run SQL, then apply(), then record the version. */
+    public void applyMigrations(String fromVersion, String toVersion) {
```
ViacheslavKlimov, line 84: These three entry points — `applyMigrations`, `runSchemaMigrations`, `runDataMigrations` — share the same `(from, to)` signature but differ in subtle, easy-to-get-wrong ways: `applyMigrations` runs SQL + `apply()` + records the version, `runSchemaMigrations` runs SQL only, and `runDataMigrations` runs `apply()` only — and which ones record progress is only discoverable from the Javadoc. A caller could reasonably reach for `runDataMigrations` expecting the SQL to run too. Would an explicit phase enum, or names that encode the matrix (e.g. `applyFull` / `applySchemaOnly` / `applyDataOnly`), make the SQL-vs-apply-vs-record behavior harder to misuse?

application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
```diff
@@ -0,0 +1,131 @@
...
+            log.info("Applied LTS data migration {}", migration.getVersion());
+        }
+    }
+
+    private List<LtsMigration> select(String fromVersion, String toVersion) {
+        LtsVersion from = LtsVersion.parse(fromVersion);
+        LtsVersion to = LtsVersion.parse(toVersion);
+        return migrations.stream()
+                .filter(migration -> {
+                    LtsVersion version = LtsVersion.parse(migration.getVersion());
```
ViacheslavKlimov: `getVersion()` returns a `String` and the version gets re-parsed into `LtsVersion` repeatedly — here in `select()` for every migration on every call, again in the `validateAndSort` comparator, and once in the constructor. The list is tiny so this isn't a perf concern, but since the version is conceptually an `LtsVersion` and is already validated as parseable at construction, caching the parsed value alongside each migration (e.g. a small record pairing the migration with its parsed version, built once in `validateAndSort`) would let `select()` read more directly and keep parsing in one place.

application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigration.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * One LTS migration, keyed by the version it ships in (e.g. "4.2.2.3").
+ * The runner runs the version's data/upgrade/lts/&lt;version&gt;/schema_update.sql by convention
+ * (if present), then calls {@link #apply()} for any programmatic work beyond SQL.
+ * Most migrations are SQL-only and only override {@link #getVersion()}.
+ */
+public interface LtsMigration {
+
+    String getVersion();
```
ViacheslavKlimov, line 26: This `String` is doubly load-bearing: it both orders/selects the migration and is interpolated into the on-disk path `data/upgrade/lts/<version>/schema_update.sql` by `runSchemaUpdate`. That coupling between a hardcoded Java string and a directory name is invisible and unenforced — if a future migration returns `"4.3.1.4"` but the SQL lives in a dir named `4.3.4` (or vice versa), `runSchemaUpdate` just logs at trace and silently runs no SQL. Might be worth a guard, e.g. a test that scans the discovered migrations and asserts each version either has a matching dir or is intentionally SQL-less, so the convention can't drift unnoticed.

application/src/main/java/org/thingsboard/server/service/install/DefaultDatabaseSchemaSettingsService.java
```diff
@@ -123,13 +128,15 @@ private String getProductFromDb() {
...
-        String[] versionParts = getPackageSchemaVersion().split("\\.");
-
-        long major = Integer.parseInt(versionParts[0]);
-        long minor = Integer.parseInt(versionParts[1]);
-        long maintenance = Integer.parseInt(versionParts[2]);
-        long patch = Integer.parseInt(versionParts[3]);
+        return toDbVersion(getPackageSchemaVersion());
+    }
 
+    private long toDbVersion(String version) {
```
ViacheslavKlimov, line 135: `toDbVersion` hand-rolls the same "split on dots, pad missing components to four" parsing that the new `LtsVersion.parse` now does, in the same module and the same change set. Since this PR introduces `LtsVersion` specifically as the version value type, could `toDbVersion` build an `LtsVersion` and read `major()/minor()/maintenance()/patch()` off it, rather than being a third hand-written parser to keep in sync?

application/src/main/java/org/thingsboard/server/service/install/update/DefaultDataUpdateService.java
```diff
@@ -52,12 +54,13 @@ public class DefaultDataUpdateService implements DataUpdateService {
...
     private final DbUpgradeExecutorService executorService;
+    private final DatabaseSchemaSettingsService schemaSettingsService;
+    private final LtsMigrationService ltsMigrationService;
 
     @Override
     public void updateData() throws Exception {
         log.info("Updating data ...");
-        //TODO: should be cleaned after each release
-
+        ltsMigrationService.runDataMigrations(schemaSettingsService.getDbSchemaVersion(), schemaSettingsService.getPackageSchemaVersion());
```
ViacheslavKlimov, line 63: This offline data-phase wiring (and the symmetric `runSchemaMigrations` call in `SqlDatabaseUpgradeService.upgradeDatabase`) is only covered by `LtsMigrationServiceTest`, which mocks the transaction manager and calls the service directly. The integration test (`LtsMigrationIntegrationTest`) only drives `applyMigrations` — the no-downtime path. Since the offline path deliberately records no per-version progress and replays the whole range on resume, it's arguably the path most worth an end-to-end test. Could the integration test also exercise `runSchemaMigrations`/`runDataMigrations` (or the upgrade services that call them) against a real DB?

ViacheslavKlimov (commented):
## Review summary

Reviewed this PR together with its three siblings ([thingsboard#15809](https://github.com/thingsboard/thingsboard/pull/15809), [thingsboard-pe#4869](https://github.com/thingsboard/thingsboard-pe/pull/4869), [thingsboard-pe#4870](https://github.com/thingsboard/thingsboard-pe/pull/4870)). This is the CE `lts-4.2` PR that introduces the core per-version migration mechanism, so the inline comments here focus on the mechanism's design and test coverage; the 4.3-delta and PE-specific findings are on the sibling PRs.

The mechanism is clean and well-tested: half-open `(from, to]` selection, idempotent SQL, one-transaction-per-migration on the no-downtime path, and a deliberate replay-the-whole-range design on the offline path. I checked the offline ordering concern (could the schema version be recorded between the schema and data phases, making `runDataMigrations` select an empty range?) and it's fine — `updateSchemaVersion()` runs only at the very end of `ThingsboardInstallService.performInstall`, so both phases see the old DB version. No correctness bugs found; the comments below are design/maintainability/coverage suggestions.

---

application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
ViacheslavKlimov: Kept the three entry points as distinct, single-responsibility methods documented in Javadoc. The two callers invoke exactly the one they need — `SqlDatabaseUpgradeService` drives the schema phase (`runSchemaMigrations`) and `DefaultDataUpdateService` drives the data phase (`runDataMigrations`) — so a phase enum or renamed pair would churn both callers without removing the fact that the offline schema and data phases are invoked separately by their own collaborators. The part worth keeping DRY, the range+family selection, is now extracted into the package-private `isInRangeForTargetFamily` predicate shared by all three. Leaving the public API stable.

application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
ViacheslavKlimov: Done. The parsed `LtsVersion` is now cached per migration in a `VersionedMigration` record built once in `validateAndSort`; `select()` and all three runners read the cached value, so each bean's version is parsed once at construction (plus `from`/`to` once per `select()` call).

application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigration.java
ViacheslavKlimov: Done. Added a bidirectional dir-to-bean consistency guard in `LtsMigrationIntegrationTest`: it asserts every on-disk `data/upgrade/lts/<version>` directory has a registered `LtsMigration` bean, and every registered bean's version has a matching directory (or is explicitly allow-listed as SQL-less). The registration-only `V4_3_1_2Migration` also now carries a Javadoc explaining why it must exist — the runner iterates the injected beans, not the on-disk directories, so a directory with a `schema_update.sql` but no bean would be silently skipped.

application/src/main/java/org/thingsboard/server/service/install/DefaultDatabaseSchemaSettingsService.java
ViacheslavKlimov: Done. `toDbVersion` (and `normalizeVersion`) now parse via `LtsVersion.parse(...)` and read `major()/minor()/maintenance()/patch()`, so there is a single version grammar instead of a second hand-rolled parser to keep in sync.

application/src/main/java/org/thingsboard/server/service/install/update/DefaultDataUpdateService.java
ViacheslavKlimov: Done. `LtsMigrationIntegrationTest` now exercises the offline path end-to-end against the real DB: it drives `runSchemaMigrations` then `runDataMigrations` over the supported range and asserts the schema effect landed, the data `apply()` ran, and that no schema version is recorded (unlike `applyMigrations`).

PR #15809, opened 2026-06-19
LTS versioned migrations

ViacheslavKlimov:
## Summary

Same per-version LTS migration mechanism as the `lts-4.2` PR (per-version migrations selected by the range `(dbVersion, packageVersion]`, discovered as `List<LtsMigration>` beans, ordered by `LtsVersion`, run by `LtsMigrationService` across the no-downtime and offline upgrade paths). The mechanism reaches this branch **by merging** the CE `lts-4.2` feature branch (preserving commit identity for the release cascade), then layers the 4.3-family migrations on top.

## CE `lts-4.3` content (on top of the merged mechanism)

- `lts/4.3.1.2/schema_update.sql` — `calculated_field.additional_info` column (idempotent `ADD COLUMN IF NOT EXISTS`).
- `lts/4.3.1.3/schema_update.sql` — iot_hub table + indexes (same body as 4.2.2.3).
- `V4_3_1_2Migration` — SQL-only.
- `V4_3_1_3Migration` — SQL + deprecates the four obsolete system bundles' widget **types** (keeping the type entities) and deletes **only** the bundle entities.

The merged-in `4.2.2.3` migration stays present and dormant for 4.3-family upgrades (out of range). Covered by `V4_3_1_3MigrationTest` and the merged `LtsMigrationIntegrationTest`.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/install/lts/V4_3_1_3Migration.java
+++ b/application/src/main/java/org/thingsboard/server/service/install/lts/V4_3_1_3Migration.java
@@ -1,0 +1,73 @@
+/**
+ * Copyright © 2016-2026 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.service.install.lts;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+import org.thingsboard.server.common.data.id.TenantId;
+import org.thingsboard.server.common.data.widget.WidgetTypeDetails;
+import org.thingsboard.server.common.data.widget.WidgetsBundle;
+import org.thingsboard.server.dao.widget.WidgetTypeService;
+import org.thingsboard.server.dao.widget.WidgetsBundleService;
+import org.thingsboard.server.queue.util.TbCoreComponent;
+
+import java.util.List;
+
+@Slf4j
+@Component
+@TbCoreComponent
+@RequiredArgsConstructor
+public class V4_3_1_3Migration implements LtsMigration {
+
+    // Match on the bundle ALIAS, not the JSON filename stem.
+    // The "industrial_widgets" bundle shipped in a misspelled file (industial_widgets.json),
+    // but the alias inside it is correctly spelled.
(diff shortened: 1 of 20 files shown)


application/src/main/java/org/thingsboard/server/service/install/lts/V4_3_1_3Migration.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public void apply() {
+        for (String alias : OBSOLETE_BUNDLE_ALIASES) {
+            deprecateTypesAndDeleteBundle(alias);
+        }
+    }
+
+    // Marks the bundle's widget types as deprecated (keeping the type entities) and deletes ONLY the bundle entity.
+    private void deprecateTypesAndDeleteBundle(String alias) {
```
ViacheslavKlimov, line 58: `V4_3_1_3Migration` is essentially identical to `V4_2_2_3Migration` — same `OBSOLETE_BUNDLE_ALIASES`, same `apply()`, same `deprecateTypesAndDeleteBundle`, even the same comment — differing only in the version string. I understand the duplication is deliberate (4.2.x installs run 4.2.2.3, 4.3.x installs run 4.3.1.3, and both families shipped these bundles), but copy-pasting the whole body means a fix to the deprecation logic has to land in two places and can silently drift. Could the shared behavior live in a small abstract base (holding the alias list and the loop) so each concrete version only declares `getVersion()`?

ViacheslavKlimov (commented):
## Review summary

Reviewed this PR together with its three siblings ([thingsboard#15808](https://github.com/thingsboard/thingsboard/pull/15808), [thingsboard-pe#4869](https://github.com/thingsboard/thingsboard-pe/pull/4869), [thingsboard-pe#4870](https://github.com/thingsboard/thingsboard-pe/pull/4870)). This is the CE `lts-4.3` PR, which merges the 4.2 mechanism and layers `V4_3_1_2`/`V4_3_1_3` on top. The core mechanism is reviewed on the 4.2 PR (#15808) — the inline comment here is the one finding specific to the 4.3 delta.

The 4.3 SQL is idempotent (`ADD COLUMN IF NOT EXISTS`, `CREATE TABLE/INDEX IF NOT EXISTS`), and the empty `V4_3_1_2Migration` bean correctly exists to trigger its `schema_update.sql`. No correctness bugs found.

## Additional findings

These apply to shared mechanism code and are commented inline on #15808 to avoid duplication: the `applyMigrations`/`runSchemaMigrations`/`runDataMigrations` naming, the repeated `LtsVersion` re-parsing in `select()`, the `getVersion()`→directory coupling with no guard, the `toDbVersion` parser duplicating `LtsVersion`, and the offline path lacking an end-to-end test.

---

application/src/main/java/org/thingsboard/server/service/install/lts/V4_3_1_3Migration.java
ViacheslavKlimov: Keeping the duplication intentionally. The per-family migration beans are deliberately self-contained so each LTS family is independent and the release-merge cascade never couples families — that self-containment is exactly what makes the target-family filter correct. With the 4.4 data step now handled inline in `updateData()` (see the cross-family thread) rather than as a new bean, the duplication is bounded to two beans per repo (`4.2.2.3` and `4.3.1.3`) and does not grow per release, so an abstract base would re-introduce cross-family coupling for a two-class duplication. Leaving as-is.

PR #15810, opened 2026-06-19
LTS versioned migrations

ViacheslavKlimov:
Forward-port of the LTS versioned-migration mechanism to `master` (4.4).

Counterpart of the LTS PRs:
- thingsboard/thingsboard#15808 (lts-4.2)
- thingsboard/thingsboard#15809 (lts-4.3)

Brings the per-version LTS migration mechanism onto `master` via merge (preserving commit identity for the release cascade):

- `LtsVersion` value type + `LtsMigration` bean interface, discovered and ordered by `LtsMigrationService`.
- Migrations are selected by the range `(dbVersion, packageVersion]` **and** restricted to the target version's family, so older-family migrations carried onto a newer branch by the release cascade stay dormant (no cross-family double-processing).
- Wired into both the no-downtime patch path (`SystemPatchApplier`) and the offline major-upgrade path (`SqlDatabaseUpgradeService` schema phase, `DefaultDataUpdateService` data phase).

On `master` (4.4) the existing 4.2.x/4.3.x migration beans are present but dormant (target family is 4.4), so this is a no-op for 4.4 upgrades until 4.4-family migrations are added. Merged cleanly onto master; full reactor build (`mvn install -pl application --also-make`) is green.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
+++ b/application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
@@ -110,7 +110,15 @@ public class LtsMigrationService {
         return migrations.stream()
                 .filter(migration -> {
                     LtsVersion version = LtsVersion.parse(migration.getVersion());
-                    return version.compareTo(from) > 0 && version.compareTo(to) <= 0;
+                    // Run only migrations whose family matches the target version. Older-family
+                    // migrations (e.g. 4.2.x) ride onto newer-family branches (e.g. 4.3.x) via the
+                    // release-merge cascade, but each branch's own family of migrations is
+                    // self-contained (newer-family copies reproduce the older schema/data changes),
+                    // so a cross-family upgrade is fully handled by the target-family migrations.
+                    // Excluding the dormant older-family beans avoids double-processing.
+                    return version.sameFamily(to)
+                            && version.compareTo(from) > 0
+                            && version.compareTo(to) <= 0;
                 })
                 .toList();
     }
--- a/application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
+++ b/application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
@@ -29,6 +29,7 @@ import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
@@ -41,11 +42,14 @@ public class LtsMigrationService {
 
     private static final String SCHEMA_UPDATE_SQL = "schema_update.sql";
 
+    /** A migration paired with its parsed version, so the version is parsed exactly once per bean. */
+    private record VersionedMigration(LtsVersion version, LtsMigration migration) {}
+
     private final JdbcTemplate jdbcTemplate;
     private final InstallScripts installScripts;
     private final DatabaseSchemaSettingsService schemaSettingsService;
     private final TransactionTemplate transactionTemplate;
-    private final List<LtsMigration> migrations;
+    private final List<VersionedMigration> migrations;
 
     public LtsMigrationService(JdbcTemplate jdbcTemplate,
                                InstallScripts installScripts,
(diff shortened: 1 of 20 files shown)


application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
```diff
@@ -0,0 +1,139 @@
...
+        return migrations.stream()
+                .filter(migration -> {
+                    LtsVersion version = LtsVersion.parse(migration.getVersion());
+                    // Run only migrations whose family matches the target version. Older-family
+                    // migrations (e.g. 4.2.x) ride onto newer-family branches (e.g. 4.3.x) via the
+                    // release-merge cascade, but each branch's own family of migrations is
+                    // self-contained (newer-family copies reproduce the older schema/data changes),
+                    // so a cross-family upgrade is fully handled by the target-family migrations.
+                    // Excluding the dormant older-family beans avoids double-processing.
+                    return version.sameFamily(to)
```
ViacheslavKlimov: This `sameFamily(to)` restriction is the key new behavior in the master port, and it's the right instinct for avoiding the cross-family double-processing the lts branches have. But combined with the supported-upgrade set it leaves a gap on the data side. `SUPPORTED_VERSIONS_FOR_UPGRADE` allows upgrading to 4.4 from any `4.3.0.x` or `4.3.1.x`, including versions *below* 4.3.1.3. The 4.3.1.x **schema** deltas are safe because `data/upgrade/basic/schema_update.sql` re-applies them unconditionally — but the 4.3.1.3 **data** migration (deprecating and removing the obsolete system widget bundles) lives only in `V4_3_1_3Migration`, which is now dormant on master (target family 4.4). So a user upgrading from, say, 4.3.1.2 or 4.3.0.x — who never ran the 4.3.1.3 data migration while on 4.3 — would skip it permanently, and nothing reproduces it. The "newer-family migrations reproduce older changes" invariant in the comment holds for schema but not for these data migrations, since there are no 4.4-family migrations yet. Is the plan to add 4.4-family copies of the still-needed data migrations, or to tighten the supported-from set to the latest 4.3.1.x? As written, the obsolete-bundle cleanup is lost for sub-4.3.1.3 upgrade sources.

application/src/test/java/org/thingsboard/server/service/install/lts/LtsMigrationServiceTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+        assertEquals(List.of("4.2.2.3", "4.2.2.4"), applied);
+        verify(jdbcTemplate).execute("SELECT 1;");
+        verify(jdbcTemplate).execute("SELECT 2;");
+        verify(schemaSettingsService).updateSchemaVersion("4.2.2.3");
+        verify(schemaSettingsService).updateSchemaVersion("4.2.2.4");
+    }
+
+    @Test
+    void skipsMigrationsOutsideTargetFamilyOnCrossFamilyUpgrade() {
```
ViacheslavKlimov, line 119: Good to see a cross-family case for the new filter. Two suggestions: (1) it exercises the family restriction only through `runDataMigrations`, but the filter lives in the shared `select()` used by all three runners — a future refactor that inlines selection could silently drop it from one path, so asserting it through `applyMigrations`/`runSchemaMigrations` too (or noting that `select()` is the single tested seam) would be safer. (2) The family rule is currently expressed only as a comment inside the filter lambda; extracting it into a named predicate (e.g. `isInRangeForTargetFamily(version, from, to)`) would make it independently unit-testable and put the intent in code. Since `LtsMigrationIntegrationTest` stays within 4.2, this unit test is the only guard on the most load-bearing new logic — a cross-family integration scenario would be stronger.

ViacheslavKlimov (commented):
## Review summary

Reviewed as the `master`/4.4 forward-port in the LTS-migration cascade, alongside its PE counterpart [thingsboard-pe#4871](https://github.com/thingsboard/thingsboard-pe/pull/4871) and the already-reviewed lts PRs (#15808, #15809). The mechanism merges cleanly and the migration beans/maps are byte-identical to lts-4.3. The one genuinely new thing here is the `sameFamily(to)` restriction in `select()`, so the inline comments focus there.

I verified the family restriction is safe on the **schema** side: although it makes the 4.3.1.x migrations dormant on 4.4, the 4.3.1.x schema deltas (`calculated_field.additional_info`, `role.excluded_permissions`) are re-applied unconditionally by `data/upgrade/basic/schema_update.sql`, and `iot_hub_installed_item` is a missing-table create. The open question is on the **data** side — see inline.

## Additional findings

These are the same framework-level points already commented on the lts PRs (the code is identical via merge), noted here only so this PR stands alone — no need to re-address if handled upstream:
- The `V4_3_1_3Migration` / `V4_2_2_3Migration` beans (and their test classes) are near-verbatim duplicates differing only in version + data — a shared abstract base would remove the copy-paste. Raised on #15809.
- `applyMigrations` / `runSchemaMigrations` / `runDataMigrations` share one signature with materially different effects; and `select()` re-parses each `LtsVersion` on every call. Raised on #15808.

Cross-PR consistency note: this `sameFamily` restriction was **not** applied on the lts-4.2/lts-4.3 PRs (#15808/#15809) — their `select()` is range-only — so those branches still cross-process older-family migrations on a cross-family offline upgrade. Worth deciding whether the restriction should be consistent across the cascade.

---

application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
ViacheslavKlimov: Fixed. On 4.4 the dormant 4.3.1.3 data migration is now run explicitly: `DefaultDataUpdateService.updateData()` calls `V4_3_1_3Migration.apply()` unconditionally after `runDataMigrations(...)`, so every supported 4.3.x to 4.4 upgrade (including sources below 4.3.1.3) gets the obsolete-bundle cleanup — and, on PE, the legacy solution-template to IoT Hub migration. It is idempotent (bundle deletion checks existence first; the solution-template step consumes-once via tenant-attribute deletion), so re-running it for a source already at 4.3.1.3 is a safe no-op. We kept it inline in the data-update flow rather than adding a 4.4-family bean, since this is a one-time release data migration, not an LTS-line patch.

application/src/test/java/org/thingsboard/server/service/install/lts/LtsMigrationServiceTest.java
ViacheslavKlimov: Done. Extracted the rule into a package-private `isInRangeForTargetFamily(version, from, to)` with a direct unit test covering the boundaries (same-family in range, older-family in numeric range, the `to` boundary, and at/below `from`). Also added cross-family-skip assertions through `applyMigrations` and `runSchemaMigrations`, not just `runDataMigrations`, since all three share `select()`.

PR #15822, opened 2026-06-22
Fixed time unit input validation and unwanted dirty state when loading a value

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/shared/components/time-unit-input.component.ts
+++ b/ui-ngx/src/app/shared/components/time-unit-input.component.ts
@@ -209,14 +209,11 @@ export class TimeUnitInputComponent implements ControlValueAccessor, Validator, 
         this.timeInputForm.patchValue(this.parseTime(sec), {emitEvent: false});
         this.modelValue = sec;
       } else {
-        this.timeInputForm.patchValue({
-          time: 0,
-          timeUnit: TimeUnit.SECONDS
-        }, {emitEvent: false});
+        this.timeInputForm.patchValue(this.secondsModel(0), {emitEvent: false});
         this.modelValue = 0;
       }
+      this.refreshTimeValidators();
     }
-    this.refreshTimeValidators();
   }
 
   validate(): ValidationErrors | null {
@@ -243,8 +240,12 @@ export class TimeUnitInputComponent implements ControlValueAccessor, Validator, 
         }
       }
     }
+    return this.secondsModel(value);
+  }
+
+  private secondsModel(time: number): TimeUnitInputModel {
     return {
-      time: value,
+      time,
       timeUnit: TimeUnit.SECONDS
     };
   }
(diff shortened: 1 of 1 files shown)


ui-ngx/src/app/shared/components/time-unit-input.component.ts
```diff
@@ -216,6 +216,7 @@ export class TimeUnitInputComponent implements ControlValueAccessor, Validator,
         this.modelValue = 0;
       }
     }
+    this.refreshTimeValidators();
```
vvlladd28: `refreshTimeValidators()` now runs on every `writeValue`, including when `sec === this.modelValue` and the `if` body is skipped entirely (nothing actually changed). Angular re-invokes `writeValue` fairly often, so each no-op write still does a `setValidators` + `updateValueAndValidity` pass. It's cheap here so not a blocker — but it reads a little surprising. Was the intent to refresh only when the patched value/unit actually changed, or is the unconditional refresh deliberate (e.g. covering inputs that changed independently)? If the latter, a one-line comment would save the next reader the head-scratch.

vvlladd28 (commented):
## Review summary

Reviewed 1 changed file in [Fixed time unit input validation and unwanted dirty state when loading a value](https://github.com/thingsboard/thingsboard/pull/15822). Left 2 comment(s) inline.

The fix itself looks correct: switching the valid branch of `writeValue` to `{emitEvent: false}` stops the CVA from calling `propagateChange` during a programmatic write (the source of the unwanted dirty state), and now matches the else-branch which already suppressed events. The explicit `refreshTimeValidators()` call also closes a real gap — validators previously only re-ran when the `timeUnit` actually changed (via `timeUnit.valueChanges`), so a write that kept the same unit left stale validators. And the `parseTime` fallback resolves a genuine desync where a fall-through returned `undefined`, `patchValue(undefined)` no-opped, yet `modelValue` was still set. The two comments below are maintainability observations, not blockers.

---

ui-ngx/src/app/shared/components/time-unit-input.component.ts
```diff
@@ -242,6 +243,10 @@ export class TimeUnitInputComponent implements ControlValueAccessor, Validator,
         }
       }
     }
+    return {
```
vvlladd28, line 247: This fallback is now the third spot in the file that hardcodes a `SECONDS`-defaulted model — the others being the `writeValue` else-branch (`{time: 0, timeUnit: TimeUnit.SECONDS}`) and the form's initial value (`timeUnit: [TimeUnit.SECONDS]`). Now that `parseTime` reliably returns a model for any numeric input, would it be cleaner to route the zero/null case through `parseTime` too, or extract a small `secondsModel(time)` helper, so the default unit lives in one place? As written, changing the default unit later means hunting down several spots.

vvlladd28 (approved):

PR #15840, opened 2026-06-25
Feature/monitoring CoAPS port and and reconnect lifecycle

pon0marev:
## Pull Request description

  - Add CoAPS (CoAP over DTLS) transport health checker using Scandium with
    AnonymousCertificateProvider — server uses WANTED mode so no client cert is required
  - Add TlsConfig supporting truststore, keystore and auto-download of server certificate
    for CoAPS; support for mutual TLS via keystore is wired but optional
  - Reconnect on every probe across all transports (CoAP, CoAPS, MQTT) so certificate
    changes are detected immediately
  - Add CoAPS latency series to the monitoring dashboard

```
2026-06-25 14:58:16,274 [main] INFO  o.t.m.service.BaseMonitoringService - Initializing CoapTransportHealthChecker for coap://coap.ххххххх.cloud:5683
2026-06-25 14:58:17,041 [main] INFO  o.t.m.service.BaseMonitoringService - Initializing CoapTransportHealthChecker for coaps://coap.ххххххх.cloud:5684
2026-06-25 14:58:17,667 [main] INFO  o.t.m.service.BaseMonitoringService - Initializing HttpTransportHealthChecker for http://ххххххх.cloud
2026-06-25 14:58:18,281 [main] INFO  o.t.m.service.BaseMonitoringService - Initializing MqttTransportHealthChecker for ssl://ххххххх.cloud:8883
2026-06-25 14:58:18,879 [main] INFO  o.t.m.ThingsboardMonitoringApplication - Scheduling initialDelay 0, fixedDelay 60000 for monitoring 'TransportsMonitoringService' 
2026-06-25 14:58:18,882 [monitoring-1-thread-1] INFO  o.t.m.service.BaseMonitoringService - Starting transports check
2026-06-25 14:58:19,084 [main] WARN  o.t.m.s.MonitoringEntityService - Dashboard is not assigned to public customer. Public link can't be generated.
2026-06-25 14:58:19,612 [monitoring-1-thread-1] INFO  o.t.m.service.MonitoringReporter - Login is OK
2026-06-25 14:58:20,438 [monitoring-1-thread-1] INFO  o.t.m.service.MonitoringReporter - WS Connect is OK
2026-06-25 14:58:20,688 [monitoring-1-thread-1] INFO  o.t.m.service.MonitoringReporter - WS Subscribe is OK
2026-06-25 14:58:20,691 [monitoring-1-thread-1] DEBUG o.t.m.s.t.i.CoapTransportHealthChecker - Connecting COAP client to coap://coap.ххххххх.cloud:5683
2026-06-25 14:58:21,078 [monitoring-1-thread-1] DEBUG o.t.m.s.t.i.CoapTransportHealthChecker - Disconnected COAP client
2026-06-25 14:58:21,174 [monitoring-1-thread-1] INFO  o.t.m.service.MonitoringReporter - *CoAP* (coap://coap.ххххххх.cloud:5683) is OK
2026-06-25 14:58:21,241 [monitoring-1-thread-1] DEBUG o.t.m.s.t.i.CoapTransportHealthChecker - Connecting COAPS client to coaps://coap.ххххххх.cloud:5684
2026-06-25 14:58:22,073 [monitoring-1-thread-1] DEBUG o.t.m.s.t.i.CoapTransportHealthChecker - Disconnected COAPS client
2026-06-25 14:58:22,153 [monitoring-1-thread-1] INFO  o.t.m.service.MonitoringReporter - *CoAPS* (coaps://coap.ххххххх.cloud:5684) is OK
2026-06-25 14:58:22,161 [monitoring-1-thread-1] DEBUG o.t.m.s.t.i.HttpTransportHealthChecker - Initialized HTTP client
2026-06-25 14:58:22,654 [monitoring-1-thread-1] INFO  o.t.m.service.MonitoringReporter - *HTTP* (http://ххххххх.cloud) is OK
2026-06-25 14:58:23,736 [monitoring-1-thread-1] DEBUG o.t.m.s.t.i.MqttTransportHealthChecker - Connecting MQTT client to ssl://ххххххх.cloud:8883
2026-06-25 14:58:23,937 [monitoring-1-thread-1] DEBUG o.t.m.s.t.i.MqttTransportHealthChecker - Disconnected MQTT client
2026-06-25 14:58:23,991 [monitoring-1-thread-1] INFO  o.t.m.service.MonitoringReporter - *MQTT* (ssl://ххххххх.cloud:8883) is OK
2026-06-25 14:58:24,636 [monitoring-1-thread-1] INFO  o.t.m.service.MonitoringReporter - Monitoring is OK
```

## General checklist

## Back-End feature checklist


smatvienko-tb:
image

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -29,55 +45,145 @@
...
+        }
+
+        String accessToken = target.getDevice().getCredentials().getCredentialsId();
+        String uri = target.getBaseUrl() + "/api/v1/" + accessToken + "/telemetry";
+        coapClient = new CoapClient(uri);
+        if (target.getBaseUrl().startsWith("coaps")) {
+            Configuration dtlsConfiguration = new Configuration();
+            dtlsConfiguration.set(DtlsConfig.DTLS_ROLE, DtlsConfig.DtlsRole.CLIENT_ONLY);
+            dtlsConfiguration.set(DtlsConfig.DTLS_USE_SERVER_NAME_INDICATION, true);
+            dtlsConfiguration.set(DtlsConfig.DTLS_VERIFY_SERVER_CERTIFICATES_SUBJECT, false);
```
smatvienko-tb: 🔒 **Security** — this disables server-certificate **subject/hostname** verification.

`DTLS_VERIFY_SERVER_CERTIFICATES_SUBJECT` defaults to `true`. Setting it `false` means the client accepts any certificate from any trusted public CA for *any* hostname — e.g. a valid Let's Encrypt cert issued for an attacker's own domain would pass, so the connection is MITM-able.

SNI is enabled on the line above, so hostname validation will work correctly. Please just **remove this line** and let it keep the secure default.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -29,55 +45,145 @@
...
+import java.util.List;
 
 @Component
 @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
 @Slf4j
 public class CoapTransportHealthChecker extends TransportHealthChecker<CoapTransportMonitoringConfig> {
 
+    private static final X509Certificate[] TRUSTED_CERTS;
+
     static {
```
smatvienko-tb, line 43: This inline trust-store loading should move to `common/util` → `SslUtil` (`monitoring` already depends on that module, and `SslUtil` already hosts the cert/key helpers).

The idiom itself is correct and canonical — Scandium doesn't read `javax.net.ssl.trustStore` and can't consume a JSSE `TrustManager`, so extracting `getAcceptedIssuers()` is the right approach — it's just reusable security plumbing rather than health-checker logic:

```java
@SneakyThrows
public static X509Certificate[] getDefaultTrustedCertificates() {
    TrustManagerFactory tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
    tmf.init((KeyStore) null); // JVM default trust store (cacerts)
    return ((X509TrustManager) tmf.getTrustManagers()[0]).getAcceptedIssuers();
}
```

Then this whole `static {}` block and the `javax.net.ssl.*` / `java.security.KeyStore` imports drop out of this class.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -29,55 +45,145 @@
...
+
+        String accessToken = target.getDevice().getCredentials().getCredentialsId();
+        String uri = target.getBaseUrl() + "/api/v1/" + accessToken + "/telemetry";
+        coapClient = new CoapClient(uri);
+        if (target.getBaseUrl().startsWith("coaps")) {
+            Configuration dtlsConfiguration = new Configuration();
+            dtlsConfiguration.set(DtlsConfig.DTLS_ROLE, DtlsConfig.DtlsRole.CLIENT_ONLY);
+            dtlsConfiguration.set(DtlsConfig.DTLS_USE_SERVER_NAME_INDICATION, true);
+            dtlsConfiguration.set(DtlsConfig.DTLS_VERIFY_SERVER_CERTIFICATES_SUBJECT, false);
+            dtlsConfiguration.setTransient(DtlsConfig.DTLS_SIGNATURE_AND_HASH_ALGORITHMS);
```
smatvienko-tb: Two things here:

1. `setTransient(...)` is a no-op — this `Configuration` is in-memory and never written to a file.
2. The list is worth **keeping but broadening**. It's load-bearing (Scandium's built-in default is only `[SHA256_WITH_ECDSA, SHA256_WITH_RSA]`, so `SHA384_WITH_ECDSA` is genuinely required for ECDSA chains), but too narrow for a general probe — a server whose chain uses `SHA384withRSA`/`SHA512withRSA` would fail the handshake. Recommend covering both families across hash sizes:

```java
List.of(
    SignatureAndHashAlgorithm.SHA256_WITH_ECDSA,
    SignatureAndHashAlgorithm.SHA384_WITH_ECDSA,
    SignatureAndHashAlgorithm.valueOf("SHA512withECDSA"),
    SignatureAndHashAlgorithm.SHA256_WITH_RSA,
    SignatureAndHashAlgorithm.valueOf("SHA384withRSA"),
    SignatureAndHashAlgorithm.valueOf("SHA512withRSA"),
    SignatureAndHashAlgorithm.INTRINSIC_WITH_ED25519)
```

`valueOf("SHA384withRSA")` covers the combos that lack a named constant.

smatvienko-tb (commented):
Thanks for adding CoAPS and the reconnect lifecycle — the health-checker refactor and per-probe reconnect are solid improvements.

Almost all of my feedback is about the DTLS setup, which builds inline what should be a small, reusable, and (in one respect) more **secure** helper. One item is a genuine security issue; the rest are simplifications. In short, the whole `coaps` branch can collapse to a one-line call against a shared factory in `common/util`'s `SslUtil` (already a dependency of `monitoring`).

### 1. 🔒 Hostname verification is disabled (please fix)
`DTLS_VERIFY_SERVER_CERTIFICATES_SUBJECT = false` (line 102) turns off subject/SAN checking. It defaults to `true`, so this line actively weakens security: the client will accept **any** certificate from **any** trusted public CA for **any** hostname — a valid Let's Encrypt cert for an attacker's own domain would pass, which is MITM-able. SNI is already enabled on the line above, so hostname validation works correctly once this line is removed. See inline comment.

### 2. Move the trust-store loading to `SslUtil`
The `TrustManagerFactory ... init((KeyStore) null) ... getAcceptedIssuers()` idiom is actually the correct, canonical way to read the JVM default trust store — Scandium doesn't read `javax.net.ssl.trustStore` and can't consume a `TrustManager`, so extracting the anchors is unavoidable. But it's reusable security plumbing, not health-checker logic. Proposed `SslUtil.getDefaultTrustedCertificates()` — inline comment has the snippet.

### 3. `AnonymousCertificateProvider` can be deleted
With `DTLS_ROLE = CLIENT_ONLY` the client presents no certificate, so the empty 24-line identity provider is redundant — just omit `setCertificateIdentityProvider(...)`.

### 4. Signature algorithms: keep, but broaden (and drop `setTransient`)
This list is load-bearing — Scandium's built-in default is only `[SHA256_WITH_ECDSA, SHA256_WITH_RSA]`, so `SHA384_WITH_ECDSA` is genuinely needed (Let's Encrypt E-series intermediates sign with ECDSA-SHA384). But it's too narrow for a general probe: an endpoint whose chain uses `SHA384withRSA`/`SHA512withRSA` would fail the handshake. Recommend a comprehensive list. `setTransient(...)` is a no-op on this in-memory `Configuration`.

### Putting it together
```java
// common/util → SslUtil.java  (reusable, secure DTLS client over the JVM default trust store)
private static final List<SignatureAndHashAlgorithm> CLIENT_SIG_ALGS = List.of(
        SignatureAndHashAlgorithm.SHA256_WITH_ECDSA,
        SignatureAndHashAlgorithm.SHA384_WITH_ECDSA,
        SignatureAndHashAlgorithm.valueOf("SHA512withECDSA"),
        SignatureAndHashAlgorithm.SHA256_WITH_RSA,
        SignatureAndHashAlgorithm.valueOf("SHA384withRSA"),
        SignatureAndHashAlgorithm.valueOf("SHA512withRSA"),
        SignatureAndHashAlgorithm.INTRINSIC_WITH_ED25519);

@SneakyThrows
public static X509Certificate[] getDefaultTrustedCertificates() {
    TrustManagerFactory tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
    tmf.init((KeyStore) null); // JVM default trust store (cacerts)
    return ((X509TrustManager) tmf.getTrustManagers()[0]).getAcceptedIssuers();
}

public static DTLSConnector defaultDtlsClientConnector() {
    Configuration config = new Configuration();
    config.set(DtlsConfig.DTLS_ROLE, DtlsRole.CLIENT_ONLY);            // no client cert required
    config.set(DtlsConfig.DTLS_USE_SERVER_NAME_INDICATION, true);      // select the correct server cert
    config.set(DtlsConfig.DTLS_SIGNATURE_AND_HASH_ALGORITHMS, CLIENT_SIG_ALGS);
    // DTLS_VERIFY_SERVER_CERTIFICATES_SUBJECT stays at its default (true) → hostname is validated
    return new DTLSConnector(DtlsConnectorConfig.builder(config)
            .setAdvancedCertificateVerifier(StaticNewAdvancedCertificateVerifier.builder()
                    .setTrustedCertificates(getDefaultTrustedCertificates())
                    .build())
            .build());
}
```
Call site in `CoapTransportHealthChecker` then becomes:
```java
if (target.getBaseUrl().startsWith("coaps")) {
    coapEndpoint = new CoapEndpoint.Builder()
            .setConnector(SslUtil.defaultDtlsClientConnector())
            .build();
    coapClient.setEndpoint(coapEndpoint);
}
```
That removes the static block, the `AnonymousCertificateProvider`, and the `javax.net.ssl` / `KeyStore` imports from this class, while making the client *more* secure (hostname validation back on) and robust against any public-CA chain. (One-time module registration — `SystemConfig`/`CoapConfig`/`DtlsConfig` — can live in `SslUtil`'s static initializer.)

Nice work overall — these are refinements on top of a good feature.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -29,55 +45,145 @@
...
+            dtlsConfiguration.set(DtlsConfig.DTLS_SIGNATURE_AND_HASH_ALGORITHMS, Arrays.asList(
+                    SignatureAndHashAlgorithm.SHA256_WITH_ECDSA,
+                    SignatureAndHashAlgorithm.SHA384_WITH_ECDSA,
+                    SignatureAndHashAlgorithm.SHA256_WITH_RSA));
+
+            DtlsConnectorConfig.Builder dtlsConfigBuilder = DtlsConnectorConfig.builder(dtlsConfiguration);
+            dtlsConfigBuilder.setCertificateIdentityProvider(new AnonymousCertificateProvider());
```
smatvienko-tb: With `DTLS_ROLE = CLIENT_ONLY` the client doesn't present a certificate, so an identity provider isn't needed — you can drop this line and delete `AnonymousCertificateProvider` below. Worth one handshake to confirm `build()` is happy without it, but this is the standard Scandium verify-only client pattern.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -29,55 +45,145 @@
...
         }
     }
 
     @Override
     protected TransportType getTransportType() {
-        return TransportType.COAP;
+        return config.getTransportType();
+    }
+
+    private static class AnonymousCertificateProvider implements CertificateProvider {
```
smatvienko-tb: This whole class can be removed — see the note on the `setCertificateIdentityProvider(...)` call above. A `CLIENT_ONLY` verify-only client presents no certificate, so an empty identity provider is redundant. (Side note: the `EC, RSA` here is the *client's own* key type, unrelated to verifying the server.)

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -29,55 +45,145 @@
...
+        if (target.getBaseUrl().startsWith("coaps")) {
+            Configuration dtlsConfiguration = new Configuration();
+            dtlsConfiguration.set(DtlsConfig.DTLS_ROLE, DtlsConfig.DtlsRole.CLIENT_ONLY);
+            dtlsConfiguration.set(DtlsConfig.DTLS_USE_SERVER_NAME_INDICATION, true);
+            dtlsConfiguration.set(DtlsConfig.DTLS_VERIFY_SERVER_CERTIFICATES_SUBJECT, false);
+            dtlsConfiguration.setTransient(DtlsConfig.DTLS_SIGNATURE_AND_HASH_ALGORITHMS);
+            dtlsConfiguration.set(DtlsConfig.DTLS_SIGNATURE_AND_HASH_ALGORITHMS, Arrays.asList(
+                    SignatureAndHashAlgorithm.SHA256_WITH_ECDSA,
+                    SignatureAndHashAlgorithm.SHA384_WITH_ECDSA,
+                    SignatureAndHashAlgorithm.SHA256_WITH_RSA));
```
smatvienko-tb: Combined fix for lines 102–107 — restores hostname verification (drops the `DTLS_VERIFY_SERVER_CERTIFICATES_SUBJECT, false` line), removes the no-op `setTransient(...)`, and broadens the signature algorithms so any public-CA chain verifies:

```suggestion
            dtlsConfiguration.set(DtlsConfig.DTLS_SIGNATURE_AND_HASH_ALGORITHMS, List.of(
                    SignatureAndHashAlgorithm.SHA256_WITH_ECDSA,
                    SignatureAndHashAlgorithm.SHA384_WITH_ECDSA,
                    SignatureAndHashAlgorithm.valueOf("SHA512withECDSA"),
                    SignatureAndHashAlgorithm.SHA256_WITH_RSA,
                    SignatureAndHashAlgorithm.valueOf("SHA384withRSA"),
                    SignatureAndHashAlgorithm.valueOf("SHA512withRSA"),
                    SignatureAndHashAlgorithm.INTRINSIC_WITH_ED25519));
```

After applying, drop the now-unused `import java.util.Arrays;`.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
smatvienko-tb: Not needed for a `CLIENT_ONLY` verify-only client (it presents no certificate) — delete this line and the `AnonymousCertificateProvider` class below:

```suggestion
```

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -29,55 +45,145 @@
...
+                List<X500Principal> issuers, ServerNames serverNames,
+                List<CipherSuite.CertificateKeyAlgorithm> keyAlgorithms,
+                List<SignatureAndHashAlgorithm> signatureAndHashAlgorithms,
+                List<XECDHECryptography.SupportedGroup> curves) {
+            return new CertificateIdentityResult(cid);
+        }
+
+        @Override
+        public void setResultHandler(HandshakeResultHandler handler) {}
     }
```
smatvienko-tb: Delete the redundant identity provider (see note on the `setCertificateIdentityProvider(...)` call):

```suggestion
```

Then remove its now-unused imports: `CertificateProvider`, `CertificateIdentityResult`, `ConnectionId`, `HandshakeResultHandler`, `XECDHECryptography`, `ServerNames`, `X500Principal`, `CipherSuite`.

smatvienko-tb (commented):
Adding applyable code suggestions for the in-file changes from the previous review (the trust-store move to `SslUtil` is cross-file, so it stays as a note there).

smatvienko-tb (commented):
## Review summary

Reviewed 12 changed files in [Feature/monitoring CoAPS port and and reconnect lifecycle](https://github.com/thingsboard/thingsboard/pull/15840). Left 15 comment(s) inline. The DTLS/wildcard-verification core looks solid — the new `WildcardAwareCertificateVerifierTest` passes locally (7/7), and the Scandium configuration/module-registration semantics and the Paho/Californium default-port claims all check out against the 3.12.1/1.2.5 sources. The main functional gap: a `coaps://` target shares its latency key and "CoAP" label with the plain `coap://` target, so no distinct CoAPS latency series is actually produced; and the new yml comment advertising port-less / `coaps://` LwM2M URLs doesn't match what the LwM2M client supports. Note the PR description looks stale relative to the final code (TlsConfig/keystore auto-download, dashboard changes, and the "*CoAPS*" log labels were removed/changed in later commits).

## Additional findings

- **monitoring/src/main/java/org/thingsboard/monitoring/service/BaseMonitoringService.java:205** — In the stale-associate cleanup, `stopHealthChecker(healthChecker)` stops the *parent* checker instead of the stale associate (`associates.get(url)`): the parent's client is destroyed and the parent's device ID is removed from `devices` (so it drops out of the WS telemetry subscription and the parent target starts failing with "No WS update arrived"), while the stale associate's client — for CoAPS a DTLS endpoint holding a socket and threads — is never shut down and leaks. Pre-existing, but this PR's cert-rotation/IP-churn focus makes the path more likely to fire; worth fixing alongside.

---

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -38,46 +41,79 @@ public class CoapTransportHealthChecker extends TransportHealthChecker<CoapTrans
...
             coapClient = null;
-            log.info("Disconnected CoAP client");
+            log.debug("Disconnected {} client", getTransportType());
         }
     }
 
     @Override
     protected TransportType getTransportType() {
-        return TransportType.COAP;
+        return config.getTransportType();
```
smatvienko-tb: `config.getTransportType()` still returns `TransportType.COAP`, so this change is a no-op — and it points at a real gap: with both a `coap://` and a `coaps://` target configured (the setup shown in the PR description's log), the two checkers produce the same `getKey()` = `coapTransport`, so both report `coapTransportRequest`/`coapTransportWsUpdate` and the later value overwrites the earlier in `MonitoringReporter.latencies` every cycle. There is no distinct CoAPS latency series, and `TransportInfo` labels both targets "CoAP" — the `*CoAPS*` labels in the PR description's log can't be produced by this code. If a separate CoAPS series is the goal (the description mentions adding one to the dashboard), the key needs to incorporate the scheme — e.g. a dedicated COAPS transport type or a scheme-aware `getKey()`. Same story for a `tcp://` + `ssl://` MQTT target pair.

monitoring/src/main/resources/tb-monitoring.yml
```diff
@@ -102,7 +108,8 @@ monitoring:
       # LwM2M request timeout in milliseconds
       request_timeout_ms: '${LWM2M_REQUEST_TIMEOUT_MS:4000}'
       targets:
-          # LwM2M transport base url, coap://DOMAIN:5685 by default
+          # LwM2M transport base url. Port is optional — if omitted, defaults by scheme (coap: 5685, coaps: 5686).
```
smatvienko-tb: This comment promises two things the LwM2M monitoring client doesn't deliver. A port-less URL actually breaks it: `Lwm2mClient.initClient()` parses the port with `StringUtils.substringAfterLast(serverUri, ":")`, which for `coap://DOMAIN` yields `//DOMAIN` and throws `NumberFormatException`, so every probe fails — and 5685/5686 aren't scheme defaults anywhere in the code (Californium's coap default would be 5683). `coaps://` is a dead end too: the client registers with `noSec(...)` and the monitoring device is created with NoSec LwM2M credentials, so no DTLS handshake can succeed. Suggest aligning this comment with what the client actually supports (explicit port, coap only) until LwM2M DTLS support lands.

common/util/pom.xml
```diff
@@ -123,6 +123,14 @@
             <groupId>com.github.ben-manes.caffeine</groupId>
             <artifactId>caffeine</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.californium</groupId>
+            <artifactId>scandium</artifactId>
```
smatvienko-tb: Adding scandium and httpclient5 here makes them transitive dependencies of nearly every module in the project (dao, application, rule-engine, rest-client, etc. all pull in `common/util`), yet the only consumer of the new DTLS code is the monitoring microservice — which already depends on scandium directly and also carries httpclient 4.x, so this puts a second HttpClient major version on its classpath. Could `defaultDtlsClientConnector()`/`WildcardAwareCertificateVerifier` live in the monitoring module (or a CoAP-specific module) instead, so the shared base module stays lean?

monitoring/src/main/java/org/thingsboard/monitoring/util/WildcardAwareCertificateVerifier.java
```diff
@@ -0,0 +1,191 @@
...
+    @Override
+    public CertificateVerificationResult verifyCertificate(ConnectionId cid, ServerNames serverNames, InetSocketAddress remotePeer,
+            boolean clientUsage, boolean verifySubject, boolean truncateCertificatePath, CertificateMessage message) {
+        CertificateVerificationResult result = delegate.verifyCertificate(cid, serverNames, remotePeer,
+                clientUsage, false, truncateCertificatePath, message);
+        if (!verifySubject || result.getException() != null || result.getCertificatePath() == null) {
+            return result;
+        }
+
+        X509Certificate leafCertificate = (X509Certificate) result.getCertificatePath().getCertificates().get(0);
```
smatvienko-tb, line 87: `StaticNewAdvancedCertificateVerifier` can return a result with an *empty* certificate path and no exception — it skips validation entirely when the certificate message is empty (the `!message.isEmpty()` branch), which happens on the server side with wanted/optional client auth. `getCertificates().get(0)` would then throw `IndexOutOfBoundsException` inside the handshake. Today the connector is CLIENT_ONLY, so a server certificate is always present, but since this class sits in `common/util` as a general-purpose verifier, an `isEmpty()` guard (returning the delegate result) would be a cheap safety net against future server-side reuse.

common/util/src/main/java/org/thingsboard/common/util/SslUtil.java
```diff
@@ -51,10 +60,23 @@ public class SslUtil {
...
+            SignatureAndHashAlgorithm.SHA256_WITH_RSA,
+            SignatureAndHashAlgorithm.valueOf("SHA384withRSA"),
+            SignatureAndHashAlgorithm.valueOf("SHA512withRSA"),
+            SignatureAndHashAlgorithm.INTRINSIC_WITH_ED25519);
+
     static {
         if (Security.getProvider(BouncyCastleProvider.PROVIDER_NAME) == null) {
             Security.addProvider(DEFAULT_PROVIDER);
         }
+        SystemConfig.register();
```
smatvienko-tb: Putting `SystemConfig.register()` in SslUtil's static initializer means every class that touches this utility — and it's used codebase-wide for plain PEM/key parsing — now triggers Californium module registration as a hidden side effect. Registration responsibility is also split between here and `CoapTransportHealthChecker`'s static block (`SystemConfig` + `CoapConfig`). Could the Californium setup be scoped to the code that actually needs it, e.g. inside `defaultDtlsClientConnector()` or a dedicated DTLS factory class?

common/util/src/main/java/org/thingsboard/common/util/SslUtil.java
```diff
@@ -134,4 +156,23 @@ public static char[] getPassword(String passStr) {
...
     }
 
+    @SneakyThrows
+    public static X509Certificate[] getDefaultTrustedCertificates() {
+        TrustManagerFactory tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+        tmf.init((KeyStore) null); // JVM default trust store (cacerts)
+        return ((X509TrustManager) tmf.getTrustManagers()[0]).getAcceptedIssuers();
+    }
+
+    public static DTLSConnector defaultDtlsClientConnector() {
```
smatvienko-tb: SslUtil has so far been a set of generic PEM/key parsing helpers; this adds a Californium-specific connector factory with a baked-in trust source and hostname-verification policy. The name also doesn't convey the most important behavioral fact — that it trusts only the JVM default cacerts — so a future caller pointing it at a server with a private-CA certificate will be surprised at handshake time. If it stays here, would a name like `jvmTrustedDtlsClientConnector()` (or javadoc stating the trust source and the wildcard-aware hostname check) make it harder to misuse?

monitoring/src/test/java/org/thingsboard/monitoring/util/WildcardAwareCertificateVerifierTest.java
```diff
@@ -0,0 +1,158 @@
...
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.Security;
+import java.security.cert.X509Certificate;
+import java.time.Instant;
+import java.util.Date;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class WildcardAwareCertificateVerifierTest {
```
smatvienko-tb, line 58: These tests only exercise the static `verifySubject` helper; the `verifyCertificate` override — the piece that actually wires into Scandium (chain validation with the delegate's subject check disabled, wrapping a mismatch into a `CertificateVerificationResult`, the `verifySubject=false` passthrough) — is untested, as are the fail-closed `CertificateParsingException` branch and the non-SNI hostname destination path via `StringUtil.toHostString`. The repo already spins up a real DTLS server in `common/coap-server`'s `CoapDtlsCertificateReloadIntegrationTest`, so an end-to-end handshake test against a wildcard-cert server seems achievable and would cover the whole verifier, not just the hostname-matching half.

monitoring/src/main/java/org/thingsboard/monitoring/service/BaseHealthChecker.java
```diff
@@ -124,6 +125,14 @@ private void checkWsUpdates(WsClient wsClient, String testValue) {
         reporter.reportLatency(Latencies.wsUpdate(getKey()), stopWatch.getTime());
     }
 
+    private void safeDestroyClient() {
+        try {
+            destroyClient();
+        } catch (Throwable e) {
+            log.warn("[{}] Failed to destroy client: {}", info, e.getMessage());
```
smatvienko-tb: The new teardown warnings (here and in `shutdownCoapClient`/`closeMqttClient`) log only `e.getMessage()`, discarding the exception type and stack trace — and for many Throwables `getMessage()` is null, producing `Failed to destroy client: null`. Since these paths fire when something is already going wrong, passing the exception as the last slf4j argument (or logging it at debug) would keep the diagnostics without adding noise.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/TransportHealthChecker.java
```diff
@@ -30,6 +30,18 @@ public abstract class TransportHealthChecker<C extends TransportMonitoringConfig
 
     @Value("${monitoring.calculated_fields.enabled:true}")
     private boolean calculatedFieldsMonitoringEnabled;
+    @Value("${monitoring.session_duration_ms:3600000}")
+    private long sessionDurationMs;
+
+    private long sessionStartTime;
+
+    protected boolean isSessionExpired() {
+        return sessionDurationMs > 0 && System.currentTimeMillis() - sessionStartTime >= sessionDurationMs;
```
smatvienko-tb: Two thoughts here: (1) `System.currentTimeMillis()` is wall-clock, so an NTP step can expire a session early or postpone reconnection past the configured duration — `System.nanoTime()` deltas are the usual choice for measuring durations. (2) The session-expiry behavior has no tests — neither the `isSessionExpired`/`recordSessionStart` contract (expiry boundary, the documented `0 = never expire` mode) nor the reconnect paths in the MQTT/CoAP `initClient` implementations. A unit test with a short duration would be cheap and would pin down the `0` semantics the yml comment advertises.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/CoapTransportHealthChecker.java
```diff
@@ -38,46 +41,79 @@ public class CoapTransportHealthChecker extends TransportHealthChecker<CoapTrans
...
+                shutdownCoapClient();
+            } else {
+                return;
+            }
+        }
+
+        String accessToken = target.getDevice().getCredentials().getCredentialsId();
+        String uri = target.getBaseUrl() + "/api/v1/" + accessToken + "/telemetry";
+        coapClient = new CoapClient(uri);
+        if (target.getBaseUrl().startsWith("coaps")) {
```
smatvienko-tb: A raw `startsWith("coaps")` on the whole URL works, but comparing the parsed URI scheme against `CoAP.COAP_SECURE_URI_SCHEME` would be self-documenting and immune to odd inputs — and the class imports `CoAP` already.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/MqttTransportHealthChecker.java
```diff
@@ -45,21 +45,28 @@ protected MqttTransportHealthChecker(MqttTransportMonitoringConfig config, Trans
...
-            options.setConnectionTimeout(config.getRequestTimeoutMs() / 1000);
-            IMqttToken result = mqttClient.connectWithResult(options);
-            if (result.getException() != null) {
-                throw result.getException();
+        if (mqttClient != null) {
+            if (mqttClient.isConnected() && !isSessionExpired()) {
+                return;
             }
-            log.debug("Initialized MQTT client for URI {}", mqttClient.getServerURI());
+            log.info("Reconnecting {} client to {}", getTransportType(), target.getBaseUrl());
```
smatvienko-tb: The session-expiry reconnect choreography is now duplicated between the MQTT and CoAP checkers: both implement the same "if client exists → if expired, log Reconnecting + tear down, else return; then connect + `recordSessionStart()` + log" sequence with identical messages. Consider hoisting it into a template method on `TransportHealthChecker` (e.g. a final `initClient()` calling abstract `isClientOpen()`/`connect()`/`closeClient()`) — that removes the copy-paste, guarantees `recordSessionStart()` can't be forgotten by the next transport, and would give LwM2M the behavior for free.

monitoring/src/main/java/org/thingsboard/monitoring/service/transport/impl/MqttTransportHealthChecker.java
```diff
@@ -45,21 +45,28 @@ protected MqttTransportHealthChecker(MqttTransportMonitoringConfig config, Trans
...
+
+        MqttConnectOptions options = new MqttConnectOptions();
+        options.setUserName(accessToken);
+        options.setConnectionTimeout(config.getRequestTimeoutMs() / 1000);
+        IMqttToken result = mqttClient.connectWithResult(options);
+        if (result.getException() != null) {
+            throw result.getException();
         }
+        recordSessionStart();
+        log.debug("Connecting {} client to {}", getTransportType(), target.getBaseUrl());
```
smatvienko-tb: This "Connecting" message is logged after `connectWithResult` has already returned, so it reads as in-progress when the connection is actually established. There's also an asymmetry: reconnects log at INFO but the first-ever connect only at DEBUG — a consistent `Connected {} client to {}` would describe reality and make session cycling visible in default logs. (Same applies to the CoAP checker.)

monitoring/src/main/java/org/thingsboard/monitoring/ThingsboardMonitoringApplication.java
```diff
@@ -49,6 +49,8 @@ public class ThingsboardMonitoringApplication {
 
     @Value("${monitoring.monitoring_rate_ms}")
     private int monitoringRateMs;
+    @Value("${monitoring.session_duration_ms:3600000}")
```
smatvienko-tb: The `${monitoring.session_duration_ms:3600000}` expression (fallback included) is now duplicated here and in `TransportHealthChecker`, and this injection exists solely to append "session duration: Xms" to a scheduling log line printed for every monitoring service — including ones that ignore the setting (HTTP, LwM2M). If the default ever changes, two `@Value` fallbacks plus the yml default must stay in sync; logging it once from `TransportHealthChecker` or centralizing the property in a config bean would avoid both.

monitoring/src/main/resources/tb-monitoring.yml
```diff
@@ -36,6 +36,9 @@ monitoring:
   monitoring_rate_ms: '${MONITORING_RATE_MS:10000}'
   # Maximum time between request to transport and WebSocket update
   check_timeout_ms: '${CHECK_TIMEOUT_MS:5000}'
+  # Transport session duration in milliseconds. Connection is reused until it expires, then reconnected.
+  # Set to 0 to keep the session open indefinitely (reconnect only on failure).
+  session_duration_ms: '${SESSION_DURATION_MS:3600000}'
```
smatvienko-tb, line 43: Only the MQTT and CoAP checkers honor `session_duration_ms` — the LwM2M checker keeps its Leshan client alive indefinitely, and LwM2M is exactly the kind of long-lived (potentially DTLS) session the cert-rotation rationale applies to; HTTP ignores it too (stateless per request). An operator setting this knob will expect it to cover all connection-oriented transports — if extending it to LwM2M is out of scope, could the comment state which transports respect it?

monitoring/src/main/java/org/thingsboard/monitoring/config/transport/TransportType.java
```diff
@@ -28,11 +28,11 @@
 public enum TransportType {
 
     MQTT("MQTT", MqttTransportHealthChecker.class),
-    COAP("CoAP",CoapTransportHealthChecker.class),
+    COAP("CoAP", CoapTransportHealthChecker.class),
```
smatvienko-tb, line 31: Several files in this PR lost their trailing newline (this one, `BaseHealthChecker.java`, `MqttTransportHealthChecker.java`, `tb-monitoring.yml`, the new test file — the diff shows `\ No newline at end of file` on each). Cosmetic, but it dirties future diffs of the last line; worth restoring before merge.

smatvienko-tb:
### Code Review: Pull Request #15840

The following issue was identified in the implementation of the `WildcardAwareCertificateVerifier`.\n\n#### 🚨 Logic Bug in Wildcard Verification
**File:** `WildcardAwareCertificateVerifier.java`  
**Method:** `matchesOverlyBroadWildcard`\n\n**Description:**
The current logic incorrectly flags legitimate multi-label public suffixes (e.g., `*.github.io`) as "overly broad" because the comparison `dnsName.substring(2).equalsIgnoreCase(destinationSuffix)` returns `true` for these domains, and they are successfully matched by the `PUBLIC_SUFFIX_MATCHER`.\n\nThis contradicts the stated requirement:\n> "...it doesn't reject multi-label public suffixes such as ... *.github.io"\n\n**Impact:**
Valid certificates using multi-label subdomains will be incorrectly flagged/rejected as overly broad, which may break legitimate connections for users on platforms like GitHub Pages or similar services.\n\n**Suggested Fix:**
Adjust the condition to ensure that only true public infrastructure (like `.com" or `.org`) is rejected when it matches a destination suffix, while allowing shared-hosting style wildcards to pass.

PR #15846, opened 2026-06-26
Standardize Capitalization of Rule Chain Link Labels

LeoMorgan113:
## Pull Request description

Added a link formatter for the UI, updated messageTypes map.
Updated links in translations.

image

## General checklist

## Back-End feature checklist


vvlladd28 (commented):
## Review summary

Reviewed 20 changed files in [Standardize Capitalization of Rule Chain Link Labels](https://github.com/thingsboard/thingsboard/pull/15846). Left 8 comment(s) inline.

The intent is display-only normalization, and the persisted `value`/`labels` are indeed preserved in the main dialog flow. However, `edge.label` is not purely presentational in this page — two flows derive persisted data from it, so the standardized display string can leak into saved rule chain metadata (see the comment on `AddRuleNodeLinkDialogComponent.add()`). The normalization is also applied only on the add-link path, so the standardized casing reverts after a reload or an edit.

## Additional findings

- **ui-ngx/src/app/modules/home/pages/rulechain/link-labels.component.ts:221** — `transformLinkLabel` matches the typed text against `label.name`, which is now the standardized form. Typing or pasting the exact stored value (e.g. `Post telemetry`) no longer matches the suggestion entry: with `allowCustom` off the input is now silently dropped (it used to add the label, since name === value before this PR), and with `allowCustom` on it creates a chip whose displayed casing differs from the identically-valued suggestion. Consider matching against both `name` and `value`.

---

ui-ngx/src/app/modules/home/pages/rulechain/rulechain-page.component.ts
```diff
@@ -1807,7 +1812,7 @@ export class AddRuleNodeLinkDialogComponent extends DialogComponent<AddRuleNodeL
   add(): void {
     this.submitted = true;
     const link: FcRuleEdge = this.ruleNodeLinkFormGroup.get('link').value;
-    this.link = {...this.link, ...link};
+    this.link = {...this.link, ...link, label: toStandardizedLinkLabel(link.label)};
```
vvlladd28: Standardizing `label` here makes the edge's display label diverge from `labels` (which keeps the raw values) — and `edge.label` isn't purely presentational in this page. Two flows turn it back into persisted data:

1. **Re-attaching an edge.** ngx-flowchart's `dragstart` captures only `edge.label` (`dragLabel = edge.label`) and `_addEdge()` builds a fresh edge without the custom `labels` array, so the `createEdge` callback reconstructs it via `edge.labels = edge.label.split(' / ')` (lines 214–217). After this change that produces the title-cased strings, and on save the connection `type` becomes e.g. `Post Telemetry` instead of the original `Post telemetry`.
2. **Creating a nested rule chain from a selection.** `outputEdge.labels = [outputEdge.label]` and the generated output node's `name: outputEdge.label` persist the display label directly.

Routing keeps working because the rule engine matches relation types case-insensitively (`RuleChainActorMessageProcessor#contains` → `equalsIgnoreCase`), but the saved metadata silently changes — exports/version-control diffs pick it up, and case-variant duplicates can accumulate (`addLabel` dedups by exact `value`). Since the intent is display-only, normalizing at render time (e.g. a pipe where edge labels are shown) would keep the model untouched and remove the leak entirely. Also note this title-cases the compound joined string (`Success / my label`) and relies on the `/` token passing through the word-caser unchanged — `link.labels.map(toStandardizedLinkLabel).join(' / ')` would be more explicit if the transform stays here.

ui-ngx/src/app/shared/models/rule-node.models.ts
```diff
@@ -80,10 +80,38 @@ export interface IRuleNodeConfigurationComponent {
...
+      }
+
+      if (LOWERCASE_WORDS.has(word.toLowerCase())) {
+        return word.toLowerCase();
+      }
+      return word.charAt(0).toUpperCase() + word.slice(1).toLowerCase();
+    }).join(' ');
+}
+
+export function normalizeLinkLabel(label: LinkLabel): LinkLabel {
```
vvlladd28, line 110: The normalization runs at scattered write points (the add-link dialog, `createEdge`'s labels map, `link-labels` `writeValue`/`prepareLabelsList`), but the two other paths that produce edge labels don't go through it: loading a saved chain builds the label raw (`label: connection.type`, rulechain-page.component.ts:650), and editing a link writes back a raw label (`RuleNodeLinkComponent.updateModel` joins raw values, then `saveRuleNodeLink` stores it). So a link added via the dialog shows "Post Telemetry" only until the page is reloaded or the link is edited — then it reverts to "Post telemetry". Applying `toStandardizedLinkLabel` at the single place edge labels are rendered (e.g. a pipe) would give uniform casing across load/add/edit and make the display-only guarantee structural rather than convention-based.

ui-ngx/src/app/modules/home/pages/rulechain/link-labels.component.ts
```diff
@@ -150,12 +150,9 @@ export class LinkLabelsComponent implements ControlValueAccessor, OnInit, OnChan
...
         value.forEach((label) => {
           if (this.allowedLabels[label]) {
-            this.labels.push(deepClone(this.allowedLabels[label]));
+            this.labels.push(deepClone(normalizeLinkLabel(this.allowedLabels[label])));
           } else {
-            this.labels.push({
-              name: label,
-              value: label
-            });
+            this.labels.push({name: toStandardizedLinkLabel(label), value: label});
```
vvlladd28: This branch handles labels that aren't in `allowedLabels` — i.e. user-defined custom labels — and title-casing them misrepresents what the user typed: `checkTemperature` shows as `Checktemperature`, `gRPC` as `Grpc`, `IoT data` as `Iot Data` (the all-caps guard doesn't cover mixed-case tokens). The chip then differs from the stored value, which is confusing when debugging why a relation looks different in an export or in debug events. Would it be safer to standardize only the well-known system labels and render custom labels verbatim?

ui-ngx/src/app/modules/home/pages/rulechain/rulechain-page.component.ts
```diff
@@ -221,6 +221,11 @@ export class RuleChainPageComponent extends PageComponent
           const allowCustomLabels = this.ruleChainService.ruleNodeAllowCustomLinks(sourceNode.component);
           const sourceRuleChainId = this.ruleChainService.ruleNodeSourceRuleChainId(sourceNode.component, sourceNode.configuration);
           this.enableHotKeys = false;
+
+          for (const key of Object.keys(labels)) {
```
vvlladd28: This loop looks redundant: `labels` is only consumed by `AddRuleNodeLinkDialogComponent`, which hands it to `tb-rule-node-link` → `tb-link-labels`, and that component already normalizes every entry in `writeValue()`/`prepareLabelsList()`. The sibling edit path (`openLinkDetails`, line 1138) passes the same `getRuleNodeSupportedLinks()` result without pre-normalizing and relies on exactly that internal normalization. If normalizing at the source is preferred, doing it once inside `RuleChainService.getRuleNodeSupportedLinks()` would cover both call sites and let this loop be dropped.

ui-ngx/src/app/shared/models/rule-node.models.ts
```diff
@@ -80,10 +80,38 @@ export interface IRuleNodeConfigurationComponent {
...
+  testScript?(debugEventBody?: DebugRuleNodeEventBody);
   [key: string]: any;
 }
 
+const LOWERCASE_WORDS = new Set(['a', 'an', 'the', 'and',
+  'but', 'or', 'nor', 'for', 'yet', 'so', 'at', 'by', 'from',
+  'in', 'of', 'on', 'to', 'up', 'with', 'as', 'into', 'per', 'via', 'than'
+]);
+
+export function toStandardizedLinkLabel(label: string): string {
```
vvlladd28, line 92: This pure function now sits in front of every link label the user sees and has several distinct branches (all-caps preservation, first word, connector words), but no spec was added. A small spec covering e.g. `RPC Request from Device`, `entity assigned to tenant`, a mixed-case custom label, and an empty string would lock the casing rules in and document them for whoever tweaks `LOWERCASE_WORDS` next. It's also worth making it tolerant of `undefined`/`null` (`label?.split(...)` with a fallback) — it's an exported shared-model util, and callers like the add-link dialog would throw here if ever reached with an edge that has no label.

ui-ngx/src/app/modules/home/pages/rulechain/link-labels.component.ts
```diff
@@ -150,12 +150,9 @@ export class LinkLabelsComponent implements ControlValueAccessor, OnInit, OnChan
       if (value) {
         value.forEach((label) => {
           if (this.allowedLabels[label]) {
-            this.labels.push(deepClone(this.allowedLabels[label]));
+            this.labels.push(deepClone(normalizeLinkLabel(this.allowedLabels[label])));
```
vvlladd28: `normalizeLinkLabel` already returns a fresh `{name, value}` literal of two strings, so the wrapping `deepClone` no longer does anything — it was only needed when the `allowedLabels` entry was pushed directly. Dropping it simplifies the line.

ui-ngx/src/app/modules/home/pages/rulechain/rulechain-page.component.ts
```diff
@@ -67,14 +67,14 @@ import {
   FcRuleNode,
   FcRuleNodeType,
   getRuleNodeHelpLink,
-  LinkLabel,
+  LinkLabel, normalizeLinkLabel,
```
vvlladd28: Minor: the new symbols are tacked onto existing lines (`LinkLabel, normalizeLinkLabel,` and `ruleNodeTypesLibrary, toStandardizedLinkLabel`), breaking the one-symbol-per-line, alphabetized style of this import block — worth letting the formatter keep it consistent.

ui-ngx/src/assets/locale/locale.constant-zh_TW.json
```diff
@@ -1928,7 +1928,7 @@
         "timeseries-placeholder": "時間序列",
         "target-entity": "目標實體",
         "attributes-propagation": "屬性傳播",
-        "attributes-propagation-hint": "每次儲存或更新這個實體視圖時，實體視圖將自動從目標實體複製指定的屬性。由於性能原因，目標實體屬性不會在每次屬性更改時傳播到實體視圖。您可以通過配置\"copy to view\"規則鏈中的規則節點，並將\"Post attributes\"和\"attributes Updated\"消息連結到新規則節點，從而啟用自動傳播。",
+        "attributes-propagation-hint": "每次儲存或更新這個實體視圖時，實體視圖將自動從目標實體複製指定的屬性。由於性能原因，目標實體屬性不會在每次屬性更改時傳播到實體視圖。您可以通過配置\"copy to view\"規則鏈中的規則節點，並將\"Post Attributes\"和\"attributes Updated\"消息連結到新規則節點，從而啟用自動傳播。",
```
vvlladd28: While fixing "Post Attributes" here, the same string still contains lowercase `"attributes Updated"` — since this line is already being touched, worth aligning it with the other locales' "Attributes Updated".

PR #15847, opened 2026-06-26
Fix chart axis ticks font size

LeoMorgan113:
## Pull Request description

Fixed calculations for the top & bottom grid offset for axes tick large font sizes.

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/lib/chart/time-series-chart.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/chart/time-series-chart.ts
@@ -846,23 +846,28 @@ export class TbTimeSeriesChart {
     }
   }
 
-  private minTopOffset(): number {
-    if (this.topPointLabels) return 25;
-    const yAxisWithTickLabels = this.yAxisList.filter(yAxis => yAxis.settings.show && yAxis.settings.showTickLabels);
-    if (yAxisWithTickLabels.length) {
-      const maxFontSize = Math.max(...yAxisWithTickLabels.map(a => a.settings.tickLabelFont?.size || 12));
-      return Math.max(10, Math.ceil(maxFontSize / 2));
+  private maxTickLabelHalfHeight(): number {
+    const axes = this.yAxisList.filter(a => a.settings.show && a.settings.showTickLabels);
+    if (!axes.length) {
+      return 0;
     }
-    return 5;
+
+    const defaultSize = defaultTimeSeriesChartYAxisSettings.tickLabelFont.size;
+    const maxFontSize = Math.max(...axes.map(a => a.settings.tickLabelFont?.size || defaultSize));
+    return Math.ceil(maxFontSize / 2);
+  }
+
+  private minTopOffset(): number {
+    if (this.topPointLabels) {
+      return 25;
+    }
+    const half = this.maxTickLabelHalfHeight();
+    return half ? Math.max(10, half) : 5;
   }
 
   private minBottomOffset(): number {
-    const yAxisWithTickLabels = this.yAxisList.filter(yAxis => yAxis.settings.show && yAxis.settings.showTickLabels);
-    let minOffset = 5;
-    if (yAxisWithTickLabels.length) {
-      const maxFontSize = Math.max(...yAxisWithTickLabels.map(a => a.settings.tickLabelFont?.size || 12));
-      minOffset = Math.max(5, Math.ceil(maxFontSize / 2));
-    }
+    const half = this.maxTickLabelHalfHeight();
+    const minOffset = half ? Math.max(5, half) : 5;
     return this.settings.dataZoom ? Math.max(45, minOffset) : minOffset;
   }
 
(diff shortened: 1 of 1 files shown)


ui-ngx/src/app/modules/home/components/widget/lib/chart/time-series-chart.ts
```diff
@@ -847,14 +847,23 @@ export class TbTimeSeriesChart {
   }
 
   private minTopOffset(): number {
-    const showTickLabels =
-      !!this.yAxisList.find(yAxis => yAxis.settings.show && yAxis.settings.showTickLabels);
-    return (this.topPointLabels) ? 25 :
-      (showTickLabels ? 10 : 5);
+    if (this.topPointLabels) return 25;
```
vvlladd28: This early return short-circuits the new font-size logic, so when `topPointLabels` is true the top offset is fixed at 25 regardless of the tick-label font. Since the whole point of this change is to keep large tick fonts from clipping, a very large y-axis tick font combined with point labels would still clip at the top. Should this be `Math.max(25, Math.ceil(maxFontSize / 2))` so it accounts for both, or are point labels guaranteed to dominate in practice?

ui-ngx/src/app/modules/home/components/widget/lib/chart/time-series-chart.ts
```diff
@@ -847,14 +847,23 @@ export class TbTimeSeriesChart {
...
+    const yAxisWithTickLabels = this.yAxisList.filter(yAxis => yAxis.settings.show && yAxis.settings.showTickLabels);
+    if (yAxisWithTickLabels.length) {
+      const maxFontSize = Math.max(...yAxisWithTickLabels.map(a => a.settings.tickLabelFont?.size || 12));
```
vvlladd28: This `filter` + `map` + `Math.max(... || 12)` computation is duplicated verbatim in `minBottomOffset()` just below. Extracting a small helper (e.g. `maxTickLabelHalfOffset()`) would keep the `12` default and the `/2` scaling in one place — right now if either ever changes, the two copies can silently drift apart.

Also, the bare constants here (`12`, the `/2` divisor, the `10`/`5`/`45` floors) read as magic numbers. A one-line comment on why `maxFontSize / 2` is the right headroom would help the next person tuning this, and since `12` mirrors the default `tickLabelFont.size` from `timeSeriesChartYAxisDefaultSettings`, referencing that constant instead of re-hardcoding it would be safer.

vvlladd28 (commented):
## Review summary

Reviewed 1 changed file in [Fix for axis ticks font size](https://github.com/thingsboard/thingsboard/pull/15847). Left 2 comment(s) inline.

The core fix is sound: the empty-array case is correctly guarded so `Math.max(...[])` never produces `-Infinity`, and the default 12px font keeps the previous offsets (top 10, bottom 5/45) so the change is backward-compatible. The comments below are about a completeness edge case relative to the PR's own goal and some duplication/readability cleanup.

---

vvlladd28:
### Optional refactor suggestion (maintainability)

The fix itself is correct, but `minTopOffset()` and `minBottomOffset()` now share the exact same font-size computation (filter shown axes with tick labels → take the max `tickLabelFont.size` defaulting to `12` → halve via `Math.ceil`). That block is copy-pasted in both methods, so the `12` default and the `/2` scaling can silently drift apart if either is ever tuned.

Extracting a small helper removes the duplication, puts the font-related logic in one place, and makes the intent explicit:

```ts
private maxTickLabelHalfHeight(): number {
  const axes = this.yAxisList.filter(a => a.settings.show && a.settings.showTickLabels);
  if (!axes.length) {
    return 0;
  }
  // a tick label is vertically centered on its gridline, so half of it overflows the grid edge
  const defaultSize = defaultTimeSeriesChartYAxisSettings.tickLabelFont.size;
  const maxFontSize = Math.max(...axes.map(a => a.settings.tickLabelFont?.size || defaultSize));
  return Math.ceil(maxFontSize / 2);
}

private minTopOffset(): number {
  if (this.topPointLabels) {
    return 25;
  }
  const half = this.maxTickLabelHalfHeight();
  return half ? Math.max(10, half) : 5;
}

private minBottomOffset(): number {
  const half = this.maxTickLabelHalfHeight();
  const minOffset = half ? Math.max(5, half) : 5;
  return this.settings.dataZoom ? Math.max(45, minOffset) : minOffset;
}
```

Benefits:
- No duplication — the `filter` / `map` / `/2` logic lives in a single method.
- The default font size references `defaultTimeSeriesChartYAxisSettings.tickLabelFont.size` instead of a hardcoded `12`, so it stays in sync with the settings default.
- A short comment documents *why* `fontSize / 2` is the right amount of headroom.
- Easier to extend later: if you ever want a precise measurement (e.g. `estimateLabelUnionRect` / `getTextRect`) instead of the `/2` heuristic, you only change `maxTickLabelHalfHeight()` and both offsets pick it up automatically.

Not blocking — just a small cleanup that keeps this offset logic easy to maintain.

vvlladd28 (approved):

Issue #15289, opened 2026-03-24
Decimal numbers show inconsistent separators


itnbenzing:
Description:
When viewing device attributes, decimal numbers are displayed using a dot (.) as the decimal separator.
However, when editing a decimal number on the device details page, the same number is shown using a comma (,).
This inconsistency can confuse users.

<img width="1649" height="885" alt="Image" src="https://github.com/user-attachments/assets/68070154-e07f-4f95-ab54-b5d5295cba58" />

Steps to Reproduce:
1. Go to a device in ThingsBoard.
2. Open the device attributes view.
3. Add a decimal number as a server attibute..
4. Observe that decimal numbers use a dot (.) in the attribute overview.
5. Click to edit the decimal number.
6. Observe that the decimal number now uses a comma (,).

Expected Behavior:
The decimal separator should be consistent across the platform. 

Professional Edition 4.3.0.1

itnbenzing:
This is tied to my feature request #15290

mykhailokornieiev:
Hi. I was unable to reproduce this behavior at all
It is not possible to save double value with a comma as a separator, and it always displays a dot separator instead

<img width="1054" height="142" alt="Image" src="https://github.com/user-attachments/assets/9a797439-6434-4a54-95ea-f544fd88eec7" />

I was able to save it with a comma only as a string, but it displayed with a comma in all cases. Please try to reproduce this behavior on a clean tenant without white labeling and in a private browser tab as well.

mykhailokornieiev:
Please ignore previous message, I was able to reproduce this behavior. Will test it and come back with update later

mykhailokornieiev:
Thank you for reporting this bug. I already added the respective ticket to our backlog for developers. 
We will review this request and will get back to you with updates as soon as they arise. 
Please follow our [Release Notes](https://thingsboard.io/docs/reference/releases/) to see all new additions and fixes to the platform in subsequent releases.

PR #15856, opened 2026-06-27
Fix decimal separator inconsistency

chopchaper42:
## Pull Request description

Fixes inconsistency in decimal number separators in device attribute table and attribute edit window.   
Original Issue #15289 

**Before:**
old

**After:**
new

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/shared/components/value-input.component.html
+++ b/ui-ngx/src/app/shared/components/value-input.component.html
@@ -68,7 +68,7 @@
     }
     @if (valueType === valueTypeEnum.DOUBLE) {
       <mat-form-field appearance="outline" subscriptSizing="dynamic" class="tb-inline-field tb-suffix-absolute number flex flex-1">
-        <input [disabled]="disabled" matInput [required]="required" name="value" type="number" step="any" #value="ngModel" [(ngModel)]="modelValue" (ngModelChange)="onValueChanged()"
+        <input [disabled]="disabled" matInput [required]="required" name="value" type="text" pattern="^-?[0-9]+(\.[0-9]+)?$" #value="ngModel" [(ngModel)]="modelValue" (ngModelChange)="onValueChanged()"
                placeholder="{{ 'value.double-value' | translate }}{{ required ? '*' : ''}}"/>
         @if (value.hasError('required') && value.control.touched) {
           <mat-icon matSuffix


ui-ngx/src/app/shared/components/value-input.component.html
```diff
@@ -68,7 +68,7 @@
     }
     @if (valueType === valueTypeEnum.DOUBLE) {
       <mat-form-field appearance="outline" subscriptSizing="dynamic" class="tb-inline-field tb-suffix-absolute number flex flex-1">
-        <input [disabled]="disabled" matInput [required]="required" name="value" type="number" step="any" #value="ngModel" [(ngModel)]="modelValue" (ngModelChange)="onValueChanged()"
+        <input [disabled]="disabled" matInput [required]="required" name="value" type="text" pattern="^-?[0-9]+(\.[0-9]+)?$" #value="ngModel" [(ngModel)]="modelValue" (ngModelChange)="onValueChanged()"
```
vvlladd28, line 71: Switching to `type="text"` changes the type of the value this component emits. With `type="number"`, Angular's `NumberValueAccessor` parsed the input into a JS number; with a text input the default accessor leaves `modelValue` as a **string**, and `updateView()` propagates it as-is. Downstream, `edit-attribute-value-panel` / `add-attribute-dialog` pass that value verbatim to `AttributeService.saveEntityAttributes`, and the backend derives the stored entry type from the JSON value type (`JsonConverter.convertToAttributes`, `TelemetryController:692`). With `type_cast_enabled: false` an edited double attribute gets silently saved as a **STRING** entry; even with the cast enabled (default), set-value widget actions and calculated-field test arguments now emit strings instead of numbers. The component's own logic also assumes numbers for DOUBLE — `detectValueType()` and `updateModelToValueType()` check `typeof === 'number'` — so a value that round-trips as a string will be mis-detected as STRING on the next edit.

Suggest normalizing in the component instead of relying on the input element's accessor: in `updateView()` (or `onValueChanged()`), when `valueType` is `DOUBLE` and the form is valid, convert with `parseFloat` before propagating, so the number-based contract holds regardless of which input element the template uses.

A few more side effects of the text+pattern approach worth a deliberate decision:
- Existing doubles whose JS string form is exponent notation (values < 1e-6 or ≥ 1e21, e.g. `1e-7`) render as `1e-7` in the text input and fail the pattern — a legitimately stored value becomes uneditable, and scientific-notation entry (previously allowed by `step="any"`) is gone.
- The regex also rejects `.5`, which the old number input accepted — `^-?(\d+(\.\d+)?|\.\d+)$` would keep that shorthand working.
- Dropping `type="number"` drops the numeric on-screen keyboard on mobile; consider adding `inputmode="decimal"` (the codebase already pairs `type="text"` with `inputmode` in `rest-api-call-config.component.html` and the 2FA dialogs). Note that some mobile keyboards render the decimal key as a comma in comma-decimal locales, which this pattern rejects — one more reason to surface the pattern error (see the comment below).
- INTEGER keeps `type="number"` while DOUBLE becomes a text field, so the two numeric siblings now behave differently (spinner arrows, keypad, native character filtering). Probably fine since integers have no decimal separator, but a short note explaining why only DOUBLE changed would keep the next maintainer from "fixing" the inconsistency in either direction.
- No test accompanies the behavior change; a small TestBed spec for the DOUBLE branch (dot accepted, comma rejected, negative values, value emitted as a number) would pin down the intended dot-only contract.

ui-ngx/src/app/shared/components/value-input.component.html
```diff
@@ -68,7 +68,7 @@
...
                placeholder="{{ 'value.double-value' | translate }}{{ required ? '*' : ''}}"/>
         @if (value.hasError('required') && value.control.touched) {
```
vvlladd28, line 73: The pattern introduces a new failure mode this branch doesn't surface: the browser no longer blocks non-numeric characters, so a user can now type `abc` or — much more likely given the locale motivation of this PR — `1,5`, and the field silently becomes invalid while `updateView()` propagates `null`; the only visible feedback here is the `required` warning. The sibling INTEGER branch already handles this: it checks `value.control.invalid` and shows `value.invalid-integer-value` for pattern failures. Could you mirror that here with a new `value.invalid-double-value` key (only `double-value` and `double-value-required` exist in `locale.constant-en_US.json` today)? Ideally the message should hint that a dot is the expected decimal separator, since comma-locale users are exactly the audience this change affects.

vvlladd28 (commented):
## Review summary

Thank you for the contribution! The goal of the fix is valid — the decimal separator rendered by `type="number"` is chosen by the browser/OS locale, not by ThingsBoard settings, so making the edit field consistent with the attribute table makes sense.

However, the current implementation introduces a functional regression: switching the input to `type="text"` changes the value emitted by the component from a JS number to a string. The backend derives the stored attribute type from the JSON value type, so with `type_cast_enabled: false` an edited double attribute gets silently saved as a STRING entry, and set-value widget actions / calculated-field test arguments start sending strings instead of numbers. There are also a few secondary issues (silent pattern-validation failures, exponent-form values becoming uneditable, lost numeric keyboard on mobile) — see the inline comments for details and suggested fixes.

Because of that, we are closing this PR for now. Please address the comments below and reopen the PR (or submit a new one referencing this review) — we'd be happy to take another look.

---

Issue #5906, opened 2022-01-17
[Feature Request] Request all shared attributes from MQTT Gateway API


adw128:
I have done some investigation through the code and I want to request all the shared attributes from thingsboard.  I've been trying to use the documentation as https://thingsboard.io/docs/reference/gateway-mqtt-api/#request-attribute-values-from-the-server, but it doesn't show you how to request all attributes. I've tried using: "*", "", ".*" to no effect.

I've searched through the existing issues but can't find any relevant ones. What's the best way to request all attributes? 

Cc: @pbrkr

volodymyr-babak:
hi @adw128 you can request all SHARED attributes from MQTT Gateway API with the next command:
**{"id": $request_id, "device": "Device A", "client": false, "keys": []}**

Please re-open the ticket if this will not work for you.

mcavalcante:
> hi @adw128 you can request all SHARED attributes from MQTT Gateway API with the next command: **{"id": $request_id, "device": "Device A", "client": false, "keys": []}**
> 
> Please re-open the ticket if this will not work for you.

Hello @volodymyr-babak ,

I tried to request all shared attributes from Arduino using MQTT, as explained by you, but it didn´t work. Does it still work for the newer ThingsBoard version (3.4.x)? 

Thanks.

ospmoon:
> > hi @adw128 you can request all SHARED attributes from MQTT Gateway API with the next command: **{"id": $request_id, "device": "Device A", "client": false, "keys": []}**
> > Please re-open the ticket if this will not work for you.
> 
> Hello @volodymyr-babak ,
> 
> I tried to request all shared attributes from Arduino using MQTT, as explained by you, but it didn´t work. Does it still work for the newer ThingsBoard version (3.4.x)?
> 
> Thanks.

I checked and it works!

if it doesn't work for you,
I can advise you a few steps:
1) subscribe to topics
subscribe("v1/gateway/attributes/response");
subscribe("v1/gateway/attributes");

2) and take a close look at the last key in the payload
there should be "keys" but not "key".
payload ="{\"id\":1, \"device\":\"device_0\", \"client\": false, \"keys\": []}";

Good luck.

eleicsjaivik:
It would be have great addition to the main document
I found it in github issues, the position of it should be in main docs.

PR #15865, opened 2026-06-29
feat(transport): unified attribute request — fetch-all per scope + scope-separated gateway response

ShvaykaD:
## Pull Request description

Unifies the device & gateway **attribute request** API across all three transports (MQTT, HTTP, CoAP) so a client can request **all keys in a scope** (client and/or shared), not just an explicit key list. Adds a scope-separated gateway response and fixes several backward-compat dead-ends in the existing API.

### Motivation
The previous attribute-request API only supported explicit key lists. An empty value was a dead-end (empty string → no result), and there was no way to say "give me everything in this scope". The gateway response also flattened client and shared attributes into a single object, so callers couldn't tell which scope a value came from.

This delivers the capability requested in #5906 (request all shared attributes from the MQTT Gateway API).

### Scope of changes

**Proto (`transport.proto`, `queue.proto`)**
- `GetAttributeRequestMsg`: add `allClientAttributes`, `allSharedAttributes`, `separateScopesResponse`.
- `GetAttributeResponseMsg`: add `separateScopesResponse`; deprecate `isMultipleAttributesRequest`.
- `GatewayAttributesRequestMsg`: add `clientKeys`, `sharedKeys`, `allClientKeys`, `allSharedKeys`; deprecate legacy `client`/`keys`.
- `onlyShared` (CoAP attribute-observe initial shared state) left untouched.

**Per-scope semantics (client / shared)**
- absent → exclude that scope
- present + empty → all keys in that scope
- present + list → those keys
- "all" wins over a specific list; both-absent keeps the legacy "fetch everything" behavior.

**Transports**
- MQTT (device): empty value now means "all in scope" instead of a dead-end.
- MQTT (gateway): new `clientKeys`/`sharedKeys` format (empty = all) returns a scope-separated `{client, shared}` response; legacy `client` + `key`/`keys` still returns `value`/`values`. Also fixes an omit-keys session crash.
- HTTP: additive `allClientKeys`/`allSharedKeys` boolean params; existing `clientKeys`/`sharedKeys` unchanged.
- CoAP: additive `allClientKeys`/`allSharedKeys` query params; existing params and `onlyShared` unchanged.

**Server**
- `DeviceActorMessageProcessor`: explicit per-scope all/specific/none resolution with all-wins; echoes `separateScopesResponse` from request to response.
- `JsonConverter.getJsonObjectForGateway`: emits `{client, shared}` when `separateScopesResponse`, else legacy `value`/`values`.

### Documentation notes
The transport API docs (MQTT/HTTP/CoAP attribute request sections) should be updated to document the new `allClientKeys`/`allSharedKeys` params, the empty-value = "all in scope" behavior, and the scope-separated gateway response shape.

**Documentation PR:** thingsboard/thingsboard.io#529

## General checklist

_Not applicable — no front-end changes._

## Back-End feature checklist

- [x] If new dependency was added: the dependency tree is checked for conflicts. _(No new dependencies.)_
- [x] If new service was added: the service is marked with corresponding @TbCoreComponent, @TbRuleEngineComponent, @TbTransportComponent, etc. _(No new services.)_
- [x] If new REST API was added: the RestClient.java was updated, issue for [Python REST client](https://github.com/thingsboard/thingsboard-python-rest-client) is created. _(No new admin REST API; only additive transport device-API params.)_
- [x] If new yml property was added: make sure a description is added (above or near the property). _(No new yml properties.)_

## Downstream consumers

This change is additive and backward-compatible — no consumer breaks. Follow-ups (full analysis tracked internally):

- **Python REST client** (`thingsboard-python-rest-client`): regenerate the device API client so the new `allClientKeys`/`allSharedKeys` params are exposed (CE + PE).
- **IoT Gateway** (`tb-mqtt-client` / `thingsboard-gateway`): optionally migrate gateway attribute requests from the now-deprecated `client`/`key`/`keys` to the new `clientKeys`/`sharedKeys` (they remain supported).
- **Device SDKs** (official device client SDKs that implement attribute requests): _nice-to-have_ — add support for the new request capabilities so a device can request **all client and/or all shared** attributes (and the empty-value/all-keys conventions) without enumerating keys — MQTT empty value; HTTP/CoAP `allClientKeys`/`allSharedKeys`.


ShvaykaD (commented):
_Superseded — this note was mistakenly framed as a re-review. See the full review here: https://github.com/thingsboard/thingsboard/pull/15865#pullrequestreview-4601215932_

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
```diff
@@ -654,36 +653,74 @@ private void onDeviceAttributesRequestJson(MqttPublishMessage msg) throws Adapto
...
+            JsonConverter.parseAttributeScope(jsonObj, "sharedKeys", b::addAllSharedAttributeNames, () -> b.setAllSharedAttributes(true));
+            requestMsg = b.build();
+        } else if (jsonObj.has("client")) {
+            // legacy format: client boolean + key/keys; keep the legacy value/values response
+            boolean clientScope = jsonObj.get("client").getAsBoolean();
+            Set<String> keys;
+            if (jsonObj.has("key")) {
+                keys = Collections.singleton(jsonObj.get("key").getAsString());
+            } else {
+                JsonArray keysArray = jsonObj.get("keys").getAsJsonArray();
```
ShvaykaD, line 671: In this legacy branch, a request with `client` present but neither `key` nor `keys` will NPE here — `jsonObj.get("keys")` returns null and `.getAsJsonArray()` throws, propagating as a raw `RuntimeException` (the method only declares `AdaptorException`, so it bypasses the adaptor-error handling and forces a connection close). Could you guard it — treat a missing `keys` as empty, or throw a clear `AdaptorException`? Low severity since the legacy format normally always carries `key`/`keys`.

ShvaykaD (commented):
## Review summary

Reviewed 18 changed files in [feat(transport): unified attribute request — fetch-all per scope + scope-separated gateway response](https://github.com/thingsboard/thingsboard/pull/15865). Left 8 comments inline. The change is additive and backward-compatible; the correctness/security pass found no blocking issues (the scope-separated gateway response reuses `toJson`, which emits only `client`/`shared`, and the actor's `resolveScopeFuture` applies all-wins precedence correctly). The inline notes are quality/design nits plus one latent NPE on the legacy gateway path.

## Additional findings

- **application/src/test/java/org/thingsboard/server/transport/coap/attributes/request/CoapAttributesRequestProtoIntegrationTest.java** — The CoAP proto variant has no tests for the new `allClientKeys`/`allSharedKeys` params. The CoAP proto adaptor delegates to the same `CoapAdaptorUtils.toGetAttributeRequestMsg()` as JSON, so the behavior holds, but it's untested under proto config — a test mirroring the JSON variant's `testRequestAllClientAttributes` would close the gap.

---

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
```diff
@@ -654,36 +653,74 @@ private void onDeviceAttributesRequestJson(MqttPublishMessage msg) throws Adapto
...
-            String deviceName = checkDeviceName(gatewayAttributesRequestMsg.getDeviceName());
-            int requestId = gatewayAttributesRequestMsg.getId();
-            boolean clientScope = gatewayAttributesRequestMsg.getClient();
-            ProtocolStringList keysList = gatewayAttributesRequestMsg.getKeysList();
-            Set<String> keys = new HashSet<>(keysList);
-            TransportProtos.GetAttributeRequestMsg requestMsg = toGetAttributeRequestMsg(requestId, clientScope, keys);
+            TransportApiProtos.GatewayAttributesRequestMsg gw = TransportApiProtos.GatewayAttributesRequestMsg.parseFrom(getBytes(mqttMsg.payload()));
+            String deviceName = checkDeviceName(gw.getDeviceName());
+            int requestId = gw.getId();
+            boolean newFormat = gw.getAllClientKeys() || gw.getAllSharedKeys()
```
ShvaykaD, line 691: The `newFormat` detection (`getAllClientKeys() || getAllSharedKeys() || getClientKeysCount() > 0 || getSharedKeysCount() > 0`) is the switch between the new and legacy gateway proto paths, but the intent — "any new-format field is set" — is implicit. A one-line comment naming that invariant would help, since a future proto field (e.g. a scope enum) would need to be remembered and added here.

common/proto/src/main/java/org/thingsboard/server/common/adaptor/JsonConverter.java
```diff
@@ -361,13 +363,68 @@ public static JsonObject getJsonObjectForGateway(
...
+    public static void parseAttributeScope(JsonObject json, String field,
+                                           Consumer<List<String>> setNames, Runnable setAll) {
+        if (!json.has(field) || json.get(field).isJsonNull()) {
+            return;
+        }
+        String value = json.get(field).getAsString();
+        if (value.trim().isEmpty()) {
+            setAll.run();
+        } else {
+            setNames.accept(Arrays.asList(value.split(",")));
```
ShvaykaD: `parseAttributeScope` trims the whole value to detect "blank = all", but doesn't trim individual names after the split — `" a, b "` is forwarded as `[" a", " b "]`, which silently matches no attribute. The Javadoc doesn't mention this. Either document that callers must send pre-trimmed names, or trim each token (`Arrays.stream(value.split(",")).map(String::trim)...`).

common/proto/src/main/java/org/thingsboard/server/common/adaptor/JsonConverter.java
```diff
@@ -361,13 +363,68 @@ public static JsonObject getJsonObjectForGateway(
...
+        }
+    }
+
+    /**
+     * Populate the client scope of an attribute request: "all" wins over a specific key list; a missing/empty
+     * list leaves the scope unset. Shared by the HTTP and CoAP query-parameter parsers.
+     */
+    public static void applyClientScope(TransportProtos.GetAttributeRequestMsg.Builder builder, boolean all, Collection<String> names) {
```
ShvaykaD, line 411: `applyClientScope` and `applySharedScope` are structurally identical — same signature and branching, differing only in which two builder setters they call. The unit tests cover both, but the "all wins over names" invariant now lives in two places. Worth collapsing into one helper parameterized by the two method references? Minor — the symmetry is also easy to keep by eye.

common/transport/http/src/main/java/org/thingsboard/server/transport/http/DeviceApiController.java
```diff
@@ -141,22 +141,22 @@ public class DeviceApiController implements TbTransportService {
...
-            @Parameter(description = "Comma separated key names for attribute with client scope", required = true , schema = @Schema(defaultValue = "state"))
+            @Parameter(description = "Comma separated key names for attribute with client scope", required = false, schema = @Schema(defaultValue = "state"))
             @RequestParam(value = "clientKeys", required = false, defaultValue = "") String clientKeys,
-            @Parameter(description = "Comma separated key names for attribute with shared scope", required = true , schema = @Schema(defaultValue = "configuration"))
-            @RequestParam(value = "sharedKeys", required = false, defaultValue = "") String sharedKeys) {
+            @Parameter(description = "Comma separated key names for attribute with shared scope", required = false, schema = @Schema(defaultValue = "configuration"))
+            @RequestParam(value = "sharedKeys", required = false, defaultValue = "") String sharedKeys,
+            @Parameter(description = "Set to true to return ALL client-scope attributes (ignores clientKeys)")
+            @RequestParam(value = "allClientKeys", required = false, defaultValue = "false") boolean allClientKeys,
```
ShvaykaD, line 149: The `@Parameter` docs for `allClientKeys`/`allSharedKeys` say "ignores clientKeys"/"ignores sharedKeys" but don't spell out the precedence when both `allClientKeys=true` and `clientKeys=...` are sent (the `all` flag wins). A short note in the description would save developers a trip to the source.

application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
```diff
@@ -567,25 +566,45 @@ public void onFailure(Throwable t) {
         }
     }
 
+    @SuppressWarnings("deprecation") // isMultipleAttributesRequest retained for the legacy gateway value/values response
+    private static void setMultipleAttributesRequest(GetAttributeResponseMsg.Builder builder, boolean multipleAttributesRequest) {
```
ShvaykaD: `setMultipleAttributesRequest` is a private one-liner that exists only to hold the `@SuppressWarnings("deprecation")` over a single delegating call. It conveys no domain intent and adds a hop at the call sites. Java only lets `@SuppressWarnings` sit on a declaration, so the alternatives are annotating the enclosing method or a local — your call whether the narrow wrapper is worth the indirection.

application/src/test/java/org/thingsboard/server/transport/coap/attributes/AbstractCoapAttributesIntegrationTest.java
```diff
@@ -89,12 +89,15 @@ public abstract class AbstractCoapAttributesIntegrationTest extends AbstractCoap
...
 
-    private static final String CLIENT_ATTRIBUTES_PAYLOAD = "{\"clientStr\":\"value1\",\"clientBool\":true,\"clientDbl\":42.0,\"clientLong\":73," +
+    protected static final String CLIENT_ATTRIBUTES_PAYLOAD = "{\"clientStr\":\"value1\",\"clientBool\":true,\"clientDbl\":42.0,\"clientLong\":73," +
             "\"clientJson\":{\"someNumber\":42,\"someArray\":[1,2,3],\"someNestedObject\":{\"key\":\"value\"}}}";
 
-    private static final String SHARED_ATTRIBUTES_PAYLOAD = "{\"sharedStr\":\"value1\",\"sharedBool\":true,\"sharedDbl\":42.0,\"sharedLong\":73," +
+    protected static final String SHARED_ATTRIBUTES_PAYLOAD = "{\"sharedStr\":\"value1\",\"sharedBool\":true,\"sharedDbl\":42.0,\"sharedLong\":73," +
             "\"sharedJson\":{\"someNumber\":42,\"someArray\":[1,2,3],\"someNestedObject\":{\"key\":\"value\"}}}";
 
+    protected static final String CLIENT_ATTRIBUTE_KEYS = "clientStr,clientBool,clientDbl,clientLong,clientJson";
```
ShvaykaD: `CLIENT_ATTRIBUTE_KEYS`/`SHARED_ATTRIBUTE_KEYS` (and the payload constants) are now defined identically here and in `AbstractMqttAttributesIntegrationTest`, which share no ancestor — so the fixtures are copied. If the attribute fixture ever changes, both copies must move together. Worth extracting to a shared test-fixtures holder both bases can reference? (May be more than warranted given the separate hierarchies.)

application/src/test/java/org/thingsboard/server/transport/coap/attributes/AbstractCoapAttributesIntegrationTest.java
```diff
@@ -193,19 +194,43 @@ protected void processJsonTestRequestAttributesValuesFromTheServer() throws Exce
         String update = getWsClient().waitForUpdate();
         assertThat(update).as("ws update received").isNotBlank();
 
-        String featureTokenUrl = CoapTestClient.getFeatureTokenUrl(accessToken, FeatureType.ATTRIBUTES) + "?clientKeys=" + clientKeysStr + "&sharedKeys=" + sharedKeysStr;
+        String featureTokenUrl = CoapTestClient.getFeatureTokenUrl(accessToken, FeatureType.ATTRIBUTES) + "?clientKeys=" + CLIENT_ATTRIBUTE_KEYS + "&sharedKeys=" + SHARED_ATTRIBUTE_KEYS;
         client.setURI(featureTokenUrl);
         validateJsonResponse(client.getMethod());
     }
 
+    protected void processJsonTestRequestAttributesWithQuery(String querySuffix, String expectedResponse) throws Exception {
```
ShvaykaD, line 179: `processJsonTestRequestAttributesWithQuery` shares ~18 lines of setup (create client, WS subscribe, post attributes, await WS update) with `processJsonTestRequestAttributesValuesFromTheServer` — only the URL suffix and response validation differ. Extracting the shared setup into a private helper would make future setup changes a single edit.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
ShvaykaD: Intentional for the legacy/deprecated gateway path, leaving as-is.

The NPE surfaces as a raw `RuntimeException`, which the publish handler catches at `MqttTransportHandler#443` and closes the connection. If we converted it to an `AdaptorException`, it would route through `sendResponseForAdaptorErrorOrCloseContext(...)`, which — when `deviceSessionCtx.isSendAckOnValidationException()` is true (or MQTT 5) — returns a `PubAck (PAYLOAD_FORMAT_INVALID)` instead of closing.

That would be a behavior change on a deprecated path: a gateway configured with "send ack on validation exception" that today relies on the session being closed on this malformed input would suddenly start receiving a PubAck and keep the connection open. The legacy format always carries `key`/`keys`, so this case is not reachable from a conformant gateway anyway — there is no reason to risk altering the long-standing close behavior. Leaving it as-is.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/session/AbstractGatewaySessionHandler.java
ShvaykaD: Reads clearly as-is; not adding a code comment here.

common/proto/src/main/java/org/thingsboard/server/common/adaptor/JsonConverter.java
ShvaykaD: Confirmed there was never any trimming: the original lts-4.2 `toStringSet` and the HTTP/CoAP parsers all split(",") with no trim. `parseAttributeScope` keeps that long-standing behavior, so this is not a regression — leaving it consistent across transports.

common/proto/src/main/java/org/thingsboard/server/common/adaptor/JsonConverter.java
ShvaykaD: Done in 6a86b70022 — collapsed into a shared `applyScope` helper parameterized by the two builder methods.

common/transport/http/src/main/java/org/thingsboard/server/transport/http/DeviceApiController.java
ShvaykaD: Reads clearly as-is — leaving the parameter descriptions.

application/src/main/java/org/thingsboard/server/actors/device/DeviceActorMessageProcessor.java
ShvaykaD: Discussed previously — keeping the narrow wrapper so the @SuppressWarnings("deprecation") stays scoped to the single deprecated call rather than the whole method.

application/src/test/java/org/thingsboard/server/transport/coap/attributes/AbstractCoapAttributesIntegrationTest.java
ShvaykaD: Done in 6a86b70022 — moved `CLIENT_ATTRIBUTE_KEYS`/`SHARED_ATTRIBUTE_KEYS` up to the shared `AbstractTransportIntegrationTest`.

application/src/test/java/org/thingsboard/server/transport/coap/attributes/AbstractCoapAttributesIntegrationTest.java
ShvaykaD: Done in 6a86b70022 — extracted the shared setup into `postAttributesAndAwaitWsUpdate`.

ShvaykaD (commented):
## Review summary

Reviewed 19 changed files in [feat(transport): unified attribute request — fetch-all per scope + scope-separated gateway response](https://github.com/thingsboard/thingsboard/pull/15865).

The change is additive and backward-compatible, and all findings from the previous review round are resolved. The one actionable item from this pass — the stale Swagger `@Schema(defaultValue="state"/"configuration")` on the now-optional `clientKeys`/`sharedKeys` HTTP params — has been addressed in `852026ff3f` (defaults dropped, descriptions point to `allClientKeys`/`allSharedKeys` for fetching a whole scope). No other issues.

---

ShvaykaD (commented):
## Review summary

Reviewed 5 changed files in commit `14f9230e58` (refactor: tidy GetAttribute response handling and parseAttributeScope). No issues found — the refactor is behavior-preserving across every consumer of the affected response fields, and the two quality nits from the prior pass (redundant `legacyRequest` local; duplicated scope-precedence rule) have been addressed. Everything is clear.

---

ViacheslavKlimov:
Leaving only the 4.4 PRs

PR #15866, opened 2026-06-29
Fixed Calculated Field test script running against stale entity values instead of current form input

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/calculated-fields/calculated-field.component.ts
+++ b/ui-ngx/src/app/modules/home/components/calculated-fields/calculated-field.component.ts
@@ -132,7 +132,7 @@ export class CalculatedFieldComponent extends EntityComponent<CalculatedFieldsTa
   onTestScript(expression?: string): Observable<string> {
     return this.cfFormService.testScript(
       this.entity?.id?.id,
-      this.entityValue,
+      this.entityFormValue(),
       this.entitiesTableConfig.getTestScriptDialog.bind(this.entitiesTableConfig),
       this.destroyRef,
       expression


ui-ngx/src/app/modules/home/components/calculated-fields/calculated-field.component.ts
```diff
@@ -132,7 +132,7 @@ export class CalculatedFieldComponent extends EntityComponent<CalculatedFieldsTa
   onTestScript(expression?: string): Observable<string> {
     return this.cfFormService.testScript(
       this.entity?.id?.id,
-      this.entityValue,
+      this.entityFormValue(),
```
vvlladd28, line 135: The change looks right — it brings this call site in line with `alarm-rules.component.ts#onTestScript` and the `additionalDebugActionConfig` above, both of which already use `entityFormValue()`. One note: nothing guards against a future regression back to the stale `entityValue`. I realize ui-ngx has essentially no component specs today, so this isn't a blocker — but the "test script uses the current form value, not the loaded entity" behavior is exactly the kind of subtle regression a small spec would catch if component testing infra ever lands.

vvlladd28 (commented):
## Review summary

Reviewed 1 changed file in [Fixed Calculated Field test script running against stale entity values instead of current form input](https://github.com/thingsboard/thingsboard/pull/15866). Left 1 comment inline.

The fix itself looks correct: passing `entityFormValue()` instead of the injected `entityValue` matches both the sibling `alarm-rules.component.ts#onTestScript` and this component's own `additionalDebugActionConfig`, which already use the live form value. The missing `id` in the form value is not a problem on this path — `getTestScriptDialog` only reads `type` and `configuration`, and the `...calculatedField` spread is guarded by `openCalculatedFieldEdit`, which is always `false` here.

---

vvlladd28 (approved):

PR #15868, opened 2026-06-29
Fixed false "Unsaved changes" popup on the Alarm rule details page when a clear condition is present

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
+++ b/ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
@@ -25,7 +25,7 @@ import {
   Validators
 } from '@angular/forms';
 import { MatDialog } from '@angular/material/dialog';
-import { isDefinedAndNotNull } from '@core/utils';
+import { isDefinedAndNotNull, isEqual } from '@core/utils';
 import { DashboardId } from '@shared/models/id/dashboard-id';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { AlarmRule, AlarmRuleCondition } from "@shared/models/alarm-rule.models";
@@ -76,6 +76,7 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
   testScript: (expression: string) => Observable<string>;
 
   private modelValue: AlarmRule;
+  private propagatedValue: AlarmRule;
 
   alarmRuleFormGroup = this.fb.group({
     condition: this.fb.control<AlarmRuleCondition | null>(null, Validators.required),
@@ -131,6 +132,7 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
       dashboardId: this.modelValue.dashboardId?.id
     } : null;
     this.alarmRuleFormGroup.patchValue(model, {emitEvent: false});
+    this.propagatedValue = this.toModel(this.alarmRuleFormGroup.value);
   }
 
   public openEditDetailsDialog($event: Event) {
@@ -161,8 +163,16 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
   }
 
   private updateModel() {
-    const value = this.alarmRuleFormGroup.value;
-    this.modelValue = {...value, dashboardId: value.dashboardId ? new DashboardId(value.dashboardId) : null} as AlarmRule;
-    this.propagateChange(this.modelValue);
+    const modelValue = this.toModel(this.alarmRuleFormGroup.value);
+    if (isEqual(modelValue, this.propagatedValue)) {
+      return;
+    }
+    this.modelValue = modelValue;
+    this.propagatedValue = modelValue;
+    this.propagateChange(modelValue);
+  }
+
+  private toModel(value: any): AlarmRule {
+    return {...value, dashboardId: value.dashboardId ? new DashboardId(value.dashboardId) : null} as AlarmRule;
   }
 }
--- a/ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
+++ b/ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
@@ -76,12 +76,11 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
   testScript: (expression: string) => Observable<string>;
 
   private modelValue: AlarmRule;
-  private propagatedValue: AlarmRule;
 
   alarmRuleFormGroup = this.fb.group({
     condition: this.fb.control<AlarmRuleCondition | null>(null, Validators.required),
-    alarmDetails: [null],
-    dashboardId: [null]
+    alarmDetails: this.fb.control<string | null>(null),
+    dashboardId: this.fb.control<string | null>(null)
   });
 
   private propagateChange = (v: any) => { };
@@ -106,8 +105,8 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
   ngOnInit() {
     this.alarmRuleFormGroup.valueChanges.pipe(
       takeUntilDestroyed(this.destroyRef)
-    ).subscribe(() => {
-      this.updateModel();
+    ).subscribe((value) => {
+      this.updateModel(value);
     });
     this.alarmRuleFormGroup.statusChanges.pipe(
       takeUntilDestroyed(this.destroyRef)
@@ -126,13 +125,12 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
   }
 
   writeValue(value: AlarmRule): void {
-    this.modelValue = value;
-    const model = this.modelValue ? {
-      ...this.modelValue,
-      dashboardId: this.modelValue.dashboardId?.id
+    const model = value ? {
+      ...value,
+      dashboardId: value.dashboardId?.id
     } : null;
     this.alarmRuleFormGroup.patchValue(model, {emitEvent: false});
-    this.propagatedValue = this.toModel(this.alarmRuleFormGroup.value);
+    this.modelValue = model ? this.toModel(model) : null;
   }
 
   public openEditDetailsDialog($event: Event) {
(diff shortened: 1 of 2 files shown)


ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
```diff
@@ -76,6 +76,7 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
   testScript: (expression: string) => Observable<string>;
 
   private modelValue: AlarmRule;
+  private propagatedValue: AlarmRule;
```
vvlladd28: The established pattern elsewhere in the codebase for suppressing redundant propagation is a single-field guard — e.g. `unit-input.component.ts` and several autocomplete components do `if (!isEqual(this.modelValue, value))` without a second field. Here `propagatedValue` is assigned in lockstep with `modelValue` in `updateModel()` and diverges only after `writeValue()` (raw written value vs. the form round-trip through `toModel()`).

I assume the second field exists because the raw written value doesn't survive the round-trip shape-wise — `dashboardId` arrives as a plain object but `toModel()` produces a `DashboardId` instance, so `isEqual` against the raw value would never match. If so, a short comment on the field explaining when the two diverge (and possibly a name like `lastSyncedValue`, since at the `writeValue()` assignment nothing has actually been propagated yet) would save the next reader from re-deriving that.

ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
```diff
@@ -161,8 +163,16 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
   }
 
   private updateModel() {
-    const value = this.alarmRuleFormGroup.value;
-    this.modelValue = {...value, dashboardId: value.dashboardId ? new DashboardId(value.dashboardId) : null} as AlarmRule;
-    this.propagateChange(this.modelValue);
+    const modelValue = this.toModel(this.alarmRuleFormGroup.value);
+    if (isEqual(modelValue, this.propagatedValue)) {
```
vvlladd28: This guard — the core of the fix — is subtle CVA behavior that a future refactor of `updateModel()`/`writeValue()` could silently regress, and nothing pins it down. I realize ui-ngx has almost no component specs today, but this logic is pure enough to cover cheaply with a TestBed harness: write a value, trigger an equal re-emission from the form, assert `propagateChange` is not called — plus the inverse (a real edit still propagates).

ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
```diff
@@ -161,8 +163,16 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
...
+      return;
+    }
+    this.modelValue = modelValue;
+    this.propagatedValue = modelValue;
+    this.propagateChange(modelValue);
+  }
+
+  private toModel(value: any): AlarmRule {
```
vvlladd28: `toModel` takes `value: any`, but both call sites pass `this.alarmRuleFormGroup.value`, so the parameter could be typed as `typeof this.alarmRuleFormGroup.value` (or a named interface for the raw form value). That keeps strict type checking and lets the compiler catch a future rename of `dashboardId`; the project TypeScript rules also discourage `any`.

vvlladd28 (commented):
## Review summary

Reviewed 1 changed files in [Fixed false "Unsaved changes" popup on the Alarm rule details page when a clear condition is present](https://github.com/thingsboard/thingsboard/pull/15868). Left 3 comment(s) inline.

---

ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
```diff
@@ -160,9 +160,16 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
     };
   }
 
-  private updateModel() {
-    const value = this.alarmRuleFormGroup.value;
-    this.modelValue = {...value, dashboardId: value.dashboardId ? new DashboardId(value.dashboardId) : null} as AlarmRule;
-    this.propagateChange(this.modelValue);
+  private updateModel(value: typeof this.alarmRuleFormGroup.value): void {
+    const modelValue = this.toModel(value);
+    if (isEqual(modelValue, this.modelValue)) {
```
vvlladd28, line 165: This guard compares against an object that shares the `condition` reference with the form, so an in-place mutation of the condition makes the comparison trivially true and the change is dropped. Chain: `writeValue` stores `modelValue.condition === value.condition` (both spreads are shallow) and `patchValue` puts that very same object into the `condition` control, so `CfAlarmRuleConditionComponent.modelValue` is the same instance. `openScheduleDialog` in that component then does `this.modelValue.schedule = result; this.updateModel();` — an in-place mutation — and propagates the same reference back. The resulting `alarmRuleFormGroup.valueChanges` emission carries a `condition` that is object-identical to `this.modelValue.condition`, so `isEqual` returns true (same-ref condition plus two distinct `new DashboardId('d1')` instances still compares equal) and `propagateChange` is skipped. Net effect: editing the alarm schedule of a clear rule (or a create rule — same component) never marks `entityForm` dirty, and `entity-details-page.component.html` gates Apply on `!detailsForm.dirty`, so the user makes the edit and the save button stays disabled. Storing a `deepClone` of the value in `modelValue` (or having the condition component replace rather than mutate its `modelValue`, the way `openFilterDialog` already does with `{...this.modelValue, ...result}`) would make the snapshot a real snapshot.

ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
```diff
@@ -125,12 +125,12 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
...
-    this.modelValue = value;
-    const model = this.modelValue ? {
-      ...this.modelValue,
-      dashboardId: this.modelValue.dashboardId?.id
+    const model = value ? {
+      ...value,
+      dashboardId: value.dashboardId?.id
     } : null;
     this.alarmRuleFormGroup.patchValue(model, {emitEvent: false});
+    this.modelValue = model ? this.toModel(model) : null;
```
vvlladd28, line 133: The baseline is now derived from the incoming `value` instead of from `this.alarmRuleFormGroup.value` after the patch (which is what the previous commit on this branch did), so the baseline's key set is whatever the caller happened to supply, while every later emission always has exactly the three form keys. lodash `isEqual` is key-count sensitive, so any missing optional key makes the guard permanently unable to match — and `alarmDetails`/`dashboardId` are both optional on `AlarmRule` (`ui-ngx/src/app/shared/models/alarm-rule.models.ts`). Concrete case: `alarm-rule-dialog.component.ts:154 addClearAlarmRule()` builds `{condition: {...}}` with no `alarmDetails`, so the baseline is `{condition, dashboardId: null}` (2 keys) while `toModel(formValue)` yields 3 — the guard is a dead branch for that dialog and the false-dirty behaviour is unchanged there. (`alarm-rules.component.ts:165` does include `alarmDetails: null`, which is why the details page appears fixed; anything loaded from the server that omits the optional fields would hit the same dead branch.) Deriving the baseline from the patched form value (e.g. `this.modelValue = this.toModel(this.alarmRuleFormGroup.value)`) keeps the two sides shape-identical by construction, and also removes the standing maintenance hazard: as written the guard only holds while `AlarmRule` and the three controls in `alarmRuleFormGroup` stay in exact lockstep, so adding a model field that isn't a control (or a control that isn't on the model) silently reintroduces the false "Unsaved changes" popup with nothing to catch it.

ui-ngx/src/app/modules/home/components/alarm-rules/cf-alarm-rule.component.ts
```diff
@@ -125,12 +125,12 @@ export class CfAlarmRuleComponent implements ControlValueAccessor, OnInit, Valid
   }
 
   writeValue(value: AlarmRule): void {
-    this.modelValue = value;
-    const model = this.modelValue ? {
-      ...this.modelValue,
-      dashboardId: this.modelValue.dashboardId?.id
+    const model = value ? {
```
vvlladd28, line 128: `writeValue(null)` sets `modelValue = null` but leaves the form untouched, because `FormGroup.patchValue(null)` returns early in Angular — nothing is cleared. That leaves the component in a state where `modelValue` says "no rule" while the form (and the nested condition component's own `modelValue`) still hold the previous rule, and the new guard can't help because `isEqual(staleModel, null)` is false. Reachable sequence: on the details page, remove the clear rule (`removeClearAlarmRule()` → `patchValue({clearRule: null})`), then delete the argument that the removed condition referenced; the condition component's `arguments` `ngOnChanges` flips `filtersArgumentsValid` true→false and calls `onValidatorChange()`, which is `condition.updateValueAndValidity()`, which bubbles up and re-emits `alarmRuleFormGroup.valueChanges` with the stale value — `updateModel` then propagates the deleted clear rule back into `entityForm`, and it comes back invalid so the user can't save at all. That bubbling emission is exactly the mechanism this PR is guarding against (see the earlier fix in `cf-alarm-rule-condition.component.ts`), so while the stale-form problem predates this change, the rework of `writeValue` seems like the right place to close it — `this.alarmRuleFormGroup.reset(undefined, {emitEvent: false})` on the null branch would keep form and `modelValue` in step.

vvlladd28 (commented):
## Review summary

Reviewed 1 changed file in [Fixed false "Unsaved changes" popup on the Alarm rule details page when a clear condition is present](https://github.com/thingsboard/thingsboard/pull/15868). Left 3 comment(s) inline.

The one to look at first is the new `isEqual` guard in `updateModel` (line 165): `writeValue` stores the incoming `condition` by reference and `patchValue` hands that same instance to the child component, so an in-place mutation like `openScheduleDialog` in `cf-alarm-rule-condition.component.ts` makes the comparison trivially true and the edit is dropped — the user changes the schedule and Apply stays disabled. Relatedly, the baseline's key set in `writeValue` (line 133) is caller-dependent while every later emission has exactly three keys, which makes the guard a dead branch for `addClearAlarmRule()` in `alarm-rule-dialog.component.ts`.

---

vvlladd28 (commented):
## Review summary

Reviewed 1 changed file in [Fixed false "Unsaved changes" popup on the Alarm rule details page when a clear condition is present](https://github.com/thingsboard/thingsboard/pull/15868). No new issues found.

HEAD is still `b79d328`, the same commit the previous automated review ran against, so nothing has changed since — the findings already left inline on `cf-alarm-rule.component.ts` still stand and are not repeated here. Everything else I looked at this pass checked out: `typeof this.alarmRuleFormGroup.value` compiles cleanly (verified against the project's TypeScript 5.9.3), the new `string | null` typing on `dashboardId` matches what `tb-dashboard-autocomplete` emits (`useIdValue` defaults to `true`), and assigning `modelValue` after `patchValue` is safe here because neither `CfAlarmRuleConditionComponent.writeValue` nor `tb-dashboard-autocomplete.writeValue` propagates synchronously.

---

PR #15878, opened 2026-06-30
Remove redundant [disabled] binding on reactive-form fields to fix NG01001 warning

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/widget-config.component.html
+++ b/ui-ngx/src/app/modules/home/components/widget/widget-config.component.html
@@ -51,7 +51,6 @@
                                               brackets="curly"
                                               subscriptSizing="dynamic"
                                               [predefinedValuesButton]="true"
-                                              [disabled]="!widgetSettings.get('showTitle').value"
                                               [predefinedValues]="predefinedValues"
                                               placeholderText="{{ 'widget-config.set' | translate }}"
                                               formControlName="title"
--- a/ui-ngx/src/app/shared/components/entity/entity-keys-list.component.html
+++ b/ui-ngx/src/app/shared/components/entity/entity-keys-list.component.html
@@ -42,7 +42,6 @@
            [matChipInputFor]="chipList"
            [matChipInputSeparatorKeyCodes]="separatorKeysCodes"
            matChipInputAddOnBlur
-           [disabled]="disabled"
            (matChipInputTokenEnd)="add($event)">
   </mat-chip-grid>
   <mat-autocomplete #keyAutocomplete="matAutocomplete"
--- a/ui-ngx/src/app/modules/home/components/widget/config/basic/button/power-button-basic-config.component.html
+++ b/ui-ngx/src/app/modules/home/components/widget/config/basic/button/power-button-basic-config.component.html
@@ -88,7 +88,6 @@
                                         subscriptSizing="dynamic"
                                         predefinedValuesButton="true"
                                         [predefinedValues]="predefinedValues"
-                                        [disabled]="!powerButtonWidgetConfigForm.get('showTitle').value"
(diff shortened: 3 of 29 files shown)


ui-ngx/src/app/shared/components/entity/entity-keys-list.component.html
```diff
@@ -42,7 +42,6 @@
            [matChipInputFor]="chipList"
```
vvlladd28, line 42: Removing this binding is safe for the form-driven path — `setDisabledState()` disables `keysListFormGroup`, which natively disables this input. But the component also declares `@Input() disabled`, and through that path the inner `key` control now stays enabled while the surrounding `mat-chip-grid`, chip remove buttons, and placeholder all render as disabled — a half-disabled state. No current caller uses the input (all four usages in `entity-view.component.html` go through `formControlName`), so this seems like a good moment to remove the `@Input()` decorator from `disabled` in `entity-keys-list.component.ts` and leave `setDisabledState` as the single source of truth.

ui-ngx/src/app/modules/home/components/widget/widget-config.component.html
```diff
@@ -58,7 +58,6 @@
                                                 brackets="curly"
                                                 subscriptSizing="dynamic"
                                                 [predefinedValuesButton]="true"
-                                                [disabled]="!widgetSettings.get('showTitle').value"
                                                 [predefinedValues]="predefinedValues"
                                                 placeholderText="{{ 'widget-config.set' | translate }}"
                                                 formControlName="title"
```
vvlladd28, line 63: This sweep covers all occurrences — I couldn't find any remaining same-element `[disabled]` + `formControlName`/`formControl` combination in ui-ngx — but nothing prevents the pattern from creeping back; it clearly proliferated by copy-paste across these widget config templates. Would it be worth adding a guard alongside this cleanup, e.g. an `@angular-eslint` template rule or a simple CI grep for `[disabled]` co-located with `formControlName`, so the next template doesn't reintroduce the NG01001 warning?

vvlladd28 (commented):
## Review summary

Reviewed 29 changed files in [Remove redundant [disabled] binding on reactive-form fields to fix NG01001 warning](https://github.com/thingsboard/thingsboard/pull/15878). Left 2 comment(s) inline.

Verified that every removed `[disabled]` binding is backed by proper FormControl `enable()`/`disable()` management: all 21 basic-config components run `updateValidators` at setup and on each `showTitle`/`showLabel`/`showSubtitle` toggle (via `BasicWidgetConfigComponent`), the settings components do the same via the `WidgetSettingsComponent` base or their own `writeValue`/`valueChanges` wiring, and `widget-config.component.ts` calls `updateWidgetSettingsEnabledState()` both on config set and on toggle changes. `tb-string-pattern-autocomplete.setDisabledState()` sets the internal control and the visual `disabled` state, so the CVA path fully replaces the removed bindings — no behavioral regression found, and no remaining same-element `[disabled]` + `formControlName` combination is left in ui-ngx.

## Additional findings

- **ui-ngx/src/app/shared/components/string-pattern-autocomplete.component.ts:70** — With this PR, no template in the codebase passes `[disabled]` into `tb-string-pattern-autocomplete` anymore; the only remaining writer of this field is `setDisabledState()`. The `@Input()` path is also only half-functional: `ngOnChanges` reacts to `predefinedValues`, `patternSymbol` and `required` but ignores `disabled`, so a binding set after init would flip the visual state without disabling the internal `selectionFormControl`. Since this PR eliminates the anti-pattern across 29 call sites, consider dropping the `@Input()` decorator here too (keeping the plain field for CVA state) so the dual disabled mechanism can't be reintroduced by the next caller.

---

vvlladd28 (approved):

PR #15885, opened 2026-07-01
Fixed overflow of value/action settings popovers

mtsymbarov-del:
## Pull Request description

Before:
Знімок екрана 2026-07-01 о 13 46 54

After:
Знімок екрана 2026-07-01 о 13 53 19

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/lib/settings/common/action/get-value-action-settings.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/settings/common/action/get-value-action-settings.component.ts
@@ -135,7 +135,7 @@ export class GetValueActionSettingsComponent implements OnInit, ControlValueAcce
         renderer: this.renderer,
         componentType: GetValueActionSettingsPanelComponent,
         hostView: this.viewContainerRef,
-        preferredPlacement: ['leftTopOnly', 'leftOnly', 'leftBottomOnly'],
+        preferredPlacement: 'leftBottom',
         context: {
           getValueSettings: this.modelValue,
           panelTitle: this.panelTitle,
@@ -147,7 +147,8 @@ export class GetValueActionSettingsComponent implements OnInit, ControlValueAcce
           targetDevice: this.targetDevice,
           widgetType: this.widgetType
         },
-        isModal: true
+        isModal: true,
+        overlayStyle: { maxHeight: '70vh', height: '100%' }
       });
       getValueSettingsPanelPopover.tbComponentRef.instance.popover = getValueSettingsPanelPopover;
       getValueSettingsPanelPopover.tbComponentRef.instance.getValueSettingsApplied.subscribe((getValueSettings) => {
--- a/ui-ngx/src/app/modules/home/components/widget/lib/settings/common/action/set-value-action-settings.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/settings/common/action/set-value-action-settings.component.ts
@@ -121,7 +121,7 @@ export class SetValueActionSettingsComponent implements OnInit, ControlValueAcce
        renderer: this.renderer,
        componentType: SetValueActionSettingsPanelComponent,
        hostView: this.viewContainerRef,
-       preferredPlacement: ['leftTopOnly', 'leftOnly', 'leftBottomOnly'],
+       preferredPlacement: 'leftBottom',
        context: {
          setValueSettings: this.modelValue,
          panelTitle: this.panelTitle,
@@ -130,7 +130,8 @@ export class SetValueActionSettingsComponent implements OnInit, ControlValueAcce
          targetDevice: this.targetDevice,
          widgetType: this.widgetType
        },
-       isModal: true
+       isModal: true,
+       overlayStyle: { maxHeight: '70vh', height: '100%' }
      });
       setValueSettingsPanelPopover.tbComponentRef.instance.popover = setValueSettingsPanelPopover;
       setValueSettingsPanelPopover.tbComponentRef.instance.setValueSettingsApplied.subscribe((setValueSettings) => {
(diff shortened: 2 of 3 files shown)


ui-ngx/src/app/modules/home/components/widget/lib/settings/common/action/get-value-action-settings.component.ts
```diff
@@ -135,7 +135,7 @@ export class GetValueActionSettingsComponent implements OnInit, ControlValueAcce
         renderer: this.renderer,
         componentType: GetValueActionSettingsPanelComponent,
         hostView: this.viewContainerRef,
-        preferredPlacement: ['leftTopOnly', 'leftOnly', 'leftBottomOnly'],
```
vvlladd28, line 138: This code was written specifically so that the popover would not move anywhere above/below, but would always be on the left.
You may need to call automatic repositioning when changing the type or see how this is implemented elsewhere.

ui-ngx/src/app/modules/home/components/widget/lib/settings/common/action/get-value-action-settings.component.ts
```diff
@@ -147,7 +147,8 @@ export class GetValueActionSettingsComponent implements OnInit, ControlValueAcce
           targetDevice: this.targetDevice,
           widgetType: this.widgetType
         },
-        isModal: true
+        isModal: true,
+        overlayStyle: { maxHeight: '70vh', height: '100%' }
```
vvlladd28, line 151: Moving the height constraint onto the overlay also moves the scroll container, and that costs the pinned footer.

In strict mode `popover.component.ts` adds `.strict-position` to `.tb-popover-inner-content` (`display: flex; overflow: hidden`) and sets `display: flex; height: 100%` on the panel host in `onComponentChange`. The panel then fills the popover, and the only scrolling element is `.tb-action-settings-panel-content` — so the title and the Cancel/Apply row stay put while the form scrolls.

Without strict mode none of that applies: the panel is laid out at its natural height and the scrolling element becomes `.tb-popover-inner-content`, which wraps the *whole* panel. The title and the Cancel/Apply buttons now scroll out of view together with the form. Easy to confirm in devtools — pick `Execute RPC` + a long attribute config and scroll inside the popover.

Same in `set-value-action-settings.component.ts:134` and `widget-action-settings.component.ts:132`.

Bounding the panel itself instead (`max-height: 90vh` on `.tb-action-settings-panel`, `flex: 1; min-height: 0` on `.tb-action-settings-panel-content`) keeps the strict placement, keeps the footer pinned, and is one SCSS file rather than three TS call sites — see `color-settings-panel.component.scss`, which already does exactly this.

vvlladd28 (commented):
## Review summary

Reviewed 3 changed files in [Fixed overflow of value/action settings popovers](https://github.com/thingsboard/thingsboard/pull/15885). Left 1 comment inline.

The change works, but it fixes the symptom at the overlay level rather than the cause at the panel level, and it costs two behaviours that the strict placement was providing: the left-only anchoring (already raised in review) and the pinned title / Apply-Cancel footer. There is an existing in-repo pattern (`color-settings-panel.component.scss`) that solves exactly this problem without giving up either.

## Why the popover truncates today (root cause)

With `['leftTopOnly', 'leftOnly', 'leftBottomOnly']` the popover runs in *strict* mode (`popover.component.ts:457-467`), which means `cdkConnectedOverlayFlexibleDimensions=true` and `cdkConnectedOverlayPush=false`. In that mode the CDK does **not** give the pane an exact height — it gives the bounding box a height equal to the gap between the trigger and the viewport edge (`_calculateBoundingBoxRect`), and `.cdk-overlay-pane { max-height: 100% }` clamps the popover to it.

The popover's height chain is then `height: auto` clamped by that max-height, so every `height: 100%` below it (`.tb-popover-content` → `.tb-popover-inner` → `.tb-popover-inner-content`) resolves to `auto`. The panel inside is laid out at its full natural height and simply gets clipped by `.tb-popover-inner { overflow: hidden }` — no scrollbar appears anywhere. That is exactly the PROD-5018 symptom: truncated, not scrollable.

`.tb-action-settings-panel-content` does have `overflow: auto`, but it never becomes a scroll container because nothing above it has a resolved height.

## Why other panels are fine

They are not structurally different — they bound themselves. `color-settings.component.ts:168` uses the same strict triple, and `color-settings-panel.component.scss` carries `min-height: 300px; max-height: 90vh` on the panel root plus `flex: 1; min-height: 0` on the scrolling section. A viewport-relative `max-height` resolves without needing the percentage chain, so the internal scroll region works while the popover stays left-only.

`action-settings-panel.component.scss` has no such bound — that is the whole difference.

## Suggested alternative

Keep `preferredPlacement: ['leftTopOnly', 'leftOnly', 'leftBottomOnly']` in all three components and change only `action-settings-panel.component.scss`:

```scss
.tb-action-settings-panel {
  width: 530px;
  max-height: 90vh;      // or min(90vh, …) to taste
  display: flex;
  flex-direction: column;
  gap: 16px;

  .tb-action-settings-panel-content {
    flex: 1;
    min-height: 0;
    overflow: auto;
    …
  }
}
```

One file instead of three, the left-only invariant is kept, the footer stays pinned, and it matches the pattern already used by the colour settings panel.

## Corner cases worth checking if the current approach is kept

- **Fallback placements.** As already noted in review, a plain `'leftBottom'` string expands to `[leftBottom, top, right, bottom, left]` (`popover.component.ts:461`). When the trigger sits near the top of the viewport — the PROD-5018 repro itself — `leftBottom` (which grows upward) will not fit, `top` will not fit, `right` is off the settings sidebar, and the popover lands on **`bottom`**, centred under the trigger and covering the settings panel it belongs to.
- **Narrow viewports.** Under `@media #{$mat-lt-md}` the panel is `90vw`, so there is never room to the left and the fallback above is the normal case, not the corner case.
- **`height: '100%'` is load-bearing, not decorative.** It is what makes `.tb-popover`'s height definite so `.tb-popover-inner-content { overflow: auto }` can scroll; drop it and `max-height: 70vh` clips silently, reproducing the original bug. The idiom is used elsewhere (`timewindow-config-dialog.component.ts:558`, `entity-versions-table.component.ts:238`), so it is consistent — just fragile enough to deserve a comment.
- **Repositioning on type change alone would not have been enough.** `_setBoundingBoxStyles` clamps with `Math.min(height, this._lastBoundingBoxSize.height)` unless `growAfterOpen` is set, and the popover template never sets `cdkConnectedOverlayGrowAfterOpen`. So a plain `updatePosition()` after the panel grows can only shrink the box, never grow it.

---

PR #15893, opened 2026-07-02
Fixed validation for Digital Gauge widget maxValue when it is zero

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/config/basic/gauge/digital-simple-gauge-basic-config.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/config/basic/gauge/digital-simple-gauge-basic-config.component.ts
@@ -30,7 +30,7 @@ import {
   getTimewindowConfig,
   setTimewindowConfig
 } from '@home/components/widget/config/timewindow-config-panel.component';
-import { formatValue, isDefined, isUndefined } from '@core/utils';
+import { formatValue, isDefined, isDefinedAndNotNull, isUndefined } from '@core/utils';
 import { Component } from '@angular/core';
 import {
   convertLevelColorsSettingsToColorProcessor,
@@ -190,7 +190,7 @@ export class DigitalSimpleGaugeBasicConfigComponent extends BasicWidgetConfigCom
   private maxValueValidation(): ValidatorFn {
     return (control: AbstractControl): ValidationErrors | null => {
       const value: string = control.value;
-      if (value) {
+      if (isDefinedAndNotNull(value)) {
         if (value < control.parent?.get('minValue').value) {
           return {maxValue: true};
         }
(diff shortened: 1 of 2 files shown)


ui-ngx/src/app/modules/home/components/widget/config/basic/gauge/digital-simple-gauge-basic-config.component.ts
```diff
@@ -190,7 +190,7 @@ export class DigitalSimpleGaugeBasicConfigComponent extends BasicWidgetConfigCom
   private maxValueValidation(): ValidatorFn {
     return (control: AbstractControl): ValidationErrors | null => {
       const value: string = control.value;
```
vvlladd28: The `: string` annotation on `value` is misleading — this control is bound to a numeric input, and the next line compares it numerically against `minValue`. Since the guard is being reworked anyway, it might be worth changing the type to `number` (or dropping the annotation and letting it infer). A guard that states the actual intent, e.g. `isNumeric(value)`, would also read more clearly than a generic null check, and the two nested `if`s could collapse into one condition. Same applies to the copy in `digital-gauge-widget-settings.component.ts`.

ui-ngx/src/app/modules/home/components/widget/config/basic/gauge/digital-simple-gauge-basic-config.component.ts
```diff
@@ -190,7 +190,7 @@ export class DigitalSimpleGaugeBasicConfigComponent extends BasicWidgetConfigCom
...
-      if (value) {
+      if (isDefinedAndNotNull(value)) {
         if (value < control.parent?.get('minValue').value) {
```
vvlladd28: Not introduced by this change, but closely related: when `minValue` is left empty its control value is `null`, and `value < null` coerces `null` to `0` — so a negative `maxValue` with an empty `minValue` shows the min/max error even though no minimum is set. (Your fix behaves fine for `0` here, since `0 < 0` is false.) Worth also guarding on `isDefinedAndNotNull` of the `minValue` value while you're in this code? Applies to both components.

vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Fixed validation for Digital Gauge widget maxValue when it is zero](https://github.com/thingsboard/thingsboard/pull/15893). Left 3 comment(s) inline.

The fix itself is correct: both `maxValue` fields are bound to `type="number"` inputs, so an empty field yields `null` (not `''`) and is still skipped by `isDefinedAndNotNull`, while `0` is now validated against `minValue` as intended. Both occurrences of the truthy-check pattern in the codebase are covered, and both components already re-validate `maxValue` on `minValue` changes. The inline comments are about deduplicating the copy-pasted validator, a misleading type annotation, and a pre-existing edge case with an empty `minValue`.

---

vvlladd28 (approved):

PR #15903, opened 2026-07-06
Added show "Not supported"/"N/A" text for non-numeric or empty telemetry values

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/lib/analogue-gauge.models.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/analogue-gauge.models.ts
@@ -17,7 +17,7 @@
 import * as CanvasGauges from 'canvas-gauges';
 import { FontSettings, getFontFamily } from '@home/components/widget/lib/settings.models';
 import { WidgetContext } from '@home/models/widget-component.models';
-import { isDefined, isDefinedAndNotNull } from '@core/utils';
+import { isDefined, isDefinedAndNotNull, isNumeric } from '@core/utils';
 import tinycolor from 'tinycolor2';
 import Highlight = CanvasGauges.Highlight;
 import BaseGauge = CanvasGauges.BaseGauge;
@@ -27,6 +27,7 @@ import { ValueFormatProcessor } from '@shared/models/widget-settings.models';
 import { UnitService } from '@core/services/unit.service';
 import { DataKey } from '@shared/models/widget.models';
 import { CustomTranslatePipe } from '@shared/pipe/custom-translate.pipe';
+import { NOT_SUPPORTED } from '@shared/models/telemetry/telemetry.models';
 
 export type AnimationRule = 'linear' | 'quad' | 'quint' | 'cycle'
                             | 'bounce' | 'elastic' | 'dequad' | 'dequint'
@@ -95,11 +96,16 @@ export abstract class TbBaseGauge<S, O extends GenericOptions> {
       const cellData = this.ctx.data[0];
       if (cellData.data.length > 0) {
         const tvPair = cellData.data[cellData.data.length - 1];
-        let value = parseFloat(tvPair[1]);
+        const origValue = tvPair[1];
+        let value = parseFloat(origValue);
         if (this.formatValue) {
           value = parseFloat(this.formatValue.format(value));
         }
-        if (value !== this.gauge.value) {
+        const valueText = !isNumeric(origValue) ? NOT_SUPPORTED : !value ? 'N/A' : '';
+        if (this.gauge.options.valueText !== valueText) {
+          this.gauge.update({valueText} as GenericOptions);
+        }
+        if (!isNaN(value) && value !== this.gauge.value) {
           if (!this.gauge.options.animation) {
             this.gauge._value = value;
           } else {
--- a/ui-ngx/src/app/modules/home/components/widget/lib/digital-gauge.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/digital-gauge.ts
@@ -21,7 +21,7 @@ import {
   DigitalGaugeSettings
 } from '@home/components/widget/lib/digital-gauge.models';
 import tinycolor from 'tinycolor2';
-import { isDefinedAndNotNull } from '@core/utils';
+import { isDefinedAndNotNull, isNumeric } from '@core/utils';
 import { prepareFontSettings } from '@home/components/widget/lib/settings.models';
 import { CanvasDigitalGauge, CanvasDigitalGaugeOptions } from '@home/components/widget/lib/canvas-digital-gauge';
 import { DatePipe } from '@angular/common';
@@ -36,6 +36,7 @@ import { UnitService } from '@core/services/unit.service';
 import { isNotEmptyTbUnits } from '@shared/models/unit.models';
 import { CustomTranslatePipe } from '@shared/pipe/custom-translate.pipe';
 import GenericOptions = CanvasGauges.GenericOptions;
+import { NOT_SUPPORTED } from '@shared/models/telemetry/telemetry.models';
 
 // @dynamic
 export class TbCanvasDigitalGauge {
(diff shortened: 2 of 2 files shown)


ui-ngx/src/app/modules/home/components/widget/lib/analogue-gauge.models.ts
```diff
@@ -95,11 +96,16 @@ export abstract class TbBaseGauge<S, O extends GenericOptions> {
...
       if (cellData.data.length > 0) {
         const tvPair = cellData.data[cellData.data.length - 1];
-        let value = parseFloat(tvPair[1]);
+        const origValue = tvPair[1];
+        let value = parseFloat(origValue);
         if (this.formatValue) {
           value = parseFloat(this.formatValue.format(value));
         }
-        if (value !== this.gauge.value) {
+        const valueText = !isNumeric(origValue) ? NOT_SUPPORTED : !value ? 'N/A' : '';
```
vvlladd28: `!value` treats a legitimate `0` as "no data". For `origValue = 0` (or `"0"`), `isNumeric()` passes, `value` is `0`, and `!0` is `true` — so the gauge shows `N/A` instead of `0`. Any device legitimately reporting zero (idle power draw, 0 °C, an empty tank) hits this.

It's worse than plain zero, because `value` here is the *formatted* value: with `decimals: 0` a reading of `0.4` formats to `"0"` → `parseFloat` → `0` → `N/A`. And since `formatValue` also unit-converts, 32 °F displayed in °C becomes `0` → `N/A`.

Suggest keying off the raw value and definedness rather than falsiness, e.g.:

```ts
const valueText = !isDefinedAndNotNull(origValue) || origValue === '' ? 'N/A'
  : !isNumeric(origValue) ? NOT_SUPPORTED : '';
```

That also matches the guard other widgets use (`battery-level-widget.component.ts:245`, `value-card-widget.component.ts:180`).

ui-ngx/src/app/modules/home/components/widget/lib/digital-gauge.ts
```diff
@@ -269,8 +270,15 @@ export class TbCanvasDigitalGauge {
           (this.gauge.options as CanvasDigitalGaugeOptions).labelTimestamp =
             filter.transform(timestamp, this.localSettings.timestampFormat);
         }
-        const value = parseFloat(tvPair[1]);
-        if (value !== this.gauge.value) {
+        const origValue = tvPair[1];
+        const value = parseFloat(origValue);
+        const options = this.gauge.options as CanvasDigitalGaugeOptions;
+        const valueText = !origValue ? 'N/A' : !isNumeric(origValue) ? NOT_SUPPORTED : '';
```
vvlladd28: Same zero problem as in `analogue-gauge.models.ts:104`, and additionally the two gauges disagree on the branch order.

`!origValue` is true for a numeric `0`. That isn't hypothetical: `EntityDataSubscription.convertValue()` (`entity-data-subscription.ts:172`) returns `Number(val)` when `Number(val).toString() === val`, so telemetry `"0"` reaches here as the number `0` → `N/A`. Note the same reading sent as `"0.0"` keeps its string form and renders correctly — so whether a zero displays depends on how the device formatted it.

The order also differs from the analogue gauge, which checks `!isNumeric()` first. For an empty/`null` value the two widgets in this same PR print different things: analogue → `Not supported!`, digital → `N/A`. Worth making both files use one shared expression so they can't drift.

ui-ngx/src/app/modules/home/components/widget/lib/digital-gauge.ts
```diff
@@ -269,8 +270,15 @@ export class TbCanvasDigitalGauge {
...
+        if (options.valueText !== valueText) {
+          options.valueText = valueText;
```
vvlladd28, line 279: `drawDigitalValue()` appends the unit symbol to whatever text it ends up with:

```ts
// canvas-digital-gauge.ts:773-774
let text = options.valueText || padValue(options.valueFormat.format(value), options.valueDec);
text += options.symbol;
```

`symbol` is set from the widget's configured units (`digital-gauge.ts:178`), so on any gauge with units this renders as `N/A%` or `Not supported!°C`. The placeholder branch should skip the symbol — e.g. `text = options.valueText ? options.valueText : padValue(...) + options.symbol;`

ui-ngx/src/app/modules/home/components/widget/lib/digital-gauge.ts
```diff
@@ -269,8 +270,15 @@ export class TbCanvasDigitalGauge {
...
+          this.gauge.update({} as CanvasDigitalGaugeOptions);
+        }
+        if (!isNaN(value) && value !== this.gauge.value) {
```
vvlladd28: Adding `!isNaN(value)` means the gauge keeps its *previous* value when the new one is non-numeric, so the bar/arc stays filled at the last good reading while the text reads `Not supported!`. Switching a key to an unsupported aggregation mid-session gives you a 75%-filled gauge labelled `Not supported!`, whereas reloading the page shows the same state at `minValue` — the same data renders two different ways.

Was leaving the needle in place deliberate? Resetting to `minValue` (or `0` progress) whenever `valueText` is non-empty would make the visual agree with the text. Same applies to `analogue-gauge.models.ts:108`.

vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Added show "Not supported"/"N/A" text for non-numeric or empty telemetry values](https://github.com/thingsboard/thingsboard/pull/15903). Left 4 comment(s) inline.

The most important one: in both gauges the "no data" branch is a falsiness check, so a legitimate telemetry value of **0** renders as `N/A` instead of `0`. `EntityDataSubscription.convertValue()` turns the string `"0"` into the number `0`, so `!origValue` (digital) and `!value` (analogue) both fire on it. On the digital gauge there is a second, separate problem: `drawDigitalValue()` appends `options.symbol` unconditionally, so with units configured the placeholder renders as `N/A%` / `Not supported!°C`.

Note also that PROD-2291 lists two problems — this PR addresses only the first (`NaN` text); the endless loading spinner on time-window change is untouched (tracked separately in CP-10212, so presumably intentional).

---

ui-ngx/src/app/modules/home/components/widget/lib/analogue-gauge.models.ts
```diff
@@ -95,17 +96,24 @@ export abstract class TbBaseGauge<S, O extends GenericOptions> {
       const cellData = this.ctx.data[0];
       if (cellData.data.length > 0) {
```
vvlladd28, line 97: The `N/A` branch can only fire when a data point actually exists and its value is `''` or null. When the key has no data points at all — `cellData.data.length === 0`, i.e. an empty time window or a key the device has never sent — `update()` returns without touching the gauge, and neither `createGaugeOptions()` here nor the digital gauge's options set an initial `value`, so canvas-gauges' default `value: 0` is what gets rendered. The user sees `0` (plus the unit symbol on the digital gauge) for "no data at all", which is arguably the most common empty case this PR is aiming at.

Handling the empty-dataset case here — and in `digital-gauge.ts:263`, which has the same `if (cellData.data.length > 0)` shape — would close that gap.

ui-ngx/src/app/modules/home/components/widget/lib/canvas-digital-gauge.ts
```diff
@@ -770,8 +770,7 @@ function drawDigitalValue(context: DigitalGaugeCanvasRenderingContext2D, options
   const textX = Math.round(baseX + width / 2);
   const textY = valueY;
 
-  let text = options.valueText || padValue(options.valueFormat.format(value), options.valueDec);
-  text += options.symbol;
+  let text = options.valueText ? options.valueText : padValue(options.valueFormat.format(value), options.valueDec) + options.symbol;
```
vvlladd28, line 773: `Not supported!` is 14 characters in a slot that normally holds a handful of digits, and nothing here adapts to that: the value font size comes only from the gauge geometry (`options.fontValueSize * bd.fontSizeFactor` in `barDimensions()`), and `drawText()` calls `fillText()` with no `maxWidth` and `textAlign = 'center'`, so the string is drawn at full width and simply spills past the bar/arc on a narrow widget or one with a large configured value font.

Did you check it at the smaller widget sizes? The analogue gauge raises the same question, since canvas-gauges sizes the value box from `valueInt`/`valueDec` rather than from `valueText`. Since the PR has no before/after screenshots it's hard to judge from the code alone — if it does overflow, scaling the font down when `options.valueText` is set (or using a shorter label) would be the fix.

vvlladd28 (commented):
## Review summary

Reviewed 3 changed files in [Added show "Not supported"/"N/A" text for non-numeric or empty telemetry values](https://github.com/thingsboard/thingsboard/pull/15903). Left 2 comment(s) inline.

All four findings from the previous automated review look addressed: the zero-falsiness bug is gone (both gauges now key off `isDefinedAndNotNull(origValue) || origValue === ''`), the branch order matches across the two files, the unit symbol is no longer appended to the placeholder, and the gauge now falls back to `minValue` instead of holding the last good reading. What remains is the case this PR doesn't cover — a key with *no* data points at all still renders `0` rather than `N/A` (`analogue-gauge.models.ts:97`) — and an open question about whether the 14-character `Not supported!` string fits inside the gauge (`canvas-digital-gauge.ts:773`).

Worth noting the PR description is still the unfilled repository template: for a purely visual change, the before/after screenshots the front-end checklist asks for would settle the second comment immediately.

---

PR #15908, opened 2026-07-07
Added show of warning toast when switching dashboard states reveals overlapping widgets in edit mode

mtsymbarov-del:
## Pull Request description

Result:
Знімок екрана 2026-07-07 о 11 52 21

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/core/services/dashboard-utils.service.ts
+++ b/ui-ngx/src/app/core/services/dashboard-utils.service.ts
@@ -33,7 +33,8 @@ import {
   DashboardStateLayouts,
   GridSettings,
   LayoutType,
-  WidgetLayout
+  WidgetLayout,
+  WidgetLayouts
 } from '@shared/models/dashboard.models';
 import { deepClone, isDefined, isDefinedAndNotNull, isNotEmptyStr, isString, isUndefined } from '@core/utils';
 import {
@@ -766,6 +767,25 @@ export class DashboardUtilsService {
       const bottom2 = widget.row + widget.sizeY;
 
       if (left < right2 && right > left2 && top < bottom2 && bottom > top2) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  public hasCollidingWidgets(widgetLayouts: WidgetLayouts): boolean {
+    if (!widgetLayouts) {
+      return false;
+    }
+    const layouts = Object.values(widgetLayouts).map(widget => ({
+      row: widget.row || 0,
+      col: widget.col || 0,
+      sizeX: widget.sizeX || 1,
+      sizeY: widget.sizeY || 1
+    }));
+    for (let i = 0; i < layouts.length; i++) {
+      const widget = layouts[i];
+      if (this.hasWidgetCollision(widget.row, widget.col, widget.sizeX, widget.sizeY, layouts.slice(i + 1))) {
         return true;
       }
     }
--- a/ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-page.component.ts
+++ b/ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-page.component.ts
@@ -43,6 +43,7 @@ import { UtilsService } from '@core/services/utils.service';
 import {
   BreakpointId,
   BreakpointInfo,
+  BreakpointLayoutInfo,
   Dashboard,
   DashboardConfiguration,
   DashboardLayoutId,
@@ -149,6 +150,7 @@ import {
   MoveWidgetsDialogResult
 } from '@home/components/dashboard-page/layout/move-widgets-dialog.component';
 import { HttpStatusCode } from '@angular/common/http';
+import { ActionNotificationShow } from '@core/notification/notification.actions';
 
 // @dynamic
 @Component({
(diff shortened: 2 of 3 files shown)


ui-ngx/src/app/core/services/dashboard-utils.service.ts
```diff
@@ -772,6 +773,25 @@ export class DashboardUtilsService {
     return false;
   }
 
+  public hasCollidingWidgets(widgetLayouts: WidgetLayouts): boolean {
```
vvlladd28, line 776: The detection is sound, but it reports a corrupted layout instead of preventing one — worth establishing where the overlap comes from before shipping a toast for it.

The likely producer is `updateLayoutSettings()` at line 823 of this same file, reached from the layout settings dialog (`manage-dashboard-layouts-dialog.component.ts:281`) and from `dashboard-page.component.ts:949`. It rescales every widget by `ratio = columns / prevColumns`, rounding position and size **independently**:

```ts
widget.col = Math.round(widget.col * ratio);
widget.sizeX = Math.max(1, Math.round(widget.sizeX * ratio));
```

`round(col*r) + round(sizeX*r)` is not `round((col + sizeX)*r)`, and the `Math.max(1, …)` floor compounds it when shrinking. Concretely, 24 → 10 columns (both well inside the dialog's `Validators.min(10)`/`max(1008)` range) with two adjacent 1-column widgets `{col: 0, sizeX: 1}` and `{col: 1, sizeX: 1}`: the first stays at `[0,1)`, the second becomes `col = Math.round(0.4167) = 0`, `sizeX = 1` → also `[0,1)`. They now fully overlap.

The fixer loop just below only repairs an overlap of exactly one unit where one of the two widgets is larger than 1 on that axis (`right - left2 === 1` with `sizeX > 1`). Neither widget above is wider than 1, so nothing is repaired and the overlap is written into `dashboard.configuration` and persisted. That loop also mutates `widget2.col`/`widget2.row` while walking every ordered pair with no convergence pass, so it can create fresh overlaps against widgets it already visited.

So the user ends up with a warning about a layout the editor itself corrupted — and since `dashboard.component.ts:233` sets `disableAutoPositionOnConflict: false`, gridster will additionally shuffle those widgets at render time.

Could this be addressed at the source instead — rescale the widget's right/bottom edge rather than its size (`sizeX = Math.round((col + sizeX) * ratio) - Math.round(col * ratio)`, clamped to ≥ 1), and re-place anything still conflicting via `widgetPossiblePosition()` (line 722), which already finds a collision-free slot? For layouts that are already broken in the DB a warning still has value, but it shouldn't be the only answer to a defect the editor keeps re-creating.

ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-page.component.ts
```diff
@@ -1106,6 +1108,17 @@ export class DashboardPageComponent extends PageComponent implements IDashboardC
     if (!this.destroyed) {
       const layoutsData = this.dashboardUtils.getStateLayoutsData(this.dashboard, state);
       if (layoutsData) {
+        if (this.isEdit && Object.keys(layoutsData).some(l => {
```
vvlladd28, line 1111: This fires on every `openDashboardState()` call in edit mode, not only on an actual state switch. `updateStates()` (line 1033) re-opens the *current* state after the "Manage dashboard states" dialog is saved (`manageDashboardStates` → line 974 → `updateStates` → `openDashboardState(targetState)`), so an author who just renames or adds an unrelated state gets an overlap warning about the state they're already looking at, with nothing having changed and nothing having been "revealed".

Would it be worth gating on the state actually changing — e.g. `state !== this.dashboardCtx.state` — so the toast matches the behaviour the PR title describes?

ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-page.component.ts
```diff
@@ -1106,6 +1108,17 @@ export class DashboardPageComponent extends PageComponent implements IDashboardC
...
+          const layout: DashboardPageLayout = this.layouts[l];
+          const breakpoint = (layout && layout.layoutCtx.breakpoint) || 'default';
+          const layoutInfo: BreakpointLayoutInfo = layoutsData[l][breakpoint] || layoutsData[l].default;
+          return layoutInfo && this.dashboardUtils.hasCollidingWidgets(layoutInfo.widgetLayouts);
+        })) {
+          this.store.dispatch(new ActionNotificationShow({
+            message: this.translate.instant('dashboard.state-widgets-overlap-warning'),
+            type: 'warn'
```
vvlladd28, line 1119: With no `duration`, `toast.directive.ts` passes `duration: undefined` into `MatSnackBarConfig` and skips its own dismiss timer (both `showSnackBar` and `showToastPanel` only arm one when `duration > 0`), so this toast stays up until the user clicks it away.

That's the house default for the other ~77 dispatches, but those describe a completed event ("failed to save…") and stay true afterwards. This one describes the state currently on screen: switch A (overlapping) → B (clean) and the warning is still sitting there, now describing a state the user has left. A short `duration` (the only other `type: 'warn'` in the app, `security.component.ts:246`, uses `1500`) would tie its lifetime to the switch that triggered it.

vvlladd28 (commented):
## Review summary

Reviewed 3 changed files in [Added show of warning toast when switching dashboard states reveals overlapping widgets in edit mode](https://github.com/thingsboard/thingsboard/pull/15908). Left 3 comment(s) inline.

The detection logic itself is correct — pairwise AABB comparison through the existing `hasWidgetCollision`, each pair tested once, and the breakpoint resolution matches exactly what `updateLayout` → `updatedLayoutForBreakpoint` renders in edit mode.

The main question is the one on `dashboard-utils.service.ts:776`: the overlap this warns about looks like it is produced by the editor itself, in `updateLayoutSettings()` in that same file, so this fixes the symptom rather than the cause. The other two comments are about when and for how long the toast shows.

---

PR #15912, opened 2026-07-08
Fixed stale dashboard state after state deletion/rename

mtsymbarov-del:
## Pull Request description

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/dashboard-page/states/entity-state-controller.component.ts
+++ b/ui-ngx/src/app/modules/home/components/dashboard-page/states/entity-state-controller.component.ts
@@ -61,7 +61,7 @@ export class EntityStateControllerComponent extends StateControllerComponent imp
 
   public init() {
     if (this.preservedState) {
-      this.stateObject = this.preservedState;
+      this.stateObject = this.syncPreservedStateWithDashboardState(this.preservedState);
       this.selectedStateIndex = this.stateObject.length - 1;
       setTimeout(() => {
         this.gotoState(this.stateObject[this.stateObject.length - 1].id, true);
@@ -83,6 +83,29 @@ export class EntityStateControllerComponent extends StateControllerComponent imp
   }
 
   protected onStatesChanged() {
+    const prevStateId = this.getStateId();
+    let i = this.stateObject.length;
+    while (i--) {
+      if (!this.stateObject[i].id || !this.states[this.stateObject[i].id]) {
+        this.stateObject.splice(i, 1);
+      }
+    }
+    if (!this.stateObject.length) {
+      const currentStateId = this.dashboardCtrl.dashboardCtx.state;
+      this.stateObject.push({
+        id: currentStateId && this.states[currentStateId] ? currentStateId : this.dashboardUtils.getRootStateId(this.states),
+        params: {}
+      });
+    }
+    this.selectedStateIndex = this.stateObject.length - 1;
+    const newStateId = this.getStateId();
+    if (newStateId !== prevStateId) {
+      this.stateIdSubject.next(newStateId);
+      if (this.syncStateWithQueryParam) {
+        this.mobileService.handleDashboardStateName(this.getStateName(this.stateObject.length - 1));
+      }
+      this.updateLocation(false);
+    }
   }
 
   protected onStateChanged() {
@@ -274,6 +297,24 @@ export class EntityStateControllerComponent extends StateControllerComponent imp
       if (!result[i].id || !this.states[result[i].id]) {
         result.splice(i, 1);
       }
+    }
+    return result;
+  }
+
+  private syncPreservedStateWithDashboardState(preservedState: StateControllerState): StateControllerState {
+    const result = preservedState.filter((stateObj) => stateObj.id && this.states[stateObj.id]);
+    const currentStateId = this.dashboardCtrl.dashboardCtx.state;
+    if (currentStateId && this.states[currentStateId] &&
+        (!result.length || result[result.length - 1].id !== currentStateId)) {
+      const stateIndex = result.map((stateObj) => stateObj.id).lastIndexOf(currentStateId);
+      if (stateIndex > -1) {
+        result.splice(stateIndex + 1);
+      } else {
+        result.push({ id: currentStateId, params: {} });
+      }
+    }
+    if (!result.length) {
+      result.push({ id: this.dashboardUtils.getRootStateId(this.states), params: {} });
     }
     return result;
   }
--- a/ui-ngx/src/app/modules/home/components/dashboard-page/states/default-state-controller.component.ts
+++ b/ui-ngx/src/app/modules/home/components/dashboard-page/states/default-state-controller.component.ts
@@ -77,6 +77,18 @@ export class DefaultStateControllerComponent extends StateControllerComponent im
   }
 
   protected onStatesChanged() {
+    const stateObj = this.stateObject[0];
+    if (stateObj && (!stateObj.id || !this.states[stateObj.id])) {
+      const currentStateId = this.dashboardCtrl.dashboardCtx.state;
+      stateObj.id = currentStateId && this.states[currentStateId]
+        ? currentStateId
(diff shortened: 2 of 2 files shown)


vvlladd28 (commented):
## Review summary

Reviewed 2 changed files in [Fixed stale dashboard state after state deletion/rename](https://github.com/thingsboard/thingsboard/pull/15912). Left 3 comment(s) inline.

The fix works for the reported scenario — `onStatesChanged()` was an empty stub in both controllers, so nothing re-emitted on `stateIdSubject` after a state id was renamed, and `DashboardStateGetter` (which subscribes to `dashboardCtx.stateId`) kept the old value until reload. But the repair is applied downstream of where the invariant is actually broken, and it lands as a third and fourth copy of logic that already exists twice in these files. The most important points: `syncPreservedStateWithDashboardState()` pushes a state with empty `params` into an *entity* state stack (entity-state-controller.component.ts:313), and the root-cause note on `updateStates()` below.

## Additional findings

- **ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-page.component.ts:1034** — This is where the invariant actually breaks, and it is worth considering fixing it here instead of repairing the consequences in each state controller. `gotoState()` in both controllers is the single place that keeps the four pieces of current-state in sync: it calls `dashboardCtrl.openDashboardState()` (updates `dashboardCtx.state` + layouts), emits `stateIdSubject`, updates the mobile state name, and rewrites the URL. `updateStates()` is the **only** caller of `openDashboardState()` anywhere that does not go through a controller's `gotoState()` — I grepped, the other two call sites are `default-state-controller.component.ts:256` and `entity-state-controller.component.ts:324`, both inside `gotoState`. So it moves `dashboardCtx.state` and the layouts while leaving `stateObject`, `stateIdSubject` and the URL pointing at the deleted id. Driving the change through the controller here (push the new `states` map, then `stateController.updateState(targetState)`) would fix it once, at the site that breaks it, instead of once per controller.

  There is a second cost to repairing it downstream: both new `onStatesChanged()` implementations read `this.dashboardCtrl.dashboardCtx.state` and depend on `updateStates()` having already called `openDashboardState()` before Angular propagates the `[states]` input. That ordering holds today only because `updateStates()` happens to be fully synchronous, and nothing at either end documents the dependency — reordering those two lines would silently reintroduce the bug.

---

ui-ngx/src/app/modules/home/components/dashboard-page/states/entity-state-controller.component.ts
```diff
@@ -278,6 +301,24 @@ export class EntityStateControllerComponent extends StateControllerComponent imp
...
+  private syncPreservedStateWithDashboardState(preservedState: StateControllerState): StateControllerState {
+    const result = preservedState.filter((stateObj) => stateObj.id && this.states[stateObj.id]);
+    const currentStateId = this.dashboardCtrl.dashboardCtx.state;
+    if (currentStateId && this.states[currentStateId] &&
+        (!result.length || result[result.length - 1].id !== currentStateId)) {
+      const stateIndex = result.map((stateObj) => stateObj.id).lastIndexOf(currentStateId);
+      if (stateIndex > -1) {
+        result.splice(stateIndex + 1);
+      } else {
+        result.push({ id: currentStateId, params: {} });
```
vvlladd28, line 313: This pushes a state into an **entity** state stack with `params: {}`, which drops the entity context that this controller exists to carry.

Reachable path: `settings.stateControllerId === 'entity'`; the user drills A → B, where B's params hold `entityId`/`entityName`. Entering edit mode preserves `[A, B]` and swaps to the default controller, whose template (`default-state-controller.component.html:19`) shows a `mat-select` over all states whenever `displayStateSelection()` is true — so the user picks unrelated state C, making `dashboardCtx.state === 'C'`. On exit, `result` is `[A, B]`, the last id is not C, `lastIndexOf('C')` is -1, so `{ id: 'C', params: {} }` is pushed and `init()` calls `gotoState('C')`. `getStateParams()` now returns `{}` and `getEntityId()` returns `undefined`, so widgets in C bound to the state entity render empty.

Note also that every other path that appends to `stateObject` (`openState`, `pushAndOpenState`, `updateState`) goes through `resolveEntity()` first; this one does not. Before this change the controller restored `[A, B]` and `gotoState(B)` navigated back to B with its params intact, so C was never entered without them.

Would keeping the preserved entry's params (or falling back to `getStateParamsByStateId(currentStateId)`) work here, rather than `{}`?

ui-ngx/src/app/modules/home/components/dashboard-page/states/entity-state-controller.component.ts
```diff
@@ -83,6 +83,29 @@ export class EntityStateControllerComponent extends StateControllerComponent imp
   }
 
   protected onStatesChanged() {
+    const prevStateId = this.getStateId();
+    let i = this.stateObject.length;
+    while (i--) {
+      if (!this.stateObject[i].id || !this.states[this.stateObject[i].id]) {
```
vvlladd28, line 89: This is now the third copy of "drop entries whose id is missing from `this.states`" in this one file — `parseState()` at lines 295-300 has the identical `while (i--)` loop, and `syncPreservedStateWithDashboardState()` at line 305 does it again with `filter`. Counting `default-state-controller.component.ts` (`parseState()` lines 245-250 and the new `onStatesChanged()`), that is five hand-maintained copies of the same normalization.

They have already drifted apart in what they fall back to, which is the concrete problem:

- `parseState()` → `getRootStateId()` only, and it *rewrites* `result[0].id` before filtering;
- the new `onStatesChanged()` → `dashboardCtx.state` first, then `getRootStateId()`;
- `syncPreservedStateWithDashboardState()` → `dashboardCtx.state` with splice/append semantics, then `getRootStateId()`.

So the same stale stack normalizes to different results depending on which entry point runs — and `reInit()` (`parseState`) and the `states` setter (`onStatesChanged`) can both run against the same input within a few milliseconds of each other. The next change to what "a valid state stack" means has to be made in five places that already disagree.

Would one `normalizeStateObject(states: StateControllerState): StateControllerState` on `StateControllerComponent` — filter invalid entries, then fall back to `dashboardCtx.state` / root — cover all three call sites here and both in the default controller?

ui-ngx/src/app/modules/home/components/dashboard-page/states/entity-state-controller.component.ts
```diff
@@ -83,6 +83,29 @@ export class EntityStateControllerComponent extends StateControllerComponent imp
...
+      });
+    }
+    this.selectedStateIndex = this.stateObject.length - 1;
+    const newStateId = this.getStateId();
+    if (newStateId !== prevStateId) {
+      this.stateIdSubject.next(newStateId);
+      if (this.syncStateWithQueryParam) {
+        this.mobileService.handleDashboardStateName(this.getStateName(this.stateObject.length - 1));
+      }
+      this.updateLocation(false);
```
vvlladd28, line 107: Two things about this block:

1. The parameter is `isStateIdChanged`, and this call sits inside `if (newStateId !== prevStateId)` — i.e. the state id *did* change, but `false` is passed. It happens to give the wanted `replaceCurrentHistoryUrl: true` in `updateStateParam()`, but the argument contradicts the branch it is in, and `updateLocation()`'s only other caller (`gotoState`, line 332) passes the genuine value. Worth a comment at minimum, since the two call sites now mean opposite things by the same flag.

2. Unlike `init()`, `onStateChanged()`, `openState()`, `pushAndOpenState()`, `updateState()`, `navigatePrevState()` and `resetState()`, this path never reaches `gotoState()`, so `dashboardCtrl.openDashboardState()` is never called. `stateIdSubject` emits the new id and the URL is rewritten, but `dashboardCtx.state` is left wherever it was — which in the removed-state case is the id that just disappeared. The default controller's version sidesteps this by treating `dashboardCtx.state` as the source of truth; this one treats its own stack as the source of truth and only consults `dashboardCtx.state` when the stack empties. Was the divergence deliberate, or should this go through `gotoState()` like every other mutation in the class?

PR #15913, opened 2026-07-08
Sync persistent RPC (v2) status updates & delete events Edge->Cloud

MazurenkoNick:
## Pull Request description

Closes https://github.com/thingsboard/thingsboard-edge/issues/210

image

### Problem

Persistent RPC (v2) state was maintained independently on the edge and the cloud. When a persistent RPC changed status or was deleted on the edge, the cloud never found out, leaving stale RPC copies out of sync between the two.

### What this does

Propagates persistent-RPC (v2) lifecycle changes from the edge to the cloud by hooking the entity lifecycle (the same mechanism used for every other synced entity), so status transitions and deletions on the edge are mirrored on the cloud.

- Status updates (Edge → Cloud): TbRpcService.save now publishes a SaveEntityEvent<Rpc>. CloudEventSourcingListener turns it into a DEVICE/RPC_CALL uplink notification carrying {requestUUID, rpcStatus, response?}; the cloud locates the matching Rpc (same id == request UUID) and updates it, guarding against stale,
out-of-order status uplinks regressing a terminal RPC.
- Delete (Edge → Cloud): TbRpcService.deleteRpc fetches-before-delete and publishes a DeleteEntityEvent<Rpc>, uplinked over the same DEVICE/RPC_CALL channel with rpcStatus = DELETED; the cloud removes its copy.
- Delete (Cloud → Edge), all statuses: RpcV2Controller.deleteRpc now routes through TbRpcService and pushes a DELETED marker to all related edges regardless of status (previously terminal RPCs were skipped); the edge's DeviceCloudProcessor removes the local copy and clears the pending actor entry for intermediate
RPCs.

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/service/rpc/TbRpcService.java
```diff
@@ -80,4 +93,29 @@ public PageData<Rpc> findAllByDeviceIdAndStatus(TenantId tenantId, DeviceId devi
...
+        rpcService.deleteRpc(tenantId, rpcId);
+        // RPC v2 (persistent) delete/abort propagation Cloud -> Edge for ANY status.
+        // Skipped when the delete originated from an edge (edge-sync context) so it is not echoed back.
+        if (rpc != null && edgeSynchronizationManager.getEdgeId().get() == null) {
+            pushRpcDeleteToEdges(tenantId, rpc);
+        }
+    }
+
+    private void pushRpcDeleteToEdges(TenantId tenantId, Rpc rpc) {
+        List<EdgeId> relatedEdgeIds = edgeService.findAllRelatedEdgeIds(tenantId, rpc.getDeviceId());
```
MazurenkoNick, line 107: double-check if it's valid. synchronise with the search mechanism of related edge in DeviceActorProcessor

application/src/main/java/org/thingsboard/server/service/rpc/TbRpcService.java
```diff
@@ -67,4 +80,29 @@ public PageData<Rpc> findAllByDeviceIdAndStatus(TenantId tenantId, DeviceId devi
...
+        if (relatedEdgeIds == null || relatedEdgeIds.isEmpty()) {
+            return;
+        }
+        ObjectNode body = JacksonUtil.newObjectNode();
+        body.put("requestUUID", rpc.getId().getId().toString());
+        body.put("rpcStatus", RpcStatus.DELETED.name());
+        for (EdgeId edgeId : relatedEdgeIds) {
+            EdgeEvent edgeEvent = EdgeUtils.constructEdgeEvent(tenantId, edgeId, EdgeEventType.DEVICE,
+                    EdgeEventActionType.RPC_CALL, rpc.getDeviceId(), body);
+            tbClusterService.onEdgeHighPriorityMsg(new EdgeHighPriorityMsg(tenantId, edgeEvent));
```
MazurenkoNick, line 104: `onEdgeHighPriorityMsg` only reaches an edge that is currently connected — `EdgeGrpcService.onEdgeHighPriorityEvent` does `sessions.get(edgeId)` and, if the session is null or disconnected, returns without doing anything. Nothing here persists the `EdgeEvent` first (`EdgeUtils.constructEdgeEvent` is a pure builder), so a persistent RPC deleted on the cloud while the edge is offline is never propagated, and the edge keeps the stale copy — the exact situation this PR is trying to fix.

`DeviceActorMessageProcessor.saveRpcRequestToEdgeQueue` can get away with the high-priority path because an RPC request is time-bounded and pointless to replay, but a delete is durable state. Every other entity delete goes through `saveEdgeEvent` → `edgeEventService.saveAsync`, so it lands in `edge_event` and is replayed on reconnect. Should the DELETED marker be persisted the same way (optionally in addition to the high-priority push, so a connected edge still gets it immediately)?

application/src/main/java/org/thingsboard/server/service/edge/rpc/processor/device/DeviceEdgeProcessor.java
```diff
@@ -125,14 +129,51 @@ private void pushDeviceCreatedEventToRuleEngine(TenantId tenantId, Edge edge, De
     @Override
     public ListenableFuture<Void> processDeviceRpcCallFromEdge(TenantId tenantId, Edge edge, DeviceRpcCallMsg deviceRpcCallMsg) {
         log.trace("[{}] processDeviceRpcCallFromEdge [{}]", tenantId, deviceRpcCallMsg);
-        if (deviceRpcCallMsg.hasResponseMsg()) {
+        if (deviceRpcCallMsg.hasRpcStatus()) {
```
MazurenkoNick, line 132: Putting `hasRpcStatus()` first means any uplink that sets both `rpcStatus` and `responseMsg` now takes the status branch and never reaches `processDeviceRpcResponseFromEdge`. That matters: `processDeviceRpcResponseFromEdge` is what pushes `FromDeviceRpcResponseActorMsg` to core, and `DeviceActorMessageProcessor.processRpcResponsesFromEdge` is what removes the entry from `toDeviceRpcPendingMap` and calls `processRpcResponseFromDeviceActor`. Skip it and the cloud-side pending RPC is never resolved — the REST caller (or rule node) waiting on a two-way persistent RPC would sit there until it times out, even though the DB row now shows SUCCESSFUL.

Since `processDeviceRpcStatusFromEdge` reads `deviceRpcCallMsg.getResponseMsg().getResponse()`, it looks like a combined message is expected. Does the edge send the terminal status and the response together, or as two separate messages? If together, the status branch probably needs to also do the response handling (or fall through to it); if separate, the `hasResponseMsg()` read in the status handler is dead.

Separately: `DeviceRpcCallMsg` is now an implicit three-way union discriminated purely by which optional fields happen to be set, and this dispatch order silently decides the meaning of ambiguous messages. The producer side had to grow matching presence checks (`body.has("method")`, the conditional `requestId`). An explicit discriminator — a `msgType` enum the way `DeviceUpdateMsg` uses `UpdateMsgType`, or a dedicated `DeviceRpcStatusMsg` — would make the routing self-evident. Was reusing `DeviceRpcCallMsg` driven by wire compatibility with existing edge builds?

application/src/main/java/org/thingsboard/server/service/edge/rpc/processor/device/DeviceEdgeProcessor.java
```diff
@@ -125,14 +129,51 @@ private void pushDeviceCreatedEventToRuleEngine(TenantId tenantId, Edge edge, De
...
+                try {
+                    edgeSynchronizationManager.getEdgeId().set(edge.getId());
+                    edgeCtx.getTbRpcService().deleteRpc(tenantId, rpcId);
+                } finally {
+                    edgeSynchronizationManager.getEdgeId().remove();
+                }
+            }
+            return Futures.immediateFuture(null);
+        }
+        RpcStatus rpcStatus = RpcStatus.valueOf(deviceRpcCallMsg.getRpcStatus());
```
MazurenkoNick, line 158: `RpcStatus.valueOf` throws `IllegalArgumentException` on anything that isn't an exact enum name — an empty string (`hasRpcStatus()` is true for `""`), or a status added by a newer edge build. That exception propagates out of `processDeviceRpcCallFromEdge` into `EdgeGrpcSession.processUplinkMsg`, which catches it and returns `Futures.immediateFailedFuture(e)` for the **entire** uplink batch. The edge will then retry the same batch, hit the same value, and fail again — a poison-pill loop that also fires an `EdgeCommunicationFailureTrigger` on every attempt.

Worth parsing defensively here (e.g. a `try`/`catch` that logs and returns `immediateFuture(null)`, or an `EnumUtils.getEnum`-style lookup with a null check) so one malformed value can't wedge the session.

application/src/main/java/org/thingsboard/server/service/edge/rpc/processor/device/DeviceEdgeProcessor.java
```diff
@@ -125,14 +129,51 @@ private void pushDeviceCreatedEventToRuleEngine(TenantId tenantId, Edge edge, De
...
+        }
+        if (!existingRpc.getStatus().isIntermediate() && rpcStatus.isIntermediate()) {
+            // guard against out-of-order status uplinks regressing a terminal RPC (e.g. stale DELIVERED after SUCCESSFUL)
+            log.debug("[{}][{}] Ignoring stale RPC status [{}] - RPC is already terminal [{}]",
+                    tenantId, rpcId, rpcStatus, existingRpc.getStatus());
+            return Futures.immediateFuture(null);
+        }
+        JsonNode response = null;
+        if (deviceRpcCallMsg.hasResponseMsg() && !StringUtils.isEmpty(deviceRpcCallMsg.getResponseMsg().getResponse())) {
+            response = JacksonUtil.toJsonNode(deviceRpcCallMsg.getResponseMsg().getResponse());
```
MazurenkoNick, line 171: `JacksonUtil.toJsonNode` throws `IllegalArgumentException` when the payload isn't valid JSON, and a device response frequently isn't (a bare `ok`, a plain number as text, etc.). Same blast radius as the `valueOf` above — it escapes into `EdgeGrpcSession.processUplinkMsg` and fails the whole uplink batch.

The local (non-edge) path already handles this explicitly, in `DeviceActorMessageProcessor.processRpcResponses`:

```java
try {
    response = JacksonUtil.toJsonNode(payload);
} catch (IllegalArgumentException e) {
    response = JacksonUtil.newObjectNode().put("error", payload);
}
```

Mirroring that here would keep the edge-routed RPC's stored response consistent with the locally-routed one, and remove the failure mode.

application/src/main/java/org/thingsboard/server/service/edge/rpc/processor/device/DeviceEdgeProcessor.java
```diff
@@ -125,14 +129,51 @@ private void pushDeviceCreatedEventToRuleEngine(TenantId tenantId, Edge edge, De
...
+                }
+            }
+            return Futures.immediateFuture(null);
+        }
+        RpcStatus rpcStatus = RpcStatus.valueOf(deviceRpcCallMsg.getRpcStatus());
+        Rpc existingRpc = edgeCtx.getTbRpcService().findRpcById(tenantId, rpcId);
+        if (existingRpc == null) {
+            return Futures.immediateFuture(null);
+        }
+        if (!existingRpc.getStatus().isIntermediate() && rpcStatus.isIntermediate()) {
```
MazurenkoNick, line 163: This only rejects terminal → intermediate. Two cases still get through:

- terminal → terminal, so a late `TIMEOUT`/`FAILED` can overwrite an already-recorded `SUCCESSFUL`;
- intermediate → intermediate, so a reordered `QUEUED` can regress a `DELIVERED`.

If the concern is genuinely stale/out-of-order uplinks, is an ordinal-style check (only accept a status that is "further along", and never change a terminal one at all) closer to what you want? If the intent is narrower — only protect the terminal state — then `if (!existingRpc.getStatus().isIntermediate())` alone would express that and also cover the terminal → terminal case.

application/src/main/java/org/thingsboard/server/service/edge/rpc/processor/device/DeviceEdgeProcessor.java
```diff
@@ -125,14 +129,51 @@ private void pushDeviceCreatedEventToRuleEngine(TenantId tenantId, Edge edge, De
...
+            return processDeviceRpcStatusFromEdge(tenantId, edge, deviceRpcCallMsg);
+        } else if (deviceRpcCallMsg.hasResponseMsg()) {
             return processDeviceRpcResponseFromEdge(tenantId, deviceRpcCallMsg);
         } else if (deviceRpcCallMsg.hasRequestMsg()) {
             return processDeviceRpcRequestFromEdge(tenantId, edge, deviceRpcCallMsg);
         }
         return Futures.immediateFuture(null);
     }
 
+    private ListenableFuture<Void> processDeviceRpcStatusFromEdge(TenantId tenantId, Edge edge, DeviceRpcCallMsg deviceRpcCallMsg) {
```
MazurenkoNick, line 143: `edge` is only used to populate the edge-sync ThreadLocal — nothing verifies that the RPC named by `requestUuid` actually belongs to a device assigned to *this* edge. Scoping is tenant-wide, so any edge in the tenant can delete or rewrite the status of any persistent RPC in that tenant if it knows (or guesses) the UUID. `processDeviceRpcResponseFromEdge` has the same shape, so this isn't a regression, but delete is destructive in a way a response isn't — would a check that `existingRpc.getDeviceId()` is related to `edge.getId()` be worth the extra lookup here?

Structurally, this method is also doing two unrelated things (delete propagation and status update) and returns `Futures.immediateFuture(null)` from five places for what is entirely synchronous work. Splitting the DELETED branch into its own `processDeviceRpcDeleteFromEdge(...)`, dispatched from `processDeviceRpcCallFromEdge` next to the other branches, would let both halves be plain `void` methods with a single future wrap at the call site — and would keep `edge` local to the path that actually needs it.

application/src/main/java/org/thingsboard/server/service/edge/rpc/processor/device/DeviceEdgeProcessor.java
```diff
@@ -125,14 +129,51 @@ private void pushDeviceCreatedEventToRuleEngine(TenantId tenantId, Edge edge, De
...
+        RpcId rpcId = new RpcId(new UUID(deviceRpcCallMsg.getRequestUuidMSB(), deviceRpcCallMsg.getRequestUuidLSB()));
+        if (RpcStatus.DELETED.name().equals(deviceRpcCallMsg.getRpcStatus())) {
+            // RPC v2 (persistent) delete/abort propagation Edge -> Cloud: remove the cloud copy.
+            // Mark the edge-sync context so TbRpcService.deleteRpc does not echo the delete back to edges.
+            if (edgeCtx.getTbRpcService().findRpcById(tenantId, rpcId) != null) {
```
MazurenkoNick, line 148: This read exists only to decide whether to call `deleteRpc`, and `deleteRpc` immediately does its own `rpcService.findById` — two selects per uplink. The pre-check isn't buying anything either, since `deleteRpc` already skips the edge push when the row is gone.

The same duplication shows up on the REST path: `RpcV2Controller.deleteRpc` already has the entity from `checkRpcId(...)` and then `tbRpcService.deleteRpc` fetches it again. A `deleteRpc(TenantId, Rpc)` overload — or having `deleteRpc` return the deleted entity — would drop both extra reads, including on non-edge installations that get nothing in return for them.

application/src/main/java/org/thingsboard/server/service/edge/rpc/processor/device/DeviceEdgeProcessor.java
```diff
@@ -125,14 +129,51 @@ private void pushDeviceCreatedEventToRuleEngine(TenantId tenantId, Edge edge, De
...
+            // guard against out-of-order status uplinks regressing a terminal RPC (e.g. stale DELIVERED after SUCCESSFUL)
+            log.debug("[{}][{}] Ignoring stale RPC status [{}] - RPC is already terminal [{}]",
+                    tenantId, rpcId, rpcStatus, existingRpc.getStatus());
+            return Futures.immediateFuture(null);
+        }
+        JsonNode response = null;
+        if (deviceRpcCallMsg.hasResponseMsg() && !StringUtils.isEmpty(deviceRpcCallMsg.getResponseMsg().getResponse())) {
+            response = JacksonUtil.toJsonNode(deviceRpcCallMsg.getResponseMsg().getResponse());
+        }
+        edgeCtx.getTbRpcService().save(tenantId, rpcId, rpcStatus, response);
```
MazurenkoNick, line 173: None of the new sync paths look covered: the status-update uplink, the terminal-status guard, the DELETED uplink, and the new Cloud→Edge delete fan-out in `TbRpcService`. `DeviceEdgeTest.testRpcCall` (`application/src/test/java/org/thingsboard/server/edge/DeviceEdgeTest.java:632`) is already a working template for the downlink direction, and that class asserts on `tbClusterService.onEdgeHighPriorityMsg(...)` elsewhere, so both directions look reachable without much scaffolding — send an uplink `DeviceRpcCallMsg` with `rpcStatus` set and assert via `GET /api/rpc/persistent/{id}` that the status changed (and 404s after a DELETED uplink), then `DELETE /api/rpc/persistent/{id}` and assert a `DeviceRpcCallMsg` with `rpcStatus=DELETED` reaches the imitator.

Could those be added here, or is the intent to cover all of it by BBT in the edge repo? The terminal-state guard in particular is pure cloud-side logic that a cheap unit test on the processor would pin down, and it's the branch most likely to be broken silently by a later refactor.

application/src/main/java/org/thingsboard/server/service/rpc/TbRpcService.java
```diff
@@ -67,4 +80,29 @@ public PageData<Rpc> findAllByDeviceIdAndStatus(TenantId tenantId, DeviceId devi
         return rpcService.findAllByDeviceIdAndStatus(tenantId, deviceId, rpcStatus, pageLink);
     }
 
+    public void deleteRpc(TenantId tenantId, RpcId rpcId) {
+        Rpc rpc = rpcService.findById(tenantId, rpcId);
+        rpcService.deleteRpc(tenantId, rpcId);
+        // RPC v2 (persistent) delete/abort propagation Cloud -> Edge for ANY status.
+        // Skipped when the delete originated from an edge (edge-sync context) so it is not echoed back.
+        if (rpc != null && edgeSynchronizationManager.getEdgeId().get() == null) {
```
MazurenkoNick, line 88: The echo suppression is ambient: whether `deleteRpc` notifies edges depends on a ThreadLocal that a caller three layers away has to remember to set — which is why `DeviceEdgeProcessor` needs a try/finally plus an explanatory comment just to call this method correctly.

The rest of the edge stack threads this explicitly instead: `sendNotificationMsgToEdge(..., EdgeId originatorEdgeId)` takes the originating edge as a parameter and `BaseEdgeProcessor.processEntityNotification` filters with `!edgeId.equals(originatorEdgeId)`. That difference is also behavioural, not just stylistic — reading the flag as a global on/off means a delete arriving from one edge suppresses the notification to *every* other edge the device is assigned to, where the `originatorEdgeId` pattern skips only the sender. A `deleteRpc(TenantId, RpcId, EdgeId originatorEdgeId)` overload (null = cloud-originated) would make the contract visible at the call site and let `DeviceEdgeProcessor` drop the ThreadLocal dance.

application/src/main/java/org/thingsboard/server/service/rpc/TbRpcService.java
```diff
@@ -67,4 +80,29 @@ public PageData<Rpc> findAllByDeviceIdAndStatus(TenantId tenantId, DeviceId devi
         return rpcService.findAllByDeviceIdAndStatus(tenantId, deviceId, rpcStatus, pageLink);
     }
 
+    public void deleteRpc(TenantId tenantId, RpcId rpcId) {
+        Rpc rpc = rpcService.findById(tenantId, rpcId);
+        rpcService.deleteRpc(tenantId, rpcId);
```
MazurenkoNick, line 85: Putting the edge propagation inside `TbRpcService.deleteRpc` makes correct edge sync depend on which delete API a future caller happens to reach for — `RpcService.deleteRpc`, `deleteEntity(...)` and the `tenantRpcRemover` in `BaseRpcService` all stay reachable and now silently skip the notification.

That's the kind of split that quietly rots. Either a short javadoc on this method stating it's the only edge-aware entry point, or moving the propagation to the `DeleteEntityEvent` listener path (`EdgeEventSourcingListener.handleEvent(DeleteEntityEvent)`) that every other entity type uses, so it can't be bypassed. If the event-sourcing route was already rejected — because RPC deletes aren't published as entity events, or because the high-priority queue is needed — a one-line note about why would save the next person the investigation.

application/src/main/java/org/thingsboard/server/service/rpc/TbRpcService.java
```diff
@@ -67,4 +80,29 @@ public PageData<Rpc> findAllByDeviceIdAndStatus(TenantId tenantId, DeviceId devi
...
+        // RPC v2 (persistent) delete/abort propagation Cloud -> Edge for ANY status.
+        // Skipped when the delete originated from an edge (edge-sync context) so it is not echoed back.
+        if (rpc != null && edgeSynchronizationManager.getEdgeId().get() == null) {
+            pushRpcDeleteToEdges(tenantId, rpc);
+        }
+    }
+
+    private void pushRpcDeleteToEdges(TenantId tenantId, Rpc rpc) {
+        List<EdgeId> relatedEdgeIds = edgeService.findAllRelatedEdgeIds(tenantId, rpc.getDeviceId());
```
MazurenkoNick, line 94: This hand-rolls a fan-out that already exists: `BaseEdgeProcessor.processNotificationToRelatedEdges(...)` walks related edges lazily via `PageDataIterableByTenantIdEntityId` with `RELATED_EDGES_CACHE_ITEMS` and skips the source edge, and `tbClusterService.sendNotificationMsgToEdge(tenantId, null, entityId, body, type, action, originatorEdgeId)` is the established entry point for "notify every edge related to this entity except the originator". Reusing either gets the paging, the `edgesEnabled` short-circuit and the originator exclusion for free.

I can see why you went direct — the high-priority path, matching `saveRpcRequestToEdgeQueue` — but that one pushes to a single known `edgeId`, so the fan-out really is new duplication. If the high-priority delivery is a hard requirement, would a shared helper (`pushHighPriorityToRelatedEdges(tenantId, entityId, type, action, body, originatorEdgeId)`) be better than a second fan-out implementation living in a non-edge service?

application/src/main/java/org/thingsboard/server/service/rpc/TbRpcService.java
```diff
@@ -67,4 +80,29 @@ public PageData<Rpc> findAllByDeviceIdAndStatus(TenantId tenantId, DeviceId devi
...
+        if (relatedEdgeIds == null || relatedEdgeIds.isEmpty()) {
+            return;
+        }
+        ObjectNode body = JacksonUtil.newObjectNode();
+        body.put("requestUUID", rpc.getId().getId().toString());
```
MazurenkoNick, line 99: This is now the fourth place that builds the untyped RPC_CALL edge-event body by raw string key (alongside `DeviceActorMessageProcessor.saveRpcRequestToEdgeQueue`, `TbSendRPCReplyNode`, and the reader in `EdgeMsgConstructorUtils.constructDeviceRpcMsg`), and the only thing keeping producers and consumer in agreement is that everyone spells `"requestUUID"` and `"rpcStatus"` identically. Pulling the key set into shared constants — or a small DTO serialized with Jackson so writer and reader share one definition — would make the next field addition a compile-time change rather than a grep.

application/src/main/java/org/thingsboard/server/service/edge/EdgeMsgConstructorUtils.java
```diff
@@ -259,20 +259,23 @@ public static DeviceRpcCallMsg constructDeviceRpcCallMsg(UUID deviceId, JsonNode
...
         return builder.build();
     }
 
     private static DeviceRpcCallMsg.Builder constructDeviceRpcMsg(UUID deviceId, JsonNode body) {
         DeviceRpcCallMsg.Builder builder = DeviceRpcCallMsg.newBuilder()
                 .setDeviceIdMSB(deviceId.getMostSignificantBits())
-                .setDeviceIdLSB(deviceId.getLeastSignificantBits())
-                .setRequestId(body.get("requestId").asInt());
+                .setDeviceIdLSB(deviceId.getLeastSignificantBits());
+        if (body.get("requestId") != null) {
```
MazurenkoNick, line 276: `requestId` is `int32 requestId = 5` in the proto — not `optional` — so skipping the builder call doesn't omit it on the wire; the edge still reads `0`, and `0` is a legitimate value from the device actor's request sequence. The receiver can't tell "no requestId" from "requestId 0".

It probably doesn't bite today since the DELETED marker is matched by `requestUUID`, but if the field is genuinely optional for this message shape, marking it `optional` in `edge.proto` (like `persisted`, `retries`, `serviceId`, and the new `rpcStatus`) would make `hasRequestId()` available and remove the ambiguity.

application/src/main/java/org/thingsboard/server/service/edge/EdgeMsgConstructorUtils.java
```diff
@@ -259,20 +259,23 @@ public static DeviceRpcCallMsg constructDeviceRpcCallMsg(UUID deviceId, JsonNode
...
             }
             builder.setResponseMsg(responseBuilder.build());
-        } else {
+        } else if (body.has("method")) {
             RpcRequestMsg.Builder requestBuilder = RpcRequestMsg.newBuilder();
             requestBuilder.setMethod(body.get("method").asText());
             requestBuilder.setParams(body.get("params").asText());
             builder.setRequestMsg(requestBuilder.build());
         }
+        // else: status-only message (rpcStatus set in constructDeviceRpcMsg) carries neither request nor response body
```
MazurenkoNick, line 268: This comment documents a branch that doesn't exist — it sits after the if/else-if chain, so a reader has to scroll back up to work out which condition it belongs to, and it'll drift the moment another shape is added. An actual `else` with a `log.trace`, or moving the note into the method javadoc alongside the other two shapes, would keep the explanation attached to the code it describes.

common/data/src/main/java/org/thingsboard/server/common/data/rpc/RpcStatus.java
```diff
@@ -16,10 +16,10 @@ public enum RpcStatus {
     DELETED(false);
 
     @Getter
-    private final boolean pushDeleteNotificationToCore;
+    private final boolean intermediate;
```
MazurenkoNick, line 19: The rename reads much better — the flag now describes the state rather than one consumer's reaction to it. Two follow-ups worth considering:

It has quietly become load-bearing for two unrelated decisions (whether to remove the RPC actor in `RpcV2Controller`, and whether an incoming edge status may overwrite the stored one), so a short javadoc stating what "intermediate" means would help whoever adds the next `RpcStatus` value get the boolean right.

And since the new call site reads `!existingRpc.getStatus().isIntermediate()` with a comment explaining that it means "terminal", a derived `isTerminal()` would let that guard read positively and drop the double negative.

MazurenkoNick (commented):
## Review summary

Reviewed 8 changed files in [Sync persistent RPC (v2) status updates & delete events Edge->Cloud](https://github.com/thingsboard/thingsboard/pull/15913). Left 17 comment(s) inline.

The overall shape — reusing the `DEVICE`/`RPC_CALL` channel and the existing edge-sync ThreadLocal — fits the codebase, and the `pushDeleteNotificationToCore` → `intermediate` rename is a genuine improvement (no stale usages left anywhere in the repo). The main things worth a look before merge:

- **Cloud→Edge deletes are dropped when the edge is offline.** `pushRpcDeleteToEdges` goes through `onEdgeHighPriorityMsg`, and `EdgeGrpcService.onEdgeHighPriorityEvent` silently no-ops when there is no connected session — which leaves behind exactly the stale RPC copy this PR is trying to eliminate.
- **`hasRpcStatus()` is checked before `hasResponseMsg()`**, so an uplink carrying both would no longer reach `processDeviceRpcResponseFromEdge` and the cloud-side pending RPC would never be resolved.
- **Two unguarded parse calls on edge-supplied input** (`RpcStatus.valueOf` and `JacksonUtil.toJsonNode`) throw `IllegalArgumentException`, which `EdgeGrpcSession.processUplinkMsg` turns into a failure of the whole uplink batch.
- **No test coverage** for any of the three new sync paths.

---

common/edge-api/src/main/proto/edge.proto
```diff
@@ -379,6 +379,7 @@ message DeviceRpcCallMsg {
   optional string additionalInfo = 12;
   optional string serviceId = 13;
   optional string sessionId = 14;
+  optional string rpcStatus = 15;
```
MazurenkoNick, line 382: Carrying a closed enum as a free-form string pushes validation to runtime — the receiver has to `RpcStatus.valueOf(...)` whatever arrives and compare terminal cases with `RpcStatus.DELETED.name().equals(...)`, so typos and unknown values only surface as exceptions. `edge.proto` already models this kind of thing as a real enum (`UpdateMsgType`), which would give you `UNRECOGNIZED` handling and the same `handleUnsupportedMsgType` treatment the other processors use.

There is a fair counter-argument — `RpcResponseMsg.error` already ships `RpcError` as a string, and a string tolerates a cloud-side enum addition better than a proto enum does against an older edge build — so if the string is deliberate that's fine; a comment on the field naming `RpcStatus` as the source of valid values would make that intent explicit.

application/src/main/java/org/thingsboard/server/service/edge/EdgeContextComponent.java
```diff
@@ -121,6 +122,9 @@ public EdgeContextComponent(List<EdgeProcessor> processors) {
     @Autowired
     private DeviceService deviceService;
 
+    @Autowired
+    private TbRpcService tbRpcService;
```
MazurenkoNick, line 126: The `@Autowired` fields here are in strict alphabetical order by type (…`DeviceProfileService`, `DeviceService`, `DomainService`, `EdgeEventService`…), and `TbRpcService` lands in the middle of the D's. Easy to fix while it's still a one-line change.

Same nit on the new imports: this file (`service.rpc.TbRpcService` inserted among the `service.edge.rpc.*` block), `RpcV2Controller` (the `Autowired` import after `ResponseEntity`, `TbRpcService` before `config.annotations`) and `DeviceEdgeProcessor` (`com.fasterxml.jackson.databind.JsonNode` after the Spring imports, `RpcId` after `TbMsgType`). In `RpcV2Controller` the new injected field is also declared above the `private static final String` API-doc constants, where the other controllers keep constants first.

PR #15917, opened 2026-07-09
Fixed timeseries entity aggregation calculated field storing numeric results as strings

ShvaykaD:
## Problem
A **Timeseries Entity Aggregation** calculated field wrote every metric result as a JSON string instead of a number, even for numeric aggregations. Stored as a string, it lands in `ts_kv.str_v`, and server-side SQL aggregation ignores it — so `AVG`/`SUM` return **no data** in widgets, while `COUNT`/`MIN`/`MAX` still return.

## Root cause
`toResult(...)` used `metricsNode.put(name, JacksonUtil.toString(value))`, which stores a `TextNode` (e.g. `"400"`) — always a string, never a number.

## Reproduce
Create a CF like `consumption = SUM(energy)`, post telemetry, then view it on a timeseries widget with a **history** timewindow: `AVG`/`SUM` show nothing, `COUNT`/`MIN`/`MAX` do. The CF debug "Result" shows the value quoted (`"400"`).

## Fix
Emit a type-preserving node: `metricsNode.set(name, JacksonUtil.valueToTree(value))`. Numeric results now go to `dbl_v`/`long_v`; genuine string `MIN`/`MAX` results stay strings. Adds unit + updated integration tests asserting the stored result is a numeric node.

## Notes
Results already written to `str_v` will not self-heal; they need a CF reprocess/backfill to become numeric. (CF reprocessing availiable only in PE)

PE PR: thingsboard/thingsboard-pe#4923

## General checklist

## Back-End feature checklist

- [x] If new dependency was added: the dependency tree is checked for conflicts. _(no new dependency)_
- [x] If new service was added: the service is marked with corresponding @TbCoreComponent, @TbRuleEngineComponent, @TbTransportComponent, etc. _(no new service)_
- [x] If new REST API was added: the RestClient.java was updated, issue for [Python REST client](https://github.com/thingsboard/thingsboard-python-rest-client) is created. _(no new REST API)_
- [x] If new yml property was added: make sure a description is added (above or near the property). _(no new yml property)_

### DIFF
--- a/application/src/test/java/org/thingsboard/server/service/cf/ctx/state/aggregation/single/EntityAggregationCalculatedFieldStateTest.java
+++ b/application/src/test/java/org/thingsboard/server/service/cf/ctx/state/aggregation/single/EntityAggregationCalculatedFieldStateTest.java
@@ -1,0 +1,179 @@
+/**
+ * Copyright © 2016-2026 The Thingsboard Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.server.service.cf.ctx.state.aggregation.single;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.thingsboard.server.actors.ActorSystemContext;
+import org.thingsboard.server.common.data.TenantProfile;
+import org.thingsboard.server.common.data.cf.CalculatedField;
+import org.thingsboard.server.common.data.cf.CalculatedFieldType;
+import org.thingsboard.server.common.data.cf.configuration.Argument;
+import org.thingsboard.server.common.data.cf.configuration.ArgumentType;
+import org.thingsboard.server.common.data.cf.configuration.ReferencedEntityKey;
+import org.thingsboard.server.common.data.cf.configuration.TimeSeriesOutput;
+import org.thingsboard.server.common.data.cf.configuration.aggregation.AggFunction;
+import org.thingsboard.server.common.data.cf.configuration.aggregation.AggKeyInput;
+import org.thingsboard.server.common.data.cf.configuration.aggregation.AggMetric;
+import org.thingsboard.server.common.data.cf.configuration.aggregation.single.EntityAggregationCalculatedFieldConfiguration;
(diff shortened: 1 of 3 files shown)


application/src/test/java/org/thingsboard/server/service/cf/ctx/state/aggregation/single/EntityAggregationCalculatedFieldStateTest.java
```diff
@@ -0,0 +1,179 @@
...
+        state.init(false);
+    }
+
+    @Test
+    void testType() {
+        assertThat(state.getType()).isEqualTo(CalculatedFieldType.ENTITY_AGGREGATION);
+    }
+
+    // A numeric aggregation result (SUM/AVG/COUNT/..., numeric MIN/MAX) must be serialized as a numeric
+    // JSON node; a genuine string result (lexical MIN/MAX over string telemetry, spec §7) must stay a string
```
ShvaykaD: The `spec §7` reference here (and on line 121) points to a document that isn't in the repo — the only such markdown is an untracked file in the author's working tree, so a future reader can't resolve `§7`. Consider inlining the actual rule (a lexical `MIN`/`MAX` result over string telemetry stays a string node) and dropping the section pointer.

application/src/test/java/org/thingsboard/server/service/cf/ctx/state/aggregation/single/EntityAggregationCalculatedFieldStateTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private static Stream<Arguments> toResultSerializationCases() {
+        return Stream.of(
+                // SUM: Number result, precision 0 -> whole-number (long) node
+                arguments("consumption", new DoubleDataEntry("consumption", 400.0), 0, true, "400"),
+                // AVG: Number result, precision 2 -> half-up rounded double node
+                arguments("avgConsumption", new DoubleDataEntry("avgConsumption", 133.335), 2, true, "133.34"),
+                // MIN/MAX over string telemetry: lexical String result -> preserved as string node (spec §7)
+                arguments("maxCode", new StringDataEntry("maxCode", "9"), 0, false, "9")
```
ShvaykaD: Minor: `maxCode` isn't defined in `getConfiguration()` (unlike `consumption`/`avgConsumption`), which could suggest `toResult` resolves metric names against the config. It doesn't — it only iterates the passed-in `results` map, so the configured metrics are just ctx construction boilerplate, not inputs to the code under test. A one-line note (or reusing a configured metric name) would spare readers the double-take.

ShvaykaD (commented):
## Review summary

Reviewed 3 changed files in [Fix Timeseries Entity Aggregation calculated field storing numeric results as strings](https://github.com/thingsboard/thingsboard/pull/15917). Left 2 comment(s) inline.

The production fix is correct and consistent with the sibling `RelatedEntitiesAggregationCalculatedFieldState` (both now use `JacksonUtil.valueToTree`); no correctness issues found, and both branches of the fix (numeric and string `MIN`/`MAX`) are covered by the new unit test. The two inline comments are minor test-readability nits.

---

ShvaykaD (commented):
## Re-review summary

Re-reviewed [Fix Timeseries Entity Aggregation calculated field storing numeric results as strings](https://github.com/thingsboard/thingsboard/pull/15917) — verified 2 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/15917#pullrequestreview-4663013312).

| Status | Count |
|--------|-------|
| ✅ Resolved | 2 |
| 💬 Acknowledged | 0 |

All previous findings have been addressed. The fix commit only reworded test comments, added the `maxCode` metric to the test configuration, and switched the string-result sample to a zero-padded `"0009"` — no new issues.

### Finding details

- ✅ **EntityAggregationCalculatedFieldStateTest.java** — dangling `spec §7` reference in the test comments — *Fixed: reworded to inline the rule (a lexical MIN/MAX result over string telemetry stays a string node).*
- ✅ **EntityAggregationCalculatedFieldStateTest.java** — `maxCode` metric used in the parameterized cases but absent from `getConfiguration()` — *Fixed: `maxCode` is now defined as a MAX metric in the test CF configuration, and the sample uses a zero-padded `"0009"` that must stay a string.*

---

Issue #12887, opened 2025-03-12
SCADA Symbols event listener


supotsaea2504:
Only click event support, need mousemove and another mouse event of tag for SCADA Symbol action.

ssakharova:
Hello @supotsaea2504,

I passed all the provided information to our developers. I created an internal prod ticket for them to decide on its necessity and possible implementation in future versions of the ThingsBoard. Please follow our [Release Notes](https://thingsboard.io/docs/reference/releases/) for updates on new additions to the platform in future releases.

However, if you’d like to see this implemented sooner, feel free to contribute, as this is not currently a priority task on our roadmap.

PR #15919, opened 2026-07-10
Added double click and right click tag action triggers for SCADA symbols

mtsymbarov-del:
## Pull Request description

Before:
Знімок екрана 2026-07-10 о 13 33 05

After:
Знімок екрана 2026-07-10 о 13 30 39

## General checklist

## Back-End feature checklist

### DIFF
--- a/ui-ngx/src/app/modules/home/components/widget/lib/scada/scada-symbol.models.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/scada/scada-symbol.models.ts
@@ -120,11 +120,23 @@ export type ScadaSymbolActionTrigger = 'click' | 'dblclick' | 'contextmenu';
 
 export const scadaSymbolActionTriggers: ScadaSymbolActionTrigger[] = ['click', 'dblclick', 'contextmenu'];
 
+// Max interval between clicks of a double click: a click action is deferred by this amount
+// when the same tag also has a dblclick action, so it can be canceled by the second click.
+const dblClickDelay = 300;
+
 export const scadaSymbolActionTriggerTranslations = new Map<ScadaSymbolActionTrigger, string>(
   [
     ['click', 'scada.tag.on-click-action'],
     ['dblclick', 'scada.tag.on-double-click-action'],
     ['contextmenu', 'scada.tag.on-right-click-action']
+  ]
+);
+
+export const scadaSymbolActionTriggerHelpIds = new Map<ScadaSymbolActionTrigger, string>(
+  [
+    ['click', 'scada/tag_click_action_fn'],
+    ['dblclick', 'scada/tag_dblclick_action_fn'],
+    ['contextmenu', 'scada/tag_contextmenu_action_fn']
   ]
 );
 
@@ -679,16 +691,29 @@ export class ScadaSymbolObject {
     for (const tag of this.metadata.tags) {
       if (tag.actions) {
         const elements = this.svgShape.find(`[tb\\:tag="${tag.tag}"]`);
+        const deferClick = !!tag.actions.click && !!tag.actions.dblclick;
         for (const trigger of Object.keys(tag.actions)) {
           const action = tag.actions[trigger];
           elements.forEach(element => {
             element.attr('cursor', 'pointer');
-            element.on(trigger, (event) => {
-              if (trigger === 'contextmenu') {
-                event.preventDefault();
-              }
-              action.action(this.context, element, event);
-            });
+            if (trigger === 'click' && deferClick) {
+              let clickTimeout: ReturnType<typeof setTimeout> = null;
+              element.on('click', (event: MouseEvent) => {
+                clearTimeout(clickTimeout);
+                if (event.detail === 1) {
+                  clickTimeout = setTimeout(() => {
+                    action.action(this.context, element, event);
+                  }, dblClickDelay);
+                }
+              });
+            } else {
+              element.on(trigger, (event) => {
+                if (trigger === 'contextmenu') {
+                  event.preventDefault();
+                }
+                action.action(this.context, element, event);
+              });
+            }
           });
         }
       }
--- a/ui-ngx/src/app/modules/home/components/widget/lib/scada/scada-symbol.models.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/scada/scada-symbol.models.ts
@@ -116,7 +116,17 @@ export type ScadaSymbolStateRenderFunction = (ctx: ScadaSymbolContext, svg: Svg)
 export type ScadaSymbolTagStateRenderFunction = (ctx: ScadaSymbolContext, element: Element) => void;
 
 // noinspection JSUnusedGlobalSymbols
-export type ScadaSymbolActionTrigger = 'click';
+export type ScadaSymbolActionTrigger = 'click' | 'dblclick' | 'contextmenu';
+
+export const scadaSymbolActionTriggers: ScadaSymbolActionTrigger[] = ['click', 'dblclick', 'contextmenu'];
+
+export const scadaSymbolActionTriggerTranslations = new Map<ScadaSymbolActionTrigger, string>(
+  [
+    ['click', 'scada.tag.on-click-action'],
+    ['dblclick', 'scada.tag.on-double-click-action'],
+    ['contextmenu', 'scada.tag.on-right-click-action']
+  ]
+);
 
 export type ScadaSymbolActionFunction = (ctx: ScadaSymbolContext, element: Element, event: Event) => void;
 export interface ScadaSymbolAction {
@@ -674,6 +684,9 @@ export class ScadaSymbolObject {
           elements.forEach(element => {
             element.attr('cursor', 'pointer');
             element.on(trigger, (event) => {
+              if (trigger === 'contextmenu') {
+                event.preventDefault();
+              }
               action.action(this.context, element, event);
             });
           });
--- a/ui-ngx/src/app/modules/home/pages/scada-symbol/metadata-components/scada-symbol-metadata-tag.component.html
+++ b/ui-ngx/src/app/modules/home/pages/scada-symbol/metadata-components/scada-symbol-metadata-tag.component.html
@@ -37,16 +37,19 @@
       </button>
     }
   </div>
-  <div class="tb-on-click-action-field">
-    @if (!disabled || tagFormGroup.get('clickAction').value) {
-      <button
-        #editClickActionButton
-        mat-button
-        color="primary"
-        (click)="editClickAction()">
-        <mat-icon>{{ disabled ? 'visibility' : (tagFormGroup.get('clickAction').value ? 'edit' : 'add') }}</mat-icon>
-        <span>{{ ( disabled ? 'action.view' : (tagFormGroup.get('clickAction').value ? 'action.edit' : 'action.add') ) | translate }}</span>
-      </button>
-    }
-  </div>
+  @for (trigger of actionTriggers; track trigger) {
+    <div class="tb-action-function-field">
+      @if (!disabled || tagFormGroup.get(actionControlName(trigger)).value) {
+        <button
+          #editActionButton
+          [attr.data-trigger]="trigger"
+          mat-button
+          color="primary"
+          (click)="editAction(trigger)">
+          <mat-icon>{{ disabled ? 'visibility' : (tagFormGroup.get(actionControlName(trigger)).value ? 'edit' : 'add') }}</mat-icon>
+          <span>{{ ( disabled ? 'action.view' : (tagFormGroup.get(actionControlName(trigger)).value ? 'action.edit' : 'action.add') ) | translate }}</span>
+        </button>
+      }
+    </div>
+  }
 </div>
--- a/ui-ngx/src/app/modules/home/pages/scada-symbol/metadata-components/scada-symbol-metadata-tag.component.ts
+++ b/ui-ngx/src/app/modules/home/pages/scada-symbol/metadata-components/scada-symbol-metadata-tag.component.ts
@@ -19,8 +19,10 @@ import {
   forwardRef,
   Input,
   OnInit,
+  QueryList,
   Renderer2,
   ViewChild,
+  ViewChildren,
   ViewContainerRef,
   ViewEncapsulation
 } from '@angular/core';
@@ -34,7 +36,11 @@ import {
   UntypedFormGroup,
   Validator
 } from '@angular/forms';
-import { ScadaSymbolTag } from '@home/components/widget/lib/scada/scada-symbol.models';
+import {
+  ScadaSymbolActionTrigger,
+  scadaSymbolActionTriggers,
+  ScadaSymbolTag
+} from '@home/components/widget/lib/scada/scada-symbol.models';
 import { TbEditorCompleter } from '@shared/models/ace/completion.models';
 import { MatButton } from '@angular/material/button';
 import { TbPopoverService } from '@shared/components/popover.service';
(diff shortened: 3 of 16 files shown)


ui-ngx/src/app/modules/home/components/widget/lib/scada/scada-symbol.models.ts
```diff
@@ -116,7 +116,29 @@ export type ScadaSymbolStateRenderFunction = (ctx: ScadaSymbolContext, svg: Svg)
...
 
 // noinspection JSUnusedGlobalSymbols
-export type ScadaSymbolActionTrigger = 'click';
+export type ScadaSymbolActionTrigger = 'click' | 'dblclick' | 'contextmenu';
+
+export const scadaSymbolActionTriggers: ScadaSymbolActionTrigger[] = ['click', 'dblclick', 'contextmenu'];
+
+// Max interval between clicks of a double click: a click action is deferred by this amount
+// when the same tag also has a dblclick action, so it can be canceled by the second click.
+const dblClickDelay = 300;
```
vvlladd28, line 125: 300 ms is shorter than the double-click threshold browsers actually use, so this deferral can let both actions run.

Chrome takes the interval from the OS: 500 ms by default on Windows (`GetDoubleClickTime()`, user-configurable 200–900 ms) and ~500 ms on macOS. With a tag that has both `click` and `dblclick` configured and 380 ms between the two clicks:

- t=0 — click #1, `detail === 1`, timer scheduled for t=300
- t=300 — timer fires, **click action executes**
- t=380 — click #2, `detail === 2`, `clearTimeout` is a no-op because the timer already ran
- t=380 — `dblclick` fires, **double click action executes**

Both run, which is what the deferral is meant to prevent, and `tag_dblclick_action_fn.md` states outright that "only the double click action is executed". A deliberate, unhurried double click lands squarely in this window.

Since the platform threshold isn't exposed to JS, the practical fix is to make the delay at least as long as the common platform default (~500 ms) so the second click always arrives first. That costs click latency for tags that configure both, which seems like the right trade — the delay only applies when the author opted into both triggers.

ui-ngx/src/app/modules/home/components/widget/lib/scada/scada-symbol.models.ts
```diff
@@ -669,13 +691,29 @@ export class ScadaSymbolObject {
...
             element.attr('cursor', 'pointer');
-            element.on(trigger, (event) => {
-              action.action(this.context, element, event);
-            });
+            if (trigger === 'click' && deferClick) {
+              let clickTimeout: ReturnType<typeof setTimeout> = null;
+              element.on('click', (event: MouseEvent) => {
+                clearTimeout(clickTimeout);
+                if (event.detail === 1) {
+                  clickTimeout = setTimeout(() => {
```
vvlladd28, line 704: This timer is never cancelled on teardown. `ScadaSymbolObject.destroy()` completes `destroy$`, stops the element timelines, destroys the value actions and removes the SVG shape, but `clickTimeout` lives in a closure scoped to this `forEach` and is unreachable from there.

So if the symbol is destroyed inside the delay window — a dashboard state change from another widget, the widget being removed in edit mode — the pending callback still fires `action.action(this.context, element, event)` against a torn-down object. Note the `takeUntil(this.destroy$)` inside `callAction` does not protect you here: `destroy$` has already completed without emitting, and `takeUntil` ignores a notifier that completes, so `valueSetter.setValue(...)` runs on a setter that `destroy()` already disposed.

Collecting the timers on the instance (e.g. a `clickTimeouts: ReturnType<typeof setTimeout>[]` field cleared in `destroy()`) would close this.

ui-ngx/src/app/modules/home/components/widget/lib/scada/scada-symbol.models.ts
```diff
@@ -116,7 +116,29 @@ export type ScadaSymbolStateRenderFunction = (ctx: ScadaSymbolContext, svg: Svg)
 export type ScadaSymbolTagStateRenderFunction = (ctx: ScadaSymbolContext, element: Element) => void;
 
 // noinspection JSUnusedGlobalSymbols
-export type ScadaSymbolActionTrigger = 'click';
+export type ScadaSymbolActionTrigger = 'click' | 'dblclick' | 'contextmenu';
+
+export const scadaSymbolActionTriggers: ScadaSymbolActionTrigger[] = ['click', 'dblclick', 'contextmenu'];
```
vvlladd28, line 121: There are now four hand-maintained places that have to agree about what a trigger is: the `ScadaSymbolActionTrigger` union, this array, `scadaSymbolActionTriggerTranslations` and `scadaSymbolActionTriggerHelpIds`. Nothing enforces that. Adding a fourth trigger and forgetting one of the Maps fails silently — `Map.get` returns `undefined`, and `scada-symbol-metadata-tag-function-panel.component.ts` then sets `panelTitle = undefined` (blank panel header) and `tagFunctionHelpId = undefined` (broken help button) with no error anywhere.

A single `Record<ScadaSymbolActionTrigger, {name: string; helpId: string}>` with the array derived via `Object.keys` makes the compiler require an entry per trigger, and collapses four edits into one.

Related, in the same area: `ScadaSymbolTag.actions` (line 153) is still `{[trigger: string]: ScadaSymbolAction}`, so the new union buys no safety at the one place it matters — `tag.actions[trigger]` and `actions[trigger] = ...` accept any string. That also leaves the runtime and the editor disagreeing: `ScadaSymbolObject` and `prepareMetadata` iterate `Object.keys(tag.actions)` and will wire up *any* key present in the metadata, while the editor (`setupTags`, `writeValue`, `updateModel`) only round-trips the three known triggers — so an action stored under an unrecognised key is executed by the widget but silently dropped the first time the symbol is saved. Narrowing to `{[trigger in ScadaSymbolActionTrigger]?: ScadaSymbolAction}` and iterating `scadaSymbolActionTriggers` at runtime too would make both sides agree.

ui-ngx/src/app/modules/home/pages/scada-symbol/metadata-components/scada-symbol-metadata-tags.component.html
```diff
@@ -20,15 +20,17 @@
     <div class="tb-form-table-header">
       <div class="tb-form-table-header-cell tb-tag-header" translate>scada.tag.tag</div>
       <div class="tb-form-table-header-cell tb-state-render-function-header" translate>scada.state-render-function</div>
-      <div class="tb-form-table-header-cell tb-on-click-action-header" translate>scada.tag.on-click-action</div>
+      @for (trigger of actionTriggers; track trigger) {
+        <div class="tb-form-table-header-cell tb-action-function-header">{{ actionTriggerTranslations.get(trigger) | translate }}</div>
```
vvlladd28, line 24: This is the part of the settings form I'd reconsider — a column per trigger makes the table pay, on every row, for options most tags will never use.

Concretely, what it costs:

- The tag column drops from `flex: 1 1 70%` to `40%` to make room. The tag chip is the row's only identifier, and it is now the element that truncates first.
- The three action headers are right-aligned inside 15% cells, and the new labels are much longer than the one they replace — "On double click action" and "On right click action" versus "On click action". At a realistic panel width those wrap to two or three lines.
- Every row renders three `Add` buttons regardless of intent. A tag with a single click action shows two buttons for triggers nobody asked for.
- It doesn't scale. Any future trigger (long press, hover) is another column and another cut to the tag column, which is already at 40%.

The same pattern lands in `scada-symbol-tooltip.components.ts` — the element tag settings popover now stacks four label/button pairs instead of two, roughly doubling the height of a popover anchored to a small SVG element.

An alternative that stays constant-width no matter how many triggers exist: keep one "Actions" column showing which triggers are configured (small chips or icons, so configured state is visible at a glance, which the current design also doesn't show without reading each button's icon), with a single button opening a menu to pick the trigger to edit. That also gives each entry a real label instead of a bare "Add".

ui-ngx/src/app/modules/home/pages/scada-symbol/metadata-components/scada-symbol-metadata-tag.component.html
```diff
@@ -37,16 +37,19 @@
...
-        <span>{{ ( disabled ? 'action.view' : (tagFormGroup.get('clickAction').value ? 'action.edit' : 'action.add') ) | translate }}</span>
-      </button>
-    }
-  </div>
+  @for (trigger of actionTriggers; track trigger) {
+    <div class="tb-action-function-field">
+      @if (!disabled || tagFormGroup.get(actionControlName(trigger)).value) {
+        <button
+          #editActionButton
+          [attr.data-trigger]="trigger"
```
vvlladd28, line 45: Each row now carries four buttons whose entire accessible name is "Add" (or "Edit"/"View") — state render plus three actions. The trigger they belong to is conveyed only by the header, and since the table is built from `div`s rather than a real `<table>`, there is no programmatic association between the header cell and the button. A screen reader user moving through the tags list hears "Add, Add, Add, Add" per tag with nothing to distinguish trigger or tag.

That trips WCAG 2.1 AA 2.4.6 / 4.1.2 and is the kind of thing AXE flags, and this PR takes it from two ambiguous buttons per row to four.

An `[attr.aria-label]` on the button combining the trigger label and the tag name would fix it, e.g. `[attr.aria-label]="(actionTriggerTranslations.get(trigger) | translate) + ': ' + tagFormGroup.get('tag').value"` (the translations map would need injecting here — it's already imported in the parent). Worth doing for the state render button in the same pass.

ui-ngx/src/app/modules/home/pages/scada-symbol/metadata-components/scada-symbol-metadata-tag.component.ts
```diff
@@ -135,15 +146,23 @@ export class ScadaSymbolMetadataTagComponent implements ControlValueAccessor, On
...
+
   editTagStateRenderFunction(): void {
     this.openTagFunction('renderFunction', this.editStateRenderFunctionButton);
   }
 
-  editClickAction(): void {
-    this.openTagFunction('clickAction', this.editClickActionButton);
+  editAction(trigger: ScadaSymbolActionTrigger): void {
+    const button = this.editActionButtons
+      .find(b => b._elementRef.nativeElement.getAttribute('data-trigger') === trigger);
```
vvlladd28, line 159: Routing through a DOM attribute to find the button feels like a step backwards from the `@ViewChild` it replaced — you write `data-trigger` into the DOM purely so you can read it back out, and reach through Material's private `_elementRef` to do it.

The failure mode is quiet: if the lookup ever misses, `if (button)` swallows it and the user clicks `Add` and nothing happens, with nothing logged. That path is reachable today through `ScadaSymbolMetadataTagsComponent.editTagAction()` → `tagComponent?.editAction(trigger)`, which is how the tooltip drives this component.

If the `@if` around the button is what pushed you here (it breaks index-based `QueryList.get(i)`), building a `Map<ScadaSymbolActionTrigger, MatButton>` from `editActionButtons.changes` would keep the association in the component rather than in the DOM.

vvlladd28 (commented):
## Review summary

Reviewed 16 changed files in [Added double click and right click tag action triggers for SCADA symbols](https://github.com/thingsboard/thingsboard/pull/15919). Left 6 comment(s) inline.

The rename from `clickAction` to the generic trigger-based API is clean and complete — I grepped the whole `ui-ngx` tree and found no leftover references to `tagHasClickAction`, `editTagClickAction`, `scadaSymbolClickActionHighlightRules` or `clickActionFunctionCompletions`. Two things are worth looking at first:

1. **`scada-symbol.models.ts:125`** — the 300 ms `dblClickDelay` is shorter than the platform double-click threshold (500 ms by default on both Windows and macOS), so a slightly slow double click runs *both* the click action and the double click action. That is exactly what the deferral exists to prevent, and `tag_dblclick_action_fn.md` promises it cannot happen.
2. **`scada-symbol-metadata-tags.component.html:24`** — the column-per-trigger table is the part of the settings form that I think needs rethinking: the tag column drops from 70% to 40%, three long headers get right-aligned into 15% cells, and every row now shows three `Add` buttons for actions most tags will never configure. Details and an alternative are in the inline comment.

Smaller process note: the PR description is only before/after screenshots — the "human-readable scope of changes" and "references specific issue" checklist items are unticked, and a reader landing on this PR later has no text explaining the 300 ms deferral or the context-menu suppression, both of which are behavior changes to the existing click action.

## Additional findings

- **ui-ngx/src/assets/help/en_US/scada/tag_click_action_fn.md** — the click action's behavior now changes when the same tag also has a double click action (it is delayed by `dblClickDelay` and can be cancelled), but that is documented only in `tag_dblclick_action_fn.md`. Someone configuring a click action and wondering why it feels laggy will open the click help and find nothing. Worth one sentence here mirroring the note you added to the double click doc.

---

vvlladd28:
## Proposal: ship `contextmenu` now, defer `dblclick`

Following up on my review — specifically the two findings on `scada-symbol.models.ts` (the 300 ms `dblClickDelay` being shorter than the platform double-click threshold, and the pending timer surviving `destroy()`). Rather than tuning the delay, I'd suggest reconsidering whether `dblclick` belongs in this PR at all.

**The two new triggers are not comparable in cost.**

| | `contextmenu` | `dblclick` |
|---|---|---|
| Interaction with the existing `click` action | none — different mouse button | conflicts, needs deferral |
| Implementation | one `preventDefault()` | timer, `deferClick`, `event.detail`, uncancelled timeout |
| Correctness findings in the review | none | both of them |
| Open design questions | none | mutual exclusivity, delay threshold, touch behaviour |

Dropping `dblclick` removes both correctness findings, the 300-vs-500 ms question and the mutual-exclusivity question in a single move. `contextmenu` is untouched by any of it.

**Neither trigger is specified.** The originating request (#12887) asked for `mousemove` and "another mouse event"; the internal ticket lists right-click and double-click as examples ("e.g."). The trigger set was a choice, not a requirement — which also means #12887 isn't actually closed by this PR, since hover isn't implemented. Worth saying so explicitly in the description either way.

**The stated goal is already met by right-click.** The justification recorded on the ticket is "quick result without excessive navigation". Right-click delivers that and is far more discoverable: people try the right mouse button looking for a context menu, whereas nobody can guess that an element responds to a double click.

**Touch is a real gap.** Double-tap is bound to zoom in mobile browsers. And if `click` and `dblclick` become mutually exclusive (below), a tag configured with only `dblclick` has no click action at all — so a symbol authored on desktop silently does nothing on a tablet.

**The honest counter-argument**, which I don't want to leave out: double-click to open a faceplate is a long-standing convention in traditional SCADA/DCS HMIs, and operators migrating from those systems do expect it. But note that in those systems a single click typically only selects, or does nothing — the convention is "double click opens details", not "click runs one command and double click runs another". That supports `dblclick` *instead of* a click action, not alongside it.

**So if `dblclick` stays**, the minimum I'd ask for is making `click` and `dblclick` mutually exclusive per tag, enforced in the editor and defensively at runtime (metadata is hand-editable in the SVG). That removes the deferral entirely, and with it both correctness findings. It's free to introduce now — no released symbol has both — and it becomes a breaking change if we ship permissive and tighten later. `contextmenu` is unaffected and should stay freely combinable with either.

**The cut is clean if you go that way:** drop `'dblclick'` from the union, the triggers array and the two maps; delete `dblClickDelay` and the `deferClick` branch (the `else` becomes the only path); remove `tag_dblclick_action_fn.md` and the `on-double-click-action` key.

Everything else in the PR stays, and stays worth having — the `clickAction` → generic-trigger refactor is exactly what makes adding any future trigger cheap, `dblclick` included whenever it comes back. As a side effect the tags table drops from five columns to four, which takes some pressure off the tag column (40% → ~55%), though the structural point in my inline comment still stands.

To be clear, this is a product call rather than a technical one — both options work. One of them just needs two decisions that nobody has made yet.

PR #15925, opened 2026-07-13
Support cross-family LTS upgrades (4.3.x to 4.4) via a family-agnostic migration selector

ViacheslavKlimov:
## Problem

We are introducing the 4.4 LTS family while continuing to support 4.3. Cross-family upgrades (any `4.3.x -> 4.4`) are always offline. The existing LTS-migration selector filtered beans to the target version's family, which on a 4.4 target excluded the real in-range `4.3.1.x` migration beans, so their schema/data changes would not run. A stop-gap ad-hoc `v4313Migration.apply()` call existed and was buggy: it ran only `4.3.1.3` and skipped `applyAfterCommit()`.

## Change

- Loosen `LtsMigrationService.select()` from `isInRangeForTargetFamily` (which had a `sameFamily(to)` clause) to a family-agnostic `isInRange` = `(from, to]`. A `4.3.x -> 4.4` offline upgrade now selects the real in-range `4.3.1.x` beans **and** the new `4.4.0.0` baseline bean, each exactly once (the half-open range skips anything the source already passed).
- Remove the ad-hoc `v4313Migration.apply()` hack (its field and import too) from `DefaultDataUpdateService`. `runDataMigrations()` now covers those beans' `apply()`.
- Introduce `V4_4_0_0Migration` + `lts/4.4.0.0/schema_update.sql` as the home for 4.4 baseline flat DDL (moved `rule_chain.notes` here). Make `basic/schema_update.sql` intentionally empty — the offline path is now fully bean-driven.

## Safety

- No-downtime path unchanged: within-family ranges select the same set as before, and `sameFamily` remains the no-downtime gate in `SystemPatchApplier`.
- Load-bearing invariant: a reproduction-duplicate bean must sit strictly below the minimum supported source (`4.3.0.0`). The only such pair (`4.2.2.3 <-> 4.3.1.3`) is below the floor, so both can never be selected together. Documented in `isInRange` and guarded by a unit test.

Reworked `LtsMigrationServiceTest` (cross-family selection + invariant guard) and `LtsMigrationIntegrationTest` (offline `4.3.x -> 4.4` schema apply; the dir/bean-sync guard now includes `4.4.0.0`); both pass.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
+++ b/application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
@@ -114,26 +114,21 @@ public class LtsMigrationService {
     private List<VersionedMigration> select(String fromVersion, String toVersion) {
         LtsVersion from = LtsVersion.parse(fromVersion);
         LtsVersion to = LtsVersion.parse(toVersion);
+        // Select every migration in the (from, to] range, regardless of family. On a cross-family offline
+        // upgrade (e.g. 4.3.x -> 4.4) this is what makes the real in-range older-family beans run: it picks the
+        // 4.3.1.x schema/data changes the source has not yet passed AND the new target-family beans, each exactly
+        // once -- the half-open (from, to] range skips anything the source already applied. One logical migration
+        // is thus authored once (one bean + one lts/<version>/schema_update.sql) and reused by both the offline
+        // and no-downtime paths; nothing is reproduced into a newer family.
+        //
+        // Load-bearing invariant: no two beans may reproduce the same change within a single supported upgrade
+        // range. A reproduction-duplicate bean (one that re-does an older bean's work on a newer-family branch)
+        // must sit STRICTLY BELOW the minimum supported upgrade source, so it can never be selected together with
+        // the bean it duplicates. The only such pair today is 4.2.2.3 <-> 4.3.1.3, and the supported-source floor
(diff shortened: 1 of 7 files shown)


application/src/main/java/org/thingsboard/server/service/install/lts/LtsMigrationService.java
```diff
@@ -115,19 +115,24 @@ private List<VersionedMigration> select(String fromVersion, String toVersion) {
...
+    // once -- the half-open (from, to] range skips anything the source already applied. One logical migration
+    // is thus authored once (one bean + one lts/<version>/schema_update.sql) and reused by both the offline
+    // and no-downtime paths; nothing is reproduced into a newer family.
+    //
+    // Load-bearing invariant: no two beans may reproduce the same change within a single supported upgrade
+    // range. A reproduction-duplicate bean (one that re-does an older bean's work on a newer-family branch)
+    // must sit STRICTLY BELOW the minimum supported upgrade source, so it can never be selected together with
+    // the bean it duplicates. The only such pair today is 4.2.2.3 <-> 4.3.1.3, and the supported-source floor
+    // is 4.3.0.0 (SUPPORTED_VERSIONS_FOR_UPGRADE), well above 4.2.2.3. LtsMigrationServiceTest guards this.
+    static boolean isInRange(LtsVersion version, LtsVersion from, LtsVersion to) {
```
ViacheslavKlimov: Maybe let's move this method to LtsVersion class?

Issue #229, opened 2017-08-02
How to determine the equipment online or offline?


yuyihan666:
How to get the connection status of the device

ashvayka:
This will be available in v1.4. For now you can implement this as a separate microservice

mfalkvidd:
When implementing this, please also consider devices using Sigfox and similar protocols. Online or offline can be a bit tricky to determine for devices that don't use persistent connections.

ashvayka:
Hi! @yuyihan666  and @mfalkvidd . This was implemented long time ago. Sorry for late update of the ticket. See https://thingsboard.io/docs/user-guide/device-connectivity-status/

PR #15964, opened 2026-07-24
Fix Nashorn Global PropertyMap leak on JS script release

MazurenkoNick:
### Fix Nashorn Global PropertyMap leak on JS script release ([#229](https://github.com/thingsboard/thingsboard-edge/issues/229))
image
image
image

### Problem

Every JS rule-node script is deployed as a uniquely named function
(`invokeInternal_<uuid>`) evaluated onto **one Nashorn `Global` object shared by
all scripts**. Releasing a script ran `invokeInternal_<uuid> = undefined;`, which
frees the function *value* but permanently leaks its *property slot* on the
Global. In the field incident behind #229 this retained **2.57 GB — 88.7 % of
reachable heap** (9,018,082 `PropertyHashMap$Element` instances) after ~10 days
of script deploy/test/release cycles, and the resulting memory pressure
prevented the Edge gRPC uplink from ever reconnecting.

### Background: how Nashorn stores an object

Nashorn splits every JS object into two pieces (same idea as V8 hidden classes):

```text
Global object (one per engine, shared by ALL scripts)    PropertyMap — the "shape" (layout only, immutable)
┌──────────────────────────────┐                         ┌───────────────────────────────────────────────┐
│ map ─────────────────────────┼────────────────────────▶│ property name            descriptor           │
│                              │                         │ ───────────────────────  ────────────────────│
│ storage (JS values, by slot):│                         │ "invokeInternal_a3f…" → { slot: 0,            │
│   slot 0 : ScriptFunction ƒ  │                         │                           configurable: false }│
│   slot 1 : undefined         │                         │ "invokeInternal_9c1…" → { slot: 1, … }        │
│   slot 2 : ScriptFunction ƒ  │                         │ "invokeInternal_e77…" → { slot: 2, … }        │
└──────────────────────────────┘                         └───────────────────────────────────────────────┘

reading a global:  name ──(map lookup)──▶ descriptor { slot: n } ──▶ storage[n]  = the actual JS value
```

The map entry's "value" is **not** the JS value — it is a `Property` descriptor
holding a slot index plus flags. The JS value lives in the object's storage
array at that slot. The map's hash-table nodes are the `PropertyHashMap$Element`
objects visible in heap dumps.

### The leak

`fn = undefined` writes `undefined` into `storage[slot]` — the `ScriptFunction`
and its compiled classes get collected (verified: generated classes were never
pinned, 90k+ unloaded in tests). But the **map entry and the slot stay**, so the
Global's current map accumulates one dead entry per script *ever* deployed,
dragging the whole predecessor chain with it. Layout grows forever; only values
were freed.

### Why plain `delete` can't fix it

How a property is *created* decides its `configurable` flag forever:

| created by                                        | descriptor              | `delete`            |
|---------------------------------------------------|-------------------------|---------------------|
| `function f() {…}` executed at global scope       | `configurable: false`   | refused (`false`)   |
| plain assignment (`this['f'] = …`)                | `configurable: true`    | works               |

Per ECMAScript, a top-level function *declaration* creates its global property
non-configurable. `delete invokeInternal_x` returns `false` and removes nothing
(verified on nashorn-core 15.4, engine and sandbox). That is why the release
path used `= undefined` in the first place.

### The fix

The declaration must never *execute at global scope*. `doEval` wraps the
generated script in an IIFE, so the declaration executes inside the IIFE's
private scope (discarded as soon as the IIFE returns), and the Global receives
only the finished function **value** through an ordinary property write —
top-level `this` is the Global object itself:

```js
this['invokeInternal_x'] = (function() {
    function invokeInternal_x(msgStr, metadataStr, msgType) { /* … */ }  // binding lives & dies in IIFE scope
    return invokeInternal_x;
})();
```

Assignment-created properties are `configurable: true`, so `doRelease` can now
actually remove the slot:

```js
delete this['invokeInternal_x'];
```

The Global switches to a map *without* that entry; predecessor maps are held
only via weak/soft history references and are reclaimed by GC. Slot count now
tracks *currently live* scripts instead of scripts-ever-deployed.

image
image

### Performance

The wrapper costs one extra function allocation, one call and one property
write, **once per script deployment** — noise compared to the parse/compile/
class-load that eval already performs. The per-message invocation path is
unchanged: `invokeFunction(name)` does the same name → slot → value lookup and
calls the byte-for-byte same function object; the IIFE no longer exists at that
point. Measured: 4,000 multi-threaded invocations in 79 ms, 1,000
single-threaded in 43 ms (existing perf tests, unchanged budget, pass).

### Evidence

| Measurement (10,000 unique script eval/release cycles)          | before (`= undefined`) | after (IIFE + `delete`) |
|-----------------------------------------------------------------|------------------------|--------------------------|
| Strong heap growth per 2,000 cycles (soft refs force-cleared)   | ~1 MB                  | 0 MB                     |
| `PropertyHashMap$Element` after 10k cycles                      | grows unbounded (field: 9.0 M) | 2,020 (flat)     |
| `PropertyMap`/`Element`/`AccessorProperty` at 10k / 20k / 30k    | grows                  | 58 / 485 / 221 — bit-identical (no creep) |
| Generated `Script$NNN` classes                                  | unloaded (never the leak) | unloaded (identical)  |

### Backward compatibility

Verified by new tests in `NashornJsInvokeServiceTest` (12/12 pass):

- invocation results, `this`/scope-chain behavior — unchanged (same function object in the same slot);
- implicit global state sharing between scripts (`x = 5` without `var`) — still lands on the shared Global, the IIFE does not capture it;
- compilation error code and error body — unchanged (`TbScriptException` still carries the original script);
- invoke-after-release — identical `No compiled script found` behavior;
- released scripts are now provably **removed** from the Global (new regression test).

Only observable difference: compile-error *positions* shift by a constant +2
lines. These positions are reported in the coordinates of the sandbox-rewritten
text (the sandbox reformats and instruments every script before compiling), so
they never pointed at the user's source line to begin with — e.g. an error on
user line 3 was reported as line 9 before this change and as line 11 after. The
constant shift is pinned by a test so a real regression in error reporting
still fails loudly.

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
+++ b/application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
@@ -17,13 +17,17 @@ package org.thingsboard.server.service.script;
 
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import delight.nashornsandbox.NashornSandbox;
+import delight.nashornsandbox.NashornSandboxes;
 import lombok.extern.slf4j.Slf4j;
 import org.junit.Assert;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.util.ReflectionTestUtils;
 import org.thingsboard.common.util.TbStopWatch;
+import org.thingsboard.script.api.RuleNodeScriptFactory;
 import org.thingsboard.script.api.ScriptType;
 import org.thingsboard.script.api.TbScriptException;
 import org.thingsboard.script.api.js.NashornJsInvokeService;
@@ -31,6 +35,7 @@ import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.controller.AbstractControllerTest;
 import org.thingsboard.server.dao.service.DaoSqlTest;
 
+import javax.script.ScriptEngine;
 import javax.script.ScriptException;
 import java.util.ArrayList;
 import java.util.List;
@@ -220,6 +225,86 @@ class NashornJsInvokeServiceTest extends AbstractControllerTest {
         assertDoesNotThrow(() -> {
             evalScript(script);
         });
+    }
+
+    @Test
+    void whenScriptIsReleased_thenGlobalPropertyIsRemoved() throws Exception {
+        UUID scriptId = evalScript("return msg;");
+        String functionName = "invokeInternal_" + scriptId.toString().replace('-', '_');
+        assertThat(evalInEngine("this.hasOwnProperty('" + functionName + "')")).hasToString("true");
+
+        invokeService.release(scriptId).get();
+
+        // the property must be deleted, not just set to undefined: an abandoned binding keeps
+        assertThat(evalInEngine("this.hasOwnProperty('" + functionName + "')")).hasToString("false");
+    }
+
+    @Test
+    void givenIifeWrappedEval_whenInvoking_thenResultIsUnchanged() throws Exception {
+        UUID scriptId = evalScript("return {doubled: msg.temperature * 2};");
+        assertThat(invokeScript(scriptId, "{\"temperature\":21}")).contains("\"doubled\":42");
+    }
+
+    @Test
+    void givenIifeWrappedEval_thenImplicitGlobalStateIsStillSharedBetweenScripts() throws Exception {
+        UUID writer = evalScript("compatSharedState = 777; return msg;");
+        UUID reader = evalScript("return {seen: (typeof compatSharedState === 'undefined') ? null : compatSharedState};");
+        invokeScript(writer, "{}");
+        assertThat(invokeScript(reader, "{}")).contains("\"seen\":777");
+    }
+
+    @Test
+    void givenIifeWrappedEval_thenCompilationErrorUnchanged() {
+        String badScript = "var a = 1;\nvar b = 2;\nreturn msg.temperature?.value;";
+
+        String legacyScript = RuleNodeScriptFactory.generateRuleNodeScript("invokeInternal_legacy", badScript,
+                "msg", "metadata", "msgType");
+        NashornSandbox legacySandbox = NashornSandboxes.create();
+        legacySandbox.allowNoBraces(false);
+        legacySandbox.allowLoadFunctions(true);
+        legacySandbox.setMaxPreparedStatements(30);
+        int legacyErrorLine = -1;
+        try {
+            legacySandbox.eval(legacyScript);
+        } catch (ScriptException e) {
+            legacyErrorLine = e.getLineNumber();
+        }
+        assertThat(legacyErrorLine).isGreaterThan(0);
+        final int expectedLine = legacyErrorLine;
+
+        assertThatThrownBy(() -> evalScript(badScript))
+                .isInstanceOf(ExecutionException.class)
+                .cause()
+                .isInstanceOf(TbScriptException.class)
+                .asInstanceOf(type(TbScriptException.class))
+                .satisfies(ex -> {
+                    assertThat(ex.getErrorCode()).isEqualTo(TbScriptException.ErrorCode.COMPILATION);
+                    assertThat(ex.getBody()).contains(badScript);
+                    assertThat(ex.getCause()).isInstanceOf(ScriptException.class);
+                    // The IIFE wrapper shifts sandbox-reported line numbers by a constant +2
+                    // (the beautifier splits the wrapper prefix onto its own lines). Reported
+                    // lines never matched the user's source anyway (the sandbox instrumentation
+                    // already offsets them), so only the small constant shift is tolerated here.
+                    assertThat(((ScriptException) ex.getCause()).getLineNumber()).isBetween(expectedLine, expectedLine + 2);
+                });
+    }
+
+    @Test
+    void whenInvokingAfterRelease_thenFailsWithNoCompiledScriptFound() throws Exception {
+        UUID scriptId = evalScript("return msg;");
+        invokeScript(scriptId, "{}");
+        invokeService.release(scriptId).get();
+        assertThatThrownBy(() -> invokeScript(scriptId, "{}"))
+                .isInstanceOf(ExecutionException.class)
+                .hasMessageContaining("No compiled script found");
+    }
+
+    private Object evalInEngine(String expression) throws Exception {
+        Object sandbox = ReflectionTestUtils.getField(invokeService, "sandbox");
+        if (sandbox != null) {
+            return ((NashornSandbox) sandbox).eval(expression);
+        }
+        return ((ScriptEngine) ReflectionTestUtils.getField(invokeService, "engine")).eval(expression);
     }
 
     private void assertThatScriptIsBlocked(UUID scriptId) {
--- a/application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
+++ b/application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
@@ -231,12 +231,13 @@ class NashornJsInvokeServiceTest extends AbstractControllerTest {
     void whenScriptIsReleased_thenGlobalPropertyIsRemoved() throws Exception {
         UUID scriptId = evalScript("return msg;");
         String functionName = "invokeInternal_" + scriptId.toString().replace('-', '_');
-        assertThat(evalInEngine("this.hasOwnProperty('" + functionName + "')")).hasToString("true");
+        assertThat(evalInEngine("this.hasOwnProperty('" + functionName + "')")).isEqualTo(true);
 
         invokeService.release(scriptId).get();
 
         // the property must be deleted, not just set to undefined: an abandoned binding keeps
-        assertThat(evalInEngine("this.hasOwnProperty('" + functionName + "')")).hasToString("false");
+        // its slot in the Global's PropertyMap forever
+        assertThat(evalInEngine("this.hasOwnProperty('" + functionName + "')")).isEqualTo(false);
     }
 
     @Test
--- a/common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
+++ b/common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
@@ -140,14 +140,21 @@ public class NashornJsInvokeService extends AbstractJsInvokeService {
 
     @Override
     protected ListenableFuture<UUID> doEval(UUID scriptId, JsScriptInfo scriptInfo, String jsScript) {
+        // A top-level function declaration creates a non-configurable property on the Nashorn Global
+        // that can never be deleted, so every eval/release cycle would grow the Global's PropertyMap
+        // shape history forever. Declaring the function inside an IIFE and assigning it to a global
+        // property keeps the property configurable, allowing doRelease() to actually delete it.
+        // The prefix stays on the first line so that script line numbers in errors are unchanged.
+        String wrappedScript = "this['" + scriptInfo.getFunctionName() + "'] = (function() { " + jsScript
+                + "\nreturn " + scriptInfo.getFunctionName() + ";\n})();";
         return jsExecutor.submit(() -> {
             try {
                 evalLock.lock();
                 try {
                     if (useJsSandbox) {
-                        sandbox.eval(jsScript);
+                        sandbox.eval(wrappedScript);
                     } else {
-                        engine.eval(jsScript);
+                        engine.eval(wrappedScript);
                     }
                 } finally {
                     evalLock.unlock();
@@ -182,10 +189,11 @@ public class NashornJsInvokeService extends AbstractJsInvokeService {
     }
 
     protected void doRelease(UUID scriptId, JsScriptInfo scriptInfo) throws ScriptException {
+        String deleteScript = "delete this['" + scriptInfo.getFunctionName() + "'];";
         if (useJsSandbox) {
-            sandbox.eval(scriptInfo.getFunctionName() + " = undefined;");
+            sandbox.eval(deleteScript);
         } else {
-            engine.eval(scriptInfo.getFunctionName() + " = undefined;");
+            engine.eval(deleteScript);
         }
     }
 
--- a/common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
+++ b/common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
@@ -144,7 +144,8 @@ public class NashornJsInvokeService extends AbstractJsInvokeService {
         // that can never be deleted, so every eval/release cycle would grow the Global's PropertyMap
         // shape history forever. Declaring the function inside an IIFE and assigning it to a global
         // property keeps the property configurable, allowing doRelease() to actually delete it.
-        // The prefix stays on the first line so that script line numbers in errors are unchanged.
+        // The prefix stays on the first line to preserve line numbering on the raw engine path;
+        // the sandbox beautifier reformats the wrapper and shifts reported error lines anyway.
         String wrappedScript = "this['" + scriptInfo.getFunctionName() + "'] = (function() { " + jsScript
                 + "\nreturn " + scriptInfo.getFunctionName() + ";\n})();";
         return jsExecutor.submit(() -> {
@@ -190,10 +191,15 @@ public class NashornJsInvokeService extends AbstractJsInvokeService {
 
     protected void doRelease(UUID scriptId, JsScriptInfo scriptInfo) throws ScriptException {
         String deleteScript = "delete this['" + scriptInfo.getFunctionName() + "'];";
-        if (useJsSandbox) {
-            sandbox.eval(deleteScript);
-        } else {
-            engine.eval(deleteScript);
+        evalLock.lock();
+        try {
+            if (useJsSandbox) {
+                sandbox.eval(deleteScript);
+            } else {
+                engine.eval(deleteScript);
+            }
+        } finally {
+            evalLock.unlock();
         }
     }
 
--- a/common/edge-api/src/main/java/org/thingsboard/edge/rpc/EdgeGrpcClient.java
+++ b/common/edge-api/src/main/java/org/thingsboard/edge/rpc/EdgeGrpcClient.java
@@ -86,7 +86,7 @@ public class EdgeGrpcClient implements EdgeRpcClient {
 
     private StreamObserver<RequestMsg> inputStream;
 
-    private static final ReentrantLock uplinkMsgLock = new ReentrantLock();
+    private final ReentrantLock connectionLock = new ReentrantLock();
 
     @Override
     public void connect(String edgeKey,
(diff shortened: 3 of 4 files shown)


common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
```diff
@@ -182,10 +189,11 @@ protected ListenableFuture<Object> doInvokeFunction(UUID scriptId, JsScriptInfo
     }
 
     protected void doRelease(UUID scriptId, JsScriptInfo scriptInfo) throws ScriptException {
+        String deleteScript = "delete this['" + scriptInfo.getFunctionName() + "'];";
```
MazurenkoNick, line 193: `doEval` takes `evalLock` around its `eval`, but `doRelease` doesn't — and this change turns release into a `PropertyMap` mutation for the first time. Under the old `fn = undefined`, `ScriptObject.set` found the existing property and wrote its spill slot, leaving the Global's map untouched. `delete this['...']` goes through `ScriptObject.deleteOwnProperty`, which swaps the Global's `map` field via `compareAndSetMap` — and that isn't a real CAS: in nashorn-core 15.4 `ScriptObject` declares `private PropertyMap map;` (plain, non-volatile) and `compareAndSetMap` is an unsynchronized read-compare-assign. `addOwnProperty` on the eval path uses the same non-atomic loop, so a release racing an eval on another `jsExecutor` thread can lose either side: a lost delete silently reinstates exactly the leak this PR fixes, and a lost add drops the just-deployed function from the Global, so every later `invokeFunction` fails even though `scriptInfoMap` still reports the script as present. The window is narrow, but `release()` is driven from actor threads while evals run concurrently on the js executor, so it looks reachable. Worth taking `evalLock` here too?

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
```diff
@@ -182,10 +189,11 @@ protected ListenableFuture<Object> doInvokeFunction(UUID scriptId, JsScriptInfo
...
         if (useJsSandbox) {
-            sandbox.eval(scriptInfo.getFunctionName() + " = undefined;");
+            sandbox.eval(deleteScript);
```
MazurenkoNick: The result of the `delete` is dropped, and in non-strict mode a refused delete just evaluates to `false` rather than throwing — so if the property ever ends up non-configurable again (someone restructures the wrapper in `doEval`, or a future `ScriptType` factory emits a genuine top-level declaration) the leak returns with no signal at all. `eval` hands back the expression's value here, so something like `if (Boolean.FALSE.equals(sandbox.eval(deleteScript))) log.warn(...)` would make that regression visible instead of silent. Worth noting the failure mode is otherwise benign: `deleteOwnProperty` calls `erasePropertyValue` before it checks configurability, so a refused delete still degrades to the old `= undefined` behavior.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
```diff
@@ -140,14 +140,21 @@ public void stop() {
 
     @Override
     protected ListenableFuture<UUID> doEval(UUID scriptId, JsScriptInfo scriptInfo, String jsScript) {
+        // A top-level function declaration creates a non-configurable property on the Nashorn Global
+        // that can never be deleted, so every eval/release cycle would grow the Global's PropertyMap
+        // shape history forever. Declaring the function inside an IIFE and assigning it to a global
+        // property keeps the property configurable, allowing doRelease() to actually delete it.
+        // The prefix stays on the first line so that script line numbers in errors are unchanged.
```
MazurenkoNick: This last comment line is contradicted by the test added in the same PR, which documents (and asserts) a constant +2 shift in sandbox-reported line numbers. The claim only holds for the raw-`engine` path; on the default sandbox path the beautifier reformats the wrapper onto its own lines. Since this is exactly the kind of statement a future reader will trust when deciding whether they can restructure the wrapper, could it say something like "keeping the prefix on the first line preserves line numbers on the raw engine path; the sandbox beautifier still shifts them by a constant"? Otherwise the two comments in the same change tell opposite stories.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
```diff
@@ -140,14 +140,21 @@ public void stop() {
...
+        String wrappedScript = "this['" + scriptInfo.getFunctionName() + "'] = (function() { " + jsScript
```
MazurenkoNick, line 149: The `this['<fn>']` property-access form is now hand-built in two places (here and the `delete this['<fn>'];` in `doRelease`), and the assignment side has to stay exactly in sync with the delete side or the leak silently comes back. Worth extracting a small private pair in this class — e.g. `wrapAsGlobalAssignment(functionName, jsScript)` and `globalDeleteScript(functionName)`, or a single template constant used by both — so the quoting/property form exists once. I don't think it belongs in `RuleNodeScriptFactory` despite that being where script text is otherwise generated: that factory is shared with `RemoteJsInvokeService`, and the remote js-executor must not receive the IIFE, so keeping the wrapper Nashorn-local is the right call — it just doesn't need to be inline string concatenation in the middle of `doEval`.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
```diff
@@ -140,14 +140,21 @@ public void stop() {
...
+                + "\nreturn " + scriptInfo.getFunctionName() + ";\n})();";
```
MazurenkoNick, line 150: The wrapper silently assumes `jsScript` is exactly one top-level declaration of a function literally named `scriptInfo.getFunctionName()` — an assumption owned by `RuleNodeScriptFactory.generateRuleNodeScript`, two modules away, with nothing at either end pointing at the other. `AbstractJsInvokeService.generateJsScript` already dispatches on `ScriptType` and is clearly meant to grow more factories; the first one that emits a function expression, a differently-named function, or any trailing statement will break the `return <functionName>;` tail here in a way that's hard to trace back. A one-line note on `generateRuleNodeScript` (or on the abstract `doEval`) stating that implementations may rely on the output being a single declaration of `functionName` would make the contract visible to whoever adds the next script type.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
```diff
@@ -140,14 +140,21 @@ public void stop() {
...
         return jsExecutor.submit(() -> {
             try {
                 evalLock.lock();
                 try {
                     if (useJsSandbox) {
-                        sandbox.eval(jsScript);
+                        sandbox.eval(wrappedScript);
                     } else {
-                        engine.eval(jsScript);
+                        engine.eval(wrappedScript);
```
MazurenkoNick, line 158: Both changed methods have two parallel backends, and only one of them is covered. `thingsboard.yml` defaults `js.local.use_js_sandbox` to `true` and the test class doesn't override it, so the `engine.eval` branches in `doEval` and `doRelease` — the raw `ScriptEngine` path, which behaves differently precisely where this change matters (no beautifier, so no line shift) — are never exercised by the new tests. A nested class with `@TestPropertySource(properties = "js.local.use_js_sandbox=false")` re-running at least `whenScriptIsReleased_thenGlobalPropertyIsRemoved` and the invoke test would close that gap cheaply, and would also give the `evalInEngine` fallback a reason to exist.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
```diff
@@ -222,6 +227,86 @@ void givenComplexScript_testCompile() {
...
 
+    @Test
+    void whenScriptIsReleased_thenGlobalPropertyIsRemoved() throws Exception {
+        UUID scriptId = evalScript("return msg;");
+        String functionName = "invokeInternal_" + scriptId.toString().replace('-', '_');
+        assertThat(evalInEngine("this.hasOwnProperty('" + functionName + "')")).hasToString("true");
+
+        invokeService.release(scriptId).get();
+
+        // the property must be deleted, not just set to undefined: an abandoned binding keeps
```
MazurenkoNick: This comment is cut off mid-sentence ("an abandoned binding keeps" …keeps what?). It's the one place explaining *why* the assertion is about `hasOwnProperty` rather than the value, so it's worth finishing — something like "…keeps its PropertyMap slot on the Global forever".

MazurenkoNick (commented):
## Review summary

Reviewed 2 changed files in [Fix Nashorn Global PropertyMap leak on JS script release](https://github.com/thingsboard/thingsboard/pull/15964). Left 10 comment(s) inline.

The diagnosis and the IIFE-plus-`delete` approach both hold up — I confirmed against `nashorn-core` 15.4 and `delight-nashorn-sandbox` 0.4.5 that `sandbox.eval(js)` with no context/bindings evaluates against the engine's ENGINE_SCOPE, so top-level `this` really is the same `Global` that `invokeFunction` resolves names on. The main thing worth a second look is that `delete` is a `PropertyMap` mutation while the old `= undefined` was not, and `doRelease` runs outside `evalLock`.

## Additional findings

- **common/script/script-api/src/main/java/org/thingsboard/script/api/js/AbstractJsInvokeService.java:95** — `doRelease(scriptId, scriptInfoMap.remove(scriptId))` can pass `null` down: `AbstractScriptInvokeService.release` gates on `isScriptPresent(scriptId)` and only then removes, so two concurrent `release()` calls for the same id both pass the check and the loser gets `null`. Pre-existing, but the new `String deleteScript = ... scriptInfo.getFunctionName() ...` line is now where it NPEs. A null guard (or treating a `null` info as already-released) would make double-release a no-op instead of a failed future.

One further nit had no anchor land: the comment at `NashornJsInvokeServiceTest.java:238` is cut off mid-sentence ("an abandoned binding keeps" …keeps what?) — completed in 970d590.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
```diff
@@ -222,6 +227,86 @@ void givenComplexScript_testCompile() {
         });
     }
 
+    @Test
+    void whenScriptIsReleased_thenGlobalPropertyIsRemoved() throws Exception {
+        UUID scriptId = evalScript("return msg;");
+        String functionName = "invokeInternal_" + scriptId.toString().replace('-', '_');
+        assertThat(evalInEngine("this.hasOwnProperty('" + functionName + "')")).hasToString("true");
```
MazurenkoNick: `hasToString("true")` asserts on the string form of whatever the engine returns, which reads oddly and would also pass for the literal string `"true"`. Nashorn hands back a `java.lang.Boolean` here, so `isEqualTo(true)` says what's meant directly; alternatively have `evalInEngine` return `boolean` (or add a `hasOwnProperty(functionName)` helper) and assert `isTrue()`/`isFalse()`, which would read better in both places.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
```diff
@@ -222,6 +227,86 @@ void givenComplexScript_testCompile() {
         });
     }
 
+    @Test
+    void whenScriptIsReleased_thenGlobalPropertyIsRemoved() throws Exception {
+        UUID scriptId = evalScript("return msg;");
+        String functionName = "invokeInternal_" + scriptId.toString().replace('-', '_');
```
MazurenkoNick, line 233: This re-derives the naming rule from `AbstractJsInvokeService.constructFunctionName`, duplicating both the `invokeInternal_` literal and the dash-to-underscore transform. The prefix is a bare literal in production too, so the cheapest fix is to pull it into a constant there (e.g. `AbstractJsInvokeService.FUNCTION_NAME_PREFIX`) and reference it from the test — a rename then shows up as a compile error rather than a puzzling `hasOwnProperty` failure.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
```diff
@@ -222,6 +227,86 @@ void givenComplexScript_testCompile() {
...
+        assertThat(invokeScript(reader, "{}")).contains("\"seen\":777");
+    }
+
+    @Test
+    void givenIifeWrappedEval_thenCompilationErrorUnchanged() {
+        String badScript = "var a = 1;\nvar b = 2;\nreturn msg.temperature?.value;";
+
+        String legacyScript = RuleNodeScriptFactory.generateRuleNodeScript("invokeInternal_legacy", badScript,
+                "msg", "metadata", "msgType");
+        NashornSandbox legacySandbox = NashornSandboxes.create();
```
MazurenkoNick, line 263: This copies three lines of the production sandbox setup from `NashornJsInvokeService.init()`, magic `30` included, so it silently drifts the moment someone tunes those values — and it spins up a second Nashorn engine per run just to recompute a number the assertion then only checks within a 3-value window. `isBetween(expectedLine, expectedLine + 2)` on line 288 also doesn't pin the constant the comment describes: a 0 or +1 shift passes just as happily, so the "fails loudly on a real regression" property isn't really there. The offset is stable — `JsSanitizer` is constructed only from `maxPreparedStatements` and `allowNoBraces`, and `maxCPUTime`/`maxMemory` affect only whether the eval runs on the monitor executor, not the injected instrumentation — so either `isEqualTo(expectedLine + 2)`, or dropping the live baseline for a named constant with this comment attached, would be simpler and stricter at once.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
```diff
@@ -222,6 +227,86 @@ void givenComplexScript_testCompile() {
...
+        UUID scriptId = evalScript("return msg;");
+        invokeScript(scriptId, "{}");
+        invokeService.release(scriptId).get();
+        assertThatThrownBy(() -> invokeScript(scriptId, "{}"))
+                .isInstanceOf(ExecutionException.class)
+                .hasMessageContaining("No compiled script found");
+    }
+
+    private Object evalInEngine(String expression) throws Exception {
+        Object sandbox = ReflectionTestUtils.getField(invokeService, "sandbox");
```
MazurenkoNick, line 304: `ReflectionTestUtils.getField` for internals is well established in this repo, so the reflection itself is fine. What isn't established is the silent null fallback: it re-implements production's `useJsSandbox` branch as "is the field null", and under the current test properties the `engine` branch is unreachable, so a typo in either field name degrades into an obscure NPE instead of a clear failure. Reading the `useJsSandbox` flag (also reachable via `ReflectionTestUtils`) and asserting the corresponding field is non-null would make the intent explicit and fail with a useful message if these fields are ever renamed.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
MazurenkoNick: Fixed in 970d590 — `doRelease` now takes `evalLock` around both branches.

Confirmed the mechanism against nashorn-core 15.4: `ScriptObject.map` is a plain non-volatile field and `compareAndSetMap` is an unsynchronized read-compare-assign, so the delete really can lose against a concurrent `addOwnProperty`. This is also the exact hazard the lock was introduced for in the first place — 953054fadc8, *"eval is not thread safe. Added lock to avoid 'No such function invokeInternalXXX' exception"*. The old `fn = undefined` was safe without the lock because assigning to an already existing property only writes its storage slot and never swaps the map, which is precisely what changes here.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
MazurenkoNick: Not adding a runtime check, on reflection.

I prototyped appending `this.hasOwnProperty('<fn>')` to the delete script and warning when it came back `true`, then dropped it: a non-configurable property can only arise from restructuring the `doEval` wrapper, and the new `NashornJsInvokeServiceLifecycleTest` now asserts removal on both backends — so a regression fails the build deterministically instead of emitting a warning nobody greps. A developer-regression signal belongs in a test, not in production code.

One correction to the comment above: a refused delete does **not** degrade to erasing the value. `ScriptObject.deleteObject` returns `false` on `!isConfigurable()` *before* reaching `deleteOwnProperty`, so `erasePropertyValue` never runs and the `ScriptFunction` stays pinned as well — a regression would be worse than the pre-PR behaviour, not equivalent to it.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
MazurenkoNick: Fixed in 970d590. It now reads:

```java
// The prefix stays on the first line to preserve line numbering on the raw engine path;
// the sandbox beautifier reformats the wrapper and shifts reported error lines anyway.
```

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
MazurenkoNick: Deferred, deliberately.

Extracting the helper rewrites the exact string that the `isBetween(expectedLine, expectedLine + 2)` assertion in the test is calibrated against — the +2 comes from the beautifier reflowing the `this['<fn>'] = (function() { ` prefix and the `\nreturn <fn>;\n})();` tail, so perturbing a newline moves the shift. The new unit test would catch a parse failure or a wrong function name, but it asserts nothing about line numbers.

Best done together with tightening that assertion, as one change verified against the DB-backed test, rather than blind.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
MazurenkoNick: Agreed — a one-line note on `generateRuleNodeScript` stating that implementations may rely on its output being a single top-level declaration of `functionName`. Not in 970d590; will follow.

common/script/script-api/src/main/java/org/thingsboard/script/api/js/NashornJsInvokeService.java
MazurenkoNick: Fixed in 970d590 — new `common/script/script-api/src/test/java/org/thingsboard/script/api/js/NashornJsInvokeServiceLifecycleTest`, parameterized over `useJsSandbox` in `{true, false}`.

Went with a unit test in the module that owns the class rather than a second `@TestPropertySource` here, for two reasons: adding `js.local.use_js_sandbox=false` to this class would flip *all* of it off the sandbox and neuter `givenComplexScript_testCompile`, which is specifically a sandbox `ScriptCPUAbuseException` regression test; and a `@Nested` variant would boot a second Spring + DB context (there are currently zero `@Nested` classes anywhere in `application/src/test`). The unit test runs 4 cases in ~2s with no DB and confirms the raw-engine path both stores and deletes the global property.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
MazurenkoNick: Fixed in 970d590 — `isEqualTo(true)` / `isEqualTo(false)`. Nashorn does return a real `java.lang.Boolean` for `hasOwnProperty`, verified on both backends via the new unit test.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
MazurenkoNick: Still open. Straightforward, but it adds production surface (a constant on `AbstractJsInvokeService`) purely for a test's benefit, so I'd rather group it with the remaining test cleanups than do it piecemeal.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
MazurenkoNick: Two separate points here, and one of them I'm retracting.

Retracting: the omitted `setExecutor` / `setMaxCPUTime` / `setMaxMemory` do **not** make this an invalid line-number baseline. `JsSanitizer` is constructed only from `maxPreparedStatements` and `allowNoBraces`, and `maxCPUTime`/`maxMemory` only decide whether the eval runs on the monitor executor — the injected instrumentation, and therefore the offset, is identical either way.

Standing: the duplicated production setup will drift, and `isBetween(expectedLine, expectedLine + 2)` doesn't pin the constant the comment describes — a 0 or +1 shift passes just as happily. Tightening to `isEqualTo(expectedLine + 2)` is the part worth doing, but not blind: if the real shift isn't 2 the test flips red, so it needs a DB-backed run to confirm first.

application/src/test/java/org/thingsboard/server/service/script/NashornJsInvokeServiceTest.java
MazurenkoNick: Still open — with one thing worth adding: the fallback is now provably dead. This class only ever runs with the sandbox, and the raw-engine path has its own coverage in `NashornJsInvokeServiceLifecycleTest`, so the branch can simply be deleted rather than made explicit.

PR #15993, opened 2026-07-31
Feature/auto refresh value chart widget

DmytroKhylko:
## Pull Request description

**Add auto-refresh for "last update ago" date in value card widgets**

Move the refresh scheduling into `LastUpdateAgeDateFormatProcessor` so every widget using the "last update ago" date format gets accurate, self-pacing updates automatically.

Previously, the relative timestamp label ("Last update X ago") only updated when new telemetry arrived. This PR makes the label re-evaluate itself at exactly the right moment — every second when the age is under a minute, every minute under an hour, every hour under a day, and daily beyond that — matching the granularity of `DateAgoPipe`.

The timer runs outside Angular's zone (`NgZone.runOutsideAngular`) to avoid triggering app-wide change detection on each tick; widgets subscribe to `tick$` and call `detectChanges()` locally instead.

Affected widgets: `tb-value-card-widget`, `tb-aggregated-value-card-widget`, `tb-signal-strength-widget`.
  
Before: the "last update" label freezes between data pushes
 
image

After: the label updates at exactly the right cadence for its current age, with no backend configuration needed
image

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/controller/SystemInfoController.java
+++ b/application/src/main/java/org/thingsboard/server/controller/SystemInfoController.java
@@ -87,6 +87,9 @@ public class SystemInfoController extends BaseController {
     @Value("${iot-hub.base-url:https://iot-hub.thingsboard.io}")
     private String iotHubBaseUrl;
 
+    @Value("${server.ws.dynamic_page_link.refresh_interval:60}")
+    private long dynamicPageLinkRefreshIntervalSec;
+
     private static final Set<String> ACCEPTED_NULLS_ORDER_STRATEGIES = Set.of("default", "nulls_first", "nulls_last");
 
     @Autowired(required = false)
@@ -188,6 +191,7 @@ public class SystemInfoController extends BaseController {
             systemParams.setIntermediateAggregationIntervalInSecForCF(tenantProfileConfiguration.getIntermediateAggregationIntervalInSecForCF());
             systemParams.setTrendzSettings(trendzSettingsService.findTrendzSettings(currentUser.getTenantId()));
         }
+        systemParams.setDynamicPageLinkRefreshIntervalSec(dynamicPageLinkRefreshIntervalSec);
         systemParams.setIotHubBaseUrl(iotHubBaseUrl);
         systemParams.setMobileQrEnabled(Optional.ofNullable(qrCodeSettingService.findQrCodeSettings(TenantId.SYS_TENANT_ID))
                 .map(QrCodeSettings::getQrCodeConfig).map(QRCodeConfig::isShowOnHomePage)
--- a/ui-ngx/src/app/modules/home/components/widget/lib/cards/aggregated-value-card-widget.component.ts
+++ b/ui-ngx/src/app/modules/home/components/widget/lib/cards/aggregated-value-card-widget.component.ts
@@ -18,6 +18,7 @@ import {
   AfterViewInit,
   ChangeDetectorRef,
   Component,
+  DestroyRef,
   ElementRef,
   Input,
   OnDestroy,
@@ -26,6 +27,7 @@ import {
   TemplateRef,
   ViewChild
 } from '@angular/core';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import {
   aggregatedValueCardDefaultSettings,
   AggregatedValueCardKeyPosition,
@@ -35,7 +37,7 @@ import {
   getTsValueByLatestDataKey
 } from '@home/components/widget/lib/cards/aggregated-value-card.models';
 import { WidgetContext } from '@home/models/widget-component.models';
-import { Observable } from 'rxjs';
+import { interval, Observable } from 'rxjs';
 import {
   autoDateFormat,
   backgroundStyle,
@@ -49,6 +51,7 @@ import {
 import { DataKey } from '@shared/models/widget.models';
 import { formatNumberValue, formatValue, isDefined, isDefinedAndNotNull, isNumeric } from '@core/utils';
 import { map } from 'rxjs/operators';
+import { getCurrentAuthState } from '@core/auth/auth.selectors';
 import { ImagePipe } from '@shared/pipe/image.pipe';
 import { DomSanitizer } from '@angular/platform-browser';
 import { TbTimeSeriesChart } from '@home/components/widget/lib/chart/time-series-chart';
@@ -121,7 +124,8 @@ export class AggregatedValueCardWidgetComponent implements OnInit, AfterViewInit
   constructor(private imagePipe: ImagePipe,
               private sanitizer: DomSanitizer,
               private renderer: Renderer2,
-              private cd: ChangeDetectorRef) {
+              private cd: ChangeDetectorRef,
+              private destroyRef: DestroyRef) {
   }
 
   ngOnInit(): void {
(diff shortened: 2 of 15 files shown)


ui-ngx/src/app/modules/home/components/widget/lib/cards/value-card-widget.component.ts
```diff
@@ -139,6 +144,18 @@ export class ValueCardWidgetComponent implements OnInit, AfterViewInit, OnDestro
     this.dateStyle = textStyle(this.settings.dateFont);
     this.dateColor = ColorProcessor.fromSettings(this.settings.dateColor);
 
+    if (this.showDate && this.settings.dateFormat?.lastUpdateAgo) {
```
vvlladd28, line 147: The tick rate here is fixed, but the label it drives is not — `DateAgoPipe` switches unit at 29s / 1 min / 1 hr / 1 day, so a constant 60s interval is simultaneously too slow and too fast depending on the age of `lastUpdateTs`:

- Between 29s and 60s the pipe renders seconds (`30 sec ago`, `31 sec ago`, …), i.e. the text changes every second. With a 60s tick that whole range is never rendered — a value arriving just after a tick shows `just now` for a full minute and then jumps straight to `1 min ago`.
- Past an hour the label changes at most once an hour (once a day beyond 24h), yet the timer keeps calling `dateFormat.update()` + `detectChanges()` every 60s forever, for every card on the dashboard.

A self-rescheduling timer fixes both ends and removes the need for a configured period: compute the delay until the text can actually change from `Date.now() - this.lastUpdateTs` against the same thresholds the pipe uses (next second below 1 min, next minute below 1 hr, next hour below 1 day, next day beyond), then `timer(delay)` and reschedule on each emission. Stale cards then cost essentially nothing, and fresh ones update at the granularity they actually display.

Worth noting too: zone.js patches `setInterval`, so each tick also triggers an application-wide change detection pass on top of the local `detectChanges()`. `ctx.ngZone.runOutsideAngular(...)` (ngZone is already on `WidgetContext`) would keep that work inside the card.

ui-ngx/src/app/modules/home/components/widget/lib/cards/value-card-widget.component.ts
```diff
@@ -184,14 +201,21 @@ export class ValueCardWidgetComponent implements OnInit, AfterViewInit, OnDestro
...
-    this.dateFormat.update(ts);
+    this.updateLastUpdateTs(ts);
     this.iconColor.update(value);
     this.labelColor.update(value);
     this.valueColor.update(value);
     this.dateColor.update(value);
     this.cd.detectChanges();
   }
 
+  private updateLastUpdateTs(ts: number) {
```
vvlladd28, line 212: The `ts > this.lastUpdateTs` guard changes behavior beyond just remembering the timestamp. `onDataUpdated()` previously called `dateFormat.update(ts)` on every path, including the `else` branch where `ts` stays undefined — and `LastUpdateAgoDateFormatProcessor.update()` resets `formatted` to `&nbsp;` for a falsy ts. That reset is now skipped, so a card that had data and then loses it (`getSingleTsValue` returns null when the datasource's `data` array is empty — e.g. an entity alias resolving to an entity without that key) renders `N/A` for the value while keeping the old date. With the new interval that stale date then keeps counting up, so the card reads as if it were still receiving data.

The max-timestamp semantics look carried over from `aggregated-value-card-widget.component.ts`, where `updateLastUpdateTs` is called inside a loop over several latest keys and picking the newest is the whole point. Here there's a single value, so the comparison doesn't buy anything — assigning unconditionally keeps the previous behavior intact:

```ts
private updateLastUpdateTs(ts: number) {
  this.lastUpdateTs = ts;
  this.dateFormat.update(ts);
}
```

ui-ngx/src/app/modules/home/components/widget/lib/cards/aggregated-value-card-widget.component.ts
```diff
@@ -168,6 +172,18 @@ export class AggregatedValueCardWidgetComponent implements OnInit, AfterViewInit
     this.dateStyle = textStyle(this.settings.dateFont);
     this.dateColor = this.settings.dateColor;
 
+    if (this.showDate && this.settings.dateFormat?.lastUpdateAgo) {
```
vvlladd28, line 175: This is the same 11 lines as in `value-card-widget.component.ts`, so any change to the cadence has to be made twice and kept in sync.

`LastUpdateAgoDateFormatProcessor` looks like the natural owner: it already holds `formatted`, and it's the only `DateFormatProcessor` whose output depends on wall-clock time, so it's the one place that knows when its own text needs to change. Putting the scheduling there would also cover the widgets this PR misses — `signal-strength-widget.component.ts` builds its `dateFormat` from `lastUpdateAgoDateFormat()` too and has exactly the same frozen label once `showDate` is enabled, and any future widget using that format would get it for free instead of needing a third copy of this block.

application/src/main/java/org/thingsboard/server/controller/SystemInfoController.java
```diff
@@ -87,6 +87,9 @@ public class SystemInfoController extends BaseController {
     @Value("${iot-hub.base-url:https://iot-hub.thingsboard.io}")
     private String iotHubBaseUrl;
 
+    @Value("${server.ws.dynamic_page_link.refresh_interval:60}")
```
vvlladd28, line 90: `server.ws.dynamic_page_link.refresh_interval` is documented in `thingsboard.yml` as *"Refresh rate of the dynamic alarm end entity data queries"* — it's the cadence at which the backend re-runs dynamic entity data queries for websocket subscriptions. Driving a client-side relative-date label from it couples two settings that have nothing to do with each other: an admin lowering it so dynamic entity queries react faster would also make every value card on every dashboard repaint that often, and an admin raising it would silently make the "Last update … ago" labels stale. Neither is what they'd be asking for.

If the cadence genuinely should be server-driven, a dedicated property would at least say what it means. Given the label's own granularity though, deriving the next update locally from the timestamp age avoids needing a property — and this whole `SystemParams` → auth-state plumbing — at all; see the comment on `value-card-widget.component.ts`.

Separately: the default here (`:60`) doesn't match the existing `@Value` for the same key in `DefaultTbEntityDataSubscriptionService` (`:6`). `thingsboard.yml` always supplies a value so it doesn't bite in practice, but wherever the property is absent (tests, embedded contexts) the two components would now disagree about the same setting.

vvlladd28 (commented):
## Review summary

Reviewed 6 changed files in [Feature/auto refresh value chart widget](https://github.com/thingsboard/thingsboard/pull/15993). Left 4 comment(s) inline.

The main points: the fixed 60s tick doesn't match the granularity of the label it drives — `DateAgoPipe` changes unit at 29s/1min/1hr/1day, so the sub-minute range is never rendered while old cards keep re-rendering forever; `server.ws.dynamic_page_link.refresh_interval` is a backend websocket query cadence and reusing it couples two unrelated settings; and in `value-card-widget.component.ts` the new `ts > lastUpdateTs` guard drops the reset that used to blank the date when a card loses its value.

---

ui-ngx/src/app/modules/home/components/widget/lib/cards/value-card-widget.component.ts
DmytroKhylko: Moved the scheduling into LastUpdateAgeDateFormatProcessor itself — it now computes the exact delay until the label text will change (next second <1 min, next minute <1 hr, next hour <1 day, daily beyond) and self-reschedules via timer(delay) run outside Angular's zone. Widgets subscribe to tick$ and call detectChanges() locally.

ui-ngx/src/app/modules/home/components/widget/lib/cards/value-card-widget.component.ts
DmytroKhylko: Fixed — the guard is removed and assignment is now unconditional, restoring the &nbsp; reset when data is lost. Note: kept the max-ts guard in aggregated-value-card-widget since updateLastUpdateTs is called in a loop over multiple keys there and picking the newest is the point.

ui-ngx/src/app/modules/home/components/widget/lib/cards/aggregated-value-card-widget.component.ts
DmytroKhylko: Done — scheduling now lives entirely in LastUpdateAgeDateFormatProcessor. Both widget components just subscribe to tick$. Also extended this to signal-strength-widget which was using the same format but had the frozen-label bug too

application/src/main/java/org/thingsboard/server/controller/SystemInfoController.java
DmytroKhylko: Agreed, it was the wrong property to reuse. Since the self-rescheduling approach derives the next update time locally from Date.now() − lastUpdateTs, no server-side config is needed at all — reverted all the Java and auth-state changes.

PR #16013, opened 2026-08-06
Add per-calculated-field execution scope for edge deployments (EDGE/CLOUD)

MazurenkoNick:
## Pull Request description

## General checklist

## Back-End feature checklist


dao/src/main/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidator.java
```diff
@@ -144,6 +157,23 @@ private void validateEntityAggregationConfiguration(TenantId tenantId, Calculate
         }
     }
 
+    private void validateComputeOn(TenantId tenantId, CalculatedField calculatedField) {
```
MazurenkoNick, line 160: This only runs when the calculated field is saved. If the device or asset is later unassigned from its last edge, the field keeps `computeOn = EDGE` and then gets computed nowhere — the cloud skips it in `CalculatedFieldManagerMessageProcessor`, and the edge no longer has the entity. Is a follow-up planned on the unassign path, or is the intent to accept that and document it? Silently producing no output is a hard failure mode for a user to diagnose; even just a warning on the CF list would help.

dao/src/main/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidator.java
```diff
@@ -144,6 +157,23 @@ private void validateEntityAggregationConfiguration(TenantId tenantId, Calculate
...
+            return;
+        }
+        if (edgeSynchronizationManager.getEdgeId().get() != null) {
+            return;
+        }
+        EntityId entityId = calculatedField.getEntityId();
+        if (EntityType.DEVICE != entityId.getEntityType() && EntityType.ASSET != entityId.getEntityType()) {
+            return;
+        }
+        if (edgeService.findRelatedEdgeIdsByEntityId(tenantId, entityId, new PageLink(1)).getData().isEmpty()) {
```
MazurenkoNick: For DEVICE/ASSET this lands in `BaseRelatedEdgesService.findEdgeIdsByEntityId`, which short-circuits with `if (!pageLink.equals(FIRST_PAGE))` and only serves from cache when the page link is exactly `FIRST_PAGE` (`new PageLink(RELATED_EDGES_CACHE_ITEMS)`). `new PageLink(1)` therefore misses that cache and issues a DB query on every EDGE-scoped CF save. Since all we need is "is it assigned to anything", passing `BaseRelatedEdgesService.FIRST_PAGE` would reuse the cache that already has proper eviction wired up. Was the smaller page size a deliberate trade to avoid materialising up to 1000 ids?

MazurenkoNick (commented):
## Review summary

Reviewed 21 changed files in [Add per-calculated-field execution scope for edge deployments (EDGE/CLOUD)](https://github.com/thingsboard/thingsboard/pull/16013), together with the paired edge PR thingsboard/thingsboard-edge#234.

The core design holds up well: `compute_on` is nullable and `ComputeOn.orDefault` maps null to `CLOUD` here and to `EDGE` in the edge fork, so existing calculated fields keep computing on whichever side already ran them. Along the way I confirmed the native repository uses `SELECT *` (so the new column is actually picked up), that `onCfDeleted` only reads the entity id off the lifecycle message (so reusing it from `onCfUpdated` is safe), that `V4_3_1_4Migration` matches the documented SQL-only migration pattern and needs no extra registration, that `CalculatedFieldEdgeProcessor` really does set the `edgeId` ThreadLocal so the validator bypass works on uplinks, and that `EdgeService`/`EdgeSynchronizationManager` are unconditional `dao` beans, so the new validator dependency won't break non-TbCore services.

Left 10 comments inline — mostly the new validator (a cache-bypassing lookup and a rule that only runs at save time), a bit of dead or copy-pasted code, and test coverage that stays at the mock level.

common/data/src/main/java/org/thingsboard/server/common/data/cf/ComputeOn.java
```diff
@@ -0,0 +1,42 @@
...
+
+    ComputeOn(String displayName) {
+        this.displayName = displayName;
+    }
+
+    public static ComputeOn orDefault(ComputeOn computeOn) {
+        return computeOn == null ? DEFAULT : computeOn;
+    }
+
+    public static boolean isComputedHere(ComputeOn computeOn) {
```
MazurenkoNick, line 37: Worth a one-line comment noting that this body (and `DEFAULT` above) is intentionally different in the thingsboard-edge fork, where both flip to `EDGE`. Read from this repo alone it looks like a tautology inside a class that has no notion of "here", so a future cleanup that inlines it into the three call sites in `CalculatedFieldManagerMessageProcessor` or rewrites it as `computeOn != EDGE` would quietly break the edge build — and there's no test here that would catch it.

common/data/src/main/java/org/thingsboard/server/common/data/cf/ComputeOn.java
```diff
@@ -0,0 +1,42 @@
...
+
+public enum ComputeOn {
+
+    CLOUD("Cloud"),
+    EDGE("Edge");
+
+    public static final ComputeOn DEFAULT = CLOUD;
+
+    @Getter
+    private final String displayName;
```
MazurenkoNick: `displayName` doesn't appear to be read anywhere — the UI takes its labels from the `calculated-fields.compute-on.*` translation keys. Looks like it came along from `CalculatedFieldType`; dropping it and the constructor would leave just the two constants and the two helpers that are actually used.

ui-ngx/src/app/core/services/calculated-field-form.service.ts
```diff
@@ -41,6 +42,7 @@ export class CalculatedFieldFormService {
       name: ['', [Validators.required, Validators.maxLength(255)]],
       entityId: [null, Validators.required],
       type: [CalculatedFieldType.SIMPLE],
+      computeOn: [defaultComputeOn],
```
MazurenkoNick, line 45: There are now two mechanisms defaulting the same value: the backend treats `null` as CLOUD (`ComputeOn.orDefault`, the null branch in `DefaultNativeCalculatedFieldRepository`, the nullable column), while the form seeds the control with CLOUD and always submits it — including when `edgesSupportEnabled` is false and the select is never rendered, and in the alarm form which has no compute-on UI at all. So the nullable tri-state collapses to a non-null value on the first save from the UI anyway, and we still pay for it with null handling in three backend places plus `computeOn ?? defaultComputeOn` in two components. Might be worth committing to one: either make the column `NOT NULL DEFAULT 'CLOUD'` and drop `orDefault`/`DEFAULT`/the `??` fallbacks, or leave the control unset unless the user picks something. Either way it ends up as less code than the current split.

ui-ngx/src/app/modules/home/components/calculated-fields/calculated-field.component.html
```diff
@@ -69,27 +69,52 @@
...
+              </mat-option>
+            }
+          </mat-select>
+        </mat-form-field>
+        @if (edgesSupportEnabled) {
+          <mat-form-field class="flex-1" appearance="outline" subscriptSizing="dynamic">
+            <mat-label>{{ 'calculated-fields.compute-on.label' | translate }}</mat-label>
+            <mat-select formControlName="computeOn">
+              <mat-select-trigger>
+                {{ ComputeOnTranslations.has(entityForm.get('computeOn').value)
```
MazurenkoNick, line 99: This `mat-select-trigger` was copied from the type select above, where the `.has(...)` guard earns its place because CF types can include PE-only values CE has no translation for. `ComputeOn` is a closed two-value enum and both values are in `ComputeOnTranslations`, so the fallback branch is unreachable — and without a custom trigger `mat-select` already renders the selected option's `matListItemTitle`. Dropping the block removes ~5 lines here and the same again in the dialog template.

dao/src/test/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidatorTest.java
```diff
@@ -58,4 +93,103 @@ public void testUpdateNonExistingCalculatedField() {
                 .hasMessage("Can't update non existing calculated field!");
     }
 
+    @Test
+    public void testEdgeOnlyRejectedWhenDeviceIsNotAssignedToEdge() {
+        givenNotAssignedToEdge(DEVICE_ID);
+
+        assertThatThrownBy(() -> validator.validateDataImpl(TENANT_ID, edgeOnlyCf(DEVICE_ID)))
+                .isInstanceOf(DataValidationException.class)
+                .hasMessage("Calculated field computed on the edge requires device to be assigned to an edge!");
```
MazurenkoNick, line 101: The new rule is only covered by mocked unit tests, and its most interesting branch — skip validation while processing an edge uplink — is verified by stubbing a `ThreadLocal` rather than going through the real path. `CalculatedFieldEdgeTest` already has a live edge and an assigned device, so an integration test looks achievable: POST an EDGE-scoped CF for a device that is *not* assigned to any edge and assert the 400, then send an uplink for such a device and assert it's accepted. Worth noting `testSendComputeOnToCloud` doesn't exercise the bypass today — `saveDeviceOnCloudAndVerifyDeliveryToEdge` assigns the device to the edge, so validation would have passed regardless of the ThreadLocal.

dao/src/test/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidatorTest.java
```diff
@@ -58,4 +93,103 @@ public void testUpdateNonExistingCalculatedField() {
...
+        givenNotAssignedToEdge(DEVICE_ID);
+        ThreadLocal<EdgeId> edgeIdHolder = new ThreadLocal<>();
+        edgeIdHolder.set(EDGE_ID);
+        given(edgeSynchronizationManager.getEdgeId()).willReturn(edgeIdHolder);
+
+        assertThatCode(() -> validator.validateDataImpl(TENANT_ID, edgeOnlyCf(DEVICE_ID))).doesNotThrowAnyException();
+    }
+
+    @Test
+    public void testEdgeAllowedWithImmediateOutputStrategy() {
```
MazurenkoNick: `validateComputeOn` only reads `getComputeOn()` and `getEntityId()`, so the output strategy can't change the outcome — this test and the strategy mutation inside `testComputeOnCloudNeverRequiresAnEdge` assert something the new code doesn't influence. If they're guarding a rule that exists only in the edge fork, that intent isn't visible from here; otherwise they duplicate `testEdgeOnlyAllowedWhenDeviceIsAssignedToEdge` and could go.

dao/src/test/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidatorTest.java
```diff
@@ -58,4 +93,103 @@ public void testUpdateNonExistingCalculatedField() {
...
+        CalculatedField defaulted = cf(DEVICE_ID, null);
+        CalculatedField cloud = cf(DEVICE_ID, ComputeOn.CLOUD);
+        ((TimeSeriesOutput) cloud.getConfiguration().getOutput()).setStrategy(new TimeSeriesImmediateOutputStrategy());
+
+        assertThatCode(() -> validator.validateDataImpl(TENANT_ID, defaulted)).doesNotThrowAnyException();
+        assertThatCode(() -> validator.validateDataImpl(TENANT_ID, cloud)).doesNotThrowAnyException();
+    }
+
+    private void givenNotAssignedToEdge(EntityId entityId) {
+        given(edgeSynchronizationManager.getEdgeId()).willReturn(new ThreadLocal<>());
```
MazurenkoNick: `givenNotAssignedToEdge` and `givenAssignedToEdge` both re-stub `getEdgeId()` with the same empty `ThreadLocal` that `setUp` already installs. Only `testEdgeOnlyAllowedDuringEdgeSynchronization` needs a different value, so both helpers can drop this line and just stub `edgeService`.

ui-ngx/src/app/core/services/calculated-field-form.service.ts
```diff
@@ -51,6 +53,7 @@ export class CalculatedFieldFormService {
       name: ['', [Validators.required, Validators.maxLength(255)]],
       entityId: [null, Validators.required],
       type: [CalculatedFieldType.ALARM],
+      computeOn: [defaultComputeOn],
```
MazurenkoNick: `buildAlarmRuleForm` gets a `computeOn` control as well, but nothing under `components/alarm-rules/` renders it — `alarm-rule-dialog.component.ts` and `alarm-rules.component.ts` both build this form and neither template has a compute-on field. So every alarm rule saved through that form now silently persists a value the user can't see or change. Either render the selector there too, or leave this form as it was so alarm rules keep sending no value and inherit the server default.

dao/src/main/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidator.java
MazurenkoNick: Not changed. Revalidating on unassign needs a hook on that path; leaving it as a follow-up rather than widening this PR.

dao/src/main/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidator.java
MazurenkoNick: Fixed. Now passes `BaseRelatedEdgesService.FIRST_PAGE`, so the lookup is served from the cache instead of hitting the DB on every EDGE-scoped save.

common/data/src/main/java/org/thingsboard/server/common/data/cf/ComputeOn.java
MazurenkoNick: Fixed. Added a class javadoc stating that `DEFAULT` and `isComputedHere` are deliberately inverted in the edge fork, and that inlining the method would break it.

common/data/src/main/java/org/thingsboard/server/common/data/cf/ComputeOn.java
MazurenkoNick: Fixed. Removed `displayName`, its constructor and the lombok import; the enum is now two constants plus the two helpers.

ui-ngx/src/app/core/services/calculated-field-form.service.ts
MazurenkoNick: Partially. Dropped the eager default from the alarm form (see the other thread). Kept the nullable column and `orDefault`, since the edge fork's CLOUD/EDGE inversion relies on null meaning "this side's default".

ui-ngx/src/app/modules/home/components/calculated-fields/calculated-field.component.html
MazurenkoNick: Not changed. The options are multi-line (`matListItemMeta`), so without a custom trigger `mat-select` renders the hint inside the collapsed field. Every template here that uses `matListItemMeta` keeps a trigger. You're right that the `.has(...)` branch is unreachable, but dropping the whole block regresses the display.

dao/src/test/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidatorTest.java
MazurenkoNick: Fixed. Added two tests to `CalculatedFieldEdgeTest`: one asserting the 400 for an unassigned device, and one sending an uplink for that same device and asserting it is accepted. The second exercises the bypass through the real uplink path rather than a stubbed ThreadLocal.

dao/src/test/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidatorTest.java
MazurenkoNick: Fixed. Dropped the test and the same irrelevant strategy mutation inside `testComputeOnCloudNeverRequiresAnEdge`.

dao/src/test/java/org/thingsboard/server/dao/service/validator/CalculatedFieldDataValidatorTest.java
MazurenkoNick: Fixed. Both helpers now stub only `edgeService`; `setUp` still installs the empty ThreadLocal.

ui-ngx/src/app/core/services/calculated-field-form.service.ts
MazurenkoNick: Fixed. Removed `computeOn` from `buildAlarmRuleForm`. Confirmed no template under `components/alarm-rules/` references it, and `buildForm()` (which keeps it) backs only the two templates that render the control.

PR #16014, opened 2026-08-07
Feature: Map appearance option to enable Ctrl/Cmd + mouse wheel map zoom and move on desktop; two finger map zoom and move on mobile

DmytroKhylko:
## Pull Request description

Adds appearance option to maps and image map to enable Ctrl/Cmd + mouse wheel map zoom and move on desktop; two finger map zoom and move on mobile
The setting is set to default `false` so there is no dedicated upgrade is needed

Before:

https://github.com/user-attachments/assets/138617a6-f29e-45d7-b058-fe27d99f3ef6

https://github.com/user-attachments/assets/dd117876-b540-4ee8-a638-af0c46e6d4e1

After:
Map settings

https://github.com/user-attachments/assets/a1657cdb-ecda-40c6-9d99-0eff896868c1

https://github.com/user-attachments/assets/59eeaa3d-641f-496d-8ee2-58758e5fe7a8

## General checklist

## Back-End feature checklist


ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet/leaflet-tb.ts
```diff
@@ -28,6 +29,8 @@ import { WEBGL_ERROR_EVENT } from '@shared/models/widget/maps/map.models';
 
 L.MarkerCluster = L.MarkerCluster.mergeOptions({ pmIgnore: true });
 
+L.Map.addInitHook('addHandler', 'gestureHandling', GestureHandling);
```
vvlladd28, line 32: `leaflet-gesture-handling` already registers this handler itself — the package `main` (`dist/leaflet-gesture-handling.min.js`) ends with `L.Map.addInitHook("addHandler", "gestureHandling", GestureHandling)`, it resolves `L` from the global that Leaflet's UMD build always sets, and the package does not declare `sideEffects: false`, so the import alone is enough.

With this extra hook Leaflet runs `addHandler` twice per map (`_initHooks` has no dedup): two `GestureHandling` instances are constructed and both land in `map._handlers`, so on a cooperative map every `wheel`, `touchstart`, `touchmove`, `touchend` and `click` on the container is processed twice and two overlay timeouts run per scroll. `map.gestureHandling` also ends up pointing at the second instance only, so a later `map.gestureHandling.disable()` would leave the first one with its listeners still attached.

Dropping this line and switching to `import 'leaflet-gesture-handling';` keeps the registration and makes it happen once.

vvlladd28 (commented):
## Review summary

Reviewed 11 changed files in [Feature: Map appearance option to enable Ctrl/Cmd + mouse wheel map zoom and move on desktop; two finger map zoom and move on mobile](https://github.com/thingsboard/thingsboard/pull/16014). Left 2 comment(s) inline.

Backward compatibility checks out: `useGestureHandling` is added to `BaseMapSettings` with a `false` default, and every entry point merges defaults over stored settings before use — `TbMap` (`mergeDeepIgnoreArray` in the constructor), the advanced settings component (`WidgetSettingsComponent.settings` setter) and the basic config (`MapWidgetBasicConfigComponent.onConfigSet`). Dashboards saved before this change therefore resolve to `false` without an upgrade script, and the only bundled JSON carrying `zoomActions` is a demo dashboard that needs no update.

The two findings worth attention: `leaflet-tb.ts:32` registers an init hook the plugin already registers itself, so every cooperative map ends up with two `GestureHandling` instances; and the setting is gated on `MapZoomAction.scroll`, which also gates off the mobile half of the feature — one-finger page scroll vs two-finger map pan is about `dragging`, not about wheel zoom.

On the conciseness of the new `TbMap.gestureHandlingOptions` getter: extracting the translated texts into the base class is the right call — both subclasses need the identical object, and resolving them through `ctx.translate` rather than the plugin's own `navigator.language` table is the correct choice for a UI whose language is user-selected. Minor, non-blocking: the getter has no explicit return type, unlike the neighbouring members, and `duration: 2000` could use a named constant. If the gating discussed below is reworked, folding the `gestureHandling` flag into the same getter would also collapse the two repeated option lines in `geo-map.ts`/`image-map.ts` into a single spread.

One process note, not a code issue: the PR description leaves "Ensure new API is documented" unchecked, and this adds a user-visible widget setting — worth a follow-up in thingsboard-ui-help.

*Not reviewed: ui-ngx/yarn.lock — no reviewable logic.*

---

ui-ngx/src/app/modules/home/components/widget/lib/settings/common/map/map-settings.component.ts
```diff
@@ -302,6 +306,9 @@ export class MapSettingsComponent implements OnInit, ControlValueAccessor, Valid
 
   private updateModel() {
     this.modelValue = this.mapSettingsFormGroup.getRawValue();
+    if (!this.modelValue.zoomActions?.includes(MapZoomAction.scroll)) {
```
vvlladd28, line 309: Why is the setting gated on `MapZoomAction.scroll` (here and in the `@if` in the template)? The plugin governs two independent things, and only one of them is about wheel zoom:

- **Desktop** — Ctrl/⌘ + wheel to zoom. This genuinely depends on `scrollWheelZoom`: with wheel zoom already off, the map does not hijack page scroll and there is nothing to make cooperative. Gating makes sense here.
- **Mobile** — one finger scrolls the page, two fingers move the map. `_handleTouch` implements this purely through `dragging` (`_disableInteractions()` on a single touch), and it never touches `touchZoom`. Neither is related to `MapZoomAction.scroll` — and note `touchZoom` is not passed in `geo-map.ts`/`image-map.ts` at all, so pinch zoom is always on regardless of `zoomActions`.

So for a map configured with, say, `zoomActions: [doubleClick, controlButtons]`, one-finger dragging still steals the dashboard's scroll on a phone — exactly what this feature fixes — but the toggle is hidden and this line clears the flag, so the fix cannot be enabled. That is the case in the PR title's own "two finger map zoom **and move** on mobile".

If the gating is there because the plugin re-enables wheel zoom behind the config's back — `_enableInteractions()` calls `this._map.scrollWheelZoom.enable()` unconditionally and `_handleMouseOver` fires it on the first hover, and `_handleScroll` does the same on Ctrl/⌘+wheel — that conflict is desktop-only (`mouseover` does not fire on touch) and can be handled where TB already extends Leaflet, in `leaflet-tb.ts`:

```ts
const TbGestureHandling = GestureHandling.extend({
  _enableInteractions() {
    this._map.dragging.enable();
    if (this._map.options.scrollWheelZoom) {
      this._map.scrollWheelZoom.enable();
    }
    this._map.tap?.enable();
  },
  _handleScroll(e) {
    if (!this._map.options.scrollWheelZoom) {
      return;
    }
    GestureHandling.prototype._handleScroll.call(this, e);
  }
});
```

Then the toggle can be shown unconditionally, the mobile behaviour works for every map, and this normalization block is no longer needed.

If the coupling is a deliberate product decision rather than a workaround, it is worth saying so in the description — and then it would still be safer to derive the flag at map creation (`this.settings.useGestureHandling && this.settings.zoomActions.includes(MapZoomAction.scroll)`) instead of post-processing the model here, since settings also arrive from imported and hand-edited dashboard JSON that never passes through this form.

ui-ngx/src/app/modules/home/components/widget/lib/maps/leaflet/leaflet-tb.ts
DmytroKhylko: You're right. The package sets window.L = exports in its own source so the global L is always available when the UMD factory runs, meaning the self-registration fires on import. Dropping the explicit addInitHook removes the duplicate.

Rather than switching to a bare import 'leaflet-gesture-handling', I kept the named GestureHandling import to be able to patch its prototype — which is also how I addressed your second finding: _enableInteractions now guards scrollWheelZoom.enable() behind this._map.options.scrollWheelZoom, and _handleScroll bails early when scroll zoom is off. This avoids re-registering a subclass via another addInitHook (which would recreate the exact double-registration problem).

ui-ngx/src/app/modules/home/components/widget/lib/settings/common/map/map-settings.component.ts
DmytroKhylko: Agreed on all points. The gating was a workaround for _enableInteractions re-enabling scroll zoom on mouseover, not a meaningful product constraint. With the prototype patch in place the coupling is no longer needed: the toggle is now shown unconditionally, the updateModel() post-processing guard is removed, and gestureHandling is passed directly from settings.useGestureHandling at map creation — so hand-edited dashboard JSON that never touches the form is handled correctly too.
 
 Also folded the gestureHandling flag into the gestureHandlingMapOptions getter (renamed, return type added, duration extracted to a named constant), which collapsed the repeated two-line pattern in geo-map.ts and image-map.ts into a single spread as you suggested.

PR #16016, opened 2026-08-07
fix(device-state): prevent stale-clock nodes from marking active devices inactive

volodymyr-babak:
## Problem

In a clustered TB Core deployment a device can become **permanently stuck showing "inactive"** even though it is actively sending telemetry.

The root cause is a partition-ownership race:

1. A TB Core node briefly loses its ZooKeeper session (network blip, long GC, etc.). The remaining nodes recalculate partition ownership and take over its partitions, while the affected node keeps operating on its **stale** view of the cluster.
2. During that window two nodes consider themselves the owner of the same device. The device's activity messages are routed to the *new* owner, so the *stale* node stops receiving them and its in-memory `lastActivityTime` freezes.
3. The stale node's periodic inactivity check then fires and persists `active=false` (device state is a shared value in `ts_kv_latest` / server attributes).
4. Ownership settles. The real owner still has `active=true` in memory; `updateActivityState` only writes `active` on a `false → true` transition, so it never re-writes the flag. The device is shown as inactive indefinitely, with no self-healing short of a restart.

## Fix

Before a node persists `active=false` in `updateInactivityStateIfExpired`, it now re-reads the **authoritative persisted `lastActivityTime`** (the source of truth). If that value shows the device is still active — because another owner is keeping it alive — the node reconciles its local state and **skips** the inactivity report.

This makes the persisted value the arbiter, so a node with a stale clock can no longer mark a still-active device inactive. The read fails safe (returns `0` → previous behavior) so a transient read error cannot suppress legitimate inactivity reporting.

## Test

`DefaultDeviceStateServiceTest#givenStaleLocalClockButFreshPersistedActivity_whenUpdateInactivityStateIfExpired_thenDoesNotPersistInactive` reproduces the scenario (stale local `lastActivityTime`, fresh persisted value) and asserts that `active=false` is **not** written and that the local clock is reconciled from the source of truth. Verified red → green; the full `DefaultDeviceStateServiceTest` suite passes.

LTS-4.3: https://github.com/thingsboard/thingsboard/pull/16018

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
+++ b/application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
@@ -559,6 +559,18 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
                     && (state.getLastInactivityAlarmTime() == 0L || state.getLastInactivityAlarmTime() <= state.getLastActivityTime())
                     && stateData.getDeviceCreationTime() + state.getInactivityTimeout() <= ts) {
                 if (partitionService.resolve(ServiceType.TB_CORE, stateData.getTenantId(), deviceId).isMyPartition()) {
+                    // Guard against stale-clock inactivity writes. After a partition rebalance a node may still hold
+                    // a device whose activity is now handled by another (owning) node; its local lastActivityTime is
+                    // then frozen and would wrongly flip the shared 'active' flag to false. Re-read the authoritative
+                    // persisted lastActivityTime; if it shows the device is still active, reconcile the local state
+                    // and skip the inactivity report.
+                    long persistedLastActivityTime = getPersistedLastActivityTime(deviceId);
+                    if (persistedLastActivityTime > state.getLastActivityTime()) {
+                        state.setLastActivityTime(persistedLastActivityTime);
+                        if (isActive(ts, state)) {
+                            return;
+                        }
+                    }
                     reportInactivity(ts, stateData);
                 } else {
                     cleanupEntity(deviceId);
@@ -567,6 +579,23 @@ public class DefaultDeviceStateService extends AbstractPartitionBasedService<Dev
         } else {
             log.debug("[{}] Device that belongs to other server is detected and removed.", deviceId);
             cleanupEntity(deviceId);
+        }
+    }
+
+    // Reads the authoritative persisted lastActivityTime (source of truth) so a node with a stale in-memory
+    // clock cannot mark a still-active device inactive. Returns 0 on failure to preserve default behavior.
+    long getPersistedLastActivityTime(DeviceId deviceId) {
+        try {
+            if (persistToTelemetry) {
+                Optional<TsKvEntry> entry = tsService.findLatest(TenantId.SYS_TENANT_ID, deviceId, LAST_ACTIVITY_TIME).get();
+                return entry.flatMap(KvEntry::getLongValue).orElse(0L);
+            } else {
+                Optional<AttributeKvEntry> entry = attributesService.find(TenantId.SYS_TENANT_ID, deviceId, AttributeScope.SERVER_SCOPE, LAST_ACTIVITY_TIME).get();
+                return entry.flatMap(KvEntry::getLongValue).orElse(0L);
+            }
+        } catch (Exception e) {
+            log.warn("[{}] Failed to re-read persisted lastActivityTime; proceeding with local state.", deviceId, e);
+            return 0L;
         }
     }
 
--- a/application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
+++ b/application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
@@ -39,6 +39,8 @@ import org.thingsboard.server.common.data.id.EntityId;
 import org.thingsboard.server.common.data.id.TenantId;
 import org.thingsboard.server.common.data.kv.AttributeKvEntry;
 import org.thingsboard.server.common.data.kv.AttributesSaveResult;
+import org.thingsboard.server.common.data.kv.BasicTsKvEntry;
+import org.thingsboard.server.common.data.kv.LongDataEntry;
 import org.thingsboard.server.common.data.msg.TbMsgType;
 import org.thingsboard.server.common.data.notification.rule.trigger.DeviceActivityTrigger;
 import org.thingsboard.server.common.msg.TbMsg;
@@ -60,6 +62,7 @@ import java.time.Duration;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
@@ -1120,6 +1123,49 @@ class DefaultDeviceStateServiceTest {
                             request.getEntries().get(0).getValue().equals(true)
             ));
         });
+    }
+
+    // After a partition rebalance a node may hold a device with a stale in-memory lastActivityTime while the true
+    // owner keeps it alive. Such a node must NOT persist active=false: it must re-check the authoritative persisted
+    // lastActivityTime and skip the inactivity report.
+    @Test
+    void givenStaleLocalClockButFreshPersistedActivity_whenUpdateInactivityStateIfExpired_thenDoesNotPersistInactive() {
+        // GIVEN persist-to-telemetry mode
+        ReflectionTestUtils.setField(service, "persistToTelemetry", true);
+        lenient().when(telemetrySubscriptionService.saveTimeseriesInternal(any())).thenReturn(Futures.immediateFuture(null));
+
+        long now = 10_000_000L;
+        long timeout = 100_000L;
+
+        // in-memory state is STALE: last activity long ago, so locally the device looks inactive
+        var state = DeviceState.builder()
+                .active(true)
+                .lastActivityTime(now - 10 * timeout)
+                .lastInactivityAlarmTime(0L)
+                .inactivityTimeout(timeout)
+                .build();
+        var stateData = DeviceStateData.builder()
+                .tenantId(tenantId).deviceId(deviceId).deviceCreationTime(0L)
+                .metaData(new TbMsgMetaData()).state(state).build();
+
+        given(partitionService.resolve(ServiceType.TB_CORE, tenantId, deviceId)).willReturn(tpi);
+
+        // the authoritative persisted lastActivityTime is FRESH (the true owner is keeping the device alive)
+        long freshLastActivityTime = now - 1;
+        lenient().when(tsService.findLatest(TenantId.SYS_TENANT_ID, deviceId, LAST_ACTIVITY_TIME))
+                .thenReturn(Futures.immediateFuture(Optional.of(
+                        new BasicTsKvEntry(freshLastActivityTime, new LongDataEntry(LAST_ACTIVITY_TIME, freshLastActivityTime)))));
+
+        // WHEN
+        service.updateInactivityStateIfExpired(now, deviceId, stateData);
+
+        // THEN the device is NOT persisted as inactive...
+        then(telemetrySubscriptionService).should(never()).saveTimeseriesInternal(argThat(request ->
+                request.getEntries().get(0).getKey().equals(ACTIVITY_STATE)
+                        && request.getEntries().get(0).getValue().equals(false)));
+        // ...and the stale local clock is reconciled from the source of truth
+        assertThat(state.getLastActivityTime()).isEqualTo(freshLastActivityTime);
+        assertThat(state.isActive()).isTrue();
     }
 
     private void mockSuccessfulSaveAttributes() {


application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -570,6 +582,23 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
...
 
+    // Reads the authoritative persisted lastActivityTime (source of truth) so a node with a stale in-memory
+    // clock cannot mark a still-active device inactive. Returns 0 on failure to preserve default behavior.
+    long getPersistedLastActivityTime(DeviceId deviceId) {
+        try {
+            if (persistToTelemetry) {
+                Optional<TsKvEntry> entry = tsService.findLatest(TenantId.SYS_TENANT_ID, deviceId, LAST_ACTIVITY_TIME).get();
+                return entry.flatMap(KvEntry::getLongValue).orElse(0L);
+            } else {
+                Optional<AttributeKvEntry> entry = attributesService.find(TenantId.SYS_TENANT_ID, deviceId, AttributeScope.SERVER_SCOPE, LAST_ACTIVITY_TIME).get();
```
ViacheslavKlimov, line 593: In the default configuration this read isn't authoritative, so the guard silently does nothing. `state.persistToTelemetry` defaults to `false`, so this is the branch that runs out of the box, and `attributesService` is `CachedAttributesService` whenever `cache.attributes.enabled` is true (default). `CachedAttributesService.find` checks the cache first, and `doSave(...)` calls `put(entityId, scope, attributeKvEntry)` on every write — so the stale node has its own last-written `lastActivityTime` sitting in that cache. With `cache.type=caffeine` (the default in thingsboard.yml, TTL 1440 min) that cache is node-local, so the "re-read the source of truth" call hands back exactly the stale value the node wrote itself, `persistedLastActivityTime > state.getLastActivityTime()` is false, and `reportInactivity` runs anyway. The fix therefore only works when `cache.type=redis`. The telemetry branch is fine, incidentally, because the cache-aside ts-latest DAO is `@SqlTsLatestAnyDaoCachedRedis` and falls back to a direct DB read otherwise. Is a hard dependency on Redis acceptable here? If not, the read needs to bypass the attribute cache (e.g. go through `attributesDao` directly, or gate the guard on the cache being distributed).

Separately, the two branches of this helper differ only in which DAO is called — the extraction (`entry.flatMap(KvEntry::getLongValue).orElse(0L)`) is copy-pasted, and since both futures resolve to an `Optional` of a `KvEntry` subtype it collapses to one future variable plus one extraction.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -559,6 +559,18 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
                     && (state.getLastInactivityAlarmTime() == 0L || state.getLastInactivityAlarmTime() <= state.getLastActivityTime())
                     && stateData.getDeviceCreationTime() + state.getInactivityTimeout() <= ts) {
                 if (partitionService.resolve(ServiceType.TB_CORE, stateData.getTenantId(), deviceId).isMyPartition()) {
+                    // Guard against stale-clock inactivity writes. After a partition rebalance a node may still hold
+                    // a device whose activity is now handled by another (owning) node; its local lastActivityTime is
+                    // then frozen and would wrongly flip the shared 'active' flag to false. Re-read the authoritative
+                    // persisted lastActivityTime; if it shows the device is still active, reconcile the local state
+                    // and skip the inactivity report.
+                    long persistedLastActivityTime = getPersistedLastActivityTime(deviceId);
```
ViacheslavKlimov, line 567: This adds a synchronous, per-device DB round-trip to paths that were previously pure in-memory work. `checkStates()` runs on `scheduledExecutor`, which `AbstractPartitionBasedService.init()` deliberately keeps single-threaded ("Should be always single threaded due to absence of locks") and which also serves `pollInitStateFromDB` on partition-change events. On a node holding 100k+ devices, any event that expires many devices at once now serializes one read per device on that single thread — at even 1ms per read that is ~100s of blocking, which overruns `defaultStateCheckIntervalInSec` and delays the partition-change handling queued on the same executor, i.e. it prolongs exactly the stale-cluster-view condition this PR is trying to eliminate. The same call also lands on `deviceActivityEventsExecutor` via `checkAndUpdateState` (`onDeviceConnect`, `onDeviceInactivityTimeoutUpdate`), a single-threaded executor serving all activity events for the node, so a mass reconnect pays a blocking read per event.

Could this be made async and batched — have `getPersistedLastActivityTime` return a `ListenableFuture` and chain `reportInactivity` on `deviceStateCallbackExecutor` (which is how `reportInactivity`/`onDeviceActivityStatusChange` already work), or collect candidate device ids per partition and issue one multi-entity read? A cheaper alternative is to scope the guard to when it can actually matter — e.g. only re-read for devices whose local state predates the last partition change — so steady-state inactivity reporting stays read-free.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -559,6 +559,18 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
...
+                    if (persistedLastActivityTime > state.getLastActivityTime()) {
```
ViacheslavKlimov, line 568: The `reportInactivity` decision is still the one taken *before* the read, and the read now widens that window from nanoseconds to a full DB round trip. Concretely: the scheduler thread evaluates `!isActive(ts, state)` and blocks in `getPersistedLastActivityTime`. While it's blocked, a real activity event arrives on `deviceActivityEventsExecutor`; `onDeviceActivity` → `updateActivityState` sets `state.setLastActivityTime(now)` in memory and kicks off an *asynchronous* save. The read (issued earlier, and racing that async save) comes back with the old value, so `persistedLastActivityTime > state.getLastActivityTime()` is now false against the freshly-bumped local value, and we fall straight into `reportInactivity` for a device that reported activity milliseconds ago. Re-checking `isActive(ts, state)` unconditionally after the read — not just inside the `persistedLastActivityTime >` branch — would close this cheaply and also covers the general case of the local state having moved on during the read.

Relatedly, `persistedLastActivityTime > state.getLastActivityTime()` followed by `state.setLastActivityTime(...)` is a check-then-act on a non-volatile `long` in a `@Data` POJO mutated from several threads, so it can also clobber a concurrently-written newer value backwards.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -559,6 +559,18 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
...
+                        state.setLastActivityTime(persistedLastActivityTime);
```
ViacheslavKlimov, line 569: The reconcile updates `lastActivityTime` but never touches `active`, and `updateInactivityStateIfExpired` is reachable from `checkStates` with `state.isActive() == false` (unlike the `checkAndUpdateState` path, which only enters here when active is true) — e.g. a state fetched from DB with persisted `active=false` and `lastInactivityAlarmTime == 0`, or a local state left at `active=false` because the `ACTIVITY_STATE` save in `onDeviceActivityStatusChange(true, ...)` failed. In that case we learn from the source of truth that the device *is* active, then `return` leaving the local state at `active=false`. Each later check re-reads and takes the same early return as long as the owner keeps the device alive, so `active` is never restored, and the device stays counted in the inactive bucket by `reportActivityStats` until an actual activity event happens to land on this node. Should this branch also drive the state back to active (e.g. via `onDeviceActivityStatusChange(true, stateData)` when `!state.isActive()`), the way `checkAndUpdateState`'s "trying to fix activity state" branch does?

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -570,6 +582,23 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
         }
     }
 
+    // Reads the authoritative persisted lastActivityTime (source of truth) so a node with a stale in-memory
+    // clock cannot mark a still-active device inactive. Returns 0 on failure to preserve default behavior.
+    long getPersistedLastActivityTime(DeviceId deviceId) {
+        try {
+            if (persistToTelemetry) {
+                Optional<TsKvEntry> entry = tsService.findLatest(TenantId.SYS_TENANT_ID, deviceId, LAST_ACTIVITY_TIME).get();
```
ViacheslavKlimov, line 590: The `.get()` here is unbounded. If the timeseries/attributes DAO stalls (connection pool exhaustion, Cassandra node flapping), the single-threaded scheduler thread parks indefinitely, which stops all further state checks *and* the partition-change processing submitted to the same executor. Note that "fails safe (returns 0)" doesn't hold for a hang: with no timeout there's no failure to fall back from, the thread just parks. The pack-fetch path shows the bounded form (`Futures.successfulAsList(deviceStateFutures).get(5, TimeUnit.MINUTES)`), and the fail-safe `return 0L` already gives sane behavior on timeout, so a bounded `get(timeout, unit)` costs nothing and caps the blast radius. (There is already one other unbounded `.get()` reachable from `checkStates` — `fetchDeviceStateDataUsingSeparateRequests(DeviceId)` — so this isn't a new class of exposure, but it is a second one on the same thread and worth not adding.)

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -570,6 +582,23 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
...
+                return entry.flatMap(KvEntry::getLongValue).orElse(0L);
+            } else {
+                Optional<AttributeKvEntry> entry = attributesService.find(TenantId.SYS_TENANT_ID, deviceId, AttributeScope.SERVER_SCOPE, LAST_ACTIVITY_TIME).get();
+                return entry.flatMap(KvEntry::getLongValue).orElse(0L);
+            }
+        } catch (Exception e) {
+            log.warn("[{}] Failed to re-read persisted lastActivityTime; proceeding with local state.", deviceId, e);
```
ViacheslavKlimov, line 597: `catch (Exception e)` swallows `InterruptedException` from `.get()` without restoring the interrupt flag. `stop()` calls `shutdownNow()` on these executors, and `initStateFromDB` cancels partition fetch tasks; once the flag is cleared here and not re-set, the enclosing `checkStates` loop (or the init pack loop, which iterates up to 1000 devices) keeps going and the next `.get()` blocks again instead of aborting promptly. Worth splitting it out: `catch (InterruptedException e) { Thread.currentThread().interrupt(); return 0L; }` before the general catch — compare `fetchDeviceStateDataUsingSeparateRequests`, which at least distinguishes `InterruptedException | ExecutionException` and rethrows.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -570,6 +582,23 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
         }
     }
 
+    // Reads the authoritative persisted lastActivityTime (source of truth) so a node with a stale in-memory
+    // clock cannot mark a still-active device inactive. Returns 0 on failure to preserve default behavior.
+    long getPersistedLastActivityTime(DeviceId deviceId) {
```
ViacheslavKlimov, line 587: This is package-private, but nothing outside the class calls it and the new test neither invokes nor stubs it directly (it stubs `tsService` instead). The other package-private methods here (`updateInactivityStateIfExpired`, `checkAndUpdateState`, `getOrFetchDeviceStateData`) are widened specifically because tests drive or spy them, so leaving this one at default visibility with no consumer suggests a test hook that never materialized. Making it `private` keeps the intent clear.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
```diff
@@ -559,6 +559,18 @@ void updateInactivityStateIfExpired(long ts, DeviceId deviceId, DeviceStateData
                     && (state.getLastInactivityAlarmTime() == 0L || state.getLastInactivityAlarmTime() <= state.getLastActivityTime())
                     && stateData.getDeviceCreationTime() + state.getInactivityTimeout() <= ts) {
                 if (partitionService.resolve(ServiceType.TB_CORE, stateData.getTenantId(), deviceId).isMyPartition()) {
+                    // Guard against stale-clock inactivity writes. After a partition rebalance a node may still hold
```
ViacheslavKlimov, line 562: The same rationale is written out three times: five lines inline here, two more on `getPersistedLastActivityTime`, and three more above the new test. One canonical explanation (on the helper method, where the behavior lives) plus a one-line pointer at the call site would read better; the rest belongs in the PR description.

Separately, "stale-clock" is a bit misleading as a name for this — the node's clock is fine, it's the in-memory `lastActivityTime` that is frozen because activity reports are going to the new owner. Something like "stale local activity state" would save the next reader from hunting for a clock-skew problem that isn't there (the test method name inherits the same wording).

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -1122,6 +1125,49 @@ void givenDeviceActivityEventHappenedBeforeAdded_whenOnQueueMsg_thenShouldSaveAc
         });
     }
 
+    // After a partition rebalance a node may hold a device with a stale in-memory lastActivityTime while the true
+    // owner keeps it alive. Such a node must NOT persist active=false: it must re-check the authoritative persisted
+    // lastActivityTime and skip the inactivity report.
+    @Test
+    void givenStaleLocalClockButFreshPersistedActivity_whenUpdateInactivityStateIfExpired_thenDoesNotPersistInactive() {
```
ViacheslavKlimov, line 1132: Only one of the four new branches is covered — telemetry mode with a fresher persisted value. Still untested: (a) the attributes branch (`attributesService.find`), which is the *default* configuration (`state.persistToTelemetry: false`) and therefore what most deployments run; (b) the fail-safe path, where the read throws and the method returns 0 so inactivity is still reported — this is the property the PR description leans on hardest; (c) the reconcile-but-still-expired path, where the persisted value is newer than local yet still older than `ts - inactivityTimeout`, so local state should be updated *and* inactivity reported.

The attributes gap is more than a coverage hole: the pre-existing `provideParametersForUpdateInactivityStateIfExpired` cases now reach `attributesService.find` on an unstubbed mock, get `null`, and NPE straight into the broad `catch (Exception)` — so they pass while silently exercising the failure path, and a real regression in that branch wouldn't be caught by anything. A `@ParameterizedTest` over `persistToTelemetry` covering fresh/stale/absent/throwing would close all of these.

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
```diff
@@ -1122,6 +1125,49 @@ void givenDeviceActivityEventHappenedBeforeAdded_whenOnQueueMsg_thenShouldSaveAc
...
+                .build();
+        var stateData = DeviceStateData.builder()
+                .tenantId(tenantId).deviceId(deviceId).deviceCreationTime(0L)
+                .metaData(new TbMsgMetaData()).state(state).build();
+
+        given(partitionService.resolve(ServiceType.TB_CORE, tenantId, deviceId)).willReturn(tpi);
+
+        // the authoritative persisted lastActivityTime is FRESH (the true owner is keeping the device alive)
+        long freshLastActivityTime = now - 1;
+        lenient().when(tsService.findLatest(TenantId.SYS_TENANT_ID, deviceId, LAST_ACTIVITY_TIME))
```
ViacheslavKlimov, line 1155: `lenient()` on this stub isn't needed — the stub *is* consumed by the code under test, and marking it lenient throws away the strict-stub check that would otherwise fail loudly if a refactor stopped reading the persisted value (which is exactly the regression this test exists to catch); the `lenient()` on `saveTimeseriesInternal` is legitimate since that one is intentionally never called. The explicit `given(partitionService.resolve(ServiceType.TB_CORE, tenantId, deviceId)).willReturn(tpi)` is also already stubbed leniently in `setUp()` and can be dropped. Finally, the `DeviceState`/`DeviceStateData` builder block is now duplicated for the 17th time in this class, which would be a good moment to extract a small fixture helper.

ViacheslavKlimov (commented):
## Review summary

Reviewed 2 changed files in [fix(device-state): prevent stale-clock nodes from marking active devices inactive](https://github.com/thingsboard/thingsboard/pull/16016). Left 10 comment(s) inline.

The most important one first: in the **default** configuration (`state.persistToTelemetry: false`, `cache.type: caffeine`) the re-read goes through `CachedAttributesService`, which serves it from the node's own local Caffeine cache — the same cache the stale node populated when it last wrote `lastActivityTime`. The "source of truth" read then hands back exactly the stale value, the guard never trips, and the fix only actually works with `cache.type=redis`. Also worth a look: the blocking per-device `.get()` is added to the single-threaded `scheduledExecutor` that also processes partition-change events (so a slow read prolongs the very stale-view window this PR targets), the `reportInactivity` decision is still the one taken *before* the read (a concurrent activity event during the round trip leads to reporting inactivity for a just-active device), and the reconcile branch fixes `lastActivityTime` but never restores `active`, so a state that is locally `active=false` stays stuck in the inactive bucket.

---

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
volodymyr-babak: This is the important one and you're right — I traced it end to end and the guard was a no-op in the default configuration.

`state.persistToTelemetry` defaults to `false` (thingsboard.yml:1039), `cache.type` defaults to `caffeine` (thingsboard.yml:648, and `TbCaffeineCacheConfiguration` is `matchIfMissing = true`), `cache.attributes.enabled` defaults to `true`. `CachedAttributesService.find` returns the cached value when present, and the write path reaches `put(entityId, scope, attributeKvEntry)` (CachedAttributesService.java:242) via `tsSubService.saveAttributesInternal`. So the stale node's own last write is sitting in its own node-local cache and gets handed straight back. Even on a cold cache it's only correct once: the first read populates the cache with a 1440-minute TTL and every later check re-reads that.

I also checked whether caffeine-in-a-cluster is simply unsupported, which would have made this moot — it isn't enforced anywhere, there's no invalidation channel on `CaffeineTbTransactionalCache`, and it's the shipped default.

Fixed by reading through `AttributesDao` directly, which bypasses the cache, with the reason recorded at the call site. The telemetry branch stays on `tsService` as you noted. The two branches are now one method returning a `ListenableFuture<Long>`, so the duplicated extraction is gone too.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
volodymyr-babak: Agreed, and fixed — the guard is now fully asynchronous. `findPersistedLastActivityTime` returns a `ListenableFuture<Long>` and `reconcileThenReportInactivity` chains the decision on `deviceStateCallbackExecutor`, so nothing blocks `scheduledExecutor` (or `deviceActivityEventsExecutor` on the `checkAndUpdateState` path). The attributes read runs on `deviceStateExecutor` since `AttributesDao.find` is synchronous.

One correction on the cheaper alternative, though — I don't think "only re-read for devices whose local state predates the last partition change" works. The code only reaches this point when `partitionService.resolve(...).isMyPartition()` is true, so the case it has to catch is the one where *this node's partition view is itself stale*: it still believes it owns the device while activity is going to the real owner. That node has not observed a partition change, so a `lastPartitionChangeTime` gate would skip exactly the scenario the guard exists for. Going async solves the same problem without that blind spot.

Worth noting the read is also more bounded than it looks: the enclosing condition requires `lastInactivityAlarmTime == 0 || lastInactivityAlarmTime <= lastActivityTime`, and `reportInactivity` sets `lastInactivityAlarmTime = ts`, so it fires once per device per inactivity transition rather than on every check interval. A mass expiry is still a burst, which is why it's off the scheduler now.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
volodymyr-babak: Confirmed and fixed. The re-check is now unconditional rather than nested inside the `persistedLastActivityTime >` branch:

```java
if (persistedLastActivityTime > state.getLastActivityTime()) {
    state.setLastActivityTime(persistedLastActivityTime);
}
// Re-check against current local state, not only the reconciled value: a real activity event may
// have arrived on another executor while the read was in flight.
if (isActive(ts, state)) { ... return; }
reportInactivity(ts, stateData);
```

On the non-volatile `long`: you're right that it's a check-then-act, and I've kept the write conditional so it can only move forward. I've deliberately not gone further here — `DeviceState` is a `@Data` POJO mutated from several executors throughout this class (`updateActivityState`, `reportInactivity`'s callback, `onDeviceActivityStatusChange`), so making this one field safe would be an inconsistent half-measure. Making `DeviceState` properly thread-safe is worth doing, but as its own change rather than inside a targeted fix.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
volodymyr-babak: Good catch — that path was genuinely reachable and would have stuck. `updateInactivityStateIfExpired` is called unconditionally from `checkStates` (line 511), so a state with `active=false` and `lastInactivityAlarmTime == 0` reaches the guard, and the old code returned without touching `active`.

The reconcile branch now drives the state back:

```java
if (isActive(ts, state)) {
    if (!state.isActive()) {
        onDeviceActivityStatusChange(true, stateData);
    }
    return;
}
```

Covered by `givenLocalStateInactiveButPersistedActivityFresh_whenUpdateInactivityStateIfExpired_thenDrivesStateBackToActive`, parameterised over both persistence modes.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
volodymyr-babak: Fixed, and by removal rather than by adding a timeout — there is no `.get()` on this path any more. The read is a `ListenableFuture` consumed via `Futures.addCallback`, so a stalled DAO parks nothing; the scheduler thread has already moved on.

You're right that "fails safe (returns 0)" never covered a hang. Noted separately that `fetchDeviceStateDataUsingSeparateRequests(DeviceId)` still has an unbounded `.get()` on the same thread — leaving that alone here since it's pre-existing and unrelated to this fix, but it's worth its own ticket.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
volodymyr-babak: Fixed by removal — the blocking `.get()` and the broad `catch (Exception)` are both gone, so there's no interrupt flag to lose. Failure now arrives through `onFailure`, which keeps the fail-safe behaviour of reporting inactivity from local state.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
volodymyr-babak: Agreed. The method no longer exists in that form; it's been split into `reconcileThenReportInactivity` and `findPersistedLastActivityTime`, both `private`. The new tests drive `updateInactivityStateIfExpired` and stub `tsService`/`attributesDao`, so nothing needs the wider visibility.

application/src/main/java/org/thingsboard/server/service/state/DefaultDeviceStateService.java
volodymyr-babak: Both fixed. There is now one javadoc on `reconcileThenReportInactivity` where the behaviour lives, a one-line pointer at the call site, and nothing above the tests.

On the naming: you're right, and it was actively misleading — the clock is fine, the in-memory activity state is stale. All in-code wording now says "local state" / "stale local activity state"; `grep -rn 'stale.clock' application/src` comes back empty, and the test methods are renamed (`givenStaleLocalStateButFreshPersistedActivity_...`). The branch name and PR title still carry the old wording since renaming the branch would close and reopen this PR — happy to retitle the PR and the Jira ticket (PROD-8824) if you'd like them consistent.

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
volodymyr-babak: Agreed on all three gaps, and the unstubbed-mock point was the sharp end of it — thanks for catching that.

Confirmed the mechanism: `attributesService.find` (now `attributesDao.find`) returns `null` on an unstubbed mock, which NPE'd into the broad catch and returned 0, so `provideParametersForUpdateInactivityStateIfExpired` was passing while exercising the read-failure path rather than the branch it names. `setUp` now carries `lenient().when(attributesDao.find(any(), any(), any(), anyString())).thenReturn(Optional.empty())`, so the default is a clean "nothing persisted" rather than an exception.

The new test is a `@ParameterizedTest` over `persistToTelemetry` covering fresh / still-expired-after-reconcile / absent / read-throws, plus the inactive-to-active reconcile from your other comment — 10 cases. Full class: 108 tests, 0 failures.

application/src/test/java/org/thingsboard/server/service/state/DefaultDeviceStateServiceTest.java
volodymyr-babak: All three addressed. The `lenient()` is gone from the persisted-value stub (it's `given(...)` now, so a refactor that stops reading the value fails loudly, which is the point of the test); the `lenient()` that remains is on `saveTimeseriesInternal`/`saveAttributesInternal` for the cases where the save is intentionally never made. The redundant `given(partitionService.resolve(...))` is dropped — `setUp` already covers it.

On the fixture: extracted `staleStateData(now, timeout, active)` plus `givenPersistedLastActivityTime(...)` and `thenPersistedInactive(...)`, so the five new cases share one builder block instead of adding five more. I haven't retro-fitted the other 16 copies in this class — worth doing, but as a separate cleanup rather than inside this fix.

PR #16017, opened 2026-08-07
Add 'Overwrite activity time' column type to device bulk import

DmytroKhylko:
## Pull Request description

 Adds support for enabling/disabling 'overwrite activity time for connected device'
  when bulk importing gateway devices via CSV, following the same pattern as IS_GATEWAY.
  
 Test csv
 ```csv
name,type,label,is_gateway,overwrite_activity_time,access_token,description
Gateway Device 1,default,GW Label 1,true,true,token-gw-001,Main gateway with overwrite enabled
Gateway Device 2,default,GW Label 2,true,false,token-gw-002,Gateway with overwrite disabled
Regular Device 1,default,Dev Label 1,false,false,token-dev-001,Non-gateway device
Gateway Device 3,default,GW Label 3,true,true,token-gw-003,Another gateway with overwrite enabled

 ```
Before: the `overwrite_activity_time` is treated as regular attribute
image

After: the `overwrite_activity_time` is treated as a separate field
image

## General checklist

## Back-End feature checklist

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/device/DeviceBulkImportService.java
+++ b/application/src/main/java/org/thingsboard/server/service/device/DeviceBulkImportService.java
@@ -101,6 +101,9 @@ public class DeviceBulkImportService extends AbstractBulkImportService<Device> {
                 case IS_GATEWAY:
                     additionalInfo.set("gateway", BooleanNode.valueOf(Boolean.parseBoolean(value)));
                     break;
+                case OVERWRITE_ACTIVITY_TIME:
+                    additionalInfo.set("overwriteActivityTime", BooleanNode.valueOf(Boolean.parseBoolean(value)));
+                    break;
             }
             device.setAdditionalInfo(additionalInfo);
         });
--- a/ui-ngx/src/app/shared/import-export/table-columns-assignment.component.ts
+++ b/ui-ngx/src/app/shared/import-export/table-columns-assignment.component.ts
@@ -87,7 +87,8 @@ export class TableColumnsAssignmentComponent implements OnInit, ControlValueAcce
           { value: ImportEntityColumnType.sharedAttribute },
           { value: ImportEntityColumnType.serverAttribute },
           { value: ImportEntityColumnType.timeseries },
-          { value: ImportEntityColumnType.isGateway }
+          { value: ImportEntityColumnType.isGateway },
+          { value: ImportEntityColumnType.overwriteActivityTime }
         );
         this.columnDeviceCredentials.push(
           { value: ImportEntityColumnType.accessToken },
--- a/common/data/src/main/java/org/thingsboard/server/common/data/sync/ie/importing/csv/BulkImportColumnType.java
+++ b/common/data/src/main/java/org/thingsboard/server/common/data/sync/ie/importing/csv/BulkImportColumnType.java
@@ -48,6 +48,7 @@ public enum BulkImportColumnType {
     SNMP_VERSION,
     SNMP_COMMUNITY_STRING,
     IS_GATEWAY,
+    OVERWRITE_ACTIVITY_TIME,
     DESCRIPTION,
     ROUTING_KEY,
     SECRET;
--- a/ui-ngx/src/app/shared/import-export/import-export.models.ts
+++ b/ui-ngx/src/app/shared/import-export/import-export.models.ts
@@ -70,6 +70,7 @@ export enum ImportEntityColumnType {
   snmpVersion = 'SNMP_VERSION',
   snmpCommunityString = 'SNMP_COMMUNITY_STRING',
   isGateway = 'IS_GATEWAY',
+  overwriteActivityTime = 'OVERWRITE_ACTIVITY_TIME',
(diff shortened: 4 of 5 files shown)


ui-ngx/src/app/shared/import-export/table-columns-assignment.component.ts
```diff
@@ -87,7 +87,8 @@ export class TableColumnsAssignmentComponent implements OnInit, ControlValueAcce
           { value: ImportEntityColumnType.sharedAttribute },
           { value: ImportEntityColumnType.serverAttribute },
           { value: ImportEntityColumnType.timeseries },
-          { value: ImportEntityColumnType.isGateway }
+          { value: ImportEntityColumnType.isGateway },
+          { value: ImportEntityColumnType.overwriteActivityTime }
```
vvlladd28, line 91: The new type is added to the dropdown, but `columnsUpdated()` is never taught about it, so it is the only non-KV device column type that stays selectable after it has already been assigned — `name`, `type`, `label`, `description`, `isGateway` and every credential type all get `disabled = true` once used.

Concretely: a user can pick "Overwrite activity time" for two different CSV columns, the dialog stays valid, and on the backend `EntityData.fields` is a `Map<BulkImportColumnType, String>`, so only the last column's value survives — the first one is silently discarded. Since the PR follows the `IS_GATEWAY` pattern everywhere else, it would be good to follow it here too:

```ts
const isSelectOverwriteActivityTime =
  this.columns.findIndex((column) => column.type === ImportEntityColumnType.overwriteActivityTime) > -1;
const overwriteActivityTimeColumnType =
  this.columnTypes.find((columnType) => columnType.value === ImportEntityColumnType.overwriteActivityTime);
if (overwriteActivityTimeColumnType) {
  overwriteActivityTimeColumnType.disabled = isSelectOverwriteActivityTime;
}
```

application/src/main/java/org/thingsboard/server/service/device/DeviceBulkImportService.java
```diff
@@ -101,6 +101,9 @@ protected void setEntityFields(Device device, Map<BulkImportColumnType, String>
                 case IS_GATEWAY:
                     additionalInfo.set("gateway", BooleanNode.valueOf(Boolean.parseBoolean(value)));
                     break;
+                case OVERWRITE_ACTIVITY_TIME:
+                    additionalInfo.set("overwriteActivityTime", BooleanNode.valueOf(Boolean.parseBoolean(value)));
```
vvlladd28, line 105: No test covers the new column type, and the back-end checklist item is left unchecked without an explanation. `DeviceControllerTest` already has the whole harness for this (`testBulkImportDeviceWithoutCredentials` and friends build a `BulkImportRequest` and assert on the saved device), so pinning this down is a handful of lines: import a row with `OVERWRITE_ACTIVITY_TIME` and assert `savedDevice.getAdditionalInfo().get("overwriteActivityTime")` is the expected boolean. Without it nothing catches a future rename of the `additionalInfo` field name, which is what `DefaultTransportService` and `AbstractGatewaySessionHandler` read via `DataConstants.OVERWRITE_ACTIVITY_TIME_PARAMETER`.

vvlladd28 (commented):
## Review summary

Reviewed 5 changed files in [Add 'Overwrite activity time' column type to device bulk import](https://github.com/thingsboard/thingsboard/pull/16017). Left 2 comment(s) inline.

The main thing to look at is `table-columns-assignment.component.ts`: unlike `isGateway`, the new column type is never disabled in `columnsUpdated()` once it has been assigned, so it can be mapped to two CSV columns at once and one of them is silently dropped on the backend.

## Additional findings

- **ui-ngx/src/assets/locale/locale.constant-en_US.json:3828** — `import.column-type.activity-time-from-gateway-device` (the line right below your new key) is dead: it was added in 1bdd7d6 for this very feature, the device-form key was renamed to `overwrite-activity-time` in d06ce55, and nothing has referenced the import-side one since. All `import.column-type.*` lookups go through the static `importEntityColumnTypeTranslations` map, so there is no dynamic reference either — and the key is still carried in 23 locale files. Since your new key replaces it in spirit, this PR is a natural place to drop it.

---

ui-ngx/src/app/shared/import-export/table-columns-assignment.component.ts
DmytroKhylko: Added the same findIndex/find/disabled pattern for overwriteActivityTime directly after the isGateway block inside the DEVICE guard.

application/src/main/java/org/thingsboard/server/service/device/DeviceBulkImportService.java
DmytroKhylko: Added testBulkImportDeviceWithOverwriteActivityTime in DeviceControllerTest, following the same harness as the neighbouring bulk import tests. It imports a row with OVERWRITE_ACTIVITY_TIME=true and asserts savedDevice.getAdditionalInfo().get("overwriteActivityTime").asBoolean() is true.

PR #16028, opened 2026-08-11
Open built-in IoT Hub items locally instead of installing duplicates

pinkevmladchy:
## Pull Request description

The IoT Hub API exposes a server-owned, read-only builtIn flag on item versions, marking content that already ships inside ThingsBoard (bundled widgets, SCADA symbols). The Hub browser still offered Install for those entries, 
  which duplicated a component the tenant already has and inflated Hub install counters.                                                                                                                                            
                                                                                                                                                                                                                                    
  This PR makes the UI route the user to the local copy instead. Wording matches thingsboard.io: CTA Open, marker Built-in.                                                                                                         
                                                                                                                                                                                                                                    
  Scope (UI only, no backend changes):                                                                                                                                                                                              
                                                                                                                                                                                                                                    
  - CTA verb — one shared helper iotHubItemActionMode() returns a mode (open / connect / install); labels for cards, detail footer and add-to-entity CTA come from mode-keyed maps. Install / Connect unchanged for everything else.
  - Open does not install — a built-in item's fqn equals the platform widget's own fqn, so the new IotHubBuiltInService matches on it (system.<fqn>, falling back to tenant.<fqn>) and navigates to the widget type in the Widgets  
  library.                                                                                                                                                                                                                          
  - Built-in marker appended to the version line of the detail header: v 1.0.0 | Built-in. No second badge.                                                                                                                         
  - Install counts hidden for built-in items everywhere they appear (catalogue card, detail header, dashboard widget-select card, home-page Hub widget); the counter and its separator derive from one boolean.                     
  - No install request — all three entry points to POST /api/iot-hub/versions/{versionId}/install are gated: IotHubActionsService.installItem / installDevice, DashboardWidgetSelectComponent.installAndAddWidget,                  
  TbIotHubAddItemDialogComponent.                      

 Missing local copy. If a sysadmin deleted the bundled widget, the Hub entry still lists it while the local component is gone. Open then shows a single dialog — "Not available in this instance … You can install it from the IoT 
  Hub instead" — and pressing Install starts the installation immediately: no second confirmation step, since the user has just confirmed. Once it is installed, the item behaves like any other installed item and can be opened   
  from the library.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                    
  To support that, IotHubInstallDialogData takes an optional skipConfirm, which opens the dialog directly on the installing step. It is ignored when the item needs input first (entity selection for calculated fields, rule chain 
  options). The installing step also got its own heading (iot-hub.installing-title — "Installing 'X'") instead of reusing the confirm title, so progress is no longer rendered as an unanswered "Install 'X'?" question — this makes
  the regular install flow clearer too.                                                                                                                                                                                             
                                                                                                                                                                                                                                    
  The existence check is intentionally lazy (performed on click, not prefetched per card), because there is no bulk fqn lookup to prefetch with: GET /api/widgetType?fqn= is one fqn per request and returns the full descriptor,   
  /api/widgetTypeFqns is scoped to a bundle the Hub payload does not name, and /api/widgetTypes returns heavy WidgetTypeInfo rows. Browsing the catalogue costs zero extra requests; a click costs the request already needed to    
  resolve the widget type id for navigation.                                                                                                                                                                                        
                                                                                                                                                                                                                                    
  Backward compatibility. builtIn is optional and read only through isBuiltInItem(), which treats absent / null / non-boolean as false — the current Hub deployment does not return the field yet, so until it does, every item     
  keeps its existing label, counter and install flow. The flag is never sent back. Non-built-in items are unaffected.                                                                                                               
                                                                                                                                                                                                                                    
  Follow-ups. No server-side builtIn filter exists (neither listing nor version search accepts it), so there is no built-in/installable toggle here — that needs Boolean builtIn in MpListingQuery / MpItemVersionQuery on the Hub  side. Only widgets ship inside the platform today; if a separate SCADA-symbol item type appears, IotHubBuiltInService.resolveLocalComponentUrl is the one place to extend. No new REST API or yml property — nothing to document. 

## General checklist

## Back-End feature checklist


ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-built-in.service.ts
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * There is no API to test a batch of fqns at once — `GET /api/widgetType?fqn=` is per fqn and
+   * carries the full descriptor, and `/api/widgetTypeFqns` is scoped to a bundle the Hub payload
+   * does not name — so this must stay a per-item check triggered by a click, never a prefetch
+   * across a catalogue page.
+   */
+  resolveLocalWidgetType(item: MpItemVersionView): Observable<IotHubLocalLookup<WidgetType>> {
+    const noMatch: IotHubLocalLookup<WidgetType> = { value: null, failed: false };
+    const fqn = item?.dataDescriptor?.fqn;
+    if (!isNotEmptyStr(fqn)) {
```
vvlladd28, line 73: A blank fqn returns `noMatch` — `failed: false` — which the callers read as "the component is not here". That contradicts the invariant stated in the `IotHubLocalLookup` doc block right above: `failed` exists to separate "the lookup did not answer" from "the component is not here", and only the latter may be offered as an install. A missing fqn is squarely the first case — we never asked the server anything.

The consequence is the duplicate this PR exists to prevent: `openLocalComponent` → `'missing'` → `confirmInstallMissingBuiltIn` → the user is told the widget "may have been deleted by the system administrator" and presses Install, while the bundled widget is sitting there untouched.

Worth noting `MpItemVersionView.dataDescriptor` is typed `any` (`iot-hub-version.models.ts:91`), so nothing at compile time guarantees a built-in widget item carries `fqn` — and `builtIn` isn't live on the Hub deployment yet, so the pairing hasn't been exercised against a real payload. Would `failed: true` be the safer default here? It costs a "could not check this component" alert in a case that shouldn't happen, versus a silent duplicate if it does.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-actions.service.ts
```diff
@@ -122,6 +189,9 @@ export class IotHubActionsService {
   }
 
   installDevice(item: MpItemVersionView): Observable<string> {
+    if (isBuiltInItem(item)) {
```
vvlladd28: If built-in content is only ever widgets, this guard can't fire — `installDevice` is only reached with `item.type === DEVICE`, so `isBuiltInItem(item)` is always false here.

It's also the one spot where widening `builtIn` to another type later would misbehave rather than just degrade: `runInstall` delegates to `installDevice` for DEVICE items, so `openOrInstallBuiltIn` → `'missing'` (non-widget types return early in `resolveLocalComponentUrl`) → user confirms → `runInstall(item, true)` → `installDevice` → this guard again. The user would be re-asked instead of ever installing, and `skipConfirm` is dropped on that path since `installDevice` doesn't forward it.

Simplest fix for both: drop this guard and have `runInstall` call `openDeviceInstallDialog(item)` directly, so the built-in decision lives only at the public entry points.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-actions.service.ts
```diff
@@ -72,7 +76,70 @@ export class IotHubActionsService {
...
+    return this.runInstall(item);
+  }
+
+  /**
+   * Opens the local copy of a built-in item, and emits true only when the caller should fall back to
+   * its own install flow — i.e. the component really is absent and the user asked to install it.
+   * A lookup that could not answer emits false: existence is unknown, so offering an install would
+   * risk the duplicate this whole flow exists to avoid.
+   */
+  openBuiltInOrConfirmInstall(item: MpItemVersionView): Observable<boolean> {
```
vvlladd28: The `boolean` here needs a four-line JSDoc to explain what `true` means ("the caller should now run its own install flow"), which is a sign the return type is doing too much implicit work — `false` conflates "opened fine", "navigation was blocked", "lookup failed", and "user declined". Returning the richer outcome (or a small named union like `'handled' | 'install-requested'`) would let each of the callers branch on something self-describing, and would make it much harder to read the flag backwards.

Related: `showBuiltInLookupFailed()` and `confirmInstallMissingBuiltIn()` are public on this service only so `DashboardWidgetSelectComponent` can hand-assemble the same ladder. If the outcome were expressive enough, that component could reuse one method instead of two escape hatches.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-built-in.service.ts
```diff
@@ -0,0 +1,151 @@
...
+      // remaining candidate cannot turn a failure into a trustworthy "not here".
+      first(lookup => !!lookup.value?.id?.id || lookup.failed, noMatch)
+    );
+  }
+
+  /** Same match as `resolveLocalWidgetType`, resolved to the info projection widget pickers need. */
+  resolveLocalWidgetTypeInfo(item: MpItemVersionView): Observable<IotHubLocalLookup<WidgetTypeInfo>> {
+    return this.resolveLocalWidgetType(item).pipe(
+      switchMap(lookup => lookup.value
+        ? this.widgetService.getWidgetTypeInfoById(lookup.value.id.id, { ignoreErrors: true, ignoreLoading: true }).pipe(
```
vvlladd28, line 97: This is two sequential round trips in the common case and three in the worst (`system.<fqn>` miss → `tenant.<fqn>` hit → info by id) — `concatMap` is sequential by design, since system has to be tried before tenant. The `getWidgetTypeInfoById` call is unavoidable (`WidgetType` has no `image` / `description` / `widgetType`, which `toWidgetInfo` needs), but the ones before it return the full `WidgetType` including `descriptor` — `templateHtml`, `templateCss`, `controllerScript`, `defaultConfig`, the settings form schemas — purely so `lookup.value.id.id` can be read and the rest thrown away. Bounded per click so it won't fall over, but on a slow link it's a visible pause before the widget is selected.

How often is the `tenant.` candidate expected to hit? Built-in content is system-scoped, so for a genuine built-in item the first candidate should match. A tenant-scoped copy that came from the Hub is an *installed* widget, which `onIotHubWidgetClicked` already intercepts via `installedWidgets` before reaching this path — so the second candidate only covers a widget the tenant hand-created under the same fqn. If that's not worth an extra round trip on the miss path, dropping it takes the worst case from three requests to two.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-utils.ts
```diff
@@ -14,13 +14,56 @@
...
+
+// CTA used on the item detail footer, where the device wording is spelled out.
+export const iotHubActionModeDetailLabels: Record<IotHubItemActionMode, string> = {
+  open: 'iot-hub.open',
+  connect: 'iot-hub.connect-device',
+  install: 'iot-hub.install'
+};
+
+// CTA used where an item is picked to be added to an entity or a dashboard.
+export const iotHubActionModeAddLabels: Record<IotHubItemActionMode, string> = {
```
vvlladd28: Three exported maps that differ in two cells total feels like more surface than the problem needs — and `iotHubActionModeAddLabels` maps `connect` and `install` to the same key, so it's really a two-case table wearing a three-case shape.

The cost shows up downstream: both `TbIotHubItemCardComponent` and `TbIotHubItemDetailDialogComponent` carry the same `actionLabel` getter whose only difference is which map the non-`add` branch reads, plus a private `actionMode` getter that exists purely to make that expression readable. Exporting one function instead — `iotHubItemActionLabel(item, context: 'card' | 'detail' | 'add')`, tables kept private — collapses each of those to a single line (`return iotHubItemActionLabel(this.item, this.mode === 'add' ? 'add' : 'card')`), drops a member from both classes, makes it impossible to reach for the wrong map, and turns "add a fourth surface" into one edit here rather than a hunt for every place a map is selected.

Separately, the naming is asymmetric: `iotHubActionModeLabels` is actually the *card* variant, so `iotHubActionModeCardLabels` would read better next to its two siblings.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-item-card.component.ts
```diff
@@ -51,6 +56,31 @@ export class TbIotHubItemCardComponent {
     private iotHubApiService: IotHubApiService
   ) {}
 
+  /**
+   * Built-in content already ships with the platform. A single flag on the card root drives
```
vvlladd28: The comment-to-code ratio across this PR is high enough to become its own maintenance burden — a three-line JSDoc on a one-line delegation to `isBuiltInItem`, a two-line comment on `showInstallCount` returning `!this.builtIn`, and in `iot-hub-built-in.service.ts` a doc block per member where several restate the code (`lookupResult`, the `builtIn` getter's "see `isBuiltInItem`").

The genuinely valuable ones are the non-obvious rationales: why 404 is treated differently from other failures, why `first()` also stops on failure, why there's no batch fqn API. Keeping those and trimming the rest would make the rationale easier to find — and would avoid comments describing rendering paths ("scroll grid, search-as-you-type") that will quietly go stale.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-item-card.component.html
```diff
@@ -17,7 +17,8 @@
 -->
 @if (isCompactLayout()) {
   <!-- Compact card (CF / RC) — Design: horizontal row, icon | content | button -->
-  <div class="tb-iot-hub-card-compact" role="button" tabindex="0" (click)="onCardClick()" (keyup.enter)="onCardClick()">
+  <div class="tb-iot-hub-card-compact" role="button" tabindex="0" [class.tb-iot-hub-built-in]="builtIn"
```
vvlladd28: `tb-iot-hub-built-in` has no style rule anywhere in the repo — grepping turns up only the three template bindings added here (this file twice, `dashboard-widget-select.component.html` once). So today these bindings are inert markup, and a reader will reasonably assume built-in cards are styled somewhere. Either add the intended rule alongside `.tb-iot-hub-item-card` / `.tb-iot-hub-card-compact`, or drop the bindings until there's a visual to attach to them.

ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-widget-select.component.ts
```diff
@@ -737,6 +755,35 @@ export class DashboardWidgetSelectComponent {
     this.cd.markForCheck();
   }
 
+  private addIotHubWidget(item: MpItemVersionView): void {
```
vvlladd28: This branch looks unreachable: `onIotHubWidgetClicked` already returns early for built-in items before opening the detail dialog, and the dialog's `addItem()` closes with the same object it was handed, so `result.item` can't be built-in here.

If it's purely defensive, `addIotHubWidget` reduces to a one-line alias for `installAndAddWidget` and could be dropped in favour of calling that directly. If the guard is load-bearing for some path I've missed, a comment naming that path would help — the current one explains *why built-ins aren't installed* rather than *why the check is needed twice*.

vvlladd28 (commented):
## Review summary

Reviewed 16 changed files in [Open built-in IoT Hub items locally instead of installing duplicates](https://github.com/thingsboard/thingsboard/pull/16028). Left 8 comment(s) inline.

The overall shape is sound — one `builtIn` reader, one action-mode helper, and every install entry point gated — and the lookup semantics (404 means "not here", anything else means "unknown") are the right call. The one place that logic doesn't hold is a built-in item whose `dataDescriptor.fqn` is blank: `resolveLocalWidgetType` reports it as `missing` rather than `failed`, which routes the user to the confirm-and-install dialog — the exact duplicate this PR exists to prevent. The rest of the comments are about the three near-identical label maps, a `tb-iot-hub-built-in` class with no style rule anywhere, and a now-dead `isBuiltInItem` guard in `installDevice`.

## Additional findings

- **ui-ngx/src/app/core/http/widget.service.ts:191** — `getWidgetType` interpolates the fqn straight into the query string (`` `/api/widgetType?fqn=${fullFqn}` ``) with no `encodeURIComponent`. That was harmless while every caller passed a locally-derived fqn, but `IotHubBuiltInService` now feeds it `item.dataDescriptor.fqn`, a value that comes from the Hub rather than from this instance. An fqn containing `&`, `+`, `#` or a space would produce a malformed request and silently resolve to "not here" — which, per this PR's own logic, is exactly the case that offers the user an install.

---

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-built-in.service.ts
pinkevmladchy: Good catch on the request count — the `getWidgetTypeInfoById` hop is indeed unavoidable, and the descriptor payload on the lookups is wasted, but there is no fqn→info endpoint to collapse it into one round trip.

On the `tenant.` candidate: I kept it, because the interception you describe depends on the installed-items list being current. In the widget picker `installedWidgets` is loaded asynchronously (and reset whenever the mode or sort changes), and on the Hub pages installed items are reloaded per page — so there is a window where a widget installed into this tenant is not yet in that list. Without the second candidate the user gets "may have been deleted by the system administrator" for a widget that is sitting in their library, which is the failure mode this PR is about. The cost is one extra 404 on the miss path only, so I traded requests for not lying to the user. I noted the reasoning in the code comment.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-actions.service.ts
pinkevmladchy: Agreed on the return type — the `boolean` is gone. It now returns `IotHubBuiltInAction = 'handled' | 'install-requested'`, so callers branch on something self-describing and the four-line JSDoc is no longer needed.

On the two public helpers: `DashboardWidgetSelectComponent` cannot reuse `openBuiltInOrConfirmInstall` even with a richer outcome, because its terminal action is different. That component does not navigate anywhere — it needs the resolved `WidgetTypeInfo` to emit `widgetSelected` and place the widget on the dashboard, so it goes through `resolveLocalWidgetTypeInfo` instead of `openLocalComponent`. What it shares with the other callers is only the two user-facing decisions (ask to install / report that the check failed), which is exactly what those two methods are. Folding them back in would mean handing the service a callback for "what to do when found", which reads worse than the current split.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-built-in.service.ts
pinkevmladchy: You are right, and this was the one real hole — fixed in f535454. A blank `fqn` now returns `failed: true`, since nothing was asked of the server and existence is unknown rather than disproven, so it can no longer route into the confirm-and-install dialog.

Related, from the review summary: `getWidgetType` now runs the fqn through `encodeURIComponent`. Worth fixing here rather than later, since this PR is the first caller to feed it a value that comes from the Hub instead of from this instance.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-actions.service.ts
pinkevmladchy: Applied as suggested in f535454 — the guard is gone and `runInstall` calls `openDeviceInstallDialog(item)` directly, so the built-in decision lives only at the public entry points. Built-in content will only ever be widgets and SCADA symbols, so the guard was dead weight; removing it also closes the same loop on the add-item dialog path (`doAddItem` → `installDeviceItem` → `installDevice`).

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-utils.ts
pinkevmladchy: Done in 71d7950 — the three maps are now one exported function, `iotHubItemActionLabel(item, context: 'card' | 'detail' | 'add')`, with the tables private to the helper. Both `TbIotHubItemCardComponent` and `TbIotHubItemDetailDialogComponent` lost their `actionMode` member and their `actionLabel` is now a single line, so the naming asymmetry went away with the exports.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-item-card.component.ts
pinkevmladchy: Fair — trimmed in 71d7950. Removed the ones that restate the code (the `builtIn` getter and its "see isBuiltInItem", the three-line block over a one-line delegation, `lookupResult`, the rendering-path list that would go stale). Kept the rationale ones: why 404 is treated differently from other failures, why `first()` also stops on a failure, why there is no batch fqn API, and why the navigation promise has to be awaited.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-item-card.component.html
pinkevmladchy: Dropped the bindings in 71d7950 — all three of them. There is no visual designed for built-in cards yet, and the per-component getters already keep the state in one place, so an inert class hook in three templates was worse than none. If we add a visual cue for built-in cards later, the rule and the binding land together.

ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-widget-select.component.ts
pinkevmladchy: Unreachable indeed — removed in 71d7950. `addIotHubWidget` is gone and the detail-dialog result handler calls `installAndAddWidget` directly, with a comment on that call site noting that built-in items return before the dialog opens, so anything added from it installs.

vvlladd28 (approved):

vvlladd28 (approved):

PR #16046, opened 2026-08-19
Fixed CVE-2026-13149, CVE-2026-13311, CVE-2026-13676, CVE-2026-14257, CVE-2026-16221, CVE-2026-18446, CVE-2026-48815, CVE-2026-54466,   CVE-2026-55388, CVE-2026-55603, CVE-2026-59869, CVE-2026-59873, CVE-2026-59874, CVE-2026-59879, CVE-2026-59880, CVE-2026-67213,   CVE-2026-67214, CVE-2026-68945, CVE-2026-69151, CVE-2026-69152, CVE-2026-69192, CVE-2026-73646 and CWE-407

mtsymbarov-del:
## Pull Request description

Also was fixed: CVE-2025-71329 and CVE-2025-71330.
## General checklist

## Back-End feature checklist


ui-ngx/package.json
```diff
@@ -137,13 +137,9 @@
...
   "resolutions": {
-    "ace-builds": "1.43.6",
-    "tinymce": "6.8.6",
-    "@babel/core": "7.28.3",
+    "@babel/core": "7.29.7",
     "esbuild": "0.28.1",
-    "vite": "7.3.5",
-    "rollup": "4.59.0",
-    "jquery.terminal/**/form-data": ">=4.0.4",
+    "less": "4.9.0",
```
vvlladd28: Shrinking this block from eight entries to four is a real improvement, and I checked each removal against the new lockfile — `tinymce`, `ace-builds`, `vite` and `jquery.terminal/**/form-data` are all genuinely redundant now. Three things about what's left:

**`rollup` looks like the one removal that shouldn't have happened.** `@angular/build@20.3.34` pins `rollup "4.59.0"` exactly, but `vite@7.3.6` asks for `rollup "^4.43.0"`, and the latest published rollup 4.x is **4.62.4**. They currently share a single 4.59.0 entry only because yarn v1 reuses a satisfied lockfile entry — the moment anything re-resolves that range (lockfile regeneration, `yarn upgrade`, a new dep pulling rollup) the caret floats to 4.62.x and you get two copies of rollup *plus* two sets of `@rollup/rollup-*` native binaries. That's precisely the duplication this PR is trying to reduce. Note this is the same shape as `esbuild`, which was correctly kept: `@angular/build` and `build-angular` pin `esbuild "0.28.1"` while `vite` asks `^0.27.0 || ^0.28.0`, and latest 0.28.x is 0.28.2 — so that resolution is load-bearing as a dedupe guard, not a no-op. Was dropping `rollup` deliberate?

**`less: "4.9.0"` is load-bearing but the exact form is a downgrade trap.** `build-angular` pins `less "4.4.0"` exactly (the lockfile header now reads `less@4.4.0, less@4.9.0:`), so this override does real work — and 4.9.0 happens to be the current latest 4.x. But because it's an exact version, the next devkit release that ships `less` 4.10+ gets pinned back *down* to 4.9.0 silently. Since this is a security floor rather than a dedupe guard, `">=4.9.0"` expresses the intent better — the same form the removed `jquery.terminal/**/form-data: ">=4.0.4"` used, and consistent with the `js-beautify/**/minimatch: "^9.0.7"` entry right below. (Worth noting the blast radius is small either way: this project compiles `.scss`, so `less` only exists because `build-angular` hard-depends on it.)

**`@babel/core: "7.29.7"` is a no-op today.** `build-angular@20.3.34` pins that exact version itself, and 7.29.7 is also the latest published 7.x, so the override changes nothing right now — it only starts mattering once 7.29.8 ships and the `^7.23.9` consumer would otherwise split off.

The common thread: two different kinds of entry live in this block — dedupe guards that must track Angular's exact pins (`esbuild`, and arguably `rollup`) and security floors that should be open-ended ranges (`less`, `minimatch`). Could each get a one-line note saying which it is and which advisory it's for? Without that, the next person can't tell a load-bearing entry from dead weight, which is plausibly how `rollup` came to be deleted.

ui-ngx/package.json
```diff
@@ -51,6 +51,7 @@
     "flot.curvedlines": "https://github.com/MichaelZinsmaier/CurvedLines.git#master",
     "font-awesome": "^4.7.0",
     "html2canvas": "^1.4.1",
+    "hugerte": "^1.0.12",
```
vvlladd28: Both new packages take caret ranges (`hugerte: ^1.0.12` here, `@hugerte/hugerte-angular: ^1.1.1` above), which is looser than what they replace — TinyMCE was `~6.8.6` *and* additionally locked by a `resolutions` entry. That matters more than usual because the app reaches into the editor's internal DOM in five places that are not part of any documented API: `.tox-hugerte` in `styles.scss`, `.tox.tox-hugerte-aux`, `.tox-tiered-menu .tox-menu`, `.tox-collection__item-label`, and `#hugerte` inside the iframe. A caret lets any 1.x minor of a young fork rearrange those on a routine `yarn install`, with no test to catch it. Given this file already pins several UI-critical deps tightly (`ace-builds` exact, `@svgdotjs/svg.js` exact, `angular-gridster2` with `~`), a tilde or exact pin would fit the existing convention better. For what it's worth both are at the latest published versions (`hugerte` 1.0.12, wrapper 1.1.1), so pinning costs nothing today.

ui-ngx/package.json
```diff
@@ -37,8 +37,8 @@
     "@ngx-translate/core": "^17.0.0",
     "@svgdotjs/svg.filter.js": "^3.0.9",
     "@svgdotjs/svg.js": "3.2.4",
+    "@hugerte/hugerte-angular": "^1.1.1",
```
vvlladd28: This landed between `@svgdotjs/svg.js` and `@svgdotjs/svg.panzoom.js` — presumably where the removed `@tinymce/tinymce-angular` line sat. The rest of the block is alphabetical, and the next `yarn add` will re-sort it and produce an unrelated diff line. `@h…` belongs up near `@flowjs`/`@iplab`.

ui-ngx/package.json
```diff
@@ -87,21 +88,20 @@
...
-    "@angular/build": "20.3.25",
-    "@angular/cli": "20.3.25",
-    "@angular/compiler-cli": "20.3.25",
-    "@angular/language-service": "20.3.25",
+    "@angular-devkit/build-angular": "20.3.34",
+    "@angular-devkit/core": "20.3.34",
+    "@angular-devkit/schematics": "20.3.34",
+    "@angular/build": "20.3.34",
+    "@angular/cli": "20.3.34",
+    "@angular/compiler-cli": "20.3.28",
```
vvlladd28, line 103: Just to record why this block now carries three different patch numbers, since it looks like an oversight and isn't: `@angular/core`/`animations`/`common`/`compiler`/`forms`/`platform-browser`/`router` plus `compiler-cli` and `language-service` all ship from angular/angular and are at 20.3.28; `cli`/`devkit`/`build`/`schematics` ship from angular-cli and are at 20.3.34; `cdk`/`material` track the components release at 20.2.14. I checked npm — 20.3.28 and 20.3.34 are each the latest published 20.x of their own train, so nothing is lagging. The risk is that the natural instinct on the next bump is to align all of them to one number, which would break the `compiler-cli` must-match-`core` invariant. A short comment here (or a line in the dependency notes) recording the three-train rule would make that upgrade a lot safer.

ui-ngx/angular.json
```diff
@@ -74,8 +74,8 @@
               },
               {
                 "glob": "**/*",
-                "input": "node_modules/tinymce",
-                "output": "/assets/tinymce/"
+                "input": "node_modules/hugerte",
```
vvlladd28, line 77: The `**/*` glob copies the entire package into `dist/assets/hugerte/`. HugeRTE 1.x is a fork of the TinyMCE 6 tree this replaces, so expect comparable weight — plugins, all four skins, both themes, and *both* `hugerte.js` and `hugerte.min.js`. The app sets `suffix: '.min'` and uses seven plugins total across the three hosts (`link`, `table`, `image`, `lists`, `fullscreen`, `code`, `autoresize`), so most of those bytes ship in the Docker image and are served-but-never-requested.

Carrying the glob over unchanged is defensible for a security PR, and I wouldn't block on it — but the migration is the cheapest moment to narrow it, since you're already re-verifying the editor works. The trade-off is that a narrower glob has to be maintained whenever a plugin is added to any editor config, which argues for doing it together with pulling the shared options into one place (see the comment on `notification-template-configuration.component.ts`) so the plugin list has a single home.

ui-ngx/src/app/modules/home/pages/notification/template/configuration/notification-template-configuration.component.ts
```diff
@@ -95,10 +95,9 @@ export class NotificationTemplateConfigurationComponent implements OnDestroy, Co
     height: 400,
     autofocus: false,
     branding: false,
-    promotion: false,
     setup: (editor) => {
       editor.on('PostRender', function() {
-        const container = document.querySelector('.tox.tox-tinymce-aux');
+        const container = document.querySelector('.tox.tox-hugerte-aux');
```
vvlladd28: This whole `setup` callback — the `PostRender` handler, the ~30-line CSS template literal, the `container.prepend(styleSheet)` — is byte-for-byte identical to the one in `editor-panel.component.ts`, and `base_url`/`suffix`/`height`/`autofocus`/`branding` are shared by all three editor hosts. This migration had to touch the same string in three files, which is exactly the cost that duplication imposes.

Could the common part live in one place — say a `defaultHugeRteOptions(overrides)` factory under `@shared/models/`, following the precedent of `@shared/models/ace/` — so each component only declares its own `plugins`/`menubar`/`toolbar`? Two other things get easier once it's centralized: the injected CSS has two `media screen and (min-width: …)` blocks missing the leading `@` (so those rules have never applied — see the review body), and moving this static CSS into `styles.scss` next to the existing `.tox-hugerte` rules would drop the imperative `document.createElement('style')` injection entirely and keep all `.tox-*` coupling in one file.

ui-ngx/src/app/modules/home/pages/mobile/common/editor-panel.component.ts
```diff
@@ -46,7 +46,7 @@ export class EditorPanelComponent implements OnInit {
   editorControl: FormControl<string>;
 
   tinyMceOptions: Partial<EditorOptions> = {
```
vvlladd28: The field is still `tinyMceOptions` after the move to HugeRTE — here, in the other two components, and in the three templates binding `[init]="tinyMceOptions"`. Now that TinyMCE is gone from `package.json` and the lockfile entirely, this name is the last thing in the repo pointing at a dependency the project no longer has, and it'll read as a mistake to whoever opens these files next. The rename is mechanical: three declarations, four usages in `ngOnInit`, three template bindings.

ui-ngx/src/app/modules/home/pages/notification/sent/sent-notification-dialog.componet.ts
```diff
@@ -79,7 +79,7 @@ export class SentNotificationDialogComponent extends
   showRefresh = false;
 
   tinyMceOptions: Record<string, any> = {
```
vvlladd28: The other two editor hosts type this as `Partial<EditorOptions>` while this one is `Record<string, any>`, so nothing in this object is type-checked — including the `promotion` key the migration had to find and remove by hand. `hugerte` ships `hugerte.d.ts` exporting `EditorOptions`, so switching to `Partial<EditorOptions>` (same import as the siblings) would have let the compiler point at options HugeRTE dropped instead of leaving them to be caught by eye. It also lines up with the project's rule against `any`.

vvlladd28 (commented):
## Review summary

Reviewed 12 changed files in [Fixed CVE-2026-13149 … and CWE-407](https://github.com/thingsboard/thingsboard/pull/16046). Left 11 comments inline.

The security-relevant parts check out. I verified independently:

- **Angular stays inside the 20 LTS line.** All 21 `@angular*`/`@angular-devkit*` entries resolved in `ui-ngx/yarn.lock` are 20.x. Framework is `20.3.28` and CLI/devkit/build is `20.3.34`, which are each the *latest published 20.x* of their respective release trains — so the skew is correct, not a mistake.
- **Both renamed patch files still apply.** `@angular+build+20.3.34.patch` was renamed with no content change; I diffed `angular-compilation.js` and `compiler-plugin.js` between `@angular/build` 20.3.25 and 20.3.34 and they are byte-identical, so the rename is safe. `@angular+core+20.3.28.patch` applies cleanly to `@angular/core@20.3.28` (`git apply --check` passes).
- **Every removed `resolutions` entry except one is genuinely redundant.** `tinymce` is gone from the tree entirely; `form-data` is gone too (`jquery.terminal` 2.45.2 → 2.46.2 dropped it), so `jquery.terminal/**/form-data` was correctly dropped; `ace-builds` is already pinned exactly in `dependencies` and has no other consumer; `vite` now has a single exact consumer (`@angular/build` → `vite "7.3.6"`). The `rollup` removal is the one I'd push back on — see the comment on the `resolutions` block.
- **The HugeRTE swap is API- and DOM-correct.** `@hugerte/hugerte-angular@1.1.1` exports `EditorModule` and `HUGERTE_SCRIPT_SRC` and keeps the `editor` selector; `hugerte@1.0.12` ships `hugerte.min.js` and `hugerte.d.ts` (exporting `EditorOptions`) at the package root; the dist genuinely uses `tox-hugerte` / `tox-hugerte-aux` (no `tox-tinymce` remains anywhere in the package); `body_id` defaults to `'hugerte'`, so `$('#hugerte', …)` is right; and `promotion` is no longer a registered option, so removing it was correct. Both new packages are at the latest published versions. A repo-wide grep found no remaining runtime references to TinyMCE — only the `tinyMceOptions` field name.
- **Lockfile consistency across the three files is good.** `brace-expansion`, `js-yaml` and `tar` were bumped consistently in `ui-ngx`, `msa/js-executor` and `msa/web-ui`, and neither `msa` lockfile gained a duplicate.

On the "shrink the version zoo" goal, the honest measurement: in `ui-ngx/yarn.lock`, packages resolved to more than one version *inside the same major* went **30 → 29** (total packages 1337 → 1351). `bytes` and `js-yaml` deduped, `negotiator` newly split. It can't do much better without new overrides — I traced the residue and nearly all of it is `@angular-devkit/build-angular` / `@angular/cli` pinning exact transitive versions (`semver "7.7.2"`, `terser "5.43.1"`, `autoprefixer "10.4.21"`, `@babel/generator "7.28.3"`, `webpack-dev-middleware "7.4.2"`, `tinyglobby "0.2.14"`, `watchpack "2.4.4"`, `npm-package-arg "13.0.0"`, `zod "4.1.13"`, `@inquirer/confirm "5.1.14"`) against caret ranges elsewhere in the tree. Collapsing those requires `resolutions` that override Angular's own pins — the opposite of this PR's direction — and none of them reach the browser bundle. Worth stating that in the PR description so it isn't re-litigated next security pass.

## Additional findings

- **`ui-ngx/src/app/modules/home/pages/mobile/common/editor-panel.component.ts:71,75`** and **`.../notification-template-configuration.component.ts:109,113`** — in the injected `styleSheet.innerText`, two blocks read `media screen and (min-width: 511px) and (max-width: 548px) {` and `media screen and (min-width: 549px) …` with the leading `@` missing. Those rules have never applied, so the intended 511–599px menu-width clamping silently doesn't happen. Pre-existing, but the migration copied the block across unchanged, so this is a cheap fix to fold in.
- **`ui-ngx/yarn.lock`** — `negotiator` is newly split within major 0: `accepts@~1.3.8` pins `negotiator "0.6.3"` exactly while the `compression` 1.7.4 → 1.8.1 bump in this PR brought `negotiator "~0.6.4"` → 0.6.4. That is the one new same-major duplicate this PR introduces. It's dev-server-only and harmless, but since duplicate-count is an explicit goal here it's worth knowing it went the wrong way by one.
- **`ui-ngx/yarn.lock`** — `@rollup/rollup-linux-x64-gnu` is duplicated at 4.46.2 and 4.59.0. The older copy comes from `ngx-translate-messageformat-compiler@7.2.0`, which carries the platform binary as a direct dependency rather than letting rollup's own optional-deps resolve it. That's a duplicated *native* binary in `node_modules`. Pre-existing and not caused by this PR, but it's the only non-Angular duplicate worth chasing — is there a newer `ngx-translate-messageformat-compiler` that drops it?
- **Verification evidence.** `ui-ngx` has no `test` target in `angular.json` and no test runner in `devDependencies`, so I'm not going to ask for unit tests that can't run. But swapping the rich-text engine for a fork — touching a DI token, an `angular.json` asset entry, and four undocumented DOM selectors — is landing with nothing recorded. A reasonable minimum for the description: a manual smoke check of all three editor hosts (mobile release-notes panel including the `disabled` variant that hides toolbar/menubar/statusbar; the notification-template editor's narrow-width menu overflow, which is exactly what the injected `.tox-tiered-menu` stylesheet and `.tox.tox-hugerte-aux` lookup exist for; and the sent-notification preview staying non-editable), plus confirmation that `dist/assets/hugerte/hugerte.min.js` is emitted and that both patches reported as applied during install.

---

ui-ngx/src/app/modules/home/pages/notification/sent/sent-notification-dialog.componet.ts
```diff
@@ -90,11 +90,10 @@ export class SentNotificationDialogComponent extends
     height: 400,
     autofocus: false,
     branding: false,
-    promotion: false,
     setup: (ed) => {
       ed.on('PreInit', () => {
         const document = $(ed.iframeElement.contentDocument);
-        const body = $('#tinymce', document);
+        const body = $('#hugerte', document);
```
vvlladd28: I verified this is correct — HugeRTE registers `body_id` with `default: 'hugerte'`, so the iframe body really does get `id="hugerte"`. But the selector is now coupled to an undocumented default of a young fork, in the one place where getting it wrong fails silently *and* unsafely: if the lookup misses, `contenteditable`/`pointerEvents`/`userSelect` never get applied and the read-only preview becomes editable. Setting `body_id: 'hugerte'` explicitly in the options above would put the selector and the value it depends on next to each other and make them impossible to drift apart.

ui-ngx/src/app/shared/shared.module.ts
```diff
@@ -258,8 +258,8 @@ export function MarkedOptionsFactory(markedOptionsService: MarkedOptionsService)
       useValue: Flow
     },
     {
-      provide: TINYMCE_SCRIPT_SRC,
-      useValue: 'assets/tinymce/tinymce.min.js'
+      provide: HUGERTE_SCRIPT_SRC,
+      useValue: 'assets/hugerte/hugerte.min.js'
```
vvlladd28: The editor's asset location is now spelled out in five places with two conventions: this token uses document-relative `assets/hugerte/…`, the three components use root-absolute `base_url: '/assets/hugerte'`, and `angular.json` declares the output path a third way. That's unchanged behaviour from the TinyMCE setup, so nothing regresses — but the mix is a footgun if the UI is ever served from a sub-path or behind a rewriting proxy, and a future rename has to be found by grep. Deriving both this value and `base_url` from one exported `HUGERTE_ASSETS_PATH` constant would make the path a declared contract instead of a repeated string. This is also the outward-facing edge of the change: anything downstream that referenced `/assets/tinymce` has to move, so it's worth a release-note line — I grepped and found no remaining references inside this repo, but consumers outside it can't be checked from here.

ui-ngx/patches/@angular+core+20.3.28.patch
```diff
@@ -1,8 +1,8 @@
 diff --git a/node_modules/@angular/core/fesm2022/debug_node.mjs b/node_modules/@angular/core/fesm2022/debug_node.mjs
-index be43067..079f100 100755
+index be00d4e..8d7ed1a 100755
```
vvlladd28, line 2: I confirmed both patches are fine: this one applies cleanly to `@angular/core@20.3.28`, and the `@angular/build` patch was renamed without content changes because `angular-compilation.js` and `compiler-plugin.js` are byte-identical between 20.3.25 and 20.3.34.

Two suggestions to keep it that way. `"prepare": "patch-package"` runs without `--error-on-fail`, so a filename that no longer matches the resolved version is just a warning in a noisy install log — the patch silently doesn't apply and the build quietly loses the `jit: true` / `ngDevMode` behaviour these exist to force. Now that there are two independent Angular version numbers to keep in sync, `--error-on-fail` would turn that into a hard stop. Second, a comment header in each patch saying *why* it exists would help — reconstructing "forces JIT and short-circuits `sideEffects` resolution" and "disables the dev-mode guard around the component-host assertion" from the diffs alone is guesswork on the next rebase.

vvlladd28 (dismissed):

vvlladd28 (approved):

PR #16047, opened 2026-08-19
Feature/tb monitoring api key auth

pon0marev:
## Pull Request description

Introduces the ability to use an API token as an alternative authentication method. This is required for users with two-factor authentication (2FA) enabled and for accounts subject to frequent password resets.

Additionally, an optional notification system has been implemented to alert users when a token is about to expire. Avoids sudden service interruptions through proactive expiration alerts.

Key Changes:
* API Token Support: Added an alternative sign-in method using API tokens instead of passwords.
* Expiration Notifications: Added an option to receive alerts prior to token expiration.

## General checklist

## Back-End feature checklist

### DIFF
--- a/monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
@@ -31,20 +31,20 @@ import java.util.Optional;
 @Component
 public class TbClient extends RestClient {
 
+    // separate from RestClient.AuthType: this also drives the WS auth path (see WsClient#authenticate), which RestClient doesn't know about
     public enum AuthMode {LOGIN, API_KEY}
 
     private final AuthMode authMode;
     private final String apiKey;
-
-    @Value("${monitoring.rest.username:}")
-    private String username;
-    @Value("${monitoring.rest.password:}")
-    private String password;
+    private final String username;
+    private final String password;
 
     public TbClient(@Value("${monitoring.rest.base_url}") String baseUrl,
                     @Value("${monitoring.rest.request_timeout_ms}") int requestTimeoutMs,
                     @Value("${monitoring.rest.auth_mode:LOGIN}") AuthMode authMode,
-                    @Value("${monitoring.rest.api_key:}") String apiKey) {
+                    @Value("${monitoring.rest.api_key:}") String apiKey,
+                    @Value("${monitoring.rest.username:}") String username,
+                    @Value("${monitoring.rest.password:}") String password) {
         super(new RestTemplateBuilder()
                 .connectTimeout(Duration.ofMillis(requestTimeoutMs))
                 .readTimeout(Duration.ofMillis(requestTimeoutMs))
@@ -53,26 +53,30 @@ public class TbClient extends RestClient {
                 authMode == AuthMode.API_KEY ? apiKey : null);
         this.authMode = authMode;
         this.apiKey = apiKey;
+        this.username = username;
+        this.password = password;
         if (authMode == AuthMode.API_KEY && StringUtils.isBlank(apiKey)) {
             throw new IllegalStateException("monitoring.rest.api_key must be set when monitoring.rest.auth_mode is API_KEY");
+        }
+        if (authMode == AuthMode.LOGIN && (StringUtils.isBlank(username) || StringUtils.isBlank(password))) {
+            throw new IllegalStateException("monitoring.rest.username and monitoring.rest.password must be set when monitoring.rest.auth_mode is LOGIN");
         }
         log.info("Starting TbClient with auth mode: {}", authMode);
     }
 
     @PostConstruct
     private void init() {
-        if (authMode == AuthMode.LOGIN && (StringUtils.isBlank(username) || StringUtils.isBlank(password))) {
-            throw new IllegalStateException("monitoring.rest.username and monitoring.rest.password must be set when monitoring.rest.auth_mode is LOGIN");
-        }
-        logIn();
+        getWsCredential();
     }
 
     public AuthMode getAuthMode() {
         return authMode;
     }
 
-    // API_KEY has no login step - proves the key works via getUser() and returns it as-is
-    public String logIn() {
+    // returns whatever WsClient#authenticate needs to hand the server next: a fresh JWT in LOGIN
+    // mode, or the api key itself in API_KEY mode - which has no login step, so this just proves
+    // the key works via getUser() and returns it as-is
+    public String getWsCredential() {
         if (authMode == AuthMode.API_KEY) {
             Optional<User> user = getUser();
             if (user.isEmpty()) {
--- a/monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
@@ -20,23 +20,36 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.stereotype.Component;
 import org.thingsboard.rest.client.RestClient;
+import org.thingsboard.server.common.data.User;
 
 import java.time.Duration;
+import java.util.Optional;
 
 @Component
 public class TbClient extends RestClient {
 
-    @Value("${monitoring.rest.username}")
+    public enum AuthMode {LOGIN, API_KEY}
+
+    private final AuthMode authMode;
+    private final String apiKey;
+
+    @Value("${monitoring.rest.username:}")
     private String username;
-    @Value("${monitoring.rest.password}")
+    @Value("${monitoring.rest.password:}")
     private String password;
 
     public TbClient(@Value("${monitoring.rest.base_url}") String baseUrl,
-                    @Value("${monitoring.rest.request_timeout_ms}") int requestTimeoutMs) {
+                    @Value("${monitoring.rest.request_timeout_ms}") int requestTimeoutMs,
+                    @Value("${monitoring.rest.auth_mode:LOGIN}") AuthMode authMode,
+                    @Value("${monitoring.rest.api_key:}") String apiKey) {
         super(new RestTemplateBuilder()
                 .connectTimeout(Duration.ofMillis(requestTimeoutMs))
                 .readTimeout(Duration.ofMillis(requestTimeoutMs))
-                .build(), baseUrl);
+                .build(), baseUrl,
+                authMode == AuthMode.API_KEY ? AuthType.API_KEY : AuthType.JWT,
+                authMode == AuthMode.API_KEY ? apiKey : null);
+        this.authMode = authMode;
+        this.apiKey = apiKey;
     }
 
     @PostConstruct
@@ -44,7 +57,21 @@ public class TbClient extends RestClient {
         logIn();
     }
 
+    public AuthMode getAuthMode() {
+        return authMode;
+    }
+
+    // for LOGIN, performs the real username/password login and returns the resulting JWT.
+    // for API_KEY, there is no login step - proves the key still works via a lightweight
+    // authenticated call instead, and returns the static key itself (nothing else to return).
     public String logIn() {
+        if (authMode == AuthMode.API_KEY) {
+            Optional<User> user = getUser();
+            if (user.isEmpty()) {
+                throw new IllegalStateException("API key authentication failed - no user returned");
+            }
+            return apiKey;
+        }
         login(username, password);
         return getToken();
     }
--- a/monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
@@ -71,9 +71,7 @@ public class TbClient extends RestClient {
         return authMode;
     }
 
-    // for LOGIN, performs the real username/password login and returns the resulting JWT.
-    // for API_KEY, there is no login step - proves the key still works via a lightweight
-    // authenticated call instead, and returns the static key itself (nothing else to return).
+    // API_KEY has no login step - proves the key works via getUser() and returns it as-is
     public String logIn() {
         if (authMode == AuthMode.API_KEY) {
             Optional<User> user = getUser();
--- a/monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
@@ -16,6 +16,8 @@
 package org.thingsboard.monitoring.client;
 
 import jakarta.annotation.PostConstruct;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.stereotype.Component;
@@ -25,6 +27,7 @@ import org.thingsboard.server.common.data.User;
 import java.time.Duration;
 import java.util.Optional;
 
+@Slf4j
 @Component
 public class TbClient extends RestClient {
 
@@ -50,10 +53,17 @@ public class TbClient extends RestClient {
                 authMode == AuthMode.API_KEY ? apiKey : null);
         this.authMode = authMode;
         this.apiKey = apiKey;
+        if (authMode == AuthMode.API_KEY && StringUtils.isBlank(apiKey)) {
+            throw new IllegalStateException("monitoring.rest.api_key must be set when monitoring.rest.auth_mode is API_KEY");
+        }
+        log.info("Starting TbClient with auth mode: {}", authMode);
     }
 
     @PostConstruct
     private void init() {
+        if (authMode == AuthMode.LOGIN && (StringUtils.isBlank(username) || StringUtils.isBlank(password))) {
+            throw new IllegalStateException("monitoring.rest.username and monitoring.rest.password must be set when monitoring.rest.auth_mode is LOGIN");
+        }
         logIn();
     }
 
--- a/monitoring/src/main/java/org/thingsboard/monitoring/client/WsClient.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/client/WsClient.java
@@ -128,7 +128,11 @@ public class WsClient extends WebSocketClient implements AutoCloseable {
         } finally {
             updateLock.unlock();
         }
-        super.send(text);
+        try {
+            super.send(text);
+        } catch (RuntimeException e) {
+            throw new IllegalStateException("Failed to send WS message" + closeInfo(), e);
+        }
     }
 
     public WsClient subscribeForTelemetry(List<UUID> devices, List<String> keys) {
@@ -149,11 +153,7 @@ public class WsClient extends WebSocketClient implements AutoCloseable {
         CmdsWrapper wrapper = new CmdsWrapper();
         wrapper.setCmds(List.of(cmd));
 
-        try {
-            send(JacksonUtil.toString(wrapper));
-        } catch (RuntimeException e) {
-            throw new IllegalStateException("Failed to send WS subscribe command" + closeInfo(), e);
-        }
+        send(JacksonUtil.toString(wrapper));
         return this;
     }
 
@@ -204,7 +204,7 @@ public class WsClient extends WebSocketClient implements AutoCloseable {
         if (!closed) {
             return "";
         }
-        return " - WebSocket closed before subscribe could complete (code=" + closeCode + ", reason=" + closeReason + ")";
+        return " - WebSocket closed (code=" + closeCode + ", reason=" + closeReason + ")";
     }
 
     private List<JsonNode> getLastMsgs() {
--- a/monitoring/src/main/java/org/thingsboard/monitoring/client/WsClient.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/client/WsClient.java
@@ -22,6 +22,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.java_websocket.client.WebSocketClient;
 import org.java_websocket.handshake.ServerHandshake;
 import org.thingsboard.common.util.JacksonUtil;
+import org.thingsboard.monitoring.data.cmd.AuthCmd;
 import org.thingsboard.monitoring.data.cmd.CmdsWrapper;
 import org.thingsboard.monitoring.data.cmd.EntityDataCmd;
 import org.thingsboard.monitoring.data.cmd.EntityDataUpdate;
@@ -138,7 +139,24 @@ public class WsClient extends WebSocketClient implements AutoCloseable {
         cmd.setLatestCmd(latestCmd);
 
         CmdsWrapper wrapper = new CmdsWrapper();
-        wrapper.setEntityDataCmds(List.of(cmd));
+        wrapper.setCmds(List.of(cmd));
+
+        send(JacksonUtil.toString(wrapper));
+        return this;
+    }
+
+    // sends the auth command as the connection's first message - required before any other
+    // command on the GENERAL WS endpoint, since it never authenticates via the ?token= query
+    // param the way the deprecated /api/ws/plugins/telemetry endpoint did
+    public WsClient authenticate(TbClient.AuthMode authMode, String tokenOrApiKey) {
+        AuthCmd authCmd = new AuthCmd();
+        if (authMode == TbClient.AuthMode.API_KEY) {
+            authCmd.setApiKey(tokenOrApiKey);
+        } else {
+            authCmd.setToken(tokenOrApiKey);
+        }
+        CmdsWrapper wrapper = new CmdsWrapper();
+        wrapper.setAuthCmd(authCmd);
         send(JacksonUtil.toString(wrapper));
         return this;
     }
--- a/monitoring/src/main/java/org/thingsboard/monitoring/client/WsClient.java
+++ b/monitoring/src/main/java/org/thingsboard/monitoring/client/WsClient.java
@@ -60,9 +60,7 @@ public class WsClient extends WebSocketClient implements AutoCloseable {
 
     private final long requestTimeoutMs;
 
-    // captured from onClose(), so a later send()/subscribe failure (e.g. WebsocketNotConnectedException)
-    // can report *why* the socket went away instead of just that it's no longer connected -
-    // most commonly, the server rejecting the auth message we send right after connecting
+    // captured from onClose() so a later subscribe failure can report why the socket went away
     private volatile boolean closed;
     private volatile int closeCode;
     private volatile String closeReason;
@@ -159,9 +157,7 @@ public class WsClient extends WebSocketClient implements AutoCloseable {
         return this;
     }
 
-    // sends the auth command as the connection's first message - required before any other
-    // command on the GENERAL WS endpoint, since it never authenticates via the ?token= query
-    // param the way the deprecated /api/ws/plugins/telemetry endpoint did
+    // must be the connection's first message - the GENERAL endpoint has no ?token= query param
     public WsClient authenticate(TbClient.AuthMode authMode, String tokenOrApiKey) {
         AuthCmd authCmd = new AuthCmd();
         if (authMode == TbClient.AuthMode.API_KEY) {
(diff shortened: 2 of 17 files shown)


smatvienko-tb (commented):
## Review summary

Reviewed 14 changed files in [Feature/tb monitoring api key auth](https://github.com/thingsboard/thingsboard/pull/16047). Left 22 comment(s) inline.

I verified the wire contracts against the server side in this same tree: `CmdsWrapper.authCmd`/`cmds` and the `"type":"ENTITY_DATA"` discriminator match `WsCommandsWrapper`, `AuthCmd.apiKey` matches the field `TbWebSocketHandler` reads, `/api/ws` is an exact-match path for `WebSocketSessionType.GENERAL`, and `/api/auth/user` plus `/api/apiKeys/{userId}` are both reachable for an API-key-authenticated tenant admin. I also ran the four new test classes locally — 23 tests, all green.

The substantive points are concentrated in `ApiKeyExpiryMonitoringService`: the `EXPIRED` alert branch looks unreachable in practice, and the `neverExpiresConfirmed` latch can permanently disable the check based on a key that isn't necessarily the one in use. The rest are design, naming, and coverage observations.

---

monitoring/src/main/java/org/thingsboard/monitoring/service/ApiKeyExpiryMonitoringService.java
```diff
@@ -0,0 +1,113 @@
...
+
+            ApiKeyInfo apiKeyInfo = enabledMatches.get(0);
+            if (apiKeyInfo.getExpirationTime() == 0) {
+                neverExpiresConfirmed = true;
+                log.info("API key '{}' has no expiration - expiry monitoring disabled for the rest of this process's lifetime", apiKeyDescription);
+                return;
+            }
+
+            long now = System.currentTimeMillis();
+            boolean expired = apiKeyInfo.getExpirationTime() <= now;
```
smatvienko-tb: I don't think the `expired == true` path can ever fire. `ApiKeyAuthenticationProvider` throws `CredentialsExpiredException("API key is expired")`, so the moment the key actually expires, every call this method makes — `getUser()` at line 67 and `getUserApiKeys()` at line 74 — comes back 401 and lands in the `catch` at line 108 as a WARN. Reaching this line with `expirationTime <= now` requires the server to still accept an expired key, which it doesn't. So the `:rotating_light: ... has EXPIRED - rotate it now` notification (and `alreadyExpiredKey_sendsNotificationWithExpiredFlag`, which only passes because the REST layer is mocked) covers a state that production won't reach; what operators will actually see is a `LOGIN` service failure from `runChecks()`.

If you want a real expired-key signal, one option is to remember the resolved `expirationTime` in a field and, when a check fails with 401/`CredentialsExpiredException`, send the EXPIRED notification if that remembered timestamp has passed. Otherwise it may be worth dropping the branch and leaning on the LOGIN alarm, so the code doesn't imply coverage it doesn't have.

Separately, `ApiKeyInfo` already ships `isExpired()` with exactly these semantics (including the `expirationTime == 0` case handled above), so `apiKeyInfo.isExpired()` would keep the two definitions of "expired" from drifting.

monitoring/src/main/java/org/thingsboard/monitoring/service/ApiKeyExpiryMonitoringService.java
```diff
@@ -0,0 +1,113 @@
...
+    private final TbClient tbClient;
+    private final NotificationService notificationService;
+
+    @Value("${monitoring.rest.api_key_description:}")
+    private String apiKeyDescription;
+    @Value("${monitoring.rest.api_key_expiry_warning_days:7}")
+    private int warningDays;
+
+    // expirationTime specifically can't be changed after creation, so this can never go stale
+    private volatile boolean neverExpiresConfirmed = false;
```
smatvienko-tb: This latch worries me more than the comment suggests. The justification holds per key — `expirationTime` is immutable after creation — but the flag isn't tied to a key: it's tied to whichever `ApiKeyInfo` line 95 happened to pick. If an operator rotates by creating a new key with the same description and the old never-expiring one is still around and enabled, a single unlucky ordering latches `neverExpiresConfirmed = true` and expiry monitoring is off for the rest of the process lifetime — for the key actually in use, silently, with only a one-time INFO line as evidence.

Given the whole thing buys you one REST call per 24 hours, I'd drop the flag and just do the lookup each time. If the INFO log is the thing you wanted to keep quiet, guard only the logging (e.g. remember the `ApiKeyId` you already logged for).

monitoring/src/main/java/org/thingsboard/monitoring/service/ApiKeyExpiryMonitoringService.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+            // descriptions aren't unique - if rotation left an old key behind with the same
+            // description, prefer enabled ones, but the choice among several enabled matches is
+            // still arbitrary - warn loudly so an operator notices instead of silently tracking
+            // the wrong key
+            if (enabledMatches.size() > 1) {
+                log.warn("Found {} enabled API keys with description '{}' - descriptions should be unique; using an arbitrary one for expiry checks", enabledMatches.size(), apiKeyDescription);
+            }
+
+            ApiKeyInfo apiKeyInfo = enabledMatches.get(0);
```
smatvienko-tb: Picking element 0 makes the outcome depend on whatever order the server's `ORDER BY` happens to return, and `multipleEnabledMatches_picksOneAndStillFunctions` then pins that undefined ordering as expected behaviour — its own comment says "if the service picked the second (soon-expiring) one instead, this would incorrectly send a notification", which is really a coin flip.

Since the point of the check is to warn early, picking the soonest expiration deterministically would be both the conservative choice and a stable thing to assert on — something like `min(comparingLong(ApiKeyInfo::getExpirationTime))` over the enabled matches, treating `0` as "never". A spurious early warning is much cheaper than a missed one.

monitoring/src/main/java/org/thingsboard/monitoring/service/ApiKeyExpiryMonitoringService.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Value("${monitoring.rest.api_key_description:}")
+    private String apiKeyDescription;
+    @Value("${monitoring.rest.api_key_expiry_warning_days:7}")
+    private int warningDays;
+
+    // expirationTime specifically can't be changed after creation, so this can never go stale
+    private volatile boolean neverExpiresConfirmed = false;
+
+    @Scheduled(initialDelay = 60_000, fixedDelay = 86_400_000)
```
smatvienko-tb: Two timings hardcoded into the annotation, in a module whose convention is env-overridable timings in `tb-monitoring.yml` (`monitoring_rate_ms`, `check_timeout_ms`, `request_timeout_ms`, and the new `api_key_expiry_warning_days` right next to it). `initialDelayString`/`fixedDelayString` with something like `${monitoring.rest.api_key_expiry_check_interval_ms:86400000}` would match the surrounding style and let an operator tighten the cadence — it would also make the "repeats every day" claim in the YAML comment verifiable instead of buried in a constant.

monitoring/src/main/java/org/thingsboard/monitoring/service/ApiKeyExpiryMonitoringService.java
```diff
@@ -0,0 +1,113 @@
...
+    void checkApiKeyExpiry() {
```
smatvienko-tb: This method is doing quite a few distinct jobs: three early-return gates, resolving the current user, listing and filtering keys by description, deciding which duplicate wins, computing the expiry window, and notifying — all inside one ~45-line `try` with nested conditionals just for the logging. Pulling the middle out as `private Optional<ApiKeyInfo> resolveMonitoredApiKey()` would leave this as gates → expiry math → notify, and would let the duplicate-resolution rules be tested directly rather than only through the notification side effect.

monitoring/src/main/java/org/thingsboard/monitoring/service/ApiKeyExpiryMonitoringService.java
```diff
@@ -0,0 +1,113 @@
...
+        if (tbClient.getAuthMode() != TbClient.AuthMode.API_KEY) {
+            return;
+        }
+        if (StringUtils.isBlank(apiKeyDescription)) {
+            log.warn("monitoring.rest.api_key_description is not set - API key expiry monitoring is disabled");
```
smatvienko-tb, line 56: This fires from a scheduled task, so anyone running API_KEY mode without a description gets the identical WARN every single day forever — the kind of line people add a log filter for. Validating it once at startup, where the other config validation already lives in `TbClient`, would surface it once and loudly, at the moment the operator can act on it.

monitoring/src/main/java/org/thingsboard/monitoring/client/WsClient.java
```diff
@@ -138,7 +147,26 @@ public WsClient subscribeForTelemetry(List<UUID> devices, List<String> keys) {
...
+    public WsClient authenticate(TbClient.AuthMode authMode, String tokenOrApiKey) {
+        AuthCmd authCmd = new AuthCmd();
+        if (authMode == TbClient.AuthMode.API_KEY) {
+            authCmd.setApiKey(tokenOrApiKey);
+        } else {
+            authCmd.setToken(tokenOrApiKey);
+        }
+        CmdsWrapper wrapper = new CmdsWrapper();
+        wrapper.setAuthCmd(authCmd);
         send(JacksonUtil.toString(wrapper));
```
smatvienko-tb, line 170: This `send()` is unguarded while the sibling one in `subscribeForTelemetry` (line 153) is wrapped to attach `closeInfo()`. If the socket dies between `connectBlocking` and here — server rejecting the upgrade path, immediate close — this throws a bare `WebsocketNotConnectedException` out of `WsClientFactory.createClient`, and the close code/reason that `closeInfo()` exists to surface is exactly what gets lost.

Since the enrichment is entirely about `send()` plus the close state, pushing the try/catch down into the overridden `send()` would give every command the same diagnostics and remove the per-method boilerplate.

Worth noting the related behaviour while you're here: auth is fire-and-forget and never confirmed, so a rejected token or API key surfaces as a `WS_SUBSCRIBE` failure rather than an auth failure. In API_KEY mode `logIn()` catches a bad key first, so it mostly self-corrects — but in LOGIN mode a JWT the WS layer rejects would point the alarm at the wrong subsystem.

monitoring/src/main/java/org/thingsboard/monitoring/client/WsClient.java
```diff
@@ -168,7 +196,15 @@ public JsonNode waitForReply() {
             log.debug("Failed to await reply", e);
         }
         log.trace("No reply arrived within {} ms", requestTimeoutMs);
-        throw new IllegalStateException("No WS reply arrived within " + requestTimeoutMs + " ms");
+        throw new IllegalStateException("No WS reply arrived within " + requestTimeoutMs + " ms" + closeInfo());
+    }
+
+    // surfaces the close reason instead of an opaque WebsocketNotConnectedException/timeout
+    private String closeInfo() {
```
smatvienko-tb, line 203: The helper is generic but the message hardcodes "before subscribe could complete", and it's also appended in `waitForReply()`, which knows nothing about subscribing. Dropping the operation from the shared part — `" - WebSocket was closed (code=..., reason=...)"` — keeps it accurate wherever it lands, since both callers already supply their own prefix.

monitoring/src/main/java/org/thingsboard/monitoring/client/WsClientFactory.java
```diff
@@ -39,8 +39,8 @@ public class WsClientFactory {
     @Value("${monitoring.ws.request_timeout_ms}")
     private int requestTimeoutMs;
 
-    public WsClient createClient(String accessToken) throws Exception {
-        URI uri = new URI(baseUrl + "/api/ws/plugins/telemetry?token=" + accessToken);
+    public WsClient createClient(TbClient.AuthMode authMode, String tokenOrApiKey) throws Exception {
```
smatvienko-tb: The two parameters have to agree with each other, and `tokenOrApiKey` rather admits it — the caller must remember to pass `tbClient.getAuthMode()` alongside whatever `tbClient.logIn()` returned, and nothing stops a future caller from pairing a JWT with `API_KEY`. `WsClientFactory` is already a Spring bean, so it could inject `TbClient` and read the mode itself, dropping the parameter. Alternatively have `TbClient` hand back a small record carrying mode + secret together so the two can't be mismatched.

monitoring/src/main/java/org/thingsboard/monitoring/client/WsClientFactory.java
```diff
@@ -52,6 +52,7 @@ public WsClient createClient(String accessToken) throws Exception {
         if (!connected) {
             throw new IllegalStateException("Failed to establish WS session");
         }
+        wsClient.authenticate(authMode, tokenOrApiKey);
```
smatvienko-tb: The two behaviours at the centre of this PR are both implemented here and neither is covered: that the client now connects to `/api/ws` instead of `/api/ws/plugins/telemetry?token=`, and that `AuthCmd` must be the very first frame, ahead of any subscribe. `WsClientTest` asserts the JSON shape of `authenticate()` on a client that never connects, which can't catch a wrong path or wrong frame ordering — and a wrong path fails in a quiet way, since `toRef()` resolves the session type by exact string match on `/api/ws` and anything else goes down the `WS_PLUGINS_ENDPOINT` branch.

`Java-WebSocket` is a compile-scope dependency, so standing up an `org.java_websocket.server.WebSocketServer` on an ephemeral port would let you assert the handshake resource descriptor is `/api/ws`, that the first frame is the `authCmd` and the second the `cmds` subscribe — and, by closing server-side mid-flight, that the new `closeInfo()` diagnostics actually contain the code and reason (currently neither branch of `closeInfo()` nor the enriched `waitForReply()`/`subscribeForTelemetry` messages have any test).

monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
```diff
@@ -16,35 +16,70 @@
...
-                    @Value("${monitoring.rest.request_timeout_ms}") int requestTimeoutMs) {
+                    @Value("${monitoring.rest.request_timeout_ms}") int requestTimeoutMs,
+                    @Value("${monitoring.rest.auth_mode:LOGIN}") AuthMode authMode,
+                    @Value("${monitoring.rest.api_key:}") String apiKey) {
         super(new RestTemplateBuilder()
                 .connectTimeout(Duration.ofMillis(requestTimeoutMs))
                 .readTimeout(Duration.ofMillis(requestTimeoutMs))
-                .build(), baseUrl);
+                .build(), baseUrl,
+                authMode == AuthMode.API_KEY ? AuthType.API_KEY : AuthType.JWT,
```
smatvienko-tb, line 52: `RestClient.AuthType {JWT, API_KEY}` already exists and is public, and `AuthMode {LOGIN, API_KEY}` maps onto it 1:1 — the price is these two parallel ternaries on the same condition plus a second enum readers have to keep straight. Is the extra enum earning its keep? I can see the argument that `auth_mode: LOGIN` reads better in the YAML than `JWT`, and that `AuthMode` is a monitoring-level concept rather than a REST-client one — but if that's the motivation, a one-line comment saying so would save the next reader the trip.

monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
```diff
@@ -16,35 +16,70 @@
...
+        this.apiKey = apiKey;
+        if (authMode == AuthMode.API_KEY && StringUtils.isBlank(apiKey)) {
+            throw new IllegalStateException("monitoring.rest.api_key must be set when monitoring.rest.auth_mode is API_KEY");
+        }
+        log.info("Starting TbClient with auth mode: {}", authMode);
     }
 
     @PostConstruct
     private void init() {
+        if (authMode == AuthMode.LOGIN && (StringUtils.isBlank(username) || StringUtils.isBlank(password))) {
```
smatvienko-tb: The two halves of the same config validation now sit in different lifecycle phases — API_KEY in the constructor, LOGIN here in `@PostConstruct` — purely because `username`/`password` are field-injected `@Value`s while `apiKey` is a constructor parameter. That asymmetry is surprising to read, and it's what forces the tests into `ReflectionTestUtils.setField` and `invokeMethod(client, "init")`. Promoting `username`/`password` to constructor parameters, like `baseUrl` and `apiKey` already are, would put all the validation in one place and turn those tests into plain constructor calls.

monitoring/src/main/java/org/thingsboard/monitoring/client/TbClient.java
```diff
@@ -16,35 +16,70 @@
...
+        }
         logIn();
     }
 
+    public AuthMode getAuthMode() {
+        return authMode;
+    }
+
+    // API_KEY has no login step - proves the key works via getUser() and returns it as-is
     public String logIn() {
```
smatvienko-tb: The comment above is honest that this no longer logs in in API_KEY mode, which suggests the name has outgrown the method — `authenticate()` or `getWsCredential()` would describe both branches.

The part I'd weigh more: `BaseMonitoringService.runChecks()` times this as `Latencies.LOG_IN` and reports failures under `MonitoredServiceKey.LOGIN`. So in API_KEY mode an existing `loginLatency` dashboard series silently switches to measuring a `/api/auth/user` round-trip, and Slack alerts still read as a login failure for what is really API-key validation. The YAML note covers the semantic change for whoever reads the config, but the metric and alert labels don't — a distinct latency/service key for the API-key check would keep the historical series meaningful.

monitoring/src/main/java/org/thingsboard/monitoring/data/notification/ApiKeyExpiryWarningNotification.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package org.thingsboard.monitoring.data.notification;
+
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class ApiKeyExpiryWarningNotification implements Notification {
+
+    private final String apiKeyDescription;
+    private final long daysLeft;
```
smatvienko-tb, line 23: The three-arg constructor is easy to call wrongly: when `expired` is true, `daysLeft` is meaningless and ignored — and the two call sites already disagree about what to put there (the service passes a negative number, the test passes 0). A boolean flag plus an argument that only matters for one value of that flag is what static factories fix: `ApiKeyExpiryWarningNotification.expired(description)` and `.expiringIn(description, daysLeft)`. That would also match the convention `AffectedService` in this same package already uses.

monitoring/src/main/java/org/thingsboard/monitoring/data/notification/ApiKeyExpiryWarningNotification.java
```diff
@@ -0,0 +1,40 @@
...
+    private final boolean expired;
+
+    @Override
+    public String getText() {
+        if (expired) {
+            return String.format(":rotating_light: API key '%s' has EXPIRED - rotate it now to restore monitoring", apiKeyDescription);
+        }
+        return String.format(":warning: API key '%s' expires in %d day(s) - rotate it soon to avoid a monitoring outage", apiKeyDescription, daysLeft);
```
smatvienko-tb: `Duration.toDays()` truncates, so for the entire last day before expiry this reads "expires in 0 day(s)" — the most urgent warning is also the least clear one. Special-casing it to "expires in less than a day" (or rendering hours) would make the final nudge the loudest. The `day(s)` construction also reads a bit mechanical; `daysLeft == 1 ? "day" : "days"` is cheap if you're touching this anyway.

monitoring/src/main/java/org/thingsboard/monitoring/data/cmd/EntityDataCmd.java
```diff
@@ -18,11 +18,13 @@
...
 import org.thingsboard.server.common.data.query.EntityDataQuery;
 
+// hand-rolled mirror of application's org.thingsboard.server.service.ws.telemetry.cmd.v2.EntityDataCmd
 @Data
 public class EntityDataCmd {
 
     private int cmdId;
     private EntityDataQuery query;
     private LatestValueCmd latestCmd;
+    private final String type = "ENTITY_DATA";
```
smatvienko-tb: Mirroring these DTOs seems like the pragmatic call — the real `WsCommandsWrapper`/`EntityDataCmd` live in the `application` module, which `monitoring` doesn't and shouldn't depend on. What's new is that the mirror now also duplicates the polymorphic discriminator as a bare string literal hand-copied out of the server's `@JsonSubTypes` table, with nothing failing at build time if the server ever renames it. Hoisting `"ENTITY_DATA"` into a named constant next to the class-level "mirror of…" comment would at least make it obvious what it shadows. Longer term, `AuthCmd`/`WsCmdType` moving down into `common/data` would let both sides share one definition — `EntityDataQuery`, used a few lines above, is precedent for WS/query-shaped DTOs living there.

monitoring/src/main/java/org/thingsboard/monitoring/data/cmd/AuthCmd.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.thingsboard.monitoring.data.cmd;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Data;
+
+// NON_NULL: some server versions reject the whole message if the inactive field is present as null
```
smatvienko-tb: This asserts something the next maintainer can't check — which versions, rejecting how? — and `CmdsWrapper` pairs the same claim with a "do not remove" directive. For what it's worth, the handler in this tree reads `cmdsWrapper.getAuthCmd()` and `CollectionUtils.isEmpty(getCmds())`, so an explicit `null` would be equivalent here. `NON_NULL` is a good default for an outbound command DTO on its own merits — don't send fields you don't mean — so I'd either say that plainly or cite the specific version and behaviour that motivated it. As written it reads as folklore nobody will dare touch.

monitoring/src/main/resources/tb-monitoring.yml
```diff
@@ -20,14 +20,40 @@ monitoring:
...
+    # Days before expiration to start sending a daily warning notification (auth_mode: API_KEY
+    # only, requires api_key_description to be set). The warning repeats every day once within
+    # this threshold until the key is rotated or the process restarts.
+    api_key_expiry_warning_days: '${REST_AUTH_API_KEY_EXPIRY_WARNING_DAYS:7}'
     # REST request timeout in milliseconds
     request_timeout_ms: '${REST_REQUEST_TIMEOUT_MS:5000}'
   ws:
-    # WebSocket url, wss://DOMAIN by default
+    # WebSocket url, wss://DOMAIN by default.
+    # Requires the monitored ThingsBoard server to be version ~3.6.2 or later: this client
```
smatvienko-tb, line 57: "~3.6.2 or later" isn't actionable for the operator this comment is aimed at — they can't tell whether their 3.6.1 deployment is fine. Either pin the version you actually verified `/api/ws` against, or drop the number and say it requires a server supporting the `/api/ws` GENERAL endpoint. Worth remembering too that this is a hard requirement rather than a tuning knob, and operators who configure purely through env vars never see the shipped YAML — a line in the release notes (or logged at startup) would reach them.

monitoring/src/test/java/org/thingsboard/monitoring/client/TbClientTest.java
```diff
@@ -0,0 +1,104 @@
...
+        org.mockito.Mockito.doNothing().when(client).login("tenant@thingsboard.org", "tenant");
+
+        String result = client.logIn();
+
+        verify(client).login("tenant@thingsboard.org", "tenant");
+        assertThat(result).isEqualTo("jwt-token");
+    }
+
+    @Test
+    public void apiKeyMode_logInCallsGetUserNotLogin() {
```
smatvienko-tb: These tests verify the branching inside `logIn()`, but not the part of the REST change that carries the most risk: that API_KEY mode wires the interceptor with `AuthType.API_KEY` and therefore sends `X-Authorization: ApiKey <key>` rather than `Bearer …`. Because `getUser()` is stubbed on a spy, no request is ever built, so a mistake in the `AuthType` ternary passed to `super(...)` would sail through green. `spring-boot-starter-test` brings `MockRestServiceServer`, which would let you assert the actual header for both modes — a much stronger check than `verify(client).getUser()`.

monitoring/src/test/java/org/thingsboard/monitoring/client/TbClientTest.java
```diff
@@ -0,0 +1,104 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+public class TbClientTest {
```
smatvienko-tb: Small style drift worth a pass: this class (and `WsClientTest`, `ApiKeyExpiryMonitoringServiceTest`) is `public class` with `public void` methods and mixes JUnit's `assertThrows` with AssertJ's `assertThat`/`assertThatThrownBy`, while the module's existing tests — `ServiceFailureNotificationTest`, `IncidentManagerTest` — and the fourth new class here are package-private, `void`, pure AssertJ. JUnit 5 doesn't need the `public`. In the same vein, `org.springframework.test.util.ReflectionTestUtils`, `org.mockito.Mockito.doNothing()` and `org.mockito.ArgumentMatchers.anyString()` are spelled out inline even though the file already has a tidy static-import block, which makes the assertions harder to scan (and `ApiKeyExpiryMonitoringServiceTest` imports `ReflectionTestUtils` normally, so it's inconsistent within the PR too).

monitoring/src/test/java/org/thingsboard/monitoring/service/ApiKeyExpiryMonitoringServiceTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+        service.checkApiKeyExpiry();
+        service.checkApiKeyExpiry();
+
+        verify(notificationService, never()).sendNotification(any());
+        verify(tbClient, times(1)).getUserApiKeys(any(), any());
+    }
+
+    @Test
+    public void daysLeftAboveThreshold_noNotification() {
```
smatvienko-tb: Two paths are still uncovered. `tbClient.getUser()` returning `Optional.empty()` has its own early return and warning but is never exercised — `@BeforeEach` always stubs a present user. And the comparison is `daysLeft <= warningDays`, yet `daysLeftAtOrBelowThreshold_...` only tests 3 against a threshold of 7, so the "at" boundary — exactly 7, the off-by-one most likely to regress — isn't asserted from either side. Both look like one-liners with the fixtures already in place.

monitoring/src/test/java/org/thingsboard/monitoring/service/ApiKeyExpiryMonitoringServiceTest.java
```diff
@@ -0,0 +1,213 @@
...
+        doReturn(new PageData<>(List.of(apiKeyInfo("tb-monitoring key", farFuture)), 1, 1, false))
+                .when(tbClient).getUserApiKeys(eq(userId), any());
+
+        service.checkApiKeyExpiry();
+
+        verify(notificationService, never()).sendNotification(any());
+    }
+
+    @Test
+    public void daysLeftAtOrBelowThreshold_notifiesEveryCall_noDeduplication() {
```
smatvienko-tb: This `doReturn(new PageData<>(List.of(...), 1, 1, false)).when(tbClient).getUserApiKeys(eq(userId), any())` incantation is repeated verbatim across eight tests with only the `ApiKeyInfo` varying, and the `1, 1, false` page metadata is noise in every one. A `givenApiKeys(ApiKeyInfo... keys)` helper alongside the existing `apiKeyInfo(...)` factories would reduce each to one line of intent. Same story in `TbClientTest`, where `spy(new TbClient("http://localhost:8080", 5000, …))` appears five times — `WsClientTest` already demonstrates the pattern with its `newClient()` helper.

PR #16056, opened 2026-08-25
fix(mqtt): quiet the client certificate probe on one-way TLS connect

ShvaykaD:
## Pull Request description

Found while testing the MQTT gateway API in one-way TLS mode.

### Problem

In one-way TLS mode the transport is configured to accept a connect without a client certificate (`needClientAuth(false)` with `wantClientAuth(true)`). JSSE reports the absent certificate by throwing `SSLPeerUnverifiedException`, which `getX509Certificate()` logged at `warn` — on an outcome the transport is configured to accept, and with no session or address to correlate it.

It fired twice per connect, because `onValidateDeviceResponse` re-probed the session.

### Approach

Log at `debug`, with the `[address][sessionId]` prefix used elsewhere in the class.

Drop the second probe. Its guard — `sslHandler == null || getX509Certificate() == null` — does not want the certificate; it wants to know whether this was a basic-credentials connect, which `processConnect` has already decided. `onValidateDeviceResponse` takes that decision as a `boolean x509Auth` parameter, passed `false` from `processAuthTokenConnect` and `true` from `processX509CertConnect`. Those are its only two call sites, and `x509Auth` is true iff `processConnect` found a certificate — whose negation is exactly the replaced condition, so return codes are unchanged. Passing it rather than storing it means a future connect path cannot forget to record the branch: the compiler asks.

The `sslHandler == null` guard moves into `getX509Certificate()`, which is now its only caller. That makes the method safe to call on its own terms and reduces `processConnect` to a plain `getX509Certificate()` followed by `if (cert != null)`, with no assignment inside the condition.

### Tests

`MqttTransportHandlerTest` covers both connect paths, asserting the resulting return codes and that the peer certificate is probed exactly once per connect. Both use MQTT 5 via `getMqttV5ConnectMessage()`: `ReturnCodeResolver` collapses `BAD_USERNAME_OR_PASSWORD` and `NOT_AUTHORIZED_5` into a single code for older clients, which would make the two cases indistinguishable. Confirmed the probe-count assertions are not vacuous — restoring the old guard fails both tests with `TooManyActualInvocations`.

Two test-only cleanups ride along: the connect-message builder is parameterized on netty's `MqttVersion` (previously the protocol version was a copy of the packet-id counter, and the protocol name was `"device"`), and `Stream.toList()` replaces `collect(Collectors.toList())` at the four call sites that only iterate the result.
## General checklist

- [x] Changes are backward compatible or upgrade script is provided. _(Log level and an internal guard only; return codes unchanged.)_

## Back-End feature checklist

- [x] If new dependency was added: the dependency tree is checked for conflicts. _(No new dependencies.)_
- [x] If new service was added: the service is marked with corresponding @TbCoreComponent, @TbRuleEngineComponent, @TbTransportComponent, etc. _(No new services.)_
- [x] If new REST API was added: the RestClient.java was updated. _(No new REST API.)_
- [x] If new yml property was added: make sure a description is added (above or near the property). _(No new yml properties.)_

### DIFF
--- a/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
+++ b/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
@@ -157,6 +157,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
     volatile InetSocketAddress address;
     volatile GatewaySessionHandler gatewaySessionHandler;
     volatile SparkplugNodeSessionHandler sparkplugSessionHandler;
+    private volatile boolean clientCertPresented;
 
     private final ConcurrentHashMap<String, String> otaPackSessions;
     private final ConcurrentHashMap<String, Integer> chunkSizes;
@@ -1055,6 +1056,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
         } else {
             X509Certificate cert;
             if (sslHandler != null && (cert = getX509Certificate()) != null) {
+                clientCertPresented = true;
                 processX509CertConnect(ctx, cert, msg);
             } else {
                 processAuthTokenConnect(ctx, msg);
@@ -1127,8 +1129,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
                 return (X509Certificate) certChain[0];
             }
         } catch (SSLPeerUnverifiedException e) {
-            log.warn(e.getMessage());
-            return null;
+            log.debug("[{}][{}] Client certificate is not available: {}", address, sessionId, e.getMessage());
         }
         return null;
     }
@@ -1326,7 +1327,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
         if (!msg.hasDeviceInfo()) {
             context.onAuthFailure(address);
             MqttConnectReturnCode returnCode = MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED_5;
-            if (sslHandler == null || getX509Certificate() == null) {
+            if (!clientCertPresented) {
                 String username = connectMessage.payload().userName();
                 byte[] passwordBytes = connectMessage.payload().passwordInBytes();
                 String clientId = connectMessage.payload().clientIdentifier();
--- a/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
+++ b/common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
@@ -157,7 +157,6 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
     volatile InetSocketAddress address;
     volatile GatewaySessionHandler gatewaySessionHandler;
     volatile SparkplugNodeSessionHandler sparkplugSessionHandler;
-    private volatile boolean clientCertPresented;
 
     private final ConcurrentHashMap<String, String> otaPackSessions;
     private final ConcurrentHashMap<String, Integer> chunkSizes;
@@ -1054,9 +1053,8 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
             deviceSessionCtx.setProvisionOnly(true);
             ctx.writeAndFlush(createMqttConnAckMsg(MqttConnectReturnCode.CONNECTION_ACCEPTED, msg));
         } else {
-            X509Certificate cert;
-            if (sslHandler != null && (cert = getX509Certificate()) != null) {
-                clientCertPresented = true;
+            X509Certificate cert = getX509Certificate();
+            if (cert != null) {
                 processX509CertConnect(ctx, cert, msg);
             } else {
                 processAuthTokenConnect(ctx, msg);
@@ -1081,7 +1079,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
                 new TransportServiceCallback<>() {
                     @Override
                     public void onSuccess(ValidateDeviceCredentialsResponse msg) {
-                        onValidateDeviceResponse(msg, ctx, connectMessage);
+                        onValidateDeviceResponse(msg, ctx, connectMessage, false);
                     }
 
                     @Override
@@ -1104,7 +1102,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
                     new TransportServiceCallback<>() {
                         @Override
                         public void onSuccess(ValidateDeviceCredentialsResponse msg) {
-                            onValidateDeviceResponse(msg, ctx, connectMessage);
+                            onValidateDeviceResponse(msg, ctx, connectMessage, true);
                         }
 
                         @Override
@@ -1123,6 +1121,9 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
     }
 
     private X509Certificate getX509Certificate() {
+        if (sslHandler == null) {
+            return null;
+        }
         try {
             Certificate[] certChain = sslHandler.engine().getSession().getPeerCertificates();
             if (certChain.length > 0) {
@@ -1323,11 +1324,11 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
         deviceSessionCtx.release();
     }
 
-    private void onValidateDeviceResponse(ValidateDeviceCredentialsResponse msg, ChannelHandlerContext ctx, MqttConnectMessage connectMessage) {
+    private void onValidateDeviceResponse(ValidateDeviceCredentialsResponse msg, ChannelHandlerContext ctx, MqttConnectMessage connectMessage, boolean x509Auth) {
         if (!msg.hasDeviceInfo()) {
             context.onAuthFailure(address);
             MqttConnectReturnCode returnCode = MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED_5;
-            if (!clientCertPresented) {
+            if (!x509Auth) {
                 String username = connectMessage.payload().userName();
                 byte[] passwordBytes = connectMessage.payload().passwordInBytes();
                 String clientId = connectMessage.payload().clientIdentifier();
--- a/common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
+++ b/common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
@@ -18,8 +18,10 @@ package org.thingsboard.server.transport.mqtt;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.mqtt.MqttConnAckMessage;
 import io.netty.handler.codec.mqtt.MqttConnectMessage;
 import io.netty.handler.codec.mqtt.MqttConnectPayload;
+import io.netty.handler.codec.mqtt.MqttConnectReturnCode;
 import io.netty.handler.codec.mqtt.MqttConnectVariableHeader;
 import io.netty.handler.codec.mqtt.MqttFixedHeader;
 import io.netty.handler.codec.mqtt.MqttMessageType;
@@ -32,9 +34,11 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.Spy;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.stubbing.Answer;
 import org.thingsboard.common.util.ThingsBoardThreadFactory;
 import org.thingsboard.server.common.data.DataConstants;
 import org.thingsboard.server.common.data.DeviceProfile;
@@ -44,11 +48,18 @@ import org.thingsboard.server.common.data.device.profile.JsonTransportPayloadCon
 import org.thingsboard.server.common.data.device.profile.MqttDeviceProfileTransportConfiguration;
 import org.thingsboard.server.common.msg.TbMsgMetaData;
 import org.thingsboard.server.common.transport.TransportService;
+import org.thingsboard.server.common.transport.TransportServiceCallback;
+import org.thingsboard.server.common.transport.auth.ValidateDeviceCredentialsResponse;
 import org.thingsboard.server.gen.transport.TransportProtos;
 import org.thingsboard.server.transport.mqtt.adaptors.JsonMqttAdaptor;
 
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLSession;
 import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
@@ -66,8 +77,12 @@ import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.is;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.BDDMockito.willAnswer;
 import static org.mockito.BDDMockito.willDoNothing;
+import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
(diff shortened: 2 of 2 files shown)


common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -157,6 +157,7 @@ public class MqttTransportHandler extends ChannelInboundHandlerAdapter implement
     volatile InetSocketAddress address;
     volatile GatewaySessionHandler gatewaySessionHandler;
     volatile SparkplugNodeSessionHandler sparkplugSessionHandler;
+    private volatile boolean clientCertPresented;
```
ShvaykaD: Two thoughts on this field.

On the mechanism: it exists only to carry a decision from `processConnect` down to `onValidateDeviceResponse`, and `onValidateDeviceResponse` has exactly two call sites — `processAuthTokenConnect` and `processX509CertConnect` — each of which already knows unambiguously which credential type it used. Passing that in as a parameter would keep the handler stateless for this concern, drop the need for `volatile`, and make the branch at line 1330 readable at the call site instead of depending on an assignment 270 lines up. The temporal coupling as written isn't visible at either end: a future connect path that forgets to set the flag silently changes the return code and nothing fails loudly.

On the name: `clientCertPresented` describes an SSL-level fact, but what's recorded is narrower — that this connect was routed down the X.509 branch. The read site is really asking "did we authenticate by certificate rather than username/password?", so something like `x509AuthUsed` would match that intent and keep a future reader from treating it as a general-purpose "is there a peer cert" flag.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -1055,6 +1056,7 @@ void processConnect(ChannelHandlerContext ctx, MqttConnectMessage msg) {
         } else {
             X509Certificate cert;
             if (sslHandler != null && (cert = getX509Certificate()) != null) {
```
ShvaykaD: Now that this branch has grown a second statement, it might be a good moment to drop the inherited assignment-inside-the-condition idiom — `X509Certificate cert = sslHandler != null ? getX509Certificate() : null;` followed by a plain `if (cert != null)` reads more directly than the declare-then-assign-in-condition dance.

Related: with the second probe gone, this is the only caller of `getX509Certificate()`, and that method dereferences `sslHandler.engine()` with no null check of its own — the guard now lives entirely here. Folding `if (sslHandler == null) return null;` into `getX509Certificate()` would make it safe to call on its own terms.

common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
```diff
@@ -252,4 +267,73 @@ public void givenMqttMessage_whenDeviceProfileMqttTransport_thenTopicAddedToMeta
...
+        given(givenTlsSession().getPeerCertificates()).willReturn(new Certificate[]{cert});
+        given(context.isSkipValidityCheckForClientCert()).willReturn(true);
+        willAnswer(unknownCredentials()).given(transportService)
+                .process(eq(DeviceTransportType.MQTT), any(TransportProtos.ValidateDeviceX509CertRequestMsg.class), any());
+
+        //when
+        handler.processConnect(ctx, getMqttV5ConnectMessage());
+
+        //then
+        assertThat(getConnAckReturnCode(), is(MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED_5));
```
ShvaykaD, line 311: The certificate path double-probed before this change just as the one-way path did, but only the one-way test asserts the probe count — so a re-introduced second probe in the X.509 branch would slip through. Holding the session from `givenTlsSession()` in a local and adding `verify(sslSession, times(1)).getPeerCertificates();` here would make the two tests symmetric.

(That would also mean unwrapping the nested `given(givenTlsSession().getPeerCertificates())` on line 291, which is worth doing anyway — stubbing inside another `given(...)` argument works here only because the inner stubbings complete before the outer one records, and that's a fragile thing to leave for the next editor.)

common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
```diff
@@ -252,4 +267,73 @@ public void givenMqttMessage_whenDeviceProfileMqttTransport_thenTopicAddedToMeta
...
+        return invocation -> {
+            TransportServiceCallback<ValidateDeviceCredentialsResponse> callback = invocation.getArgument(2);
+            callback.onSuccess(ValidateDeviceCredentialsResponse.builder().build());
+            return null;
+        };
+    }
+
+    // MQTT 5 is required: ReturnCodeResolver collapses BAD_USERNAME_OR_PASSWORD and NOT_AUTHORIZED_5 into
+    // a single NOT_AUTHORIZED code for older clients, which would make the two cases indistinguishable.
+    MqttConnectMessage getMqttV5ConnectMessage() {
```
ShvaykaD: This is a near-verbatim copy of `getMqttConnectMessage()` above — same fixed header, same payload, differing only in the variable header's protocol name and version. Could the existing helper be parameterized (e.g. `getMqttConnectMessage(String protocolName, int version)`, with both call sites delegating) so the payload and fixed-header construction live in one place?

ShvaykaD (commented):
## Review summary

Reviewed 2 changed files in [fix(mqtt): quiet the client certificate probe on one-way TLS connect](https://github.com/thingsboard/thingsboard/pull/16056). Left 4 comment(s) inline.

The core substitution holds up: `!clientCertPresented` is truth-table-equivalent to the old `sslHandler == null || getX509Certificate() == null` across all three reachable states (no TLS, TLS without a peer cert, TLS with a peer cert), `onValidateDeviceResponse` has only the two connect call sites so the flag is always set before it runs, and the volatile write on the event loop happens-before the read on `transportCallbackExecutor`. I found no correctness or security issues. `MqttTransportHandlerTest` passes locally (8 tests, 0 failures). The comments below are about how the state is carried and about test symmetry.

## Additional findings

- **common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java:133** — The pre-existing `getMqttConnectMessage()` passes `packedId.incrementAndGet()` as the MQTT *protocol version* (and `"device"` as the protocol name). It happens not to matter because the only test using it stubs out `processConnect`, but it reads like a copy-paste of the packet-id counter and will mislead the next person who reuses the helper for a test that actually cares about the version.

---

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
ShvaykaD: Applied in 4649fae4f9 — both halves.

The flag is now a `boolean x509Auth` parameter on `onValidateDeviceResponse`, passed `false` from `processAuthTokenConnect` and `true` from `processX509CertConnect`. That settles the temporal-coupling concern the way you framed it: a future connect path can no longer forget to record the branch, because the compiler asks. The field and its `volatile` are gone.

On the name — with the decision passed in rather than stored, `x509Auth` is what it became, so the "is there a peer cert" misreading has nowhere to take hold.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
ShvaykaD: Applied in 4649fae4f9, taking the two suggestions together rather than separately — the second one subsumes the first.

`getX509Certificate()` now opens with `if (sslHandler == null) { return null; }`, so it is safe on its own terms as you suggested. With the guard living inside, the caller no longer needs the ternary either — it reduces to:

```java
X509Certificate cert = getX509Certificate();
if (cert != null) {
    processX509CertConnect(ctx, cert, msg);
} else {
    processAuthTokenConnect(ctx, msg);
}
```

One guard in one place instead of a guard plus an assignment-in-condition.

common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
ShvaykaD: Applied in 4649fae4f9. The X.509 test holds the session in a local, stubs `getPeerCertificates()` on its own line, and asserts `verify(sslSession, times(1)).getPeerCertificates()`; renamed to `...thenKeepsGenericReturnCodeAndProbesPeerCertOnce` to match.

Confirmed the assertion is not vacuous: temporarily restoring the old `sslHandler == null || getX509Certificate() == null` guard fails **both** connect tests with `TooManyActualInvocations: Wanted 1 time`. Before this change only the one-way test would have caught it, which was your point.

common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
ShvaykaD: Applied in 4649fae4f9, with one deviation from the literal suggestion. Rather than `getMqttConnectMessage(String protocolName, int version)`, the helper takes the netty `MqttVersion` enum and derives both fields from it:

```java
MqttConnectMessage getMqttConnectMessage() {
    return getMqttConnectMessage(MqttVersion.MQTT_3_1_1);
}

MqttConnectMessage getMqttConnectMessage(MqttVersion version) {
    ...
    new MqttConnectVariableHeader(version.protocolName(), version.protocolLevel(), ...);
}
```

That folds in the additional finding at the same time: the two values can no longer disagree, and the protocol version stops being a copy of the packet-id counter. `getMqttV5ConnectMessage()` is deleted; both connect tests call `getMqttConnectMessage(MqttVersion.MQTT_5)` and the `ReturnCodeResolver` note moved to the first call site.

The no-arg overload now yields `MQTT`/`4` instead of `device`/`N`. Its only caller stubs out `processConnect`, so nothing observes the values — `MqttTransportHandlerTest` is 8 tests, 0 failures.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -1322,11 +1324,11 @@ public void doDisconnect() {
         deviceSessionCtx.release();
     }
 
-    private void onValidateDeviceResponse(ValidateDeviceCredentialsResponse msg, ChannelHandlerContext ctx, MqttConnectMessage connectMessage) {
+    private void onValidateDeviceResponse(ValidateDeviceCredentialsResponse msg, ChannelHandlerContext ctx, MqttConnectMessage connectMessage, boolean x509Auth) {
```
ShvaykaD, line 1327: Passing the state as a parameter is a clear improvement over the field, but a trailing positional boolean makes the two call sites read as `onValidateDeviceResponse(msg, ctx, connectMessage, false)` / `(..., true)`, where nothing at the call site says what the flag means — and the commit message itself calls it "the credential type".

Would `DeviceCredentialsType` (`X509_CERTIFICATE` vs `MQTT_BASIC`, both already in `common/data`) be worth it here instead? It costs nothing at the two call sites, self-documents them, and leaves room for a third credential kind if the refinement logic ever needs to distinguish access token from basic. If you'd rather keep it a boolean because it's a private single-use method, that's a defensible trade-off — in that case naming it for the question the read site actually asks (`certificateAuth`) would carry most of the benefit.

ShvaykaD (commented):
## Re-review summary

Re-reviewed [fix(mqtt): quiet the client certificate probe on one-way TLS connect](https://github.com/thingsboard/thingsboard/pull/16056) — verified 4 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/16056#pullrequestreview-5030035802).

| Status | Count |
|--------|-------|
| ✅ Resolved | 4 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 0 |

Also found 3 new issue(s) in the fix commit, commented inline.

All four findings were addressed in code rather than argued away, and the replacement is cleaner than what was suggested: with the `sslHandler == null` guard living inside `getX509Certificate()`, `processConnect` reduces to a plain `getX509Certificate()` + `if (cert != null)` with no ternary. The `volatile boolean clientCertPresented` field is gone with no leftover references, `onValidateDeviceResponse` has exactly the two call sites (`false` from `processAuthTokenConnect`, `true` from `processX509CertConnect`), and `!x509Auth` remains truth-table-equivalent to the old `sslHandler == null || getX509Certificate() == null` — so return codes are still unchanged, and the compiler now enforces what a forgotten field assignment used to leave silent. `getMqttV5ConnectMessage()` is deleted with no dangling references, and `MqttTransportHandler` has no subclasses that could depend on the changed signature. I found no new correctness or security issues in the fix commit. `MqttTransportHandlerTest` passes in a clean worktree build: 8 tests, 0 failures.

### Finding details

- ✅ **common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java:1327** — the `volatile boolean clientCertPresented` field carried a decision 270 lines down to `onValidateDeviceResponse` with the temporal coupling invisible at both ends; suggested a parameter, and a name describing "authenticated by certificate" rather than "a peer cert exists" — *Fixed in code: field and its `volatile` removed, `boolean x509Auth` parameter added.*
- ✅ **common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java:1124** — the assignment-inside-condition idiom in `processConnect`, and `getX509Certificate()` dereferencing `sslHandler.engine()` with no null check of its own once it became the sole caller — *Fixed in code: guard folded into `getX509Certificate()`, which let the caller collapse to a plain `if (cert != null)`.*
- ✅ **common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java:310** — only the one-way test asserted the probe count, so a re-introduced second probe in the X.509 branch would have slipped through; the nested `given(givenTlsSession().getPeerCertificates())` stubbing was also fragile — *Fixed in code: session held in a local, stubbed on its own line, and `verify(sslSession, times(1)).getPeerCertificates()` added.*
- ✅ **common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java:136** — `getMqttV5ConnectMessage()` was a near-verbatim copy of `getMqttConnectMessage()`, differing only in protocol name and version — *Fixed in code: helper parameterized on netty's `MqttVersion`, which also resolves the previous review's additional finding that the protocol version was a copy of the packet-id counter — the two values can no longer disagree. Verified the no-arg overload's only caller stubs out `processConnect`, so the `MQTT`/`4` values it now yields are never observed.*

---

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
```diff
@@ -1121,14 +1121,16 @@ public void onError(Throwable e) {
     }
 
     private X509Certificate getX509Certificate() {
+        if (sslHandler == null) {
```
ShvaykaD, line 1124: Folding the guard in here reads much better than the old assignment-inside-condition. One gap it opens up, though: the plain-TCP path (`sslHandler == null`, i.e. every non-TLS MQTT listener) is now a branch of *this* method, and nothing in `MqttTransportHandlerTest` exercises it — `setUp()` always constructs the handler with a mocked `SslHandler`, so both new tests take the TLS path.

Since the responsibility for "no TLS at all" just moved out of `processConnect` and into here, a third test building `new MqttTransportHandler(context, null)` and asserting the connect goes down the auth-token route would pin that down cheaply and keep the three routes symmetric.

common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
```diff
@@ -252,4 +272,69 @@ public void givenMqttMessage_whenDeviceProfileMqttTransport_thenTopicAddedToMeta
...
\ No newline at end of file
+    @Test
+    public void givenOneWayTlsAndUnknownCredentials_whenProcessConnect_thenRefinesReturnCodeAndProbesPeerCertOnce() throws Exception {
+        //given
+        SSLSession sslSession = givenTlsSession();
+        given(sslSession.getPeerCertificates()).willThrow(new SSLPeerUnverifiedException("peer not authenticated"));
+        willAnswer(unknownCredentials()).given(transportService)
+                .process(eq(DeviceTransportType.MQTT), any(TransportProtos.ValidateBasicMqttCredRequestMsg.class), any());
+
+        // MQTT 5 is required in both connect tests: ReturnCodeResolver collapses BAD_USERNAME_OR_PASSWORD and
```
ShvaykaD: Parameterizing the connect-message builder was the right call, but the rationale comment ended up in an odd spot: it's a comment about *both* tests that now lives inside the first one, wedged between the `//given` block and the `//when` marker. If someone later deletes or reorders this test, the explanation disappears from the other one.

A thin named delegate would keep the comment attached to the thing it explains without reintroducing the duplication this round removed — e.g. `MqttConnectMessage getMqttV5ConnectMessage() { return getMqttConnectMessage(MqttVersion.MQTT_5); }` with the comment on it, and both tests calling that.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
ShvaykaD: Fair — withdrawing this one.

The previous round asked for the state to be passed as a parameter, and it is. Which *type* that parameter takes was never the finding, and for a private method with two call sites a boolean is a legitimate choice; pulling `DeviceCredentialsType` into the transport handler would add an import and a mapping for no functional gain. That was taste presented as a finding, on a PR whose scope is a log level and a removed double probe.

common/transport/mqtt/src/main/java/org/thingsboard/server/transport/mqtt/MqttTransportHandler.java
ShvaykaD: Agreed, withdrawing — and your reasoning is stronger than "out of scope."

I checked the two forms against each other: with `sslHandler == null` the old code short-circuited before ever calling `getX509Certificate()`, and the new code calls it and returns `null` at the added guard. Both land in `processAuthTokenConnect`, so the PR introduces no new untested behavior — that path was untested before this change and behaves identically after it.

My comment leaned on "the responsibility moved into this method," which is true about where the code lives but not about what it does, and a request for new test coverage needs the latter. Withdrawn.

common/transport/mqtt/src/test/java/org/thingsboard/server/transport/mqtt/MqttTransportHandlerTest.java
ShvaykaD: Here I'd push back, and the comment is now moved rather than deleted.

The fact it records is load-bearing. `ReturnCodeResolver.getConnectionReturnCode` collapses *both* `CONNECTION_REFUSED_BAD_USERNAME_OR_PASSWORD` and `CONNECTION_REFUSED_NOT_AUTHORIZED_5` into a single `CONNECTION_REFUSED_NOT_AUTHORIZED` for any non-MQTT-5 client, and `createMqttConnAckMsg` runs every return code through it. So if someone later swaps `getMqttConnectMessage(MqttVersion.MQTT_5)` for the no-arg helper, both tests still pass — asserting the same collapsed code — and the one-way test quietly stops proving that the refinement happened at all.

That isn't derivable from the test logic on screen: a reader sees `MQTT_5` and two different expected codes, and nothing tells them the version is what keeps those two codes distinguishable. It's a fact about a different class, two hops from the assertion.

So rather than drop it, it now hangs off a delegate, which also answers the placement concern that prompted the original comment:

```java
// MQTT 5 is required: ReturnCodeResolver collapses BAD_USERNAME_OR_PASSWORD and NOT_AUTHORIZED_5 into
// a single NOT_AUTHORIZED code for older clients, which would make the two cases indistinguishable.
MqttConnectMessage getMqttV5ConnectMessage() {
    return getMqttConnectMessage(MqttVersion.MQTT_5);
}
```

Both connect tests call `getMqttV5ConnectMessage()`, the explanation sits on the choice it justifies instead of inside one of the two tests, and the duplication the previous round removed does not come back — the delegate is one line over the parameterized builder. `MqttTransportHandlerTest` is 8 tests, 0 failures.

PR #16082, opened 2026-08-28
Fix reconnect race where the closing edge session cancels the new session's edge event check

MazurenkoNick:
## Pull Request description

## General checklist

## Back-End feature checklist


application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -310,21 +311,20 @@ public void updateEdge(TenantId tenantId, Edge edge) {
 
     @Override
     public void deleteEdge(TenantId tenantId, EdgeId edgeId) {
-        EdgeGrpcSession session = sessions.get(edgeId);
-        if (session != null && session.isConnected()) {
+        EdgeGrpcSession toRemove = detachSession(edgeId, EdgeGrpcSession::isConnected);
```
MazurenkoNick: Two things about this predicate.

First, the `isConnected` guard means that if the registered session is present but already disconnected, `detachSession` returns null and we fall out of the method entirely — so the session stays in `sessions` for an edge that no longer exists, and `cleanUp()` never runs, leaving the Kafka topic and consumer group behind. That behaviour predates this PR, but now that the detach is atomic anyway, is the liveness check still buying anything? Detaching unconditionally here (`current -> true`) would let you clean up in both cases.

Second, `Predicate<EdgeGrpcSession>` is now carrying two quite different meanings across the two call sites: in `onEdgeDisconnect` it's the identity check that makes the whole fix work ("is the registered session still mine?"), while here it's a liveness filter. Read cold, `detachSession(edgeId, EdgeGrpcSession::isConnected)` looks like it's asserting the race guard when it isn't. A `detachSession(EdgeId edgeId, UUID sessionId)` overload for the identity case — the one every future caller will reach for — would put the important semantics in the signature.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -387,9 +383,12 @@ private void onEdgeConnect(EdgeId edgeId, EdgeGrpcSession edgeGrpcSession) {
         save(tenantId, edgeId, LAST_CONNECT_TIME, lastConnectTs);
         edgeIdServiceIdCache.put(edgeId, serviceInfoProvider.getServiceId());
         pushRuleEngineMessage(tenantId, edge, lastConnectTs, TbMsgType.CONNECT_EVENT);
-        cancelScheduleEdgeEventsCheck(edgeId);
         edgeEventsMigrationProcessed.putIfAbsent(edgeId, Boolean.FALSE);
         scheduleEdgeEventsCheck(edgeGrpcSession);
+        if (replaced != null) {
+            // destroy is slow (stops the kafka edge event consumer), keep it off the atomic path
```
MazurenkoNick: This rule now lives as three copies of the same one-line comment, each explaining only the call it happens to sit next to. That documents the symptom rather than the rule, and it tells a future reader nothing about the other things that also must not migrate into a `sessions.compute` block — `save(...)`, `pushRuleEngineMessage(...)`, `cleanUp()` (which does a Kafka topic + consumer-group delete), or acquiring `newEventLock`. Worth stating once instead: a javadoc line on `destroySession` noting it can block for seconds (`awaitConsumerTermination` waits up to 5s), plus one line on `attachSession`/`detachSession` saying only cheap non-blocking work may run inside the remapping function — then the three inline repeats can go.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -451,55 +450,62 @@ private void scheduleEdgeEventsCheck(EdgeGrpcSession session) {
         EdgeId edgeId = session.getEdge().getId();
         TenantId tenantId = session.getEdge().getTenantId();
 
-        cancelScheduleEdgeEventsCheck(edgeId);
-
-        if (sessions.containsKey(edgeId)) {
-            ScheduledFuture<?> edgeEventCheckTask = edgeEventProcessingExecutorService.schedule(() -> {
+        Runnable edgeEventCheck = () -> {
```
MazurenkoNick: Hoisting this into a named local is already an improvement over the inline lambda, but the method is still doing two unrelated jobs at two different altitudes: defining the entire event-processing loop (lock handling, high-priority events, migration, the async `processEdgeEvents` callback, four separate re-arm calls) and then the arming/currency bookkeeping. Pulling the body out into a private `runEdgeEventCheck(EdgeGrpcSession session)` would leave `scheduleEdgeEventsCheck` as a ~10-line "arm the next check if this session is still current", which is the part a reader comes here for, and would make the `compute` block the focus instead of a tail after 45 lines. While you're in there — the four identical `scheduleEdgeEventsCheck(session)` calls in mutually exclusive branches are easy to get out of sync; the two synchronous ones could collapse into a single re-arm after the if/else, leaving only the future-callback ones.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -451,55 +450,62 @@ private void scheduleEdgeEventsCheck(EdgeGrpcSession session) {
...
+            try {
+                final Lock newEventLock = sessionNewEventsLocks.computeIfAbsent(edgeId, id -> new ReentrantLock());
```
MazurenkoNick: With this change the `sessionNewEventsLocks.computeIfAbsent(...)` + `lock()` / `try` / `finally unlock()` idiom is copy-pasted in four places (`deleteEdge`, `onEdgeConnect`, `updateSessionEventsFlag`, and here). Three of them were touched in this PR anyway, so it seems like a good moment to extract a `private void withNewEventsLock(EdgeId edgeId, Runnable action)` — or at minimum a `getNewEventsLock(edgeId)` accessor. That removes the chance of a future copy forgetting the `finally`, and gives you one place to document how this lock relates to the new `sessions.compute` critical sections, which is now the subtlest thing in this class.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -451,55 +450,62 @@ private void scheduleEdgeEventsCheck(EdgeGrpcSession session) {
...
-                            if (Boolean.TRUE.equals(edgeEventsMigrationProcessed.get(edgeId))) {
-                                Futures.addCallback(session.processEdgeEvents(), new FutureCallback<>() {
-                                    @Override
-                                    public void onSuccess(Boolean newEventsAdded) {
-                                        if (Boolean.TRUE.equals(newEventsAdded)) {
-                                            log.trace("[{}][{}] new events added. set session new events flag to true", tenantId, edgeId.getId());
-                                            sessionNewEvents.put(edgeId, true);
-                                        }
-                                        scheduleEdgeEventsCheck(session);
+                    if (Boolean.TRUE.equals(sessionNewEvents.get(edgeId))) {
```
MazurenkoNick: The currency guard was added to the *re-arm*, but not to the *execution* of the check, and I think that leaves a variant of the same bug open.

Consider: `sessionNewEvents[edge] == true`, task T1 (belonging to S1) starts running and blocks on `newEventLock`. S2 connects: `attachSession` cancels T1's future, but `cancel(true)` on an already-started task doesn't stop it. `onEdgeConnect` wins the lock, sets the flag to `true`, unlocks, and arms T2 for S2. T1 then acquires the lock, sees `true`, flips it to `false` here, and runs `processHighPriorityEvents()` / `processEdgeEvents()` **against S1** — the session that's being torn down. Its own re-arm is correctly skipped (`current != session`), but T2 subsequently wakes up, sees `false`, and just re-arms without doing any work.

Net effect: the reconnected edge's pending backlog is never fetched by S2, and it stays unsent until some unrelated event arrives and `updateSessionEventsFlag` lifts the flag again — the same "edge silently stops delivering" symptom this PR is fixing, just reached via a different interleaving. Since the flag is keyed by edge rather than by session, the check body can't tell whose flag it's consuming.

Would it work to apply the same guard at the top of the runnable — bail out (without touching `sessionNewEvents`) if `sessions.get(edgeId) != session`? An interrupted/superseded task then leaves the flag for the session that actually owns the edge.

MazurenkoNick (commented):
## Review summary

Reviewed 1 changed file in [Fix reconnect race where the closing edge session cancels the new session's edge event check](https://github.com/thingsboard/thingsboard/pull/16082). Left 14 comment(s) inline.

The core idea — making "swap the registered session" and "cancel the outgoing session's check" a single `sessions.compute` step, and re-arming only while the arming session is still current — closes the race described in the title, and keeping `destroySession` off the atomic path is the right call. The comments below are mostly about the parts of the same state machine the fix doesn't yet cover: the check body itself still has no currency guard, so a dying session's in-flight task can consume the `sessionNewEvents` flag that `onEdgeConnect` just set for the new session; `edgeIdServiceIdCache.evict` changed reachability; and the zombie sweep still mutates `sessions` directly instead of going through the new helper. There are also no tests for any of the new branches.

## Additional findings

- **application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java:105** — `sessionNewEvents` is a plain `HashMap`, but the lock guarding it (`sessionNewEventsLocks`, line 104) is *per edge*. Two threads handling two different edges take two different locks and mutate the same non-concurrent map at the same time — that's enough to lose entries or corrupt the table on resize. Since this PR now leans harder on this flag being reliable across a reconnect, it seems worth fixing here: either make it a `ConcurrentHashMap` (the per-edge lock would still give you the read-modify-write atomicity you need) or guard it with one shared lock.
- **application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java:109** — `zombieSessions` is a plain `ArrayList`, and `destroySession` (line 611) appends to it from whatever thread happens to be tearing a session down — gRPC handler threads via `onEdgeDisconnect`, the connect path, `deleteEdge` — while the scheduled `cleanupZombieSessions` walks it with `removeIf` (line 730). Unsynchronized concurrent `add` plus iteration can drop entries or throw `ConcurrentModificationException` inside the cleanup task, which would silently kill the sweep for that tick. A `CopyOnWriteArrayList` or a concurrent queue would be a cheap fix given how rarely it's written.
- **application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java:529** — `cancelScheduleEdgeEventsCheck` only removes the map entry *inside* the `!isCancelled() && !isDone()` branch, so whenever the future has already completed the stale entry is left behind in `sessionEdgeEventChecks`. After a detach that entry is never cleaned up (only the next `attachSession` for the same edge overwrites it), so the map keeps an entry per edge that has ever connected. Moving `sessionEdgeEventChecks.remove(edgeId)` out of the inner `if` would make the method actually mean "there is no armed check for this edge any more", which is what all three new call sites are relying on.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -451,55 +450,62 @@ private void scheduleEdgeEventsCheck(EdgeGrpcSession session) {
...
+                        sessionNewEvents.put(edgeId, false);
+                        session.processHighPriorityEvents();
+                        processEdgeEventMigrationIfNeeded(session, edgeId);
+                        if (Boolean.TRUE.equals(edgeEventsMigrationProcessed.get(edgeId))) {
+                            Futures.addCallback(session.processEdgeEvents(), new FutureCallback<>() {
+                                @Override
+                                public void onSuccess(Boolean newEventsAdded) {
+                                    if (Boolean.TRUE.equals(newEventsAdded)) {
+                                        log.trace("[{}][{}] new events added. set session new events flag to true", tenantId, edgeId.getId());
+                                        sessionNewEvents.put(edgeId, true);
```
MazurenkoNick: This `put` runs on `ctx.getGrpcCallbackExecutorService()`, which is a real thread pool (`GrpcCallbackExecutorService extends AbstractListeningExecutor`), so unlike every other access to `sessionNewEvents` it happens without holding `newEventLock` — and `sessionNewEvents` is a plain `HashMap` (line 105). So this is an unsynchronized write racing with the locked read/write a few lines above and with `updateSessionEventsFlag`. Worst case it's a lost "new events arrived" signal, which again means a stalled edge until the next event nudges the flag.

It's pre-existing rather than introduced here, but the whole block moved in this diff and the PR is specifically about this flag being reliable, so it seems worth pulling into the lock (or handling as part of making the map concurrent — see the note on line 105).

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -451,55 +450,62 @@ private void scheduleEdgeEventsCheck(EdgeGrpcSession session) {
...
+        // cancelling the previous one, so that a concurrent connect/disconnect can neither lose the
+        // re-arm nor cancel a check that belongs to a session other than the one it is tearing down
+        sessions.compute(edgeId, (id, current) -> {
+            if (current != session) {
+                log.debug("[{}] Session is not current anymore and edge event check schedule must not be started [{}]",
+                        tenantId, edgeId.getId());
+                return current;
+            }
+            cancelScheduleEdgeEventsCheck(edgeId);
+            sessionEdgeEventChecks.put(edgeId, edgeEventProcessingExecutorService.schedule(edgeEventCheck,
```
MazurenkoNick: This is the per-edge periodic path, re-armed every `noRecordsSleepInterval` for every connected edge, and the re-arm now runs inside a `sessions.compute` remapping function — so each tick takes the map's bin lock and, while holding it, calls `ScheduledFuture.cancel(true)` and `schedule(...)`, nesting the scheduler's internal lock inside the map's. It's cheap in absolute terms, but two small things would shrink the section:

- The `edgeEventCheck` closure is built before the currency check, so on the "not current anymore" path it's allocated and thrown away on every tick. Constructing it after the guard (or moving the body to a method reference) avoids that.
- On the normal re-arm path, `cancelScheduleEdgeEventsCheck(edgeId)` is cancelling the very task that's calling it — the map still holds the running task's future, which isn't `isDone()` yet, so `cancel(true)` fires and self-interrupts the scheduler thread. It happens to be harmless (nothing after it blocks interruptibly, and `ThreadPoolExecutor` clears the flag before the worker's next task), but it's a confusing thing to leave in a hot loop. Skipping the cancel when the armed future is the caller's own would make the intent clearer.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -532,24 +538,20 @@ private void cancelScheduleEdgeEventsCheck(EdgeId edgeId) {
...
-            newEventLock.lock();
-            try {
-                sessionNewEvents.remove(edgeId);
-            } finally {
-                newEventLock.unlock();
-            }
-            destroySession(toRemove);
+        EdgeGrpcSession toRemove = detachSession(edgeId, current -> current.getSessionId().equals(sessionId));
+        if (toRemove != null) {
+            // sessionNewEvents is deliberately left as is: clearing it here could wipe the flag a
```
MazurenkoNick, line 553: The reasoning here makes sense, but it's worth spelling out the consequence in the comment: with this `remove` gone, `sessionNewEvents` is now pruned only in `deleteEdge`, so it retains an entry for every edge that has ever connected to this core node for the process lifetime — joining `sessionNewEventsLocks` and `edgeEventsMigrationProcessed`, which already never shrink. Each entry is tiny so nothing falls over, but that's three permanently growing `EdgeId`-keyed maps with no eviction path other than edge deletion, and the next person to look at a heap dump will diagnose it as a leak.

Since the only reason to keep the entry is the narrow reconnect window, could the cleanup move somewhere the race can't bite — e.g. into the periodic `cleanupZombieSessions` sweep, dropping state for edges with no entry in `sessions`? Failing that, a sentence saying the retention is deliberate and bounded by edge count would save someone the investigation.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -532,24 +538,20 @@ private void cancelScheduleEdgeEventsCheck(EdgeId edgeId) {
...
+            // reconnecting session has just set, and updateSessionEventsFlag only lifts FALSE, not a
+            // missing entry, which would leave the check loop idle forever. deleteEdge still clears it
             sessionsById.remove(sessionId);
+            edgeIdServiceIdCache.evict(edgeId);
```
MazurenkoNick, line 557: Moving the evict from the end of the method into this branch changes when it runs, and the `else` branch is now reachable in cases where the cache probably should still be cleared. Not evicting when a *superseded* session closes is clearly the right fix — the current session on this node legitimately owns the mapping. But the `else` branch is also taken when there's no registered session at all, e.g. after `cleanupZombieSessions` removed it, or after `deleteEdge` detached it and the gRPC stream closes afterwards. In those cases the entry now survives, pointing at this service for an edge (or a deleted edge) that has no session here.

A miss falls back to `broadcastEdgeNotification` (`DefaultTbClusterService:557`), so a stale entry is strictly worse than no entry — it narrows delivery to one node instead of broadcasting. Would it be safer to evict in the `else` branch too when `sessionsById.remove(sessionId)` came back null (nothing of ours left for that edge)?

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -564,7 +566,41 @@ private void onEdgeDisconnect(Edge edge, UUID sessionId) {
...
         }
-        edgeIdServiceIdCache.evict(edgeId);
+    }
+
+    /**
+     * Registers the session as the current one for the edge and cancels the edge event check armed by
+     * the session it replaces, as a single atomic step. Returns the replaced session, or null if there
+     * was none - the caller is responsible for destroying it outside of this critical section.
+     */
+    private EdgeGrpcSession attachSession(EdgeId edgeId, EdgeGrpcSession session) {
```
MazurenkoNick, line 590: Stepping back from the fix itself: this class keeps five parallel `EdgeId`-keyed maps (`sessions`, `sessionEdgeEventChecks`, `sessionNewEvents`, `sessionNewEventsLocks`, `edgeEventsMigrationProcessed`), and after this PR each has a different lifecycle owner and a different set of rules about where it may be touched — `sessions` only via attach/detach, `sessionEdgeEventChecks` only from inside a `compute` block, `sessionNewEvents` only under its per-edge lock and pruned only in `deleteEdge`, the other two never pruned. That's a lot of implicit knowledge to carry, and it's the reason these races keep needing hand-written comments to explain.

Would it be worth collapsing them into one per-edge state holder (session + newEvents flag + check future + migration flag) in a single `ConcurrentMap`? A single `compute` on that map would then cover the whole state transition, attach/detach become trivial, and the lock discipline is enforced by the shape of the code rather than by comments. Definitely a bigger change than this fix warrants — but if reconnect races keep resurfacing here it's probably the cheaper long-term option. (Note `EdgeSessionState` is already taken by a class in this package, so a different name would be needed.)

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -564,7 +566,41 @@ private void onEdgeDisconnect(Edge edge, UUID sessionId) {
...
+        });
+        return replaced.get();
+    }
+
+    /**
+     * Detaches the edge's registered session and cancels its edge event check as a single atomic step,
+     * but only if that session satisfies shouldDetach. Returns the detached session, or null if the
+     * predicate rejected it - most importantly when another session has taken the edge over in the
+     * meantime, in which case nothing is touched and a reconnect racing with a teardown cannot lose
+     * the check it has just armed. The caller destroys the returned session outside the atomic step.
```
MazurenkoNick, line 609: Both helpers hand back a nullable session with an obligation attached: the caller must destroy it, and must do so outside the critical section. That contract lives only in this javadoc, and the three call sites already discharge it three different ways — `deleteEdge` also calls `cleanUp()`, `onEdgeConnect` destroys at the very end of the method, `onEdgeDisconnect` in the middle. Forgetting the follow-up leaks a Kafka consumer plus its executor, and does so silently.

Would it be easier to use correctly if the helpers took the follow-up as a parameter — `attachSession(edgeId, session, Consumer<EdgeGrpcSession> onReplaced)` / `detachSession(edgeId, predicate, Consumer<EdgeGrpcSession> onDetached)` — and invoked it after `compute` returns? Then "outside the lock" is guaranteed in one place and callers can't forget. If you'd rather keep the plain return, `Optional<EdgeGrpcSession>` would at least make the null case explicit at the call site.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -564,7 +566,41 @@ private void onEdgeDisconnect(Edge edge, UUID sessionId) {
...
+     */
+    private EdgeGrpcSession detachSession(EdgeId edgeId, Predicate<EdgeGrpcSession> shouldDetach) {
```
MazurenkoNick, line 614: This PR adds no tests, and the behaviours it introduces are exactly the ones that stay invisible in production until an edge silently stops delivering: the predicate-rejected path in `detachSession` (another session took the edge over mid-teardown), the `current != session` early return in `scheduleEdgeEventsCheck`, the compare-and-remove in `cleanupZombieSessions`, and `onEdgeDisconnect` no longer clearing `sessionNewEvents`. None are covered today — the only tests near this code are `EdgeGrpcSslTest` and `KafkaEdgeGrpcSessionTest`, and the latter tests the session, not this service.

Both harnesses you'd need already exist: `EdgeImitator` has `connect()`/`disconnect()` (`EdgeImitator.java:145`/`:155`) and `DeviceProfileEdgeTest:153` already uses that pattern, so the headline scenario — disconnect, immediately reconnect, assert queued events still arrive — looks achievable as an integration test. And the `ReflectionTestUtils` style of `KafkaEdgeGrpcSessionTest` makes the branch-level cases (attach/detach with a mismatched session, zombie cleanup after a reconnect) unit-testable without standing up gRPC. Given the next refactor of these maps is very likely to reintroduce this, a regression test looks like it pays for itself quickly.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -564,7 +566,41 @@ private void onEdgeDisconnect(Edge edge, UUID sessionId) {
...
+        AtomicReference<EdgeGrpcSession> detached = new AtomicReference<>();
+        sessions.compute(edgeId, (id, current) -> {
+            if (current == null || !shouldDetach.test(current)) {
+                return current;
+            }
+            detached.set(current);
+            cancelScheduleEdgeEventsCheck(edgeId);
```
MazurenkoNick, line 621: `cancelScheduleEdgeEventsCheck` now has an unwritten precondition: it's only correct when called from inside `sessions.compute(edgeId, ...)` for that same edge, since that's what makes the cancel atomic with respect to the arming in `scheduleEdgeEventsCheck`. All three new call sites honour it, but the method itself says nothing — so the obvious next change ("the edge is going away, let me just cancel its check here") quietly reintroduces the race this PR fixes. Worth either a short javadoc stating the requirement or a name that carries it.

Related: the `@PreDestroy destroy()` method still hand-rolls the same cancel-and-remove loop over `sessionEdgeEventChecks` outside that discipline. Harmless at shutdown, but it's now the one place that duplicates this logic and diverges from the new rule, so folding it onto the helper (or noting why shutdown is exempt) would keep the invariant uniform.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -686,7 +722,9 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
 
     private void cleanupZombieSessions() {
         try {
-            tryToDestroyZombieSessions(getZombieSessions(sessions.values()), s -> sessions.remove(s.getEdge().getId()));
+            // compare-and-remove: the edge may have reconnected since the zombie scan, and that newer
+            // session must not be evicted by the cleanup of the one it replaced
+            tryToDestroyZombieSessions(getZombieSessions(sessions.values()), s -> sessions.remove(s.getEdge().getId(), s));
```
MazurenkoNick: The compare-and-remove is the right instinct, but this is the one place left that mutates `sessions` without going through the new helper, and it loses both of the properties the helper provides.

- No `cancelScheduleEdgeEventsCheck`, so the zombie's armed future survives the removal and fires once against a session whose consumer has already been nulled out by `destroy()`. The new currency guard does stop the loop re-arming, so it's one wasted (probably exception-logging) tick rather than a leak — but it's avoidable.
- `ConcurrentMap.remove(key, value)` reads as an identity CAS, but it isn't: `EdgeGrpcSession` is annotated `@Data`, so it has a Lombok-generated field-by-field `equals`, and CHM only tries identity first before falling back to `equals`. It happens to behave identically here (the first compared field is a `ConcurrentLinkedQueue`, which uses identity equality, so two distinct sessions differ immediately), but that's an accident of field ordering in another class — someone reordering fields or adding `@EqualsAndHashCode` to a subclass could turn this into a wrong removal.

Both go away with `detachSession(s.getEdge().getId(), current -> current == s)`, which is exactly what the new helper is for.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
MazurenkoNick: Agreed, done. The three inline copies are gone and the rule is stated once:

- `destroySession` now has a javadoc line saying it can block for seconds (`KafkaEdgeGrpcSession.destroy()` gives the consumer executor up to 5s to terminate) and must never run inside a `sessions` remapping function.
- `attachSession` documents the rule itself: only cheap non-blocking work inside the remapping function, since it holds the map's bin lock for the edge - and names what belongs outside it (`destroySession`, `cleanUp()`, `save(...)`, `pushRuleEngineMessage(...)`, acquiring `newEventLock`).
- `detachSession` points at the same rule.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
MazurenkoNick: Good catch, this is real. Fixed.

One correction to the proposed placement though: the guard has to go *inside* `newEventLock`, not at the top of the runnable. Before the lock it only moves the race - the task reads `sessions.get(edgeId) == S1`, then blocks on the lock, and `onEdgeConnect` still raises the flag in between.

Inside the lock it is airtight, because `attachSession` publishes the new session *before* `onEdgeConnect` takes `newEventLock` to raise the flag. So if the check holds the lock and still sees its own session, the flag-raise provably has not happened yet; and if it sees a different session, it bails without touching the flag and the new session consumes it on its own tick.

```java
newEventLock.lock();
try {
    if (sessions.get(edgeId) != session) {
        return; // leave the flag for whoever owns the edge now
    }
    if (Boolean.TRUE.equals(sessionNewEvents.get(edgeId))) { ...
```

It deliberately does not re-arm on the bail-out path - the session that owns the edge already has its own check armed.

Covered by `EdgeGrpcServiceReconnectTest.checkOfASupersededSessionLeavesTheFlagForTheSessionThatOwnsTheEdge`.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
MazurenkoNick: Fair point, but leaving it as is. The self-interrupt is harmless for the reasons you list, and skipping the cancel only when the armed future is the caller's own would add a branch to the same `compute` block that the rest of this PR is trying to keep minimal. Happy to revisit if it ever grows teeth.

The closure allocation on the "not current anymore" path is gone, though - the body moved into `processEvents(...)` and the schedule now takes a method-reference-style lambda constructed after the guard.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
MazurenkoNick: Added `EdgeGrpcServiceReconnectTest` - plain unit tests (Mockito + `ReflectionTestUtils`, no gRPC, no Spring context, ~1.3s for the class):

- `checkIsArmedForTheCurrentSession` / `checkIsNotArmedForASupersededSession`
- `disconnectOfTheCurrentSessionDetachesItAndCancelsItsCheck` / `disconnectOfASupersededSessionKeepsTheReconnectedSessionAndItsCheck`
- `disconnectDoesNotClearTheNewEventsFlag`
- `checkOfTheCurrentSessionConsumesTheFlagAndReArms` / `checkOfASupersededSessionLeavesTheFlagForTheSessionThatOwnsTheEdge`
- `zombieCleanupCancelsTheCheckOfTheSessionItRemoves` / `zombieCleanupKeepsASessionThatReconnectedDuringTheSweep`

The reconnect-during-sweep case is made deterministic by stubbing `zombie.destroy()` to install the replacement session, which is exactly the window between the zombie scan and the removal.

Each guard was mutation-tested to confirm the tests actually bite:

| reverted | test that fails |
| --- | --- |
| execution-side currency guard | `checkOfASupersededSessionLeavesTheFlagForTheSessionThatOwnsTheEdge` |
| zombie cleanup via `detachSession` | `zombieCleanupCancelsTheCheckOfTheSessionItRemoves` |
| re-arm currency guard + unconditional detach | `checkIsNotArmedForASupersededSession`, `disconnectOfASupersededSessionKeepsTheReconnectedSessionAndItsCheck` |
| `sessionNewEvents.remove` on disconnect | `disconnectDoesNotClearTheNewEventsFlag` |

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
MazurenkoNick: Both points check out, fixed - `cleanupZombieSessions` now uses `detachSession(s.getEdge().getId(), current -> current == s)`.

On the armed future: with the new currency guard in `processEvents` it can no longer consume the new events flag, so the blast radius was one exception-logging tick plus a stale entry in `sessionEdgeEventChecks` - but it was avoidable, and this was the last path mutating `sessions` outside the helper.

On `remove(key, value)`: agreed it reads as an identity CAS but is not one. It happens to behave like identity today only because the first field Lombok compares is a `ConcurrentLinkedQueue`, which is an accident of field ordering in another class. `current == s` makes it explicit.

Covered by `EdgeGrpcServiceReconnectTest.zombieCleanupCancelsTheCheckOfTheSessionItRemoves` and `zombieCleanupKeepsASessionThatReconnectedDuringTheSweep`.

PR #16101, opened 2026-09-01
Validate prevUri redirect parameter

ViacheslavKlimov:
The `prevUri` request parameter is stored in a cookie and later concatenated to the platform base URL to build the post-login redirect (OAuth2 login flow) and the mail settings redirect (`AdminController` mail OAuth2 callback). Neither the write nor the read site validated it, so the value could point the redirect at another host.

`prevUri` is now accepted only when it is a relative in-app path, which is what the UI actually sends (`router.url`): a single leading `/`, and no whitespace, control characters or backslashes. That leaves no room for a scheme, userinfo or authority, and blocks protocol-relative (`//`, `/\`) paths and redirect header splitting.

The check is applied both where the cookie is written and where it is read, in `HttpCookieOAuth2AuthorizationRequestRepository` / `Oauth2AuthenticationSuccessHandler` and in `AdminController`. An invalid value is ignored, and the redirect falls back to the base URL (or `/settings/outgoing-mail` for the mail flow).


ViacheslavKlimov:
Superseded by #16102, which targets lts-4.2 with an identical change and will merge forward into lts-4.3.

PR #16109, opened 2026-09-03
Report every conflicting entity before installing a solution template

pinkevmladchy:
Installing a solution template validated only rule chains, profiles and dashboards upfront, so a template with a clashing customer, asset, device or edge died mid install on a DB constraint and had to be rolled back. And the report only said that *some* entities already exist, without naming them.

- customers, assets, devices and edges are validated before anything is created, so every conflict is reported at once instead of failing on the first one
- each conflict is named, and the same sentence is no longer repeated per entity type:

```
Some entities of the solution template already exist. Rename or delete them and install the template again:

- **Asset profile**: 'Silo'
- **Asset**: 'Silo 1'
```

- rule chains are looked up by exact name and dashboards by exact title — the paged text search matches with `contains`, so a different entity could be reported as the conflict
- the install dialog no longer repeats "Failed to install '<item>'" inside the error details, since it is already shown right above them

Scoped to the solution template install only: the existing entity-save error messages are left untouched.


application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +300,117 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+        return "'" + value + "'";
+    }
+
+    /**
+     * Dashboard titles are not unique per tenant, so there is no lookup by title: the text search narrows the
+     * candidates down and the title is matched exactly here, to not report a dashboard whose title merely contains
+     * the title of the template dashboard.
+     */
+    private DashboardInfo findDashboardByTitle(TenantId tenantId, String title) {
+        return dashboardService.findDashboardsByTenantId(tenantId, new PageLink(DASHBOARD_TITLE_SEARCH_PAGE_SIZE, 0, title)).getData().stream()
```
ViacheslavKlimov, line 406: The javadoc above says "there is no lookup by title", but `DashboardService.findFirstDashboardInfoByTenantIdAndName(TenantId, String)` is exactly that — it resolves to `DashboardInfoRepository.findFirstByTenantIdAndTitle`, an exact-title query, and `findFirst` already handles the non-uniqueness the comment is worried about.

As written this only looks at the first 100 rows of an `ilike '%title%'` search (`DashboardInfoRepository.findByTenantId` matches with `CONCAT('%', :searchText, '%')`), so a tenant with more than 100 dashboards whose titles contain the template's title can have the exactly-matching one fall outside the page, and the conflict then goes unreported — which is at odds with the "list every conflict up front" contract this validation is meant to provide.

Switching to `dashboardService.findFirstDashboardInfoByTenantIdAndName(tenantId, title)` gives the exact semantics you want, drops both `DASHBOARD_TITLE_SEARCH_PAGE_SIZE` and the post-filter, and turns it into one indexed lookup per dashboard instead of scanning up to 100 rows.

ViacheslavKlimov (commented):
## Review summary

Reviewed 26 changed files in [Name the conflicting entity in "already exists" errors](https://github.com/thingsboard/thingsboard/pull/16109). Left 1 comment inline.

The centralisation through `EntityConflictMessages` and the up-front solution-template validation both look solid. The one thing worth a look is `findDashboardByTitle` in `DefaultSolutionService` — its javadoc says there is no lookup by title, but `DashboardService.findFirstDashboardInfoByTenantIdAndName` is exactly that, and the paged text-search fallback it uses instead can miss a genuine conflict.

---

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Good catch, you're right — I missed that method. Switched to findFirstDashboardInfoByTenantIdAndName, so the page size and the post-filter are gone. Fixed in 2d0c8e4784.

dao/src/main/java/org/thingsboard/server/dao/asset/AssetProfileServiceImpl.java
```diff
@@ -159,7 +163,7 @@ public AssetProfile saveAssetProfile(AssetProfile assetProfile, boolean doValida
             handleEvictEvent(new AssetProfileEvictEvent(assetProfile.getTenantId(), assetProfile.getName(),
                     oldAssetProfile != null ? oldAssetProfile.getName() : null, null, assetProfile.isDefault()));
             checkConstraintViolation(t,
-                    Map.of("asset_profile_name_unq_key", ASSET_PROFILE_WITH_SUCH_NAME_ALREADY_EXISTS,
```
irynamatveieva, line 149: Could we scope this improvement to the solution template install and leave the existing service messages alone? These are the 400 body of every entity-save endpoint. That's a silent break for anyone matching the text.

`DefaultSolutionService` doesn't reference `EntityConflictMessages` at all, so it's just dropping the DAO changes — the template validation and the conflict report stay.

pinkevmladchy: Agreed, scoped it down — dropped all the service message changes and EntityConflictMessages with them, so only the template install validation and the conflict report are left. Same in #16117.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Heads up: the branch was rewritten to scope the PR down to the template install, so 2d0c8e4784 is gone — the exact title lookup is part of the single commit now.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,102 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+        });
+        List<DeviceDefinition> devices = loadListOfEntitiesIfFileExists(tempDir, "devices.json", new TypeReference<>() {
+        });
+        List<EdgeDefinition> edges = loadListOfEntitiesIfFileExists(tempDir, "edges.json", new TypeReference<>() {
+        });
+
+        // Insertion ordered, so that the reported sections keep the order the entities are provisioned in.
+        Map<EntityType, List<HasName>> conflicts = new LinkedHashMap<>();
+        collectConflicts(conflicts, EntityType.RULE_CHAIN, ruleChains, ReferenceableEntityDefinition::getName,
+                name -> firstOrNull(ruleChainService.findTenantRuleChainsByTypeAndName(tenantId, RuleChainType.CORE, name)));
```
ShvaykaD: For rule chains the validated name comes from `rule_chains.json`'s `name`, but `provisionRuleChains` never uses that field — the created rule chain takes its name from `ruleChain.name` inside `rule_chains/<file>.json`. Is the definition `name` guaranteed to mirror the name in the file? If they can diverge, this now checks a name that will never be created (the old `contains` search had the same blind spot, but exact matching makes a divergence a guaranteed miss rather than a partial one). Reading the name out of the loaded rule chain JSON — the same source `provisionRuleChains` uses — would keep the check honest for whatever the template actually creates.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,102 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+    /**
+     * One markdown list item per entity type, so that the names of the conflicting entities are what the user reads,
+     * instead of the same sentence repeated for every type.
+     */
+    private void appendConflicts(StringBuilder details, EntityType entityType, List<String> conflictDescriptions) {
+        if (conflictDescriptions.isEmpty()) {
+            return;
         }
+        details.append("- **").append(entityType.getNormalName()).append("**: ")
+                .append(conflictDescriptions.stream().distinct().collect(Collectors.joining(", "))).append(System.lineSeparator());
```
ShvaykaD: The details now end with a single line separator after the last bullet, where the old format ended with `;` + a blank line. `handlePlanResult` appends ` ' ' + install-rollback-partial` to this message on a partial rollback, so that warning becomes a lazy continuation of the last list item and renders inside the bullet instead of as its own paragraph (only for a `SOLUTION_TEMPLATE` root, which is the one case rendered through `tb-iot-hub-markdown`). Ending the details with a blank line would keep anything appended downstream in its own paragraph.

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -952,7 +952,7 @@ public InstallPlanResult installPlan(SecurityUser user, InstallPlan plan, JsonNo
                         boolean rolledBack = rollbackInstalledItems(user, rollbackIds);
                         result.setSuccess(false);
                         result.setRolledBack(rolledBack);
-                        result.setErrorMessage("Failed to install '" + entry.getName() + "': " + e.getMessage());
+                        result.setErrorMessage(e.getMessage());
```
ShvaykaD: `e.getMessage()` can be null (an NPE from deep in the install path, for instance). The failing entry's `errorMessage` is null too, so the dialog's `find(entry => !entry.root && entry.errorMessage)` doesn't match and the user is left with just "Failed to install '<root>'." — the failing dependency is no longer named anywhere. Before this change the composed prefix at least identified the entry. Something like `StringUtils.defaultIfEmpty(e.getMessage(), e.getClass().getSimpleName())` would keep both the entry and the result message non-empty.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
```diff
@@ -339,6 +339,12 @@ export class TbIotHubInstallDialogComponent extends DialogComponent<TbIotHubInst
     if (!result.success) {
       this.state = 'error';
       let message = result.errorMessage || this.translate.instant('iot-hub.install-error', { name: this.item.name });
+      // The title already names the item being installed, so only a failing dependency has to be
+      // named here - the service reports the failure itself and does not compose this sentence.
+      const failedDependency = (result.entries ?? []).find(entry => !entry.root && entry.errorMessage);
+      if (failedDependency) {
+        message = this.translate.instant('iot-hub.install-dependency-error', { name: failedDependency.name }) + ' ' + message;
```
ShvaykaD: On the fallback path (server sent no `errorMessage`) `message` is the generic `iot-hub.install-error` rendered with the *root* item's name, so prefixing this sentence produces "Failed to install the dependency 'Dep': Failed to install <item>", which names the wrong thing after the colon. Worth only prefixing when `result.errorMessage` was actually present, or giving the fallback its own message.

ShvaykaD (commented):
## Review summary

Reviewed 6 changed files in [Report every conflicting entity before installing a solution template](https://github.com/thingsboard/thingsboard/pull/16109). Left 4 comment(s) inline.

The validation rework is sound: every new lookup (`findCustomerByTenantIdAndTitle`, `findAssetByTenantIdAndName`, `findDeviceByTenantIdAndName`, `findEdgeByTenantIdAndName`) maps onto a real tenant-scoped unique constraint, all names are used verbatim at provision time except customer titles (which is exactly what `isRandomizedAtInstall` guards — `randomize()` only ever substitutes `$random*` tokens), and dropping the backend prefix lines up with the dialog already rendering "Failed to install '<item>'." above the details. The comments below are about the name the rule chain check actually looks up, the null-message edge on the new `errorMessage`, and the markdown the details now produce.

---

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Confirmed — provisionRuleChains builds the chain from ruleChainJson.get("ruleChain"), so the definition name was never the created one. The check now reads the name out of the rule chain file, same source as the install, and skips definitions whose file is missing (provisionRuleChains skips those too). Covered by testValidateSolutionUsesTheRuleChainNameFromTheRuleChainFile.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Right, the appended warning ended up inside the last bullet. The details now end with a blank line, and the test asserts it.

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
pinkevmladchy: Good catch. Both the entry and the result message now fall back to the exception's simple name when getMessage() is null, so the dialog can always find the failing entry. Pinned by installPlan_failureWithoutMessage_reportsTheExceptionType.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
pinkevmladchy: Fixed — the dependency prefix is only added when the server actually sent an error message, so the generic fallback (which names the root item) is left alone.

pinkevmladchy:
Sorry for the churn on this branch — I force-pushed it several times while scoping the PR down, which wiped the SHAs your comments referenced and made "changes since your review" span the whole PR instead of just the fixes.

The branch now carries the same tree in three commits, one per review step, so incremental review works again:

- `report every conflicting entity before installing a solution template` — the scoped change
- `compose the failed dependency sentence in the dialog, not in the service` — the untranslatable sentence moved to the UI
- `address review comments on the install validation` — the four findings from the last pass

The earlier SHAs (`2d0c8e4784` and the ones before it) are still reachable from the force-push events in the timeline if you need the old diff. No more rewrites from here — follow-ups will be plain commits on top.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -177,6 +178,11 @@
 @Slf4j
 public class DefaultSolutionService implements SolutionService {
 
+    private static final String BLANK_LINE = System.lineSeparator() + System.lineSeparator();
+    private static final String CONFLICTS_INTRO =
```
ShvaykaD, line 170: This is user-facing copy hardcoded in English on the server, while every other string in this dialog goes through the translation files — this PR itself adds `iot-hub.install-dependency-error` to the locale bundle. The entity type labels from `getNormalName()` are in the same boat. Since the server has to interpolate the entity names anyway, one option is to return the conflicts structurally (type → names) and let the dialog render the sentence from a locale key. If that's too much for this fix it's at least worth a note, because as written the conflict report can never be localized. Related: `device-install-conflict-exists` (`locale.constant-en_US.json:3921`) is already the vocabulary for this same situation in the sibling device-install flow.

ShvaykaD (commented):
## Re-review summary

Re-reviewed [Report every conflicting entity before installing a solution template](https://github.com/thingsboard/thingsboard/pull/16109) — verified 4 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/16109#pullrequestreview-5131013294).

| Status | Count |
|--------|-------|
| ✅ Resolved | 4 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 0 |

All four previous findings are addressed. Found 13 new issue(s), commented inline.

The branch now carries three commits, and `56866d30c8` reproduces exactly the tree the previous review saw, so the fix delta is the single commit `0d275e0d2c` (83 insertions, 8 deletions across 5 files). Of the new comments, four sit on that fix commit — `DefaultIotHubService.java:953`, `DefaultSolutionService.java:325` and `:348`, and `iot-hub-install-dialog.component.ts:346`. The other nine are on code that was already present at the previous review; they are gaps that pass missed, not regressions introduced by the fixes.

The one worth acting on before merge is `iot-hub-install-dialog.component.ts:346`: the failing-entry lookup can match a MISSING plan entry, which carries an `errorMessage`, `root === false` and a null `name`.

### Finding details

- ✅ **DefaultSolutionService.java** — rule chain validated under the `rule_chains.json` name rather than the name the install creates — *Fixed in code: `ruleChainName` now reads `ruleChain.name` out of the rule chain file, the same source `provisionRuleChains` uses, and skips definitions whose file is missing. Covered by `testValidateSolutionUsesTheRuleChainNameFromTheRuleChainFile`.*
- ✅ **DefaultSolutionService.java** — details ended without a blank line, so the dialog's appended warning rendered inside the last bullet — *Fixed in code: `details.append(System.lineSeparator())` after the list, asserted by `endsWith(sep + sep)` in the new test.*
- ✅ **DefaultIotHubService.java** — `e.getMessage()` can be null, leaving both the entry and the result message blank — *Fixed in code: `StringUtils.isEmpty(e.getMessage()) ? e.getClass().getSimpleName() : e.getMessage()` applied to both. Pinned by `installPlan_failureWithoutMessage_reportsTheExceptionType`.*
- ✅ **iot-hub-install-dialog.component.ts** — the dependency prefix was applied on the fallback path, which names the root — *Fixed in code: the prefix is only added when `result.errorMessage` was actually present.*

---

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -177,6 +178,11 @@
...
+            "Some entities of the solution template already exist. Rename or delete them and install the template again:";
+    private static final String RANDOM_NAME_PLACEHOLDER_PREFIX = "$random";
```
ShvaykaD: Two naming things that could mislead later. The constant is called `..._PREFIX` but it's used with `contains()`, and legitimately so — a customer title like `Customer $random` has the placeholder at the end, so `PREFIX` describes neither the check nor the data. And `isRandomizedAtInstall(String name)` reads like a general predicate over any entity name, while it only holds for customer titles; the javadoc explains that, but the signature invites someone to reuse it for devices or assets, where the opposite is true. `RANDOM_PLACEHOLDER` and `isRandomizedCustomerTitle` would put the constraint in the name rather than in the doc.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,122 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
-                    alreadyExistingEntities.computeIfAbsent(EntityType.DASHBOARD, key -> new ArrayList<>()).add(savedDashboards.get(0));
-                }
+        List<AssetDefinition> assets = loadListOfEntitiesIfFileExists(tempDir, "assets.json", new TypeReference<>() {
+        });
+        List<DeviceDefinition> devices = loadListOfEntitiesIfFileExists(tempDir, "devices.json", new TypeReference<>() {
+        });
+        List<EdgeDefinition> edges = loadListOfEntitiesIfFileExists(tempDir, "edges.json", new TypeReference<>() {
+        });
+
+        // Insertion ordered, so that the reported sections keep the order the entities are provisioned in.
```
ShvaykaD, line 314: The comment doesn't match the code: `doInstallSolution` provisions rule chains → device profiles → asset profiles → customers → assets → devices → dashboards → … → edges, while the collect order here puts DASHBOARD before ASSET and DEVICE. Either move the dashboard call after the device one, or reword to say the map is just insertion-ordered for a stable report. A comment stating an invariant the code doesn't hold is worse than no comment, since the next person will trust it.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,122 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+        Map<EntityType, List<HasName>> conflicts = new LinkedHashMap<>();
+        collectConflicts(conflicts, EntityType.RULE_CHAIN, ruleChains, definition -> ruleChainName(tempDir, definition),
+                name -> firstOrNull(ruleChainService.findTenantRuleChainsByTypeAndName(tenantId, RuleChainType.CORE, name)));
```
ShvaykaD: `ruleChainName` now opens the rule chain file to read the name the install will actually create — and the type lives in that same JSON. `provisionRuleChains` builds the chain with `treeToValue(ruleChainJson.get("ruleChain"), RuleChain.class)`, so a template shipping an EDGE rule chain gets created as EDGE but is validated against `RuleChainType.CORE` here and never matches. Since the file is already parsed, returning the type alongside the name and passing it to `findTenantRuleChainsByTypeAndName` would close that. The old paged search made the same CORE assumption, so it isn't new — but this is now the one place that has the file open.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,122 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+                name -> deviceProfileService.findDeviceProfileByName(tenantId, name));
+        collectConflicts(conflicts, EntityType.ASSET_PROFILE, assetProfiles, AssetProfile::getName,
+                name -> assetProfileService.findAssetProfileByName(tenantId, name));
+        collectConflicts(conflicts, EntityType.CUSTOMER, customers, CustomerDefinition::getName,
+                title -> isRandomizedAtInstall(title) ? null : customerService.findCustomerByTenantIdAndTitle(tenantId, title).orElse(null));
+        collectConflicts(conflicts, EntityType.DASHBOARD, dashboards, DashboardDefinition::getName,
+                title -> dashboardService.findFirstDashboardInfoByTenantIdAndName(tenantId, title));
+        collectConflicts(conflicts, EntityType.ASSET, assets, AssetDefinition::getName,
+                name -> assetService.findAssetByTenantIdAndName(tenantId, name));
+        collectConflicts(conflicts, EntityType.DEVICE, devices, DeviceDefinition::getName,
```
ShvaykaD: Validation is now one point query per definition, run serially, and devices/assets are the two lists that can realistically be large in a template — the old code only walked rule chains, profiles and dashboards, which are a handful. A 5k-device template issues 5k selects before a single row is inserted. It's still proportional to the inserts the install does anyway, so it may well be acceptable, but two cheap improvements would help: dedupe the names before looking them up (`distinct()` is currently applied only when formatting the output, so a template listing the same name twice queries twice), and consider whether the definition lists loaded here could be handed to `doInstallSolution` instead of every `provisionX` re-reading and re-parsing the same JSON files a second time.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,122 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+                name -> edgeService.findEdgeByTenantIdAndName(tenantId, name));
+
+        if (conflicts.isEmpty()) {
+            return null;
+        }
+        StringBuilder details = new StringBuilder(CONFLICTS_INTRO).append(BLANK_LINE);
+        conflicts.forEach((entityType, entities) -> appendConflicts(details, entityType,
+                entities.stream().map(HasName::getName).map(DefaultSolutionService::quoted).toList()));
+
+        // the dialog appends its own sentences to these details, so the list has to end with a blank line
```
ShvaykaD: This bakes a presentation contract into the service: the markdown plus a trailing blank line exist only because `handlePlanResult` concatenates its own sentences onto the same string, and the contract is enforced only by an assertion in a unit test. Nothing on the Angular side hints that whatever it appends must be preceded by a blank line, so the next person editing either end can break the rendering silently. Would it be cleaner for the dialog to keep the server message and its own appended sentences as separate blocks — separate DOM elements, or joined with an explicit `\n\n` on the client — so the server just produces a markdown list and doesn't have to know how the message gets assembled?

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,122 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+        details.append(System.lineSeparator());
+
+        SolutionInstallResponse solutionInstallResponse = new SolutionInstallResponse();
+        solutionInstallResponse.setSuccess(false);
+        solutionInstallResponse.setDetails(details.toString());
+        return solutionInstallResponse;
+    }
+
+    private <T> void collectConflicts(Map<EntityType, List<HasName>> conflicts, EntityType entityType, List<T> definitions,
```
ShvaykaD: Five of the eight call sites pass an `EntityType` and a `::getName` that the definition already carries — `CustomerDefinition`, `DashboardDefinition`, `AssetDefinition`, `DeviceDefinition` and `EdgeDefinition` all reach `EntityDefinition` through `BaseEntityDefinition`, which declares both `getName()` and `getEntityType()`. An overload taking `List<? extends EntityDefinition>` plus the lookup would drop those two redundant arguments and remove the copy-paste hazard of pairing, say, `EntityType.ASSET` with the `devices` list — nothing in the current signature would catch that. The rule chain and the two profile lists would keep the explicit form, which is fine since they genuinely need it.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,122 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+    private static boolean isRandomizedAtInstall(String name) {
+        return name.contains(RANDOM_NAME_PLACEHOLDER_PREFIX);
+    }
+
+    /**
+     * One markdown list item per entity type, so that the names of the conflicting entities are what the user reads,
+     * instead of the same sentence repeated for every type.
+     */
+    private void appendConflicts(StringBuilder details, EntityType entityType, List<String> conflictDescriptions) {
+        if (conflictDescriptions.isEmpty()) {
```
ShvaykaD: This guard can't fire — `collectConflicts` only creates a list via `computeIfAbsent` at the moment it adds an element, so the map never holds an empty list. The formatting duties are also split oddly: the caller maps entities to names and quotes them, and the callee then dedupes and joins the already-formatted strings. Doing name → dedupe → quote → join in one place (or deduping when collecting, which would also cut the duplicate lookups) would read more directly.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,71 +297,122 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+        Path ruleChainPath = tempDir.resolve("rule_chains").resolve(definition.getFile());
+        if (!Files.exists(ruleChainPath)) {
+            // provisionRuleChains logs and skips such a definition, so there is nothing to validate
+            return null;
+        }
+        JsonNode ruleChain = JacksonUtil.toJsonNode(ruleChainPath).get("ruleChain");
+        return ruleChain != null && ruleChain.hasNonNull("name") ? ruleChain.get("name").asText() : null;
+    }
+
+    private static <T> T firstOrNull(Collection<T> data) {
```
ShvaykaD: A generic helper for a single call site, where the codebase already has an inline idiom for exactly this — `BaseRuleChainService.java:558` writes `findTenantRuleChainsByTypeAndName(...).stream().findFirst()`. Inlining `.stream().findFirst().orElse(null)` at the one lambda would keep the class smaller and match the existing pattern; the helper earns its place once a second caller appears.

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
```diff
@@ -947,12 +948,15 @@ public InstallPlanResult installPlan(SecurityUser user, InstallPlan plan, JsonNo
                     } catch (Exception e) {
                         log.error("[{}] Cascade install failed at entry {} ({}): {}", tenantId,
                                 entry.getName(), entry.getVersionId(), e.getMessage(), e);
-                        resultEntry.setErrorMessage(e.getMessage());
+                        // an exception without a message would leave both the entry and the result blank,
+                        // and the dialog would have nothing to name the failing entry by
+                        String failure = StringUtils.isEmpty(e.getMessage()) ? e.getClass().getSimpleName() : e.getMessage();
```
ShvaykaD: `ExceptionUtil.getMessage(Throwable)` in `common/util` is this exact fallback — message if non-empty, otherwise `getClass().getSimpleName()`. Reusing it would let the two-line comment go away too, since the helper name says what it does. Minor: `failure` reads like a `Throwable` rather than a `String` — `failureMessage` would be clearer.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
```diff
@@ -338,7 +338,15 @@ export class TbIotHubInstallDialogComponent extends DialogComponent<TbIotHubInst
   private handlePlanResult(result: InstallPlanResult): void {
     if (!result.success) {
       this.state = 'error';
-      let message = result.errorMessage || this.translate.instant('iot-hub.install-error', { name: this.item.name });
+      const serverError = result.errorMessage;
+      let message = serverError || this.translate.instant('iot-hub.install-error', { name: this.item.name });
+      // The title already names the item being installed, so only a failing dependency has to be named
+      // here - and only when the server reported the failure, since the fallback above names the root.
+      const failedDependency = serverError
+        ? (result.entries ?? []).find(entry => !entry.root && entry.errorMessage) : undefined;
```
ShvaykaD: This can match an entry the cascade never touched. `installPlan` copies every processed entry through `cloneEntry`, which carries `errorMessage` over, and the MISSING entries built by `missingEntry()` (`DefaultIotHubService.java:877`) always carry one — "Item not found or not published" / "Item descriptor is incomplete" — with `root` false and `name` null. A plan containing a missing dependency is installable (the plan step only warns via `install-plan-missing-warning`), so if such an entry precedes the failing one, the dialog renders "Failed to install the dependency '': <real failure>" — wrong item, empty name. The failing entry keeps its `WILL_INSTALL` status (only `errorMessage` is set on it), so `entry.status === InstallPlanEntryStatus.WILL_INSTALL && entry.errorMessage` would pin it to the right one.

More broadly, identifying the failing entry by the truthiness of a free-text field is what forces the backend's class-name fallback in this same PR — the new test comment says as much ("the dialog looks the failing entry up by its error message, so it must never be blank"). That's a fragile contract across two codebases. A `FAILED` value on `InstallPlanEntry.Status` would make this an explicit query and let the message field stay optional, even if the class-name fallback is kept for the user-visible text.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
```diff
@@ -338,7 +338,15 @@ export class TbIotHubInstallDialogComponent extends DialogComponent<TbIotHubInst
...
+      if (failedDependency) {
+        message = this.translate.instant('iot-hub.install-dependency-error', { name: failedDependency.name }) + ' ' + message;
```
ShvaykaD: The new key is a sentence fragment ending in a colon that gets glued to the server message (and possibly to `install-rollback-partial` after it). Translators get no control over ordering or punctuation, and languages that put the subject last end up with something awkward. A single key with two parameters — `"install-dependency-error": "Failed to install the dependency '{{name}}': {{error}}"` — keeps the whole sentence in the bundle and removes the manual `+ ' ' +`.

application/src/test/java/org/thingsboard/server/service/solutions/DefaultSolutionServiceTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+        SolutionInstallResponse result = service.validateSolution(tenantId, tempDir);
+
+        assertThat(result).isNotNull();
+        assertThat(result.getDetails().lines().toList())
+                .containsSubsequence(CONFLICTS_INTRO, "- **Dashboard**: 'Overview'");
+    }
+
+    @Test
+    void testValidateSolutionUsesTheRuleChainNameFromTheRuleChainFile() throws IOException {
```
ShvaykaD, line 133: Good that the new validation got a test at all, but several new paths are uncovered. Four of the eight entity types are exercised (customer, device, dashboard, rule chain) while DEVICE_PROFILE, ASSET_PROFILE and EDGE — added by this PR just like the others — are not, including the profile lookups that also come from a directory rather than a single file. `ruleChainName` has two new fallback branches (the referenced file doesn't exist; the file has no `ruleChain`/`name` node) that no test enters, and the `Objects.requireNonNullElseGet(..., ArrayList::new)` change to `loadListOfEntitiesIfFileExists` (a JSON file containing `null`) has none either. The `distinct()` dedupe in `appendConflicts` is likewise untested. Unit tests are a fair fallback here since there's no controller-level harness for solution installs, but the parallel cases deserve the same consistency as the ones that are covered. Minor: this file uses `testXxx` naming while the sibling `DefaultIotHubServiceTest` uses `method_scenario_expectation`.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
pinkevmladchy: This one was a real bug, thanks — a MISSING entry does carry an errorMessage with a null name, so it could hijack the sentence. The lookup now also requires WILL_INSTALL. Adding a FAILED status to InstallPlanEntry.Status would be the sturdier contract, but that changes the plan model shared with the marketplace client, so I'd rather do it outside this fix.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
pinkevmladchy: Done — install-dependency-error now takes {{name}} and {{error}}, so the whole sentence stays in the bundle, and the rollback warning is appended by the dialog as its own paragraph.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Fixed — the type is read from the same rule chain file as the name, defaulting to CORE when absent, so an EDGE chain is looked up among EDGE chains. Covered by testValidateSolutionUsesTheRuleChainTypeFromTheRuleChainFile.

application/src/main/java/org/thingsboard/server/service/iot_hub/DefaultIotHubService.java
pinkevmladchy: Switched to ExceptionUtil.getMessage(e) and renamed the variable to failureMessage; the comment went away with it.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Renamed to RANDOM_PLACEHOLDER and isRandomizedCustomerTitle — the constraint reads off the signature now.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Moved the dashboard check after the device one, so the order matches doInstallSolution and the comment is true again.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Names are deduplicated before the lookup now (a Set per list), so a repeated name queries once. Handing the parsed definitions to doInstallSolution so the provisionX methods stop re-reading the same files is a bigger refactor than this fix should carry — worth its own change.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Agreed, the contract was the wrong way round. The service no longer appends a trailing separator: the dialog joins its own sentences with a blank line, so the markdown list is all the server produces.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Added that overload — the five EntityDefinition lists now take the type and the name from the definition itself, and only the two profile lists and the rule chains keep the explicit form.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Right, the guard was dead. Formatting is one step now (entities → quoted names → join) and the dedupe moved to collection.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Inlined .stream().findFirst().orElse(null) and dropped the helper.

application/src/test/java/org/thingsboard/server/service/solutions/DefaultSolutionServiceTest.java
pinkevmladchy: Added tests for both profile types and edges, the rule chain type, both unresolvable-name branches of the rule chain lookup, the deduplicated lookup and a json file holding null. Left the testXxx naming as is to keep this file internally consistent.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Agreed in principle, and it is the same gap as every other server-side message in the platform: the backend has no localization at all (the only MessageSource is for email subjects, and Accept-Language is never read), while parseHttpErrorMessage shows error.message verbatim. Returning the conflicts structurally and rendering them from a locale key is the right fix, but it is a contract change plus frontend work that this PR should not grow into — noting it here so it is not lost, and device-install-conflict-exists is a good starting vocabulary.

ShvaykaD (commented):
## Re-review summary

Verified the 13 findings from the [previous review](https://github.com/thingsboard/thingsboard/pull/16109#pullrequestreview-5132157034) against `6fe4852f9b`.

| Status | Count |
|--------|-------|
| ✅ Resolved | 12 |
| 💬 Acknowledged | 1 |
| ❌ Unresolved | 0 |

All addressed — resolving those threads. The `WILL_INSTALL` guard is exactly right: `installPlan` only ever sets `errorMessage` on the entry it failed on, and that entry keeps `WILL_INSTALL`, so a MISSING entry can no longer hijack the sentence. The rule chain type now comes from the same file as the name, and the trailing-blank-line contract is genuinely gone rather than relocated — the service emits only the markdown list and the dialog joins its own sentences.

I also checked the two deferrals rather than taking them at face value, and both hold. The backend really has no request-scoped localization (the only `MessageSource` consumer is `DefaultMailService`, and nothing reads `Accept-Language`), so the hardcoded conflict copy is a platform-wide gap, not one this PR introduces. And reading the rule chain type with `RuleChainType.valueOf` adds no new failure mode, since `JacksonUtil` doesn't enable `READ_UNKNOWN_ENUM_VALUES_AS_NULL` and `provisionRuleChains` would throw on the same input.

### Non-blocking notes

Nothing below can produce wrong behaviour — no inline comments this round, just things worth knowing.

- **`DefaultSolutionService.java:153`** — `import java.util.Collection;` came in with `firstOrNull` and is orphaned now that the helper is inlined. Nothing else in the file uses it.
- **`DefaultSolutionServiceTest.java`** — the dedupe test covers the `EntityDefinition` overload only. Worth a second look because `distinct()` moved out of `appendConflicts` in this commit, so those `checked` sets are now the only thing keeping a name from appearing twice in the report. The realistic case is device/asset profiles, whose list is assembled from `device_profiles.json` **plus** the `device_profiles` directory and so genuinely can hold the same name twice — that path goes through the other overload, with its own set, untested. `verifyNoInteractions` in `testValidateSolutionWithoutEntityFiles` also still lists only the original five mocks.
- Optional polish, take it or leave it: the rule chain dedupe key concatenates the enum with the name (`type + name`), which is only unambiguous because no `RuleChainType` constant is a prefix of another; the rule chain lookup lambda names its parameter `ignored` and closes over the outer `name` instead; the customer lookup returns `null` both for "does not exist" and for "skip, randomized at install"; and there are now three near-identical `Set<String> checked` loops.

Looks good to merge.

---

ShvaykaD (dismissed):

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -177,6 +178,12 @@
 @Slf4j
 public class DefaultSolutionService implements SolutionService {
 
+    private static final String BLANK_LINE = System.lineSeparator() + System.lineSeparator();
+    private static final String CONFLICTS_INTRO =
+            "Some entities of the solution template already exist. Rename or delete them and install the template again:";
+    private static final String RANDOM_PLACEHOLDER = "$random";
+    private static final int MAX_REPORTED_CONFLICTS = 10;
```
ShvaykaD: Two small things about this constant. Neither changes behaviour.

**The name is misleading.** `MAX_REPORTED_CONFLICTS` sounds like it limits how many conflicts we look for. It doesn't — validation still queries every single name, and the cap only shortens the message that gets printed. Something like `MAX_LISTED_CONFLICT_NAMES_PER_TYPE` would say that, and would also make clear that the limit applies per entity type rather than to the report as a whole.

**The value is hardcoded, unlike every other limit here.** `maxUncompressedArchiveBytes`, `maxUncompressedEntryBytes`, `maxArchiveEntryCount` and `maxInstallTimeoutMs` are all settings, declared with `@Value("${iot-hub....}")` just a few lines below. This one needs a rebuild to change. If you'd ever want the full list for a large template, `@Value("${iot-hub.max-reported-conflicts:10}")` would follow the pattern the class already uses.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,73 +298,145 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+        collectConflicts(conflicts, devices, name -> deviceService.findDeviceByTenantIdAndName(tenantId, name));
+        collectConflicts(conflicts, dashboards, title -> dashboardService.findFirstDashboardInfoByTenantIdAndName(tenantId, title));
+        collectConflicts(conflicts, edges, name -> edgeService.findEdgeByTenantIdAndName(tenantId, name));
+
+        if (conflicts.isEmpty()) {
+            return null;
+        }
+        StringBuilder details = new StringBuilder(CONFLICTS_INTRO).append(BLANK_LINE);
+        conflicts.forEach((entityType, entities) -> appendConflicts(details, entityType,
+                entities.stream().map(entity -> "'" + entity.getName() + "'").toList()));
```
ShvaykaD: A cleanup suggestion, not a problem.

`conflicts` is a `Map<EntityType, List<HasName>>`, so it holds the entire `Device` / `Customer` / `DashboardInfo` / `Edge` object that each lookup returned. The only thing we ever do with those objects is call `getName()` on them, right here on this line.

If `collectConflict` stored `existing.getName()` and the map became `Map<EntityType, List<String>>`, the `HasName` bound could disappear from the four signatures that currently thread it through, and the type itself would make clear that the map exists only to build the message. The lookup functions can keep returning entities — only what we hold on to changes.

Safe to skip if you prefer keeping the entities available for future use.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,73 +298,145 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+     */
+    private static boolean isRandomizedCustomerTitle(String title) {
+        return title.contains(RANDOM_PLACEHOLDER);
+    }
+
+    /**
+     * One markdown list item per entity type, so that the names of the conflicting entities are what the user reads,
+     * instead of the same sentence repeated for every type.
+     */
+    private void appendConflicts(StringBuilder details, EntityType entityType, List<String> conflictDescriptions) {
```
ShvaykaD: **The javadoc is now incomplete, and that's the part I'd actually fix.** It explains why there's one list item per entity type, but says nothing about the cap. Truncating the names is the more surprising thing this method does now, so someone who trusts the comment won't realise names can be dropped. One sentence covers it.

**The rest is a judgement call, so take it or leave it.** The formatting of a single line is split across two places again: the caller adds the quotes, and this method adds the bullet, the separator, the cap and the "and N more" tail. That's the same split the last review asked to consolidate — it came back because the cap needed somewhere to live. Taking `List<HasName>` and quoting in here would put the whole line's presentation in one method, but I understand why it ended up this way.

Minor: this method touches no instance state, so it could be `static`, like `isRandomizedCustomerTitle` just above it.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
```diff
@@ -291,73 +298,145 @@ public void deleteSolution(TenantId tenantId, SolutionTemplateInstalledItemDescr
...
+        details.append("- **").append(entityType.getNormalName()).append("**: ")
+                .append(conflictDescriptions.stream().limit(MAX_REPORTED_CONFLICTS).collect(Collectors.joining(", ")));
+        if (conflictDescriptions.size() > MAX_REPORTED_CONFLICTS) {
+            details.append(" and ").append(conflictDescriptions.size() - MAX_REPORTED_CONFLICTS).append(" more");
```
ShvaykaD: **This is the one I'd like your opinion on rather than a change request** — it's a product decision, not a code problem.

The intro sentence tells the user "Rename or delete them and install the template again". For the names we didn't print, they can't do that. They fix the ten they can see, reinstall, get the next ten, and repeat. With a badly clashing template that's a lot of round trips, and it takes some of the shine off the promise in the PR description that every conflict is reported at once instead of one at a time.

Separately, the cap is per entity type with no overall limit. A template that clashes on customers, both profile types, rule chains, dashboards, assets, devices and edges can still print eighty names. So the message isn't really bounded — it's divided by eight.

**In favour of keeping it as is:** the total count is shown, so the user knows the scale immediately, and nobody with thousands of clashes is going to resolve them ten at a time anyway — they'll clean up in bulk or use a different tenant. So the cap may well be the right trade-off.

If you do keep it, two things would help: `and 4990 more (5000 in total)` reads a little clearer, and a line in the intro saying the remaining ones are listed once these are resolved would stop the instruction from being misleading. A single name budget shared across all types, instead of ten per type, is also worth a thought.

application/src/test/java/org/thingsboard/server/service/solutions/DefaultSolutionServiceTest.java
```diff
@@ -0,0 +1,275 @@
...
+    void testValidateSolutionTreatsAJsonNullFileAsEmpty() throws IOException {
+        writeEntitiesFile("devices.json", "null");
+
+        assertThat(service.validateSolution(tenantId, tempDir)).isNull();
+        verifyNoInteractions(deviceService);
+    }
+
+    @Test
+    void testValidateSolutionCapsTheNamesItLists() throws IOException {
+        String definitions = IntStream.rangeClosed(1, 12)
```
ShvaykaD: **A real coverage gap, though a small one.** This test uses 12 names, so it only exercises the `size > MAX` side of the new `if`. The case that's missing is exactly 10 — that's the one that would catch a `>=` typo printing "and 0 more". 11 is worth having too, since that's where the tail reads "and 1 more". A parameterized test over 10, 11 and 12 would cover the whole branch for roughly what this single case already cost.

**And a maintenance point.** The value 12 is written out three times, and the ten expected names are typed by hand, so the cap's value now lives in two places that have to be kept in sync by eye. If anyone bumps the constant later, this test fails with a long string diff instead of a clear message. Building the expectation from the generated names — join the first ten, then append `" and " + (total - 10) + " more"` — with the total in one local variable would make it obvious what the test is checking.

Unrelated to the cap: the "create a `Device` with this name, then stub `findDeviceByTenantIdAndName`" block now appears in four tests in this class. A small `existingDevice(String name)` helper would take care of that.

application/src/test/java/org/thingsboard/server/service/solutions/DefaultSolutionServiceTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+        assertThat(result).isNotNull();
+        assertThat(result.getDetails().lines().toList()).containsSubsequence(
+                "- **Device profile**: 'thermostat'",
+                "- **Asset profile**: 'building'",
+                "- **Edge**: 'Main edge'");
+    }
+
+    @Test
+    void testValidateSolutionLooksTheSameNameUpOnce() throws IOException {
```
ShvaykaD, line 175: **Still unresolved:** thanks for extending `verifyNoInteractions` to all eight mocks — that was half of the earlier note. The other half is still open, so here it is again with more detail on why it matters.

This is the only test for deduplication, and it uses `devices.json`. That path goes through the `List<? extends EntityDefinition>` overload of `collectConflicts`. But the case where a duplicate name actually arises is the profile lists: `deviceProfiles` is `device_profiles.json` **plus** every file in the `device_profiles` directory, merged into one list (lines 307-312), so the same name genuinely can appear twice there. Those lists go through the *other* overload, which keeps its own separate `Set<String> checked` that no test currently touches.

The gap matters more after this commit than it did before. `distinct()` used to be applied when the message was formatted; it isn't there any more, so those `checked` sets are now the only thing preventing a name from being listed twice in the report.

The worst outcome is a duplicated name in an error message, so this isn't urgent — but it's a short test: put the same profile name in both `device_profiles.json` and the `device_profiles` directory, then assert one lookup and one mention.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
```diff
@@ -331,20 +331,32 @@ export class TbIotHubInstallDialogComponent extends DialogComponent<TbIotHubInst
 
   private handleApiError(err: any): void {
     this.state = 'error';
-    this.errorMessage = err?.error?.message || err?.message ||
-      this.translate.instant('iot-hub.install-error', { name: this.item.name });
+    this.errorMessage = err?.error?.message || err?.message || '';
```
ShvaykaD, line 320: Style only — nothing behaves differently either way.

"There are no details to show" is now expressed four times in three different ways: `result.errorMessage ?? ''` on line 311, `|| ''` here on 334, `serverError ?? ''` on line 343, and then line 357 tests for emptiness a fourth time to decide how to attach the rollback warning.

Picking one idiom, and letting the join deal with the empty case, would state the convention once instead of restating it at each site — something like `[message, warning].filter(Boolean).join('\n\n')` for the warning.

On `??` versus `||`: `errorMessage` is declared as a non-optional `string` on the result type, so `??` reads as though it might be null when the type says it can't be. `||` matches the declaration more honestly.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
```diff
@@ -331,20 +331,32 @@ export class TbIotHubInstallDialogComponent extends DialogComponent<TbIotHubInst
...
   }
 
   private handlePlanResult(result: InstallPlanResult): void {
     if (!result.success) {
       this.state = 'error';
-      let message = result.errorMessage || this.translate.instant('iot-hub.install-error', { name: this.item.name });
+      const serverError = result.errorMessage;
+      // The description above the details already names the item, so nothing is rendered here when the server
```
ShvaykaD: Tiny one. This comment and the one three lines below it both open with "The description above … already names the item", so the same reasoning is given twice in a row — it looks like the new comment was added above the existing one instead of being merged into it.

This first comment also mostly restates the code: `serverError ?? ''` and the new `@if (errorMessage)` in the template already make it clear that nothing renders when the server sent no message. Merging the two would leave just the point that *isn't* obvious from reading the code — why only a failing dependency needs to be named here.

ShvaykaD (commented):
## Re-review summary

Re-reviewed [Report every conflicting entity before installing a solution template](https://github.com/thingsboard/thingsboard/pull/16109) — verified 3 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/16109#pullrequestreview-5133624327).

| Status | Count |
|--------|-------|
| ✅ Resolved | 1 |
| 💬 Acknowledged | 1 |
| ❌ Unresolved | 1 |

Also found 7 new issue(s) in the fix commits, commented inline.

**No bugs this round.** I went through the new commit (`address the PE review notes on the install report`) looking for correctness problems and found none. Specifically, I checked and confirmed: the eight lookups really are exact-match, so the "looked up by exact name" claim in the description holds; `Customer.getName()` returns the title, so the report prints the same string that was matched; an empty rule chain file gives back a `MissingNode` rather than `null`, so the existing `ruleChain == null` guard covers it; and every caller of `handleApiError` copes with an empty `errorMessage`, because Angular always fills in `HttpErrorResponse.message`. Also worth knowing: `iot-hub.install-error` is still used by `iot-hub-add-item-dialog.component.ts:104`, so dropping its two uses here did not leave the key orphaned.

Everything commented inline is quality, test coverage, or user-facing wording. One of them (the 10-name cap) is a product decision I'd like an opinion on rather than a change request; two are test gaps; the rest are small and safe to skip if you disagree. I've said which is which on each comment.

### Finding details

- ✅ **DefaultSolutionService.java:153** — orphaned `import java.util.Collection;` left behind when the `firstOrNull` helper was inlined — *Fixed in code.*
- ❌ **DefaultSolutionServiceTest.java:184** — the dedupe test covers only one of the two `collectConflicts` overloads, and not the one where duplicate names actually occur — *Half addressed: `verifyNoInteractions` now lists all eight mocks, but the profile-overload dedupe is still uncovered. Commented inline.*
- 💬 **DefaultSolutionService.java** — optional polish: the `type + name` dedupe key, the `ignored` lambda parameter, a `null` customer lookup meaning two different things, and three near-identical `Set<String> checked` loops — *All four are still present. They were offered as take-it-or-leave-it and left as is, which is a fair call.*

## Additional findings

- **`ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-update-dialog.component.html:35-41`** and **`iot-hub-update-dialog.component.ts:107,112`** — this is the most useful thing in the review, so it's worth reading even though it's outside the diff. `TbIotHubUpdateDialogComponent` is the update-flow twin of the install dialog, and it still has both of the problems this PR just fixed in the install dialog. Its details block renders unconditionally, so an empty `errorMessage` leaves an empty markdown block sitting under the heading. And its fallback message is `iot-hub.update-error` ("Failed to update {{name}}"), which repeats the `iot-hub.update-error-message` ("Failed to update '{{name}}'.") line shown directly above it — the exact repetition the PR description says it removes. This is not theoretical for solution templates: `DefaultIotHubService.java:435` throws `RuntimeException(response.getDetails())` when a `SOLUTION_TEMPLATE` update fails, and `updateItemVersion` turns that into `UpdateItemVersionResult.error(e.getMessage())`, so the new conflict report reaches this dialog too. Since the report format is what changed in this PR, applying the same two fixes here would keep the update path consistent with the install path.
- **`application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java:301`** — the `//TODO: check other entities.` at the top of `validateSolution` predates this PR, but the PR closes most of it: customers, assets, devices, edges, both profile types, dashboards and rule chains are all covered now. Worth either deleting it or naming what is actually still unchecked, so the next reader doesn't have to re-derive the list.

---

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Renamed to `MAX_LISTED_NAMES_PER_TYPE` — the old name did read as a limit on what we validate, and the new one carries the per-type part too.

Left it a constant rather than a `@Value`. The four settings below it are protections an operator has to be able to tune per deployment: archive size, entry count, install timeout. This one decides how long one sentence of one error message is. Making it configurable means a key in `thingsboard.yml` and a line in the docs for something nobody will ever set, so I'd rather not add the surface. If it ever needs tuning it is a one-line change in the same place.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Done — `conflicts` is a `Map<EntityType, List<String>>` now and `collectConflict` stores `existing.getName()`, so the `HasName` bound is gone from the four signatures and only stays where the lookups genuinely return entities. It also lines this class up with the PE version, where that map already held strings.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Javadoc now says the list is capped and points at the note that tells the user so, and the method is `static`.

On moving the quoting inside: it would cost the shared method. In PE the same `appendConflicts` also prints the entity-group section, and those items are already formatted — `'Sensors' (Type: Device, Owner: Tenant)` — so quoting in here would double-quote them and the two editions would need different methods. Keeping the quoting at the call site is what lets both use one identical method, so I left the split as is.

application/src/main/java/org/thingsboard/server/service/solutions/DefaultSolutionService.java
pinkevmladchy: Kept the cap, with both of the things you suggested:

- the tail now reads `and 4990 more (5000 in total)`;
- when anything was left out, the report ends with `Only the first 10 names of each type are listed. The rest are reported the next time the template is installed.`, so the instruction in the intro is no longer misleading. It is only appended when something was actually truncated, so the usual one- or two-conflict report is unchanged.

I did not go for a single budget shared across types. With ten devices spending it, the one conflicting dashboard would fall out of the report — and that is the name the user most needs to see. Per-type keeps a line for every type that clashes, and with the totals the user can see how big each pile is.

application/src/test/java/org/thingsboard/server/service/solutions/DefaultSolutionServiceTest.java
pinkevmladchy: Done. It is a `@ParameterizedTest` over 10, 11 and 12, the expected line is built from the generated names, and the cap lives in one constant in the test (`LISTED_NAMES_PER_TYPE`) next to the note text it has to match. 10 asserts the note is absent, 11 and 12 assert the tail and the note.

Also added `existingDevice(name)` and used it in the tests that were building the device by hand.

application/src/test/java/org/thingsboard/server/service/solutions/DefaultSolutionServiceTest.java
pinkevmladchy: Added the test: the same profile name in `device_profiles.json` and in the `device_profiles` directory, asserting one lookup and one mention.

One correction on the reasoning, though — there was never a `distinct()` in the formatting, neither before this commit nor after it. `appendConflicts` has always joined the list it was given, and the `checked` sets in the collectors have always been the only dedup. The gap you are pointing at is real, it just is not new in this commit.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
pinkevmladchy: Done — `|| ''` at all three sites, and the warning is attached with `[message, warning].filter(Boolean).join('\n\n')`, so the empty case is handled once by the join instead of being re-tested at each site. Agreed on `??` versus `||` given that the field is a non-optional `string`.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-install-dialog.component.ts
pinkevmladchy: Merged — one comment now, keeping only the part that is not obvious from the code: why just a failing dependency has to be named here.

ShvaykaD (commented):
## Re-review summary

Re-reviewed [Report every conflicting entity before installing a solution template](https://github.com/thingsboard/thingsboard/pull/16109) — verified 8 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/16109#pullrequestreview-5151201455).

| Status | Count |
|--------|-------|
| ✅ Resolved | 8 |

All previous findings have been addressed, and **no new bugs this round** — nothing left to block on. Three of the findings had a sub-point you declined with a reason I agree with: the `@Value` for the cap, the single shared name budget, and moving the quoting inside `appendConflicts`. The last one in particular — I hadn't known PE passes already-formatted entity-group items through the same method, which does rule the refactor out.

### What I checked in the fix commits

I went through both fix commits looking for correctness problems and found none. Specifically confirmed: the truncation note lands after a blank line, so it renders as its own paragraph rather than being absorbed into the markdown list; the note's `size() > MAX` condition and the one inside `appendConflicts` agree, so the note never appears without a truncated list (or the reverse); `LISTED_NAMES_PER_TYPE` is a compile-time constant, so the `@ValueSource` compiles; `{@value #MAX_LISTED_NAMES_PER_TYPE}` and `{@link #TRUNCATION_NOTE}` both resolve on private members of the same class; `|| ''` and `?? ''` are equivalent for a non-optional `string`, and `handleApiError` was already on `||`, so all four sites now read the same; and the new profile test really does exercise the directory source — `loadListOfEntitiesFromDirectory` resolves `tempDir/device_profiles` with no `entities` segment, which is exactly where `writeFile` puts it and what `provisionDeviceProfiles` reads.

One correction on the `distinct()` exchange, since it changes what the `checked` sets are load-bearing for: it *was* there. `appendConflicts` had `conflictDescriptions.stream().distinct()` in 1a15b5c, 56866d3 and 0d275e0, and it went away in 6fe4852. So the collectors' `checked` sets did become the only dedup at that commit, which is what makes the profile-overload test you just added worth having.

### Optional polish — no bugs, take or leave

- **`DefaultSolutionService.java:345`** — the "was anything truncated" decision now lives in two places: here, and inside `appendConflicts` as `conflictDescriptions.size() > MAX_LISTED_NAMES_PER_TYPE`. They agree today, but nothing ties them together, so a change to one comparison would silently produce a note with nothing truncated, or a truncated list with no note. Having `appendConflicts` return whether it truncated and OR-ing that across the `forEach` would make the note fall out of the same decision that did the truncating, and drop the extra pass over `conflicts.values()`.
- **`DefaultSolutionService.java:186`** — wording only. "The rest are reported the next time the template is installed" holds only if the user resolves the ten that were listed; reinstall without changing anything and the same ten come back, which reads as though the note didn't work. `Resolve the listed ones and install again to see the rest.` says the same thing without the promise.
- **`DefaultSolutionServiceTest.java:246`** — this one is on me. Deriving the expectation is what I asked for, and it does fix the sync problem, but it trades it for another: the test now rebuilds the expected line with the same `limit` → quote → `joining(", ")` pipeline `appendConflicts` uses, so a formatting change mirrored into the test would still pass. Keeping one case that asserts the literal line (the `+2` case, exactly the string that was there before) and letting the parameterization cover only the count arithmetic and the note would get both properties.

### Still open from the last review

Neither addressed nor replied to, restating so they aren't lost: `TbIotHubUpdateDialogComponent` still renders its details block unconditionally and still falls back to `iot-hub.update-error`, so the update path keeps both problems this PR fixed on the install path — and `DefaultIotHubService.java:435` routes solution-template failures through it, so it gets this new conflict report too. And the `//TODO: check other entities.` at the top of `validateSolution` is now mostly done by this PR, so it's worth deleting or narrowing to what's actually left.

---

pinkevmladchy:
Thanks for the two out-of-diff notes — I checked both.

**The TODO is in.** `//TODO: check other entities.` now names what is actually left: customer users (unique by email), alarm rules and calculated fields. Everything else it used to stand for — customers, assets, devices, edges, dashboards, rule chains and both profile types — is validated before anything is created.

**The update dialog stays as it is, deliberately.** Your reading of it is right, including that the report reaches it: `updateItemVersion` re-installs the template, so `installSolution` can fail validation and `UpdateItemVersionResult.error(e.getMessage())` carries the same details into `TbIotHubUpdateDialogComponent`, where the block renders unconditionally and the `iot-hub.update-error` fallback repeats the line above it. Same two problems, same fix.

I am still not doing it here. This PR was explicitly narrowed to the solution template install dialog and its error after the first review pass — the earlier version touched the `DataValidationException` texts of every entity type and had to be reverted for exactly that reason. Widening it again to the update flow, even for two lines that I agree with, walks back that decision. I would rather keep this PR to the one window and take the update dialog separately, where it can be reviewed as its own change.

ShvaykaD (dismissed):

ShvaykaD (dismissed):

ShvaykaD (approved):

ShvaykaD:
## TeamCity `test transports` failure — root cause

This is about the TeamCity **test transports (ThingsBoard)** build [#2105 / 130673](https://builds.thingsboard.io/buildConfiguration/ThingsBoard_TestTransports/130673) (agent EU11, 2026-09-11 09:08–09:22, `Tests failed: 165 (165 new), passed: 301, ignored: 1`).

All 165 failures are the same cascade. The real error is one line:

```
Error creating bean with name 'MqttTransportService': Invocation of init method failed
Caused by: java.net.BindException: Address already in use
    at sun.nio.ch.Net.bind0(Native Method)
```

The MQTT transport couldn't bind its port. That kills the Spring context, and once Spring hits `ApplicationContext failure threshold (1) exceeded`, every remaining test class is failed without being retried — which is why 165 tests report the same `IllegalStateException` and not a real assertion.

## Timeline within the run

The whole suite runs in a single surefire JVM (`ForkedBooter`). **301 tests passed first** — so the MQTT port was free at the start and only became unavailable part-way through. The first context load failure is `AttributesPublishTest` at 09:21:22; everything after that, including `MqttServerSideRpcSequenceOnAckIntegrationTest` at 09:21:29, is a downstream victim of the same threshold, not an independent failure.

Several transport test classes carry their own `@TestPropertySource`, so the suite builds more than one Spring context over the run. The failure appears at one of those rebuilds: a new context tries to bind the MQTT port while a previous one has not released it yet. Which context was still holding the port is not visible in the build log, so that last step is inference — what the log proves directly is the bind failure and the cascade.

## On the re-run

I have already triggered a re-run of the transport tests, **but I am not confident it will fix anything.** The port collision is timing- and agent-dependent, not deterministic — the same build may well go green on the next attempt without anything having changed, or fail again the same way.

Two things suggest this is environmental rather than anything in this PR: this branch already passed `test transports` twice (builds 129621 and 129994, `569 passed`), and the same failure shape appeared the same day on unrelated branches — `pull/15941` (`168 failed / 338 passed`) and `release-4.2` (`148 failed / 318 passed`). `release-4.2` itself went from `566 passed` on 09-08 to `148 failed` on 09-11 with no transport test changes in between.

So a green re-run would be a reprieve, not a fix. The underlying flake — a fixed MQTT bind port shared across test contexts — is worth addressing separately.

PR #16114, opened 2026-09-04
Restrict platform mail for tenants on a restricted profile

ViacheslavKlimov:
Restricts what a tenant on a configured "restricted" profile can do with the platform mailer, so
ThingsBoard Cloud's Free tier stops being usable as a disposable spam relay.

## The problem

A Free tenant could send arbitrary subject and body to arbitrary recipients through the system
mailer, under the platform's own domain and DKIM signature — a `to-email` rule node plus a
`send-email` node is all it takes. Recipient reputation damage lands on the platform, not the abuser.

## The mechanism

One predicate, `TbTenantProfileCache.isRestricted(TenantId)`, driven by
`security.restricted_tenant_profiles` (env `SECURITY_RESTRICTED_TENANT_PROFILES`), matched exactly
and case-sensitively against the tenant profile *name*. It gates four enforcement points:

1. **Recipient allow-list** (`RecipientValidator`) — system-relayed mail from a restricted tenant may
   only be addressed to activated users of that tenant, across to/cc/bcc, with `From` forced to the
   system address. A tenant sending through its own SMTP is exempt from the recipient policy (its own
   server, its own reputation) but stays subject to the per-tenant volume limit.
2. **Activation-link concealment** (`UserController`) — a restricted tenant cannot read a user's
   activation link, nor flip `userCredentialsEnabled` on a user who has never activated. Re-enabling
   an already-activated user still works, so lockout recovery is unaffected.
3. **Email-change block** (`UserDataValidator`) — a restricted tenant's admin cannot change a user's
   email address through the validating save path.
4. **Verified self-service email change** (`DefaultEmailChangeService`) — a user changing their own
   address must confirm a one-time code mailed to the *new* address; applying the change invalidates
   the session.

Points 2 and 3 exist to keep point 1's allow-list non-forgeable: the allow-list keys on
`credentialsEnabled`, so any path that lets a tenant mark an arbitrary address "activated" would
whitelist it as a recipient.

## Edge

Free tenants can create edges, and the edge gRPC uplink was an unguarded write path to
`user_credentials`. Two fixes:

- **Uplink** — for a restricted tenant, `enabled`, `activateToken` and `activateTokenExpTime` are
  never taken from `UserCredentialsUpdateMsg`. Where no credentials row exists (the common case for
  edge-created users), a fresh row is built rather than persisting the message's object, so a
  client-chosen `UserCredentialsId` can no longer repoint an existing row through the dao's PK upsert.
- **Downlink** — `constructUserCredentialsUpdatedMsg` previously serialised `UserCredentials`
  verbatim, handing the cloud-generated `activateToken` back to the tenant-controlled edge. It now
  emits an explicit field projection that omits all four one-time token fields. This applies to every
  tenant, not just restricted ones: no edge has a legitimate use for those tokens.

Edge user sync keeps working — the processor's own duplicate-address rename is preserved, while
tenant-driven email changes stay blocked. A user an edge creates on a restricted tenant is now mailed
its activation link directly, so the address owner remains the only party who ever sees the token.

## Rate limiting

Per-tenant mail rate limiting (`mail.per_tenant_rate_limits`) applies to both the system and own-SMTP
paths. `RateLimitExceededException` and `TbRateLimitsException` now map to `429 TOO_MANY_REQUESTS`
instead of a generic 500; on `POST /api/user` the created user is still rolled back, so a retry after
the limit refills starts clean.

## Auditing

A self-service email change now writes an `EMAIL_CHANGED` audit entry recording both the old and new
address, and mails a notice to the previous address. A notice that fails to send never rolls back or
fails the change itself.

## Deployment — required before this has any effect

`security.restricted_tenant_profiles` defaults to **empty**, so this changes nothing for existing
deployments on upgrade. Cloud must set `SECURITY_RESTRICTED_TENANT_PROFILES=Free`, **identically on
every cluster node** — nodes that disagree enforce inconsistently.

Volume enforcement additionally depends on `mail.per_tenant_rate_limits`, which also ships empty.
Until it is set, a restricted tenant's user-invite and password-reset mail is unmetered. Anyone who
sets it should know the unauthenticated `POST /api/noauth/resetPasswordByEmail` shares that
per-tenant bucket and answers `200` regardless of outcome, so a third party who knows one address of
a tenant can drain that tenant's allowance with no visible signal — watch the logs, not the status
codes.

`cache.specs.emailVerificationCodes` TTL must stay above
`security.email_verification.code_lifetime_seconds`, or an expired code surfaces as "no pending
change" instead of "expired".

## Behaviour changes for restricted tenants

- The display-name recipient form (`Ops Team <a@b.c>`) is rejected; use the bare address.
- Notification-centre email to an invited-but-not-yet-activated user fails `PERMISSION_DENIED`.
- Creating a user always sends the activation mail, regardless of the `sendActivationMail` flag.
- A SYSADMIN cannot correct a typo'd user email on a restricted tenant through the validating path
  either; the workaround is to move the tenant off the restricted profile, or delete and recreate.

## Follow-ups deliberately not in this PR

- **`ApiUsageLimitsExceededException` still surfaces as 500.** It shares the
  `AbstractRateLimitException` base with the two types now mapped to 429, but mapping it would move
  `POST /api/alarm` from 500 to 429 for a tenant over its alarm quota. 429 is arguably right there —
  it just needs its own decision rather than a free ride on a security fix.
- **The unrestricted edge path still persists a client-chosen `UserCredentialsId`**, so the PK-upsert
  row-repointing hole remains open for non-restricted tenants. Pre-existing, same defect class as the
  restricted-path fix above.
- Neither `UserServiceImpl.findUserById` nor `JpaUserCredentialsDao.findByUserId` is tenant-scoped,
  so an edge can name a user id from another tenant. Pre-existing; the downlink strip closes the leak
  regardless.
- `construct*UpdatedMsg` helpers other than the user-credentials one were not audited for the same
  secrets-to-edge class of bug.

## Note for the Edge repository

The downlink strip is unconditional, so a cloud-created, not-yet-activated user now mirrors to an
edge with no `activateToken`. If tb-edge has a local activation flow reading that field, it needs a
corresponding change — that repo was not in scope here and could not be checked.


ViacheslavKlimov:
Closing: this work is moving to Professional Edition only.

The enforcement points touch code that has diverged substantially between the editions — the mail
service in particular, where PE resolves per-tenant SMTP configuration and computes its own
system-vs-external sender discriminator. Landing the same feature in both editions would mean
maintaining two materially different implementations of the same security property, which is not a
trade worth making here.

No part of this is abandoned; it ships in PE.

PR #16117, opened 2026-09-07
Report every conflicting entity before installing a solution template

pinkevmladchy:
Same change as #16109, for `lts-4.3` (the branches conflict, so it cannot be merged forward).

Installing a solution template validated only rule chains, profiles and dashboards upfront, so a template with a clashing customer, asset, device or edge died mid install on a DB constraint and had to be rolled back. And the report only said that *some* entities already exist, without naming them.

- customers, assets, devices and edges are validated before anything is created, so every conflict is reported at once instead of failing on the first one
- each conflict is named, and the same sentence is no longer repeated per entity type:

```
Some entities of the solution template already exist. Rename or delete them and install the template again:

- **Asset profile**: 'Silo'
- **Asset**: 'Silo 1'
```

- rule chains are looked up by exact name and dashboards by exact title — the paged text search matches with `contains`, so a different entity could be reported as the conflict
- the install dialog no longer repeats "Failed to install '<item>'" inside the error details, since it is already shown right above them

Scoped to the solution template install only: the existing entity-save error messages are left untouched.


pinkevmladchy:
Sorry for the churn on this branch — I force-pushed it several times while scoping the PR down, which wiped the SHAs your comments referenced and made "changes since your review" span the whole PR instead of just the fixes.

The branch now carries the same tree in three commits, one per review step, so incremental review works again:

- `report every conflicting entity before installing a solution template` — the scoped change
- `compose the failed dependency sentence in the dialog, not in the service` — the untranslatable sentence moved to the UI
- `address review comments on the install validation` — the four findings from the last pass

The earlier SHAs (`2d0c8e4784` and the ones before it) are still reachable from the force-push events in the timeline if you need the old diff. No more rewrites from here — follow-ups will be plain commits on top.

PR #16121, opened 2026-09-08
Fixed forced SSL on the JS executor Kafka connection when SASL is enabled

zzzeebra:
tb-js-executor forced TLS on the Kafka connection whenever SASL was enabled, so brokers on non-SSL channels (`SASL_PLAINTEXT`, `PLAINTEXT`) were unreachable:

```
Connection error: Client network socket disconnected before secure TLS connection
was established
    at TLSSocket.onConnectEnd (node:_tls_wrap:1732:19)
```

Two causes:

**1. SSL was derived from nothing.** Enabling SASL set `ssl = true` unconditionally, without looking at the security protocol. While tb-node evaluates `security.protocol` (`PLAINTEXT`, `SSL`, `SASL_PLAINTEXT`, `SASL_SSL`), js-executor had no such option at all.

**2. Boolean flags were parsed by truthiness.** `node-config` returns environment overrides as strings, and `Boolean("false")` is `true`. Setting `TB_KAFKA_SSL_ENABLED=false` therefore *enabled* SSL. `kafka.use_confluent_cloud` had the same defect.

## Changes

- Added `kafka.confluent.security.protocol`, using the same `TB_QUEUE_KAFKA_CONFLUENT_SECURITY_PROTOCOL` variable and `SASL_SSL` default that tb-node already uses, so the two services are configured alike. SSL is now derived from it instead of being forced.
- Moved the decision into `resolveSslEnabled(configuredSsl, useSasl, securityProtocol)`, in the style of the existing `resolveCompressionType`. It matches the four supported protocols; anything else keeps SSL enabled and logs a warning naming the valid values, so a mistyped or blank protocol cannot put SASL credentials on an unencrypted connection. An explicit `kafka.ssl.enabled=true` still wins over the protocol, matching tb-node, where `TbKafkaSettings.configureSSL()` overwrites `security.protocol` with `"SSL"`.
- Read the protocol through a `config.has` guard. `default.yml` ships as a dpkg/rpm conffile (`packaging/js/build.gradle`), so a preserved copy from an older release has no such key and an unguarded `config.get` throws on startup. The `SASL_SSL` fallback keeps the previous behaviour on those installs.
- Parse `kafka.ssl.enabled` and `kafka.use_confluent_cloud` through a single `isTrue` helper, and quote both defaults to match the rest of `default.yml`. The helper accepts the values tb-node accepts: both variables are shared with it, and Spring binds them through `StringToBooleanConverter`, which reads `true`, `on`, `yes` and `1` as true.
- Log the resulting SSL and SASL decision, which until now was only observable as a TLS handshake failure.

Confluent Cloud is unaffected: `docker/queue-confluent.env` already sets `SASL_SSL`, which still resolves to SSL enabled.

## Testing

`test/kafkaTemplate.test.ts` covers the derivation without a broker — string parsing (`"false"` is not truthy, `TRUE` and stray whitespace are accepted, and the tb-node value set is honoured), the SSL protocols enabling TLS, the plaintext protocols leaving it off, `ssl.enabled` overriding the protocol, the protocol being ignored when SASL is off, and unusable values keeping TLS on. `npm test` now runs `test/*.test.ts`; 11 tests pass.

End to end, `KafkaTemplate.init()` against a local KRaft broker on `PLAINTEXT://localhost:9092` (`bitnamilegacy/kafka:4.0`, listeners matching `docker/kafka.env`):

| Case | Before | After |
|---|---|---|
| nothing set | connects | connects |
| `TB_KAFKA_SSL_ENABLED=false` | **TLS handshake error** | connects |
| `TB_KAFKA_SSL_ENABLED=true` | SSL on | SSL on |
| `USE_CONFLUENT_CLOUD=true`, protocol `SASL_SSL` | SSL on | SSL on |
| `USE_CONFLUENT_CLOUD=true`, protocol `SASL_PLAINTEXT` | **TLS handshake error** | plaintext connection, SASL handshake attempted |
| `USE_CONFLUENT_CLOUD=true`, protocol `SSL_SASL` (typo) | SSL on | warning logged, SSL kept on |
| `default.yml` without the new key, `USE_CONFLUENT_CLOUD=true` | n/a | starts, falls back to `SASL_SSL` |

On the last row, the same run with an unguarded `config.get` fails with `Configuration property "kafka.confluent.security.protocol" is not defined`.

### Live SASL brokers

`KafkaTemplate.init()` against a KRaft broker with a `SASL_PLAINTEXT` listener (PLAIN, `bitnamilegacy/kafka:4.0`):

| Case | Result |
|---|---|
| protocol `SASL_PLAINTEXT`, correct credentials | `SSL enabled: false, SASL enabled: true`; connects, `js_eval.requests` created |
| same, wrong password | `SASL PLAIN authentication failed: Invalid username or password` |
| same, `TB_KAFKA_SSL_ENABLED=true` | `Client network socket disconnected before secure TLS connection was established` |

The wrong-password run is what shows the SASL handshake is reached rather than skipped. The third row reproduces the reported error on that same broker, now only when SSL is asked for explicitly.

Against a second broker with a `SASL_SSL` listener (self-signed CA, SAN `localhost`):

| Case | Result |
|---|---|
| protocol `SASL_SSL` | `SSL enabled: true, SASL enabled: true`; connects |
| protocol not set at all | `SSL enabled: true, SASL enabled: true`; connects |
| protocol `SASL_PLAINTEXT` | SSL off, connection closed by the broker |
| CA file not supplied | `unable to verify the first certificate` |

The second row is the one that matters for existing Confluent Cloud installs: they never set a protocol, because the option did not exist, and the `SASL_SSL` default keeps TLS on for them. The last row only reflects the self-signed CA used here — Confluent's certificate chains to a public CA in Node's default store, so no CA file is needed in production.

### Boolean values

`TB_KAFKA_SSL_ENABLED` and `TB_QUEUE_KAFKA_USE_CONFLUENT_CLOUD` resolved on this branch and on its base:

| Value | tb-node | Before | After |
|---|---|---|---|
| `true`, `TRUE` | true | on | on |
| `on`, `yes`, `1` | true | on | on |
| `false`, `0`, `no`, `off` | false | **on** | off |
| unrecognized | startup failure | on | off |

Only the third row changes for a deployment that works today, and it changes towards tb-node. The values in rows one and two keep their meaning, so no working install loses SSL or SASL on upgrade.

### Preserved conffiles

`default.yml` and `custom-environment-variables.yml` are both `configurationFile()` entries in `packaging/js/build.gradle`, so an upgrade can preserve either one. New code, `USE_CONFLUENT_CLOUD=true`, protocol `SASL_PLAINTEXT`:

| Config files after the upgrade | Result |
|---|---|
| both new | SSL off, SASL on — the fix applies |
| both preserved | SSL on — previous behaviour, no startup failure |
| new `default.yml`, preserved `custom-environment-variables.yml` | SSL on — the fix silently does not apply |
| preserved `default.yml`, new `custom-environment-variables.yml` | SSL off, SASL on — the fix applies |

`custom-environment-variables.yml` is the file that decides: without its new line the variable maps nowhere. Repeating the second row with the `config.has` guard removed fails with `Configuration property "kafka.confluent.security.protocol" is not defined`.

The same two upgrades as a real package install: `tb-js-executor.deb` built from this branch and from its base, installed in turn in a Debian 12 container. `DEBIAN/conffiles` in the built package lists both YAML files, and dpkg acts on that:

| Upgrade | dpkg |
|---|---|
| configs untouched | `Installing new version of config file` for both — the new keys land and the fix applies |
| configs edited locally | `The default action is to keep your current version` — the edits survive, the new keys are absent, and `default.yml.dpkg-new` / `custom-environment-variables.yml.dpkg-new` are written alongside |

The second row is the "both preserved" row above, reached through dpkg rather than through `NODE_CONFIG_DIR`.

### Rule chain execution

A monolith against the `SASL_PLAINTEXT` broker above with `JS_EVALUATOR=remote`, and a `script` node — JavaScript, not the default TBEL — branched off `Message Type Switch` on `Post telemetry`:

| Step | Result |
|---|---|
| saving the rule chain | `Processing compile request` → `Sending success compile response` |
| posting telemetry | `Processing invoke request` → `Sending success invoke response`; the node's `OUT` event carries the field the script adds, and the reading is stored |
| js-executor stopped, telemetry posted again | the node fails with `java.util.concurrent.TimeoutException: Script timeout!` |
| js-executor restarted, telemetry posted again | `NOT_FOUND` on the first invoke — the script cache is empty after a restart — then success once tb-node resends the body |

The third row is what rules out the script having been evaluated inside tb-node: without a reachable js-executor there is no execution at all.

Not covered: starting the service from the installed package. The packaged binary targets `amd64` and exits with Node's internal evaluation failure under x86_64 emulation on an arm64 host — so does the binary built from the base commit, so it is the emulation rather than this change — and confirming it needs a real amd64 machine. The startup path itself is covered above through `NODE_CONFIG_DIR`. Also not covered: Confluent Cloud itself, which the `SASL_SSL` broker above stands in for.

### DIFF
--- a/msa/js-executor/config/default.yml
+++ b/msa/js-executor/config/default.yml
@@ -42,13 +42,15 @@ kafka:
   connectionTimeout: "1000" # Time in milliseconds to wait when establishing a connection to a Kafka broker
   compression: "none" # Message compression codec for the producer: gzip, lz4 or none
   topic_properties: "retention.ms:604800000;segment.bytes:52428800;retention.bytes:104857600;partitions:100;min.insync.replicas:1" # Semicolon-separated Kafka topic configuration properties applied on topic creation
-  use_confluent_cloud: false # Set to true to enable Confluent Cloud-specific configuration (SSL + SASL)
+  use_confluent_cloud: "false" # Set to true to enable SASL authentication (Confluent Cloud or any SASL-enabled broker); SSL is derived from confluent.security.protocol
   client_id: "kafkajs" # Kafka client identifier; inject the pod name to easily identify the client via kafka-consumer-groups.sh
   ssl:
-    enabled: false # Enable SSL/TLS for Kafka broker connections
+    enabled: "false" # Enable SSL/TLS for Kafka broker connections
   confluent:
     sasl:
       mechanism: "PLAIN" # SASL mechanism for Confluent Cloud authentication: PLAIN, SCRAM-SHA-256, or SCRAM-SHA-512
+    security:
+      protocol: "SASL_SSL" # Protocol used to communicate with brokers, applied only when use_confluent_cloud is true. Valid values are: PLAINTEXT, SSL, SASL_PLAINTEXT, SASL_SSL
 
 # Logging configuration
 # Controls log verbosity, output directory, and log file naming pattern for the JS executor service.
--- a/msa/js-executor/queue/kafkaTemplate.ts
+++ b/msa/js-executor/queue/kafkaTemplate.ts
@@ -72,6 +72,34 @@ export class KafkaTemplate implements IQueue {
                 return CompressionTypes.None;
         }
     }
+    // node-config returns environment overrides as raw strings, so a plain Boolean()
+    // would treat "false" as true.
+    private isTrue(value: any): boolean {
+        return String(value).trim().toLowerCase() === 'true';
+    }
+
+    private resolveSslEnabled(configuredSsl: boolean, useSasl: boolean, securityProtocol: any): boolean {
+        if (configuredSsl) {
+            return true;
+        }
+        if (!useSasl) {
+            return false;
+        }
+        switch (String(securityProtocol).trim().toUpperCase()) {
+            case 'SSL':
+            case 'SASL_SSL':
+                return true;
+            case 'PLAINTEXT':
+            case 'SASL_PLAINTEXT':
+                return false;
+            default:
+                // Keep TLS on: an unrecognized value must not silently send SASL
+                // credentials over an unencrypted connection.
+                this.logger.warn('Unknown kafka.confluent.security.protocol value "%s"; keeping SSL enabled. Supported values: PLAINTEXT, SSL, SASL_PLAINTEXT, SASL_SSL.', securityProtocol);
+                return true;
+        }
+    }
+
     private partitionsConsumedConcurrently = Number(config.get('kafka.partitions_consumed_concurrently'));
 
     private kafkaClient: Kafka;
@@ -91,8 +119,11 @@ export class KafkaTemplate implements IQueue {
         const kafkaBootstrapServers: string = config.get('kafka.bootstrap.servers');
         const queuePrefix: string = config.get('queue_prefix');
         const requestTopic: string = queuePrefix ? queuePrefix + "." + config.get('request_topic') : config.get('request_topic');
-        const useConfluent = config.get('kafka.use_confluent_cloud');
-        const enabledSsl = Boolean(config.get('kafka.ssl.enabled'));
+        const useConfluent = this.isTrue(config.get('kafka.use_confluent_cloud'));
+        const configuredSsl = this.isTrue(config.get('kafka.ssl.enabled'));
+        const securityProtocol = config.has('kafka.confluent.security.protocol')
+            ? config.get('kafka.confluent.security.protocol') : 'SASL_SSL';
+        const enabledSsl = this.resolveSslEnabled(configuredSsl, useConfluent, securityProtocol);
         const groupId:string =  queuePrefix ? queuePrefix + ".js-executor-group" : "js-executor-group";
         this.logger.info('Kafka Bootstrap Servers: %s', kafkaBootstrapServers);
         this.logger.info('Kafka Requests Topic: %s', requestTopic);
@@ -119,8 +150,10 @@ export class KafkaTemplate implements IQueue {
                 username: config.get('kafka.confluent.username'),
                 password: config.get('kafka.confluent.password')
             };
-            kafkaConfig['ssl'] = true;
+            this.logger.info('Kafka SASL security protocol: %s', securityProtocol);
         }
+
+        this.logger.info('Kafka SSL enabled: %s, SASL enabled: %s', enabledSsl, useConfluent);
 
         if (enabledSsl) {
             const certFilePath: string = config.has('kafka.ssl.cert_file') ? config.get('kafka.ssl.cert_file') : '';
--- a/msa/js-executor/config/custom-environment-variables.yml
+++ b/msa/js-executor/config/custom-environment-variables.yml
@@ -52,6 +52,8 @@ kafka:
   confluent:
     sasl:
       mechanism: "TB_QUEUE_KAFKA_CONFLUENT_SASL_MECHANISM"
+    security:
+      protocol: "TB_QUEUE_KAFKA_CONFLUENT_SECURITY_PROTOCOL"
     username: "TB_QUEUE_KAFKA_CONFLUENT_USERNAME"
     password: "TB_QUEUE_KAFKA_CONFLUENT_PASSWORD"
 
--- a/msa/js-executor/package.json
+++ b/msa/js-executor/package.json
@@ -7,7 +7,7 @@
   "bin": "server.js",
   "scripts": {
     "pkg": "tsc && pkg -t node22-linux-x64 --output ./target/thingsboard-js-executor-linux ./target/src && pkg -t node22-win-x64 --no-bytecode --public-packages \"*\" --public --output ./target/thingsboard-js-executor-win.exe ./target/src && node install.js",
-    "test": "mkdir -p target/surefire-reports && node --require ts-node/register --test --test-reporter=spec --test-reporter-destination=stdout --test-reporter=junit --test-reporter-destination=target/surefire-reports/TEST-js-executor.xml test/jsExecutor.test.ts",
+    "test": "mkdir -p target/surefire-reports && node --require ts-node/register --test --test-reporter=spec --test-reporter-destination=stdout --test-reporter=junit --test-reporter-destination=target/surefire-reports/TEST-js-executor.xml test/*.test.ts",
     "start": "nodemon --watch '.' --ext 'ts' --exec 'ts-node server.ts'",
     "start-prod": "nodemon --watch '.' --ext 'ts' --exec 'NODE_ENV=production ts-node server.ts'",
     "build": "tsc"
--- a/msa/js-executor/test/kafkaTemplate.test.ts
+++ b/msa/js-executor/test/kafkaTemplate.test.ts
@@ -1,0 +1,74 @@
+///
+/// Copyright © 2016-2026 The Thingsboard Authors
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+
+import { describe, test } from 'node:test';
+import assert from 'node:assert/strict';
+import { KafkaTemplate } from '../queue/kafkaTemplate';
+
+// Same suite name as jsExecutor.test.ts so every case lands under
+// <testsuite name="js-executor"> in the JUnit XML.
+describe('js-executor', () => {
+
+const template = new KafkaTemplate();
+const resolveSsl = (configuredSsl: boolean, useSasl: boolean, protocol?: any): boolean =>
+    (template as any).resolveSslEnabled(configuredSsl, useSasl, protocol);
+const isTrue = (value: any): boolean => (template as any).isTrue(value);
+
+test('kafka ssl.enabled is parsed as a string, not by truthiness', () => {
+    // node-config hands back environment overrides as strings; Boolean("false") is true
+    assert.equal(isTrue('false'), false, '"false" must not enable SSL');
+    assert.equal(isTrue(false), false);
+    assert.equal(isTrue('true'), true);
+    assert.equal(isTrue(true), true);
+    assert.equal(isTrue('TRUE'), true, 'value must be case-insensitive, like the tb-node flags');
+    assert.equal(isTrue(' true '), true);
+    assert.equal(isTrue('yes'), false);
+    assert.equal(isTrue(undefined), false);
+});
+
+test('kafka ssl is enabled for the SSL security protocols', () => {
+    assert.equal(resolveSsl(false, true, 'SASL_SSL'), true);
+    assert.equal(resolveSsl(false, true, 'SSL'), true);
+    assert.equal(resolveSsl(false, true, 'sasl_ssl'), true, 'protocol must be case-insensitive');
+});
+
+test('kafka ssl stays off for the plaintext security protocols', () => {
+    assert.equal(resolveSsl(false, true, 'SASL_PLAINTEXT'), false);
+    assert.equal(resolveSsl(false, true, 'PLAINTEXT'), false);
+    assert.equal(resolveSsl(false, true, ' sasl_plaintext '), false);
+});
+
+test('kafka ssl.enabled overrides the security protocol', () => {
+    assert.equal(resolveSsl(true, true, 'SASL_PLAINTEXT'), true,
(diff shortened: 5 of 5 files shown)


zzzeebra (commented):
## Review summary

Reviewed 3 changed files in [Fixed forced SSL on the JS executor Kafka connection when SASL is enabled](https://github.com/thingsboard/thingsboard/pull/16121). Left 5 comment(s) inline.

The core fix is right and the env var name / `SASL_SSL` default match `thingsboard.yml` exactly, so the two services stay configured alike. The main things worth a second look are the resolution semantics — `enabledSsl` is OR-ed rather than assigned, so `TB_KAFKA_SSL_ENABLED=true` still forces TLS on top of an explicit `SASL_PLAINTEXT` — and the fact that an unrecognized protocol value now silently resolves to plaintext, which is the direction that leaks SASL credentials rather than the direction that fails loudly. The new key is also read with `.trim()` directly on `config.get(...)`, unlike the two boolean reads above it that go through `String(...)`.

## Additional findings

- **msa/js-executor/api/jsInvokeMessageProcessor.ts:44** — `const useSandbox = config.get('script.use_sandbox') === 'true';` has the same class of defect this PR is fixing, and it fails in the dangerous direction: the comparison is exact, so `SCRIPT_USE_SANDBOX=TRUE` (or `True`, or a value with stray whitespace) evaluates to `false` and silently selects the non-sandbox path that `default.yml` documents as "DANGEROUS BY DESIGN … provides no isolation and exposes the host process". An operator trying to *enable* the sandbox can end up disabling it. `slowQueryLogBody` on line 47 has the same pattern, though the consequence there is harmless. If a shared boolean parser comes out of this PR, these are the call sites that most need it.

---

msa/js-executor/queue/kafkaTemplate.ts
```diff
@@ -119,9 +119,14 @@ export class KafkaTemplate implements IQueue {
                 username: config.get('kafka.confluent.username'),
                 password: config.get('kafka.confluent.password')
             };
-            kafkaConfig['ssl'] = true;
+            const securityProtocol: string = config.has('kafka.confluent.security.protocol')
+                ? config.get('kafka.confluent.security.protocol') : 'SASL_SSL';
+            this.logger.info('Kafka SASL security protocol: %s', securityProtocol);
+            enabledSsl = enabledSsl || securityProtocol.trim().toUpperCase().endsWith('SSL');
```
zzzeebra: Two things about this line.

OR-ing means `kafka.ssl.enabled` can still override an explicit protocol: with `use_confluent_cloud=true`, `TB_KAFKA_SSL_ENABLED=true` and `TB_QUEUE_KAFKA_CONFLUENT_SECURITY_PROTOCOL=SASL_PLAINTEXT`, TLS is forced on even though the protocol says plaintext — i.e. the exact failure this PR is fixing is still reachable, just via a different combination. Would a plain assignment inside this branch be better, letting the protocol be authoritative whenever SASL is on? It looks backward compatible: the default is `SASL_SSL`, so anyone who doesn't touch the protocol keeps SSL enabled exactly as before, and the only behavior that changes is the combination that is currently self-contradictory.

Second, `endsWith('SSL')` fails open. An unsupported or mistyped value (`SSL_SASL`, `SASL_SS`, `sasl+ssl`) doesn't end in `SSL`, so it resolves to "no TLS" and SASL credentials go out over an unencrypted connection with nothing in the log to say why. Before this change a typo failed safe, because SSL was unconditional. `resolveCompressionType` a few lines up sets the opposite precedent — it warns on an unknown value and names the supported ones — so matching against the four documented protocols with a `logger.warn` on anything else would be both safer and more consistent with the file.

Both of these, plus the fact that `enabledSsl` had to become a mutable `let` reassigned ~30 lines from its declaration, point the same way: a small `private resolveSslEnabled(sslEnabled: boolean, useSasl: boolean, securityProtocol: string): boolean` in the style of `resolveCompressionType` would keep `enabledSsl` a `const`, put the whole rule in one readable place, and make it testable without a broker.

msa/js-executor/queue/kafkaTemplate.ts
```diff
@@ -119,9 +119,14 @@ export class KafkaTemplate implements IQueue {
                 username: config.get('kafka.confluent.username'),
                 password: config.get('kafka.confluent.password')
             };
-            kafkaConfig['ssl'] = true;
+            const securityProtocol: string = config.has('kafka.confluent.security.protocol')
```
zzzeebra: The two reads above go through `String(...)` before comparing, but this one calls `.trim()` straight on the `config.get` result. `config.has` only checks for `undefined`, so a key that is present but blank in a preserved `default.yml` (`protocol:` with no value) passes the guard, `get` returns `null`, and `null.trim()` throws at startup — the same failure mode the guard was added to avoid. A non-string value (`protocol: 1`) would throw too. Wrapping it in `String(...)` like the other two reads would close that off and keep the three parses consistent.

Separately: `msa/js-executor` does have working test infrastructure — `test/jsExecutor.test.ts` on `node:test`, wired into `npm test` with a JUnit reporter — but none of the new behavior is covered. `SASL_SSL`/`SSL` enabling TLS, `SASL_PLAINTEXT`/`PLAINTEXT` leaving it off, `"false"` no longer being truthy, and this fallback when the conffile lacks the key are four cheap assertions once the derivation is a pure function. Given the bug being fixed here is precisely `Boolean("false")`, a couple of tests would keep it from coming back.

msa/js-executor/queue/kafkaTemplate.ts
```diff
@@ -91,8 +91,8 @@ export class KafkaTemplate implements IQueue {
         const kafkaBootstrapServers: string = config.get('kafka.bootstrap.servers');
         const queuePrefix: string = config.get('queue_prefix');
         const requestTopic: string = queuePrefix ? queuePrefix + "." + config.get('request_topic') : config.get('request_topic');
-        const useConfluent = config.get('kafka.use_confluent_cloud');
-        const enabledSsl = Boolean(config.get('kafka.ssl.enabled'));
+        const useConfluent = String(config.get('kafka.use_confluent_cloud')).trim().toLowerCase() === 'true';
```
zzzeebra: This string-to-boolean coercion is written out twice here and is a third distinct style within the same service — `api/jsInvokeMessageProcessor.ts` uses `config.get('script.use_sandbox') === 'true'` for `use_sandbox` and `slow_query_log_body`. Since node-config hands back env overrides as raw strings everywhere, would a single `parseBoolean(value: any): boolean` next to `isNotEmptyStr`/`isString` in `api/utils.ts` be worth it, with all four call sites routed through it? One convention for the service beats leaving two half-robust variants for the next person to choose between.

msa/js-executor/config/default.yml
```diff
@@ -42,13 +42,15 @@ kafka:
...
+  use_confluent_cloud: "false" # Set to true to enable SASL authentication (Confluent Cloud or any SASL-enabled broker); SSL is derived from confluent.security.protocol
   client_id: "kafkajs" # Kafka client identifier; inject the pod name to easily identify the client via kafka-consumer-groups.sh
   ssl:
-    enabled: false # Enable SSL/TLS for Kafka broker connections
+    enabled: "false" # Enable SSL/TLS for Kafka broker connections
   confluent:
     sasl:
       mechanism: "PLAIN" # SASL mechanism for Confluent Cloud authentication: PLAIN, SCRAM-SHA-256, or SCRAM-SHA-512
+    security:
+      protocol: "SASL_SSL" # Protocol used to communicate with brokers. Valid values are: PLAINTEXT, SSL, SASL_PLAINTEXT, SASL_SSL
```
zzzeebra: The comment matches `application/src/main/resources/thingsboard.yml` verbatim, which is good for consistency, but the wording implies the protocol is authoritative and here it isn't quite: the key is only read inside the `use_confluent_cloud` branch, so setting `security.protocol: SSL` on its own does nothing at all. Worth spelling out that it only applies when `use_confluent_cloud` is `true`? That's the part an operator is most likely to get wrong, since the surrounding keys read like standalone switches.

msa/js-executor/config/default.yml
```diff
@@ -42,13 +42,15 @@ kafka:
   connectionTimeout: "1000" # Time in milliseconds to wait when establishing a connection to a Kafka broker
   compression: "none" # Message compression codec for the producer: gzip, lz4 or none
   topic_properties: "retention.ms:604800000;segment.bytes:52428800;retention.bytes:104857600;partitions:100;min.insync.replicas:1" # Semicolon-separated Kafka topic configuration properties applied on topic creation
-  use_confluent_cloud: false # Set to true to enable Confluent Cloud-specific configuration (SSL + SASL)
+  use_confluent_cloud: "false" # Set to true to enable SASL authentication (Confluent Cloud or any SASL-enabled broker); SSL is derived from confluent.security.protocol
```
zzzeebra, line 45: Broadening this flag's documented meaning from "Confluent Cloud" to "any SASL-enabled broker" is the right call and the comment reads well. It does leave a plain-Kafka SASL user configuring keys named `kafka.confluent.*` for a broker that has nothing to do with Confluent — reusing the tb-node env var names is clearly deliberate and worth keeping, so the naming isn't the problem. But `thingsboard.yml` still describes the same flag as "Enable/Disable using of Confluent Cloud"; aligning that comment too would stop operators concluding the two services behave differently.

msa/js-executor/queue/kafkaTemplate.ts
zzzeebra: Fixed both. Verified the blank-value path: `config.has` returns true for a null value, `get` returns null, and `null.trim()` throws at startup. Worth noting `String(...)` alone is not enough — `String(null).trim()` is `"null"`, which does not end in `SSL`, so it turns the crash into a silent plaintext connection. The derivation now matches the four supported values instead, which covers null, blank, non-string and mistyped values in one place.

Tests added in `test/kafkaTemplate.test.ts`, with `npm test` widened to `test/*.test.ts`. Six cases: string parsing (`"false"` not truthy, `TRUE` and whitespace accepted), SSL protocols enabling TLS, plaintext protocols leaving it off, `ssl.enabled` overriding the protocol, the protocol being ignored without SASL, and unusable values keeping TLS on.

msa/js-executor/config/default.yml
zzzeebra: Fixed — the comment now says the protocol is applied only when `use_confluent_cloud` is true.

msa/js-executor/queue/kafkaTemplate.ts
zzzeebra: Fail-open is fixed, and the extraction is done — `resolveSslEnabled(configuredSsl, useSasl, securityProtocol)` in the style of `resolveCompressionType`, so `enabledSsl` is a `const` again. Unrecognized values now keep TLS on and log a warning naming the supported protocols. The divergence was worse than a style issue: `ProducerConfig` defines `security.protocol` with `ConfigDef.CaseInsensitiveValidString.in(...)`, so a typo makes tb-node fail to start with a `ConfigException`, while js-executor was silently downgrading to plaintext.

Keeping the OR, though. `kafka.ssl.enabled` is an explicit opt-in that also carries the cert/key/ca files, so it should win over the protocol, and that is exactly what tb-node does: `TbKafkaSettings.configureSSL()` runs after the Confluent block and overwrites `security.protocol` with `"SSL"` whenever `ssl.enabled` is true. A plain assignment would make the two services disagree, which is the opposite of what this PR is for. It is also not the same bug resurfacing: the bug was TLS being forced on operators who never asked for it, whereas here they asked for it explicitly. There is a test pinning this precedence.

msa/js-executor/queue/kafkaTemplate.ts
zzzeebra: The duplication is gone within this file — both reads go through one `isTrue(value)` helper, which the new tests cover directly for `"false"`, `TRUE` and stray whitespace.

Not moving it to `api/utils.ts` and routing the other two call sites through it here. `use_sandbox` controls script isolation, so changing how its value is parsed changes security behaviour and belongs in its own change with its own testing. Two notes on the finding for that PR: the exact-comparison bug there is real, but it is not silent — `api/jsExecutor.ts:30` logs a warning naming the risk on every startup with the sandbox off. And the fix cannot be a pure refactor, because `SCRIPT_USE_SANDBOX=TRUE` currently disables the sandbox and would start enabling it, which needs a release note.

msa/js-executor/config/default.yml
zzzeebra: Not changing `thingsboard.yml`. The concern it addresses does not hold: after this PR the two services behave identically through that flag — same env var, same `SASL_SSL` default, same "SASL on, protocol decides TLS" rule — so the differing comment wording is not a behavioural difference and cannot produce a wrong configuration.

The wording there is loose in the same way ours was, but tb-node has an escape hatch js-executor lacks: `other-inline` / `TB_QUEUE_KAFKA_OTHER_PROPERTIES` (`thingsboard.yml:1826`) passes arbitrary Kafka properties, including `security.protocol` and `sasl.*`, without the Confluent flag at all. Broadening the comment mattered here because there was no other route; there, there is one. That leaves a cosmetic edit to another service config on an LTS branch, so it stays out.

PR #16134, opened 2026-09-10
Fixed unauthorized access to attributes and time series of other tenants

zzzeebra:
`AccessValidator.validate` never calls `onFailure` for a denial — it reports the outcome by passing a non-OK `ValidationResult` to `onSuccess`. Four places did not inspect that result, so a denial was indistinguishable from a grant.

### Cross-tenant reads

- **`DefaultWebSocketService`** — all four `getAttributesFetchCallback` overloads fetched attributes in `onSuccess` without looking at the result, so `attrSubCmds` returned the attributes of any entity to any authenticated user who knew its id. They now delegate to the existing `on(success, failure)` helper, which is what the `tsSubCmds` and `historyCmds` paths already use.
- **`DefaultTbTelemetryService.getTimeseries`** — the same defect on `GET /api/plugins/telemetry/{entityType}/{entityId}/values/timeseries`, which returned another tenant's time series with HTTP 200. A non-OK result now completes the future exceptionally, so the existing `AccessValidator.handleError` mapping produces the correct 403/404. This was the last hand-rolled `FutureCallback<ValidationResult>` in the codebase; everything else goes through `ValidationCallback`, `HttpValidationCallback` or `on(...)`.

### Denied writes that still executed

- **`AccessValidator`** — `validateDeviceProfile`, `validateAssetProfile`, `validateApiUsageState` (two branches) and `validateOtaPackage` called `callback.onSuccess(accessDenied(...))` and then fell through to `callback.onSuccess(ok(entity))`. Over REST the second result is masked (`DeferredResult.setResult` ignores the second call, so the client still gets 403) but the action itself ran — a cross-tenant attribute write on those entity types succeeded behind the 403. Five `return` statements added.

### Alarm status subscriptions

- **`DefaultWebSocketService.handleWsAlarmsStatusCmd`** — `alarmStatusCmds` accepted a client-supplied `originatorId` with no permission check at all. It now validates `Operation.READ` on the originator, matching what `AlarmController` does for the REST alarm endpoints.
- **`JpaAlarmDao.findActiveOriginatorAlarms`** — received a `tenantId` and never passed it into the query, which selected purely by `originator_id`. The query now filters by `a.tenant_id`.

Because the permission check makes this handler asynchronous, the deferred work also got the error contract of `handleCommands` (including its `TbRateLimitsException` branch), a session-liveness guard so a subscription is not created for a session that closed while validation was in flight, and refusals are sent as `AlarmStatusUpdate` rather than a v1 `TelemetrySubscriptionUpdate` so the UI routes them to the alarm status subscriber instead of dropping them.

### Not affected

`entityDataCmds`, `alarmDataCmds` and `entityCountCmds` enforce tenant and customer isolation in SQL (`DefaultEntityQueryRepository.buildPermissionQuery`, `DefaultAlarmQueryRepository`), for every filter type including `SINGLE_ENTITY`. No changes were needed there.

### Tests

Four cross-tenant tests, one per defect: `WebsocketApiTest.testAttributesSubscription_differentTenant`, `WebsocketApiTest.testAlarmStatusSubscription_differentTenant`, `TelemetryControllerTest.testSaveAttributesForEntityOfDifferentTenant`, `TelemetryControllerTest.testGetTimeseriesForEntityOfDifferentTenant`. Each was verified to fail against the unpatched code — the timeseries one returns HTTP 200 with the other tenant's data, and the attribute-write one shows the attribute actually landing on the victim's entity while the response is 403. `testAttributesSubscription_tenantAdmin` was added for the authorized path.

`WebsocketApiTest` (22) and `TelemetryControllerTest` (9) pass in full.

### Behaviour changes worth knowing

- A customer user can no longer read device or asset profile attributes over WS (`CustomerUserPermissions.profilePermissionChecker` grants `READ` only), nor the attributes of the Tenant entity (`validateTenant` rejects customer users outright). A system administrator can no longer read a tenant device's attributes over WS. All of these have always been answered with 403 over REST, so the WS path only ever served them because it ignored the validation result — but a customer-facing dashboard bound to a "Current Tenant" alias that reads server attributes over WS will start reporting an error, so it belongs in the upgrade notes. The v1 attribute subscription is still used by the attributes-table widget datasource (`attribute-datasource.ts`), `utils.service.ts`, action widgets and the device connectivity dialog.
- The timeseries paths (`tsSubCmds`, `historyCmds`) already went through `on(...)` before this PR and therefore already behaved this way, so attributes are now merely consistent with telemetry.
- An `alarmStatusCmds` whose originator type is outside `AccessValidator.validate`'s switch (`DASHBOARD`, `QUEUE`, ...) now gets `BAD_REQUEST` instead of a normal `active: false` reply. Previously such a command would have thrown `IllegalStateException("Not Implemented!")` once the permission check was in place, which would answer `INTERNAL_ERROR` and log an ERROR with a stack trace on every attempt.
- A denied attribute subscription now comes back as `UNAUTHORIZED` instead of `INTERNAL_ERROR`/"Failed to fetch attributes!". `AccessValidator` only produces `accessDenied`/`entityNotFound`, so the `UnauthorizedException` branch that was meant to report this never matched; the no-keys variant reported `INTERNAL_ERROR` unconditionally.
- A system administrator subscribing to `alarmStatusCmds` is now refused explicitly. Such a subscription only ever returned the initial fetch (live updates are partitioned by the subscription's tenant id), and with the tenant filter added here that fetch would have come back empty.
- Validation denials on attribute subscriptions are logged at debug instead of error, so a client cannot flood the error log with denied or stale-entity subscriptions.

### Follow-up (not in this PR)

WS subscriptions that are registered in a callback after asynchronous validation (`attrSubCmds`, `tsSubCmds`, `alarmStatusCmds`) can create a context after the command was already cancelled, because unsubscribe runs inline. Closing that needs either cancellation state per cmdId or moving the post-validation step back onto the session's serial inbound path, and it affects all three commands, so it is left for a separate change. There is no cross-tenant exposure in it. The disconnect half is already handled here by the session-liveness guard, with the existing 60s `cleanupIfStale` sweep covering the remaining window.

The defects are present on `lts-4.3` and `master` as well and need a forward port.

### DIFF
--- a/application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
+++ b/application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
@@ -53,6 +53,8 @@ import org.thingsboard.server.common.msg.tools.TbRateLimitsException;
 import org.thingsboard.server.dao.attributes.AttributesService;
 import org.thingsboard.server.dao.tenant.TbTenantProfileCache;
 import org.thingsboard.server.dao.timeseries.TimeseriesService;
+import org.thingsboard.server.exception.AccessDeniedException;
+import org.thingsboard.server.exception.EntityNotFoundException;
 import org.thingsboard.server.exception.UnauthorizedException;
 import org.thingsboard.server.queue.discovery.TbServiceInfoProvider;
 import org.thingsboard.server.queue.util.TbCoreComponent;
@@ -77,6 +79,7 @@ import org.thingsboard.server.service.ws.telemetry.cmd.v1.TimeseriesSubscription
 import org.thingsboard.server.service.ws.telemetry.cmd.v2.AlarmCountCmd;
 import org.thingsboard.server.service.ws.telemetry.cmd.v2.AlarmDataCmd;
 import org.thingsboard.server.service.ws.telemetry.cmd.v2.AlarmStatusCmd;
+import org.thingsboard.server.service.ws.telemetry.cmd.v2.AlarmStatusUpdate;
 import org.thingsboard.server.service.ws.telemetry.cmd.v2.CmdUpdate;
 import org.thingsboard.server.service.ws.telemetry.cmd.v2.EntityCountCmd;
 import org.thingsboard.server.service.ws.telemetry.cmd.v2.EntityDataCmd;
@@ -266,8 +269,31 @@ public class DefaultWebSocketService implements WebSocketService {
     }
 
     private void handleWsAlarmsStatusCmd(WebSocketSessionRef sessionRef, AlarmStatusCmd cmd) {
-        if (validateCmd(sessionRef, cmd)) {
-            entityDataSubService.handleCmd(sessionRef, cmd);
+        if (!validateCmd(sessionRef, cmd, () -> {
+            if (cmd.getOriginatorId() == null) {
+                throw new IllegalArgumentException("Originator id is empty!");
+            }
+        })) return;
+
+        try {
+            accessValidator.validate(sessionRef.getSecurityCtx(), Operation.READ, cmd.getOriginatorId(),
+                    on(r -> executor.submit(() -> {
+                                if (!msgEndpoint.isOpen(sessionRef.getSessionId())) {
+                                    return;
+                                }
+                                try {
+                                    entityDataSubService.handleCmd(sessionRef, cmd);
+                                } catch (TbRateLimitsException e) {
+                                    log.debug("{} Failed to handle WS cmd: {}", sessionRef, cmd, e);
+                                } catch (Exception e) {
+                                    sendError(sessionRef, cmd.getCmdId(), SubscriptionErrorCode.INTERNAL_ERROR, e.getMessage());
+                                    log.error("{} Failed to handle WS cmd: {}", sessionRef, cmd, e);
+                                }
+                            }),
+                            t -> sendAlarmStatusError(sessionRef, cmd.getCmdId(), SubscriptionErrorCode.UNAUTHORIZED, t.getMessage())));
+        } catch (IllegalStateException e) {
+            sendAlarmStatusError(sessionRef, cmd.getCmdId(), SubscriptionErrorCode.BAD_REQUEST,
+                    "Unsupported originator type: " + cmd.getOriginatorId().getEntityType());
         }
     }
 
@@ -492,7 +518,7 @@ public class DefaultWebSocketService implements WebSocketService {
 
             @Override
             public void onFailure(Throwable e) {
-                log.error(FAILED_TO_FETCH_ATTRIBUTES, e);
+                logAttributesFetchFailure(e);
                 TelemetrySubscriptionUpdate update;
                 if (e instanceof UnauthorizedException) {
                     update = new TelemetrySubscriptionUpdate(cmd.getCmdId(), SubscriptionErrorCode.UNAUTHORIZED,
@@ -604,7 +630,7 @@ public class DefaultWebSocketService implements WebSocketService {
 
             @Override
             public void onFailure(Throwable e) {
-                log.error(FAILED_TO_FETCH_ATTRIBUTES, e);
+                logAttributesFetchFailure(e);
                 sendError(sessionRef, cmd.getCmdId(), SubscriptionErrorCode.INTERNAL_ERROR, FAILED_TO_FETCH_ATTRIBUTES);
             }
         };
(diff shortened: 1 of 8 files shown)


application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
```diff
@@ -266,8 +269,31 @@ private void handleWsAlarmCountCmd(WebSocketSessionRef sessionRef, AlarmCountCmd
...
-            entityDataSubService.handleCmd(sessionRef, cmd);
+        if (!validateCmd(sessionRef, cmd, () -> {
+            if (cmd.getOriginatorId() == null) {
+                throw new IllegalArgumentException("Originator id is empty!");
+            }
+        })) return;
+
+        try {
+            accessValidator.validate(sessionRef.getSecurityCtx(), Operation.READ, cmd.getOriginatorId(),
+                    on(r -> executor.submit(() -> {
```
zzzeebra, line 283: Beyond the cancel-after-validation race you explicitly deferred, this hop onto the 50-thread `executor` also removes the serialization that `handleCmd(AlarmStatusCmd)` was relying on. Inbound frames for one session are drained under `inboundMsgQueueProcessorLock` in `TbWebSocketHandler.SessionMetaData.tryProcessInboundMsgs`, so `DefaultTbEntityDataSubscriptionService.handleCmd` used to run strictly serially per session and its `getSubCtx(...) == null` → `createSubCtx(...)` check-then-act was safe. Now two `AlarmStatusCmd`s with the same cmdId (a duplicate cmd in one wrapper, or a resubscribe right after an unsubscribe) can be validated concurrently, both observe `ctx == null`, and both run `createSubCtx`: each calls `ctx.createSubscription(cmd)` (registering a distinct subscription in `localSubscriptionService`) and then `sessionSubs.put(cmdId, ctx)`, where last-write-wins. The losing context is no longer reachable from `subscriptionsBySessionId`, so `cancelSubscription`/`cleanupAndCancel` can never stop it — it keeps pushing duplicate `AlarmStatusUpdate`s for that cmdId until the session closes and `cancelAllSessionSubscriptions` runs. A `computeIfAbsent` on the cmdId (or doing the create under a per-session lock) would close this.

application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
```diff
@@ -266,8 +269,31 @@ private void handleWsAlarmCountCmd(WebSocketSessionRef sessionRef, AlarmCountCmd
...
+                                if (!msgEndpoint.isOpen(sessionRef.getSessionId())) {
+                                    return;
+                                }
+                                try {
+                                    entityDataSubService.handleCmd(sessionRef, cmd);
+                                } catch (TbRateLimitsException e) {
```
zzzeebra: This try/catch is a verbatim copy of the per-cmd error handling in `handleCommands` (TbRateLimitsException -> log.debug, Exception -> sendError(INTERNAL_ERROR) + log.error), needed only because `handleCmd` now runs on the WS executor and escapes the outer try. The two blocks now have to stay in sync: the next change to the generic WS cmd error policy in `handleCommands` (a new exception type, a different error code or update shape) will silently keep the old behavior for alarm-status cmds only. Worth extracting into a small private helper called from both places. Note also that this copy reports internal failures via `sendError`, i.e. a v1 `TelemetrySubscriptionUpdate`, while the two refusal paths added in this same method deliberately use `sendAlarmStatusError`/`AlarmStatusUpdate` — an alarm-status client gets two different update shapes depending on which failure it hits.

application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
```diff
@@ -492,7 +518,7 @@ public void onSuccess(List<AttributeKvEntry> data) {
 
             @Override
             public void onFailure(Throwable e) {
-                log.error(FAILED_TO_FETCH_ATTRIBUTES, e);
+                logAttributesFetchFailure(e);
```
zzzeebra, line 521: Now that a denied validation actually reaches this callback, the error code the client gets is wrong. `AccessValidator` only ever produces `ValidationResult.accessDenied(...)` / `entityNotFound(...)` — `ValidationResult.unauthorized(...)` and `internalError(...)` are never constructed anywhere in the tree — so `ValidationCallback.getException` hands us an `AccessDeniedException` or `EntityNotFoundException`, never an `UnauthorizedException`. The `e instanceof UnauthorizedException` branch is therefore dead for exactly the case this PR makes reachable, and a cross-tenant attribute subscription is reported as `INTERNAL_ERROR` with "Failed to fetch attributes!". The no-keys variant `handleWsAttributesSubscription` is worse — it unconditionally sends `INTERNAL_ERROR`. Contrast with the new alarm-status path, which correctly reports `UNAUTHORIZED`. Worth mapping `AccessDeniedException`/`EntityNotFoundException` to `SubscriptionErrorCode.UNAUTHORIZED` here (the same predicate `logAttributesFetchFailure` already uses), otherwise API clients can't tell a permission problem from a server fault. Note the new `testAttributesSubscription_differentTenant` only asserts `errorCode != NO_ERROR`, so it doesn't pin this either way.

application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
```diff
@@ -895,72 +921,50 @@ private ListenableFuture<List<AttributeKvEntry>> mergeAllAttributesFutures(List<
...
-            @Override
-            public void onSuccess(@Nullable ValidationResult result) {
-                List<ListenableFuture<List<AttributeKvEntry>>> futures = new ArrayList<>();
-                for (AttributeScope scope : AttributeScope.values()) {
-                    futures.add(attributesService.find(tenantId, entityId, scope, keys));
-                }
-
-                ListenableFuture<List<AttributeKvEntry>> future = mergeAllAttributesFutures(futures);
-                Futures.addCallback(future, callback, MoreExecutors.directExecutor());
+    private FutureCallback<ValidationResult> getAttributesFetchCallback(final TenantId tenantId, final EntityId entityId, final List<String> keys, final FutureCallback<List<AttributeKvEntry>> callback) {
```
zzzeebra, line 929: Heads-up on a same-tenant compatibility effect that isn't in the description: honouring the validation result here doesn't only block cross-tenant reads, it also blocks same-tenant combinations that the per-resource permission checkers happen not to grant for `READ_ATTRIBUTES`. Concretely, a customer user subscribing to attributes of a `DEVICE_PROFILE`/`ASSET_PROFILE` now fails (`CustomerUserPermissions.profilePermissionChecker` is a `GenericPermissionChecker(Operation.READ)` only), and so does a customer user subscribing to attributes of the Tenant entity (`validateTenant` rejects customer users outright) — both previously returned their own tenant's data. The mitigating factor is that the timeseries paths already went through `on(...)` before this PR and therefore already behaved this way, so attributes are now merely consistent with telemetry. Still worth calling out in the release notes, since customer-facing dashboards bound to a "Current Tenant" alias that read server attributes over WS will start erroring.

dao/src/main/java/org/thingsboard/server/dao/sql/alarm/AlarmRepository.java
```diff
@@ -402,11 +402,12 @@ String updateAlarm(@Param("t_id") UUID tenantId, @Param("a_id") UUID alarmId, @P
     int deleteTypeIfNoAlarmsExist(@Param("tenantId") UUID tenantId, @Param("types") Set<String> types);
 
     @Query(value = "SELECT a.id FROM alarm a " +
-            "WHERE a.originator_id = :originatorId " +
+            "WHERE a.tenant_id = :tenantId AND a.originator_id = :originatorId " +
```
zzzeebra, line 405: `TbAlarmStatusSubCtx.fetchActiveAlarms` passes `subscription.getTenantId()`, which is `sessionRef.getSecurityCtx().getTenantId()`. For a tenant admin or customer user that equals the alarm's `tenant_id`, but for a SYS_ADMIN session it is `TenantId.SYS_TENANT_ID`. That combination is reachable: `validateTenant` returns `ValidationResult.ok(null)` for sysadmin, so an `AlarmStatusCmd` whose originator is a `TenantId` passes the new `Operation.READ` check and then queries with `tenant_id = SYS_TENANT_ID` while the alarm rows carry the real tenant's id — the initial fetch now returns empty and the widget reports `active=false` instead of the tenant's active alarms, silently. (Live updates for such a subscription probably never arrived anyway, since `DefaultSubscriptionManagerService` partitions by `sub.getTenantId()`, so only the initial fetch regresses — but `testAttributesSubscription_sysAdmin` shows sysadmin WS subscriptions on a tenant entity are a supported scenario.) Is a sysadmin alarm-status subscription meant to work? If not, denying it explicitly in the handler would be clearer than returning an empty set.

zzzeebra (commented):
## Review summary

Reviewed 7 changed files in [Fixed unauthorized access to attributes and time series of other tenants](https://github.com/thingsboard/thingsboard/pull/16134). Left 5 comment(s) inline.

The bypasses themselves look correctly closed. The two worth looking at first are both in `DefaultWebSocketService`: making the alarm-status handler asynchronous drops the per-session serialization that `DefaultTbEntityDataSubscriptionService.handleCmd`'s check-then-act on `cmdId` relied on, which can leave an orphaned subscription that no `unsubscribe` can reach; and now that denials actually reach the attribute-subscription `onFailure`, they are reported to the client as `INTERNAL_ERROR` rather than `UNAUTHORIZED`, because `AccessValidator` never produces an `UnauthorizedException`. There is also a same-tenant behaviour change (customer users reading profile/tenant attributes over WS) that is broader than the cross-tenant fix the description promises.

---

application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
zzzeebra: Fixed: the cmdId slot is now reserved with `putIfAbsent`, and the thread that loses the race stops its own context. Kept create-then-publish ordering on purpose — publishing an empty ctx first would NPE in `TbAlarmStatusSubCtx.stop()` on the null `subscription`.

application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
zzzeebra: Fixed: extracted `handleCmdFailure(sessionRef, cmd, e)`, called from both `handleCommands` and the deferred task. Internal failures for alarm-status now go out as `AlarmStatusUpdate` too, so the client no longer gets two shapes.

application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
zzzeebra: Fixed: both variants now report `UNAUTHORIZED` via the same `isValidationFailure` predicate `logAttributesFetchFailure` uses, and the test asserts the exact code.

application/src/main/java/org/thingsboard/server/service/ws/DefaultWebSocketService.java
zzzeebra: Added to the PR description, including the Tenant-entity case and the note that REST always answered 403 for these. Upgrade notes when 4.2.x ships.

dao/src/main/java/org/thingsboard/server/dao/sql/alarm/AlarmRepository.java
zzzeebra: Fixed: sysadmin alarm-status subscriptions are now denied explicitly instead of returning an empty set.

PR #16138, opened 2026-09-10
Keep IoT Hub markdown table columns from overlapping

pinkevmladchy:
Markdown tables in the IoT Hub instructions could not cope with long values. In the "Solution instructions" dialog of an installed solution template, logins and e-mails overflowed their own column and ran into the text of the next one, and in cells with a copy button the value also ran under the copy icon.

Three separate reasons, each fixed in `iot-hub-markdown.component.scss`:

- `th, td` had no word breaking at all, so a token wider than the column simply overflowed — cells now use `overflow-wrap: anywhere`, with `overflow-wrap: break-word` kept in front of it as a fallback for engines without `anywhere`
- cells holding a copy button reserved no room for it: the icon is 38px wide and offset 3px from the right edge, so the text ran underneath it — `div.code-wrapper` now keeps 44px of padding on the right
- `table-layout: fixed` gave every column the same width (~169px in a 900px dialog), so an e-mail like `alta.sweeney@thingsboard.io` had to wrap mid-word while the Password and User Group columns stood half empty — the table is now sized by its content

`anywhere` rather than plain `break-word` on purpose: `break-word` keeps the min-content width of a cell at its longest token, so with auto layout the sum of the column minimums exceeds the dialog width, the table cannot shrink and gets clipped by its own `overflow: hidden` in a narrow window. `anywhere` drops the min-content width to a single character, so the table always fits.

Dropping `&.auto { table-layout: auto; }` is intentional — auto is the default now, so the class became a no-op here. The `{:auto}` marker in `marked-options.service.ts` still strips itself from the header and adds `class="auto"`, and in the base markdown styles the class works as before.

Scoped to the IoT Hub markdown component: `shared/components/markdown.component.scss` has the same latent issue but is left untouched.

The file is identical in PE, so it merges forward into PE and into `lts-4.3` without conflicts — no separate PE PR is needed.

### DIFF
--- a/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
+++ b/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
@@ -204,9 +204,19 @@ $gallery-border-hover-color: #2a7dec;
     }
 
     th, td {
+      // Long unbreakable tokens (logins, e-mails, passwords) must never overflow their
+      // own column. `anywhere` also drops the cell min-content width down to a single
+      // character, so auto layout can always squeeze the table into the available
+      // width instead of overflowing (and getting clipped by `overflow: hidden`).
+      overflow-wrap: break-word; // fallback for engines without `anywhere`
+      overflow-wrap: anywhere;
+
       div.code-wrapper {
         display: inline-block;
         width: 100%;
+        // Reserve room for the absolutely positioned copy icon (38px wide, offset 3px
+        // from the right edge) so the cell text does not run underneath it.
+        padding-right: 44px;
 
         button.clipboard-btn {
           top: -6px;
@@ -275,11 +285,10 @@ $gallery-border-hover-color: #2a7dec;
       margin-top: 20px;
       margin-bottom: 20px;
       overflow: hidden;
-      table-layout: fixed;
-
-      &.auto {
-        table-layout: auto;
-      }
+      // Columns are sized to their content: with fixed layout every column got the
+      // same width, so cells with long values (e-mails, logins, tokens) had to wrap
+      // mid-word while short columns kept a lot of unused space.
+      table-layout: auto;
 
       > thead {
         background-color: $table-header-background-color;
--- a/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
+++ b/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
@@ -211,6 +211,16 @@ $gallery-border-hover-color: #2a7dec;
       overflow-wrap: break-word; // fallback for engines without `anywhere`
       overflow-wrap: anywhere;
 
+      // Text is not the only thing that can outgrow a column: a preformatted block keeps
+      // its whole line as min-content width, which auto layout cannot shrink, so the table
+      // would push past its container and get cut off by a host that hides horizontal
+      // overflow (iot-hub-item-detail-dialog's .dlg-body). Wrap it inside the cell instead.
+      // Prism sets `white-space: pre` on both the block and the code inside it, so both
+      // have to be relaxed for the cell to shrink.
+      pre, pre > code {
+        white-space: pre-wrap;
+      }
+
       div.code-wrapper {
         display: inline-block;
         width: 100%;
--- a/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
+++ b/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
@@ -3,6 +3,10 @@
 $text-color: rgba(0,0,0,0.76);
 $link-color: #2a7dec;
 $copy-button-color: #2a7dec;
+// Geometry of the copy icon chip, shared by the chip itself and by the space table
+// cells reserve for it, so the two cannot drift apart.
+$copy-icon-width: 38px;
+$copy-icon-offset: 3px;
 
 $code-block-border-color: rgba(42, 125, 236, .2);
 $code-default-color: #212529;
@@ -169,10 +173,10 @@ $gallery-border-hover-color: #2a7dec;
         div {
           background-color: #fff;
           position: absolute;
-          width: 38px;
+          width: $copy-icon-width;
           height: 28px;
-          top: 3px;
-          right: 3px;
+          top: $copy-icon-offset;
+          right: $copy-icon-offset;
           padding: 8px 10px 0;
 
           img {
@@ -224,9 +228,10 @@ $gallery-border-hover-color: #2a7dec;
       div.code-wrapper {
         display: inline-block;
         width: 100%;
-        // Reserve room for the absolutely positioned copy icon (38px wide, offset 3px
-        // from the right edge) so the cell text does not run underneath it.
-        padding-right: 44px;
+        // Reserve room for the absolutely positioned copy icon, plus the same gap on its
+        // far side, so the cell text does not run underneath it and the reservation
+        // follows the chip if its geometry ever changes.
+        padding-right: $copy-icon-width + 2 * $copy-icon-offset;
 
         button.clipboard-btn {
           top: -6px;


vvlladd28 (commented):
## Review summary

Reviewed 1 changed file in [Keep IoT Hub markdown table columns from overlapping](https://github.com/thingsboard/thingsboard/pull/16138). Left 2 comment(s) inline.

The one to look at first is the removal of `table-layout: fixed`: under auto layout an oversized image in a table cell can expand its column past the container, and inside `iot-hub-item-detail-dialog`'s `overflow-x: hidden` body that content is cut off rather than scrollable. The second comment is about the hand-derived `44px` staying in sync with the copy icon's own dimensions.

---

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
```diff
@@ -275,11 +285,10 @@ $gallery-border-hover-color: #2a7dec;
       margin-top: 20px;
       margin-bottom: 20px;
       overflow: hidden;
-      table-layout: fixed;
```
vvlladd28, line 278: Dropping `table-layout: fixed` also drops the hard guarantee that the table never gets wider than its container. With fixed layout, `width: 100%` was the used width no matter what was in the cells; with auto layout the used width becomes `max(100%, min-content)`, and `overflow-wrap: anywhere` only collapses the min-content of *text* — it does nothing for replaced or non-wrappable boxes.

The concrete case is images: this stylesheet sets `img { max-width: 100% }` (line 113) at `.tb-markdown-view` level, so it applies inside table cells. In an auto-layout table the cell's containing block is indefinite while intrinsic widths are computed, so a percentage `max-width` on a replaced element doesn't constrain the intrinsic contribution — a wide screenshot in a markdown table cell should now expand its column to the image's natural width instead of being scaled down as it was under fixed layout. Secondary: the new `padding-right: 44px` puts a floor of ~44px + td padding under every copy-code cell (24px at normal widths, 8px under the `max-width: 400px` query), so a table with several copy-code columns has a real min-content floor at the narrowest dialog sizes.

Where that overflow lands differs by host. In `solution-install-dialog` the markdown sits in `mat-dialog-content`, so the worst case is a horizontal scrollbar. But `iot-hub-item-detail-dialog.component.scss`'s `.dlg-body` sets `overflow-y: auto; overflow-x: hidden;` and wraps both `.dlg-description` and `.dlg-readme` — there an over-wide table is cut off with no way to scroll to it. Note the table's own `overflow: hidden` already clipped overflowing cell text before this change, so this isn't wholly new; what changes is that an oversized image now blows out the column instead of scaling to fit. Would it be worth pairing this with `max-width: 100%` on the table, or wrapping markdown tables in an `overflow-x: auto` container, so the failure mode is a scrollbar rather than unreachable content?

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
```diff
@@ -204,9 +204,19 @@ $gallery-border-hover-color: #2a7dec;
...
+      // width instead of overflowing (and getting clipped by `overflow: hidden`).
+      overflow-wrap: break-word; // fallback for engines without `anywhere`
+      overflow-wrap: anywhere;
+
       div.code-wrapper {
         display: inline-block;
         width: 100%;
+        // Reserve room for the absolutely positioned copy icon (38px wide, offset 3px
+        // from the right edge) so the cell text does not run underneath it.
+        padding-right: 44px;
```
vvlladd28: The 44px here is 38 + 2*3, derived by hand from the copy-icon chip declared ~45 lines above in this same file (`div.code-wrapper button.clipboard-btn div { width: 38px; ... right: 3px; }`, lines 172/175). Nothing links the two, so the next person who nudges that chip's width or its right offset gets the text running under the icon again — with no compile error or test to catch it. The comment you added documents the derivation, which helps, but the neighbouring values in that same chip block have already drifted against the sibling fork `ui-ngx/src/app/shared/components/markdown.component.scss` (there the chip `div` carries `padding: 10px 10px 0` vs `8px 10px 0` here, and `button.clipboard-btn` has `right: 32px` / `right: 0` in the td override vs `right: 0` / no override here), so these numbers do get edited in isolation. Would you consider hoisting them, e.g. `$copy-icon-width: 38px; $copy-icon-offset: 3px;`, using them in the chip rule, and writing `padding-right: $copy-icon-width + 2 * $copy-icon-offset;` so the reservation follows the button automatically?

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
pinkevmladchy: Measured this one before changing anything, and the trigger turned out to be a different one than the image.

**The image case does not reproduce.** `width: 100%` gives the table a definite width, so auto layout distributes columns inside it and the percentage `max-width` on the image resolves against a definite column. A 1600px SVG in a cell, container widths 900 / 600 / 400 / 300 / 200px, Chrome and Firefox agree:

```
auto @900: table=896 img=676   auto @400: table=396 img=266
auto @600: table=596 img=430   auto @300: table=296 img=184
```

The table equals its container at every width and the image scales down, exactly as it did under fixed layout (only smaller: 281px at 900px).

**`max-width: 100%` on the table would not have helped where it does break.** A min-content floor wins over `max-width`, so the declaration is a no-op in both directions — measured with a floor present: `auto + max-width:100%` in a 396px host still lays out at 701px and is still clipped, same as without it.

**What actually breaks is a preformatted block in a cell.** Prism sets `white-space: pre` on both `pre[class*="language-"]` and the `code` inside it, so such a cell keeps the whole command line as its min-content width — 701px inside a 396px `.dlg-body`, cut off with no way to scroll to it, which is the failure mode you describe. Fixed in d0415fb by relaxing both to `pre-wrap` inside cells, in the same spirit as the `anywhere` above it: let the content wrap instead of forcing the table to grow. After it, with the real compiled stylesheet and three tables (copy-code users, a code block, a 1600px image):

```
@900: table=848  @400: table=348   CLIPPED=false, cell overflows=0
@660: table=608  @320: table=268   CLIPPED=false, cell overflows=0
```

**On the copy-code floor:** two copy-code columns in a five-column table stay well inside the narrowest sizes — 274px table in a 316px host at 320px, no clipping. The 44px is now derived rather than hand-written, see the other thread.

The trade-off worth naming: a wrapped Prism block desynchronises its line-number gutter. That only affects code blocks authored inside table cells, and today those are clipped and unreadable anyway, so wrapping strictly improves them.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss
pinkevmladchy: Agreed, and the drift you point at in the sibling fork makes the case. Hoisted in 0d54014:

```scss
$copy-icon-width: 38px;
$copy-icon-offset: 3px;
```

The chip rule now reads `width: $copy-icon-width; top: $copy-icon-offset; right: $copy-icon-offset;`, and the cell reservation is `padding-right: $copy-icon-width + 2 * $copy-icon-offset` — icon width plus the same gap on either side, so moving or resizing the chip carries the reservation with it. Compiled output is unchanged (`padding-right: 44px`), verified against the emitted CSS and in the browser (`getComputedStyle(...).paddingRight === "44px"`).

vvlladd28 (commented):
## Re-review summary

Re-reviewed [Keep IoT Hub markdown table columns from overlapping](https://github.com/thingsboard/thingsboard/pull/16138) — verified 2 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/16138#pullrequestreview-5167691718).

| Status | Count |
|--------|-------|
| ✅ Resolved | 2 |
| 💬 Acknowledged | 0 |

All previous findings have been addressed. No new issues found in the fix commits.

### Finding details

- ✅ **ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss:306** — Dropping `table-layout: fixed` removes the hard guarantee that the table never exceeds its container, so an unshrinkable min-content floor would be clipped unreachably by `.dlg-body`'s `overflow-x: hidden`. — *Fixed in code.* The measurements disprove the image mechanism I proposed, and the correction is right on both counts: `width: 100%` does give the table a definite width for percentage resolution, and `max-width: 100%` would indeed have been a no-op since a table's min-content floor wins over it. The real trigger — Prism's `white-space: pre` on both the block and its inner `code` — is addressed at line 225 by relaxing both to `pre-wrap`. That works because the `overflow-wrap: anywhere` above it is an inherited property, and `anywhere` (unlike `break-word`) does reduce min-content size, so cell content now collapses to a single character rather than a whole command line. I checked the remaining floor candidates in this stylesheet: no `width` declarations on `th`/`td` ever depended on fixed layout, and the only `white-space: nowrap` (line 460, `.tb-image-tooltip`) is absolutely positioned and so contributes nothing to intrinsic width.
- ✅ **ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-markdown.component.scss:234** — The `44px` reservation was derived by hand from the copy-icon chip declared ~45 lines above, with nothing linking the two, and the neighbouring chip values have already drifted against the sibling fork. — *Fixed in code.* `$copy-icon-width` / `$copy-icon-offset` are hoisted at lines 8-9 and now drive both the chip rule (lines 176-179) and the reservation, so the two can no longer drift. Arithmetic is preserved: 38 + 2×3 = 44px.

Also verified the PR description's claim that dropping `&.auto` is safe: with `auto` now the default for these tables, the class is a no-op in this stylesheet.

---

vvlladd28 (approved):

PR #16139, opened 2026-09-10
Refresh the installed items count after an install, update or delete

pinkevmladchy:
The IoT Hub items page read its installed-items counter once, on init. Installing, updating or deleting an item from the browse list embedded in that page refreshed the list itself but not the counter above it, so the number stayed stale until the page was reloaded.

- `tb-iot-hub-browse` now emits `installedItemsChanged` whenever it refreshes its own installed-items state, which already happens after every install, update and delete
- the items page listens for it and reloads the counter, so the count matches the list without a reload

Merges forward without conflicts: the three touched files are identical in `lts-4.2`, `lts-4.3` and `master`, and identical to their PE counterparts, so a single PR here is enough — no PE PR needed.

### DIFF
--- a/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-browse.component.ts
+++ b/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-browse.component.ts
@@ -51,6 +51,11 @@ export class TbIotHubBrowseComponent implements OnInit, AfterViewInit, OnDestroy
   @Input() mode: 'default' | 'add' = 'default';
   @Input() fixedSubType: string;
   @Output() addItem = new EventEmitter<MpItemVersionView>();
+  /**
+   * Fired whenever an install, update or delete changed what this tenant has installed, so a host
+   * page showing its own installed-items counter can refresh it instead of waiting for a reload.
+   */
+  @Output() installedItemsChanged = new EventEmitter<void>();
   @Input() set activeType(value: ItemType) {
     if (value && value !== this._activeType) {
       const wasInit = !!this._activeType;
@@ -643,6 +648,7 @@ export class TbIotHubBrowseComponent implements OnInit, AfterViewInit, OnDestroy
         this.installedItemCounts = counts;
       });
     }
+    this.installedItemsChanged.emit();
   }
 
   private loadFilterInfo(): void {
--- a/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-browse.component.ts
+++ b/ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-browse.component.ts
@@ -632,7 +632,11 @@ export class TbIotHubBrowseComponent implements OnInit, AfterViewInit, OnDestroy
     });
   }
 
-  private reloadInstalledItems(): void {
+  /**
+   * Re-reads what is installed for the active type and announces the change. Public so a host page
+   * that opens the item detail dialog itself can refresh this grid through the same path.
+   */
+  reloadInstalledItems(): void {
     const config = {ignoreLoading: true};
     const pageLink = new PageLink(10000, 0);
     if (this.activeType === ItemType.WIDGET) {
--- a/ui-ngx/src/app/modules/home/components/iot-hub/device-install-dialog/device-install-dialog.component.ts
+++ b/ui-ngx/src/app/modules/home/components/iot-hub/device-install-dialog/device-install-dialog.component.ts
@@ -95,6 +95,12 @@ export class TbDeviceInstallDialogComponent extends DialogComponent<TbDeviceInst
   wizardSteps: WizardStep[] = [];
   wizardStarted = false;
   reviewMode = false;
+  /**
+   * Set once the server has an installed item for this device. From that moment every way out of
+   * the wizard — including the header X and the Cancel button on a still-pending later step —
+   * has to report 'installed', or the caller keeps showing pre-install counts and badges.
+   */
+  private installRegistered = false;
 
   // Variable resolution state
   formValues: Record<string, any> = {};
@@ -357,7 +363,7 @@ export class TbDeviceInstallDialogComponent extends DialogComponent<TbDeviceInst
   }
 
   cancel(): void {
-    this.dialogRef.close(false);
+    this.dialogRef.close(this.installRegistered ? 'installed' : false);
   }
 
   retryEntitySteps(step: WizardStep): void {
@@ -809,6 +815,7 @@ export class TbDeviceInstallDialogComponent extends DialogComponent<TbDeviceInst
           { ignoreLoading: true }
         )
       );
+      this.installRegistered = true;
     } catch (_e) {
       // Non-critical — entities are created, tracking registration failed
       console.error('Failed to register device install', _e);
--- a/ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-items-page.component.html
+++ b/ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-items-page.component.html
@@ -35,7 +35,8 @@
   <!-- Browse component (without tabs) -->
   <tb-iot-hub-browse
     [activeType]="config.type"
-    [hideTabs]="true">
+    [hideTabs]="true"
+    (installedItemsChanged)="loadInstalledCount()">
   </tb-iot-hub-browse>
 
   <!-- Become a Creator footer -->
--- a/ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-items-page.component.ts
+++ b/ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-items-page.component.ts
@@ -109,7 +109,7 @@ export class TbIotHubItemsPageComponent implements OnInit {
     window.open(this.iotHubApiService.baseUrl + '/signup', '_blank');
   }
 
-  private loadInstalledCount(): void {
+  loadInstalledCount(): void {
     this.iotHubApiService.getInstalledItemsCount(this.config.type, { ignoreLoading: true }).subscribe(count => {
       this.installedItemsCount = count;
     });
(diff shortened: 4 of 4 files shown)


ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-browse.component.ts
```diff
@@ -643,6 +648,7 @@ export class TbIotHubBrowseComponent implements OnInit, AfterViewInit, OnDestroy
         this.installedItemCounts = counts;
       });
     }
+    this.installedItemsChanged.emit();
```
vvlladd28, line 655: Hooking the emit into `reloadInstalledItems()` is the right seam, but the premise that this method "already runs after every install" doesn't quite hold for the Device library page, so the counter can still be left stale there. `installItem`/`updateItem`/`deleteInstalledItem` only call it when the dialog result is `'installed'`/`'updated'`/truthy, and the device wizard (`device-install-dialog.component.ts`) calls `registerDeviceInstall(...)` as soon as the entity steps finish (line ~806), then auto-advances to the next wizard step. The header close button is rendered unconditionally and always calls `cancel()`, which closes with `false` (the footer's Close/Cancel switch on `allProgressDone`, but the X does not). So if the user dismisses the wizard with the X — or with Cancel on a still-pending later step — instead of the final Close, an installed item exists on the server, `reloadInstalledItems()` never runs, no `installedItemsChanged` is emitted, and both the card's install badge and the new page counter stay stale until a reload. Worth deciding whether the wizard should resolve to `'installed'` once `registerDeviceInstall` has succeeded (e.g. a flag set there, read by `cancel()`), so the new output fires for that path too.

vvlladd28 (commented):
## Review summary

Reviewed 3 changed files in [Refresh the installed items count after an install, update or delete](https://github.com/thingsboard/thingsboard/pull/16139). Left 1 comment(s) inline.

The wiring itself looks sound: `reloadInstalledItems()` is the single chokepoint for all four mutation paths in the browse component, so the emit fires exactly once per confirmed install/update/delete and costs one extra count GET. The inline comment is about one install path that never reaches that chokepoint (the device wizard closing via the header X after `registerDeviceInstall` already succeeded). Two further entry points that still leave this page's counter stale are noted below — both pre-existing, but they bound what "the count matches the list without a reload" actually delivers today.

## Additional findings

- **ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-item-detail-dialog.component.ts:253** — There's a second delete/update path that the new output won't cover. For the count-based types the items page shows (DEVICE, CALCULATED_FIELD, ALARM_RULE, RULE_CHAIN) the detail dialog renders the "Installed items: N" button when `installedItemsCount > 0`, which opens `TbIotHubInstalledItemsDialogComponent` → `tb-iot-hub-installed-items-table`, whose `deleteItem()`/`updateItem()` mutate installed items and then only reload the table's own page. That dialog's result is discarded here (`.subscribe()` with no handler) and the detail dialog itself isn't closed, so when the detail dialog eventually closes it resolves to `undefined`; `TbIotHubBrowseComponent.openItemDetail` therefore skips `reloadInstalledItems()`, nothing is emitted, and the page counter keeps showing the pre-delete number. Could this dialog propagate a `'deleted'`/`'updated'` outcome (e.g. have the nested dialog close with a flag and close/refresh the detail dialog accordingly) so the new refresh path covers it?

- **ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-items-page.component.ts:150** — The deep-link entry point on this same page opens the item detail dialog directly and throws away its result, so an install/update/delete performed from the auto-opened dialog refreshes neither the counter nor the browse grid — exactly the staleness this PR is fixing for the other path. `TbIotHubItemResolverComponent.openOnTypePage()` navigates here with `history.state.openItem` and `preview: false` for published items, so the install/update/delete buttons in that dialog are live. Since `loadInstalledCount()` is public now anyway, this could be handled here with the same result check the browse component uses, e.g. `.subscribe(result => { if (result === 'installed' || result === 'updated' || result === 'deleted') { this.loadInstalledCount(); } })` — though the browse grid would still need its own nudge to look right, which may argue for routing the deep-link open through the browse component instead.

---

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-browse.component.ts
pinkevmladchy: Confirmed — the device wizard registers the install as soon as the entity steps finish (`registerDeviceInstall` in `runEntitySteps`), while the header X calls `cancel()` unconditionally and the footer Cancel is still rendered until every progress step reports done. Both resolved to `false`, so nothing reloaded even though an installed item already existed on the server.

Fixed in e3e2b7f: `installRegistered` is set right after `registerDeviceInstall` succeeds, and `cancel()` now closes with `'installed'` when it is set. Registration failures keep closing with `false`, since no installed item exists in that case.

One visible consequence worth flagging: when the wizard is opened from the item detail dialog, `installDevice()` closes that dialog on an `'installed'` result, so dismissing the wizard with X after registration now closes both. That matches what `done()` already did.

pinkevmladchy:
On the two additional findings:

**Deep link on the items page** (`iot-hub-items-page.component.ts`) — fixed in e3e2b7f. Rather than calling `loadInstalledCount()` directly, which would have left the grid stale as you noted, `TbIotHubBrowseComponent.reloadInstalledItems()` is now public and the page holds a `@ViewChild` to it. The deep-linked dialog's result is fed back through the grid, which then emits `installedItemsChanged` and refreshes the counter through the same path as every other mutation. One chokepoint instead of two.

**Nested installed-items dialog** (`iot-hub-item-detail-dialog.component.ts:253`) — the chain checks out: the table's `deleteItem()`/`updateItem()` only call `loadData()`, `TbIotHubInstalledItemsDialogComponent.close()` resolves with no value, and `openInstalledItemsDialog()` discards it, so a delete there leaves both the counter and the card badge stale.

Leaving that one out of this PR. Propagating it means introducing a new dialog outcome that `TbIotHubBrowseComponent`, `TbIotHubHomeComponent` and `TbIotHubSearchComponent` all have to learn to handle — a change to a shared result contract, wider than refreshing the count after an install. Happy to do it as a follow-up if you would rather it land now.

vvlladd28 (commented):
## Re-review summary

Re-reviewed [Refresh the installed items count after an install, update or delete](https://github.com/thingsboard/thingsboard/pull/16139) — verified 3 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/16139#pullrequestreview-5167660534).

| Status | Count |
|--------|-------|
| ✅ Resolved | 2 |
| 💬 Acknowledged | 1 |

All previous findings have been addressed, and no new issues in the fix commits.

I traced every consumer of the device-install dialog result, since `cancel()` now returns `'installed'` instead of `false`: `TbIotHubItemDetailDialogComponent.installDevice()` (:197, closes the detail dialog — the consequence you already flagged), `TbIotHubAddItemDialogComponent.installDeviceItem()` (:109) and `TbIotHubBrowseComponent.installItem()` (:565). All three are correct under the new value — the add-item dialog is a single-add dialog that closes on every successful add anyway, so the X path now simply matches the `done()` path instead of silently skipping `config.updateData()`. `reviewDevice()` callers are unaffected, since review mode goes through `showCompletedEntitySteps()` and never registers.

### Finding details

- ✅ **ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-browse.component.ts:655** — The device wizard registered the install as soon as the entity steps finished, but dismissing it with the header X (or Cancel on a still-pending later step) closed with `false`, so `reloadInstalledItems()` never ran and the counter/badge stayed stale. — *Fixed in code.* `installRegistered` is set right after `registerDeviceInstall` succeeds and `cancel()` now closes with `'installed'`. Verified the fix is complete for this path: the dialog is opened with `disableClose: true`, so `cancel()` is the only dismissal route (no ESC/backdrop hole), and review mode goes through `showCompletedEntitySteps()` rather than `runEntitySteps()`, so the flag correctly stays `false` there.
- ✅ **ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-items-page.component.ts:150** — The deep-link entry point opened the item detail dialog and discarded its result, so an install/update/delete from the auto-opened dialog refreshed neither the counter nor the browse grid. — *Fixed in code.* Routing it through the now-public `reloadInstalledItems()` via `@ViewChild` rather than calling `loadInstalledCount()` directly is the better of the two options I offered — it keeps one chokepoint and leaves the grid consistent. The `@ViewChild` is resolved by the time the callback runs (it fires after the resolution request and the dialog close, both well past `ngAfterViewInit`).
- 💬 **ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-item-detail-dialog.component.ts:253** — The nested installed-items dialog discards its outcome, so deletes/updates performed there leave the page counter stale. — *Developer: confirmed the chain, but deferring — propagating it would mean a new dialog outcome that `TbIotHubBrowseComponent`, `TbIotHubHomeComponent` and `TbIotHubSearchComponent` all have to handle, i.e. a change to a shared result contract wider than this PR; offered as a follow-up.* Agreed that it belongs in a follow-up, though the scope argument is a little stronger than it needs to be — see below.

## Additional findings

- **ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-item-detail-dialog.component.ts:253** — On the deferral rationale, for whoever picks up the follow-up: a delete wouldn't actually need a new outcome on the shared contract. `TbIotHubBrowseComponent` (:554) and `TbIotHubSearchComponent` (:220) already handle `'installed' | 'updated' | 'deleted'`, and `TbIotHubHomeComponent` (:295) already handles `'deleted'`, so the nested dialog could resolve with a flag and have `openInstalledItemsDialog()` close the detail dialog with the existing `'deleted'`/`'updated'` value, with no consumer changes at all. The real obstacle is that the result is only delivered through `afterClosed()`, so any outcome forces the detail dialog shut — which is the wrong behaviour when the user deletes one of several installed items and wants to stay on the item. Refreshing `installedItemsCount` in place (and only signalling the host on close) is the part that needs new plumbing, so deferring is still the right call.

- **ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.ts:295** — Pre-existing gap in that same result contract: this consumer checks `result === 'installed' || result === 'deleted'` and omits `'updated'`, while the browse, search and (as of this PR) items-page consumers all handle the full triple. So updating an installed item from the detail dialog opened off the IoT Hub home page leaves `installedItemsCount` and the per-type installed badges on that page stale until a reload — the same class of staleness this PR fixes elsewhere. Adding `|| result === 'updated'` there looks like all it needs.

---

vvlladd28 (approved):

PR #16143, opened 2026-09-11
Receive mail/SMS/notification uplink requests from edge (master)

MazurenkoNick:
## Pull Request description

Master port of #15940 (lts-4.2) and #15941 (lts-4.3).

Delegates mail / SMS / notification sending from the edge to the cloud, and stops syncing admin settings to the edge. The edge cannot resolve tenant/system provider configuration on its own: the provider secrets are encrypted with the system key, which never leaves the cloud. Instead of shipping those settings down, the edge now packages the call and the cloud performs the send.

**Scope of changes (cloud side):**

- `edge.proto` — new `SendEmailUplinkMsg`, `SendSmsUplinkMsg`, `SendNotificationUplinkMsg` (each carrying the JSON-serialized request payload), wired into `UplinkMsg` as fields 35-37. The cloud always resolves the tenant from the authenticated edge session and ignores the `tenantId*` fields on these messages.
- `EdgeUplinkMessageDispatcher` — routes the three new uplink message types to `EdgeRequestsService`.
- `EdgeRequestsService` / `DefaultEdgeRequestsService` — new `processSendEmailMsg` / `processSendSmsMsg` / `processSendNotificationMsg`, which resolve the provider config on the cloud and perform the send.
- `AdminSettingsEdgeProcessor` — admin settings are no longer synced to the edge, apart from the tenant-scoped `general` and `connectivity` keys.
- `EdgeSyncCursor` — no longer seeds the admin settings fetchers.

Admin settings already persisted on existing edges are purged by the edge-side migration, not from the cloud.

**Documentation:** no user-facing changes to document. Mail/SMS/notification behaviour on the edge is unchanged from the user's point of view; only the execution location moves to the cloud.

### Crosslinks

| | CE | PE | Edge CE |
|---|---|---|---|
| lts-4.2 | #15940 | thingsboard/thingsboard-pe#4934 | thingsboard/thingsboard-edge#227 |
| lts-4.3 | #15941 | thingsboard/thingsboard-pe#4935 | thingsboard/thingsboard-edge#228 |
| master | this PR | thingsboard/thingsboard-pe#5111 | — |

## General checklist

## Back-End feature checklist


MazurenkoNick (commented):
## Review summary

Re-reviewed after the two hardening commits landed. Everything substantive from the earlier pass is addressed — only minor, non-blocking items remain.

### Addressed

| | |
|---|---|
| **Provider credentials could reach the cloud log** | `JacksonUtil.fromString(s, clazz)` builds its message as `"...: " + string`, so a parse failure put the whole payload into the exception — and for `TEST_MAIL` / `SEND_TEST_SMS` that payload carries the SMTP password and the Twilio/AWS SNS/SMPP secrets. `parseSendRequest` now goes through `IGNORE_UNKNOWN_PROPERTIES_JSON_MAPPER` directly, so the document never reaches the logger, and unknown fields no longer fail the parse if a field is renamed or removed cloud-side while an older edge still sends it. |
| **Delegated sends were acked before they ran** | `submitEdgeSend` returned `Futures.immediateFuture(null)` and swallowed the task's exception, so `EdgeGrpcSession` replied `success=true` regardless. On the edge that ack is a durable commit — `pendingMsgMap.remove(...)` followed by `updateQueueStartTsSeqIdOffset(...)` — so a failed activation or password-reset mail was silently unrecoverable. It now returns the real future, so a failure rejects the uplink and the edge's existing retry path applies. |
| **Blocking sends on the shared DB-callback pool** | SMTP / SMS / FCM calls ran on `DbCallbackExecutorService`, which the rule engine uses platform-wide. They now run on `MailExecutorService` / `SmsExecutorService` / `NotificationExecutorService`, so a stalled provider degrades its own channel instead of rule-engine DB callbacks. |
| **Unbounded send duration** | Wrapped in `Futures.withTimeout` at `edges.rpc.delegated_send_timeout_sec` (default 20s). This keeps the uplink answered inside the edge's `cloud.uplink_pack_timeout_sec` (60s), which matters now that the ack is real — an unanswered pack makes the edge resend everything unacked and deliver the same mail twice. It also caps the one timeout the edge controls, since `TEST_MAIL` reads its own from the edge-supplied config. |
| **Transient FCM failures were indistinguishable from dead tokens** | The blanket `catch (Exception e)` meant mobile push was the only channel that didn't benefit from the change above — an FCM outage, quota breach or auth failure was reported as delivered. Now only `UNREGISTERED` / `INVALID_ARGUMENT` / `SENDER_ID_MISMATCH` are skipped (at `debug`), and everything else propagates. |

### Minor — non-blocking

- **`EdgeSyncCursor:52`** — the comment says `'general'/'connectivity' are system-scoped and the edge auto-generates its own`, but `AdminSettingsEdgeProcessor` deliberately syncs exactly those keys when they are *tenant*-scoped, and rejects the system ones. Reading `EdgeSyncCursor` alone, the allow-list looks like dead code. Worth pointing the comment at `EDGE_SYNCED_SETTINGS_KEYS` rather than restating the rule. Same copy in `AbstractEdgeTest`.
- **Initial sync vs. incremental** — with the fetchers removed, a tenant's `general`/`connectivity` reach an edge only when they are next edited, so a freshly provisioned edge and a long-lived one can differ until then. It degrades rather than breaks (the read falls back to system settings), so this is a question rather than a defect: is delivering them at full sync worth a tenant-scoped fetcher?
- **`AdminSettingsEdgeTest` was removed** and replaced by a mock-only unit test of the allow-list. The end-to-end path — save a tenant `general` and see it reach a connected edge, save `mail` and see it not — now has no integration coverage.
- **No tests on the fixes above.** The ack change is the one that regressed silently and is the cheapest to pin down: make `mailService.sendEmail` throw in `SendEmailEdgeTest` and assert the `EdgeImitator` receives `success=false`.

### Notes, no action needed

- `SEND_BASIC` maps to `MailService.sendEmail(TenantId, String, String, String)`, which has no production caller in this repo, `thingsboard-edge`, or `thingsboard-pe`. Harmless, just worth knowing the enum value is currently unreachable.
- Stale FCM tokens still accumulate on the edge — the cloud sees `UNREGISTERED` but the uplink has no return path to prune them. Bounded in practice (one per token rotation), so not worth solving here, but it will not self-correct the way it does on a non-edge deployment.

Proto field numbers (35–37) match the PE definition exactly, and `EdgeMailRequest`/`EdgeSmsRequest`/`EdgeNotificationRequest` dispatch is correct across all ten mail methods.

PR #16148, opened 2026-09-15
IoT Hub: group the hero search popup, and give the search page filters

ShvaykaD:
## Pull Request description

The IoT Hub hero popup groups by content type, and the search page gains a filter panel.

**The popup** fetches one flat page of ten rows by install count and groups it in the component, so a type with real matches gets no section whenever the head types fill the page — and an absent section cannot be told apart from "nothing of this type matched". Measured against the catalogue: `temperature` returns 71 matches across five types, and a flat top-16 shows no calculated field and no alarm rule under either ordering; on `gateway` neither ordering surfaces one of the 12 matching widgets.

It now asks the Hub's sectioned endpoint — one section per type, four rows each, carrying that type's total — and renders the sections it is given, in the order the server sends them. The header links to that type page with the current query, carrying `+N more ›` when more exists; the footer names the page it opens rather than an ordering. A failed request leaves the panel empty for the next keystroke to replace rather than ending the search subscription.

**The search page** keeps its flat paginated grid and gains what it was missing beside the type pages:

- A filter panel — item type, category, use case. Type is new, this being the one page that mixes them. There is no catalogue-wide `filterInfo` endpoint, so the six per-type answers are merged client-side. Counts are deliberately not shown: `filterInfo` takes no text query, so beside a searched result they would be the whole catalogue's numbers describing a different answer.
- A page size that follows the column count, read from a hidden probe carrying the grid's own classes: `max(12, columns × 3)`. A fixed fifteen ends a page mid-row at four and six columns.
- One card layout. Calculated fields, alarm rules and rule chains render as a compact horizontal row that needs a wide cell; the type pages narrow the grid for them, a mixed grid cannot, and those rows truncate their own titles.
- The type pages' frame — one even padding on the card, the component full bleed — replacing four breakpoints' worth of negative margins, each written to cancel that padding for one element.

**Relevance** is a sort option on all four surfaces that carry a search field — search page, type pages, creator profile and the dashboard widget picker — and the default on each. With text it ranks the answer; without it the Hub substitutes the install count, so each opens on the order it opened on before, and nothing switches on the field state.

Needs thingsboard/iot-hub#153 deployed first: the popup reads `GET /api/versions/published/grouped`, and every surface here sends `sortProperty=relevance`.

Checked: `tsc --noEmit` clean over the app sources and eslint clean on the changed files; UI behaviour exercised by hand against a running Hub.

## Related pull requests

- thingsboard/iot-hub#153 — the Hub backend, which adds the endpoints and the sort this PR reads.
- thingsboard/thingsboard.io#661 — the website, which does the same for the hero search popup on `/iot-hub/`.

## What it looks like

- **One word, sectioned.** Devices lead; the types below them are reachable rather than crowded off the page.

08-platform-popup-temperature

- **The same popup, scrolled.** Solution templates, then widgets with `+37 more` — the section a flat top-16 showed nothing of.

08b-platform-popup-scrolled

- **Narrowed to one section.** `indoor temperature card`, exact name first.

09-platform-popup-three-words

- **The search page.** A filter panel it did not have — item type, category, use case — with `Most relevant` as the default, and 8 results rather than a page of near-misses.

10-platform-search-page

- **A type page.** `temperature card` over widgets alone: 24 results, the exact name first.

11-platform-widgets

## General checklist

- [x] You have reviewed the guidelines document.
- [ ] Labels that classify your pull request have been added.
- [ ] The milestone is specified and corresponds to fix version.
- [x] Description references specific issue — TBIOH-33 / TBIOH-32 (internal).
- [x] Description contains brief notes about what needs to be added to the documentation: nothing user-facing to document here; the Hub-side contract is documented in thingsboard/iot-hub#153.
- [x] Changes are backward compatible or upgrade script is provided — UI only. Not backward compatible with a Hub that does not yet carry thingsboard/iot-hub#153: `sortProperty=relevance` is the default on all four surfaces, and the old backend passes it into a native query's `ORDER BY` against a view with no such column, so those requests fail rather than falling back. #153 deploys first.
- [x] Similar PR is opened for PE version to simplify merge. Crosslinks between PRs added — `lts-4.3` counterpart is #16161. PE counterparts: thingsboard/thingsboard-pe#5136 (`lts-4.2`) and thingsboard/thingsboard-pe#5137 (`lts-4.3`). Being confined to `ui-ngx` did not make PE unaffected: the merge conflicted in four files where PE carries its own `ceOnly` handling and colour token, all resolved on those branches.

- [x] Screenshots with affected component(s) are added — above, captured against a Hub running the backend branch.
- [x] If you change the widget or other API, ensure it is backward-compatible or upgrade script is present — no widget or public API is changed; the new calls are reads of new Hub endpoints.
- [x] Ensure new API is documented — no new platform API.

## Back-End feature checklist

Not applicable: this PR is UI only. The endpoints it reads are added in the paired iot-hub PR, where their tests live.


ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.ts
```diff
@@ -96,6 +144,148 @@ export class TbIotHubSearchComponent implements OnInit, OnDestroy {
...
+  /**
+   * There is no catalogue-wide filterInfo endpoint - it answers per item type - so the six
+   * answers are merged here, the way the site merges the same six at build time. Counts are
+   * deliberately not rendered: filterInfo takes no text query, so beside a searched result they
+   * would be the whole catalogue's numbers pretending to describe this answer.
+   */
+  private loadFilterInfo(): void {
+    const config = { ignoreLoading: true, ignoreErrors: true };
+    forkJoin(
+      FACET_ITEM_TYPES.map(type => this.iotHubApiService.getFilterInfo(type, config))
```
vvlladd28: `forkJoin` is all-or-nothing: if any one of the six `getFilterInfo` calls fails, the whole stream errors and `error: () => {}` swallows it, so `typeOptions` is never assigned either — the panel renders an empty "Item type" expansion panel with no checkboxes, and Category/Use case disappear entirely. One type's endpoint returning a 500 takes out the filter feature even though the other five answered.

The `i?.categories` and `i?.useCases` guards just below suggest partial results were the intent, but `forkJoin` can never deliver them — those branches are unreachable as written. A per-request `catchError(() => of(null))` would give you what the guards already expect:

```ts
FACET_ITEM_TYPES.map(type =>
  this.iotHubApiService.getFilterInfo(type, config).pipe(catchError(() => of(null)))
)
```

That also makes `typeOptions` — which needs no server data at all — independent of whether any request succeeded.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.html
```diff
@@ -85,37 +85,54 @@ <h1 class="tb-iot-hub-hero-title">{{ 'iot-hub.home-title' | translate }}</h1>
...
-            <mat-optgroup [label]="getSearchGroupLabel(group.type)">
-              @for (item of group.items; track item.id) {
-                <mat-option [value]="searchText" (click)="openItemDetail(item)">
-                  @if (isCompactType(item.type)) {
-                    <div class="tb-search-popup-icon-square" [style.background]="item.color || '#048ad3'">
-                      <tb-icon class="tb-search-popup-icon-inner">{{ getCompactIcon(item) }}</tb-icon>
+            <mat-optgroup>
+              <span class="tb-search-popup-group-header" role="button" tabindex="0"
+                    (click)="navigateToSection(group.type)"
+                    (keyup.enter)="navigateToSection(group.type)">
```
vvlladd28: This header is mouse-only. `mat-autocomplete` keeps DOM focus in the input and walks its options with an active-descendant key manager, so a `tabindex="0"` span inside the panel is never focused and `(keyup.enter)` cannot fire — Enter while the panel is open selects the active `mat-option` instead. Tab doesn't help either, since it closes the panel.

That matters more than usual here because, per the PR description, this header is deliberately the *only* route from the popup into a type page (no per-section "see all" row). So keyboard and screen-reader users lose that route entirely, and `+N more` with it.

There's also an ARIA-shape issue: the panel is a `listbox` and `mat-optgroup` is its `group`, whose children are expected to be options — a `role="button"` among them is not a shape assistive tech handles predictably. The existing footer has the same pattern, so this isn't new, but it is newly load-bearing. Would making each section header a real `mat-option` that navigates on select work? It would inherit the key manager for free.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-facet-list.component.ts
```diff
@@ -0,0 +1,83 @@
...
+})
+export class TbIotHubFacetListComponent {
+
+  /** Translation key for the facet's heading. */
+  @Input() label: string;
+
+  @Input() options: FilterParamInfo[] = [];
+
+  /** Keys the host currently has selected. Read, never written. */
+  @Input() selected = new Set<string>();
```
vvlladd28, line 31: The host mutates this Set in place (`toggle()` does `set.add`/`set.delete` on the same instance), so the reference the child holds never changes. That works today only because this component runs on default change detection.

The trigger is concrete and likely: the moment anyone adds `ChangeDetectionStrategy.OnPush` here — which `.claude/rules/thingsboard-ui.md` asks for on every component — the checkboxes silently stop reflecting clicks, with no error and nothing in the template to hint at why. Having the host emit a new Set (`this.activeTypes = new Set(next)`) or taking the selection as a plain `string[]` removes the trap and costs nothing at these sizes.

ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-widget-select.component.ts
```diff
@@ -88,7 +88,19 @@ export class DashboardWidgetSelectComponent {
...
 
+  /**
+   * Relevance leads, and is therefore the default (iotHubSelectedSortIndex starts at 0). This
+   * panel has a search field, which is the only thing relevance ranks by; with the field empty
+   * the backend substitutes the install count, so the picker opens on the order it always had.
+   *
+   * scadaFirst is unaffected: the backend prepends it ahead of the caller's sort AND ahead of
+   * relevance, so a SCADA context still lists SCADA widgets first.
+   *
+   * Fourth copy of this list — the others are in TbIotHubSearchComponent, TbIotHubBrowseComponent
```
vvlladd28: This comment is accurate, and that's the problem worth acting on: the list is now maintained in three places in this repo (here, `TbIotHubBrowseComponent:138`, `TbIotHubSearchComponent:98`) behind three structurally identical interfaces (`WidgetSelectSortOption`, and `SortOption` declared separately in the other two — all three are `{ value: string; label: string; direction: Direction }`). This PR is itself the triggering event: adding one option meant editing three arrays, and the next sort key will too. Miss one and that surface quietly keeps a different default, which is exactly the class of bug the relevance-as-default change would hide.

`'relevance'` is also a bare literal here and in the browse component while `iot-hub-home.component.ts` and `iot-hub-search.component.ts` each define their own `const RELEVANCE` — four spellings of one value.

Exporting one `SortOption` interface and one `IOT_HUB_SORT_OPTIONS` (plus `RELEVANCE`) from `@shared/models/iot-hub/iot-hub-item.models.ts` collapses all of it, and lets these 11 lines of comment — and the 9 near-identical lines on `iot-hub-browse.component.ts:129` — go away rather than needing to be kept in step too. The thingsboard.io copy stays a hand-sync regardless; no need to document that at each site.

vvlladd28 (commented):
## Review summary

Reviewed 19 changed files in [IoT Hub: group the hero search popup, and give the search page filters](https://github.com/thingsboard/thingsboard/pull/16148). Left 12 comments inline.

The structural work is sound — the sectioned endpoint, the CSS-variable padding frame replacing four breakpoints of negative margins, and the column-probe page size all hold up (the probe reads a viewport-driven column count, so it stays correct despite sitting outside the results column, and `lt-lg`/`gt-md` tile without a gap). The most important findings are in `iot-hub-search.component.ts:204`, where `forkJoin` over six `getFilterInfo` calls makes one failing request empty the whole filter panel — the `i?.categories` guards below it can never fire — and in `iot-hub-home.component.html:91`, where the new section header is reachable only by mouse inside the autocomplete panel.

On the specific things you asked about — redundancy, comment length, conventions — details below, plus a batch of small leftovers inline.

## Additional findings

**Comment length.** 212 of the ~1,000 added source lines are comments (21%, excluding nothing but the 9 SPDX headers). A good share of them explain *why the code is not something else* rather than what it does, which is the kind that ages worst and cannot be checked from this repo: the 11-line block on `dashboard-widget-select.component.ts:91` and its 9-line twin on `iot-hub-browse.component.ts:129`, the 6-line note on `MpItemVersionGroupedQuery` about the `PageLink` it does not have, the 4-line note on `getPublishedVersionsGrouped` about why it is its own method, and several that restate the line under them (`/** total - items.length, floored at 0. */`). Worth keeping: the ones stating a contract a reader cannot infer — the empty-sets note in `fetchResults`, the leading-`&` note in `filtersToQuery`, and the probe rationale. The rest would survive being cut to one line, and three of them stop being needed at all if the duplicated lists they apologise for are merged.

**Component conventions.** `TbIotHubFacetListComponent` is written with `@Input()`/`@Output()` decorators, default change detection, getters for derived state and `[(ngModel)]`, while `.claude/rules/thingsboard-ui.md` asks for `OnPush`, signals, `input()`/`output()` and reactive forms. Worth weighing against the counter-evidence before acting: none of the 17 existing `iot-hub` components use `OnPush` or signals, and `ngModel` appears in 5 of them, so the new file matches its neighbours exactly — adopting the documented style here would make it the one file that differs. If the decision is to leave it as is, the `selected` Set comment below still applies, since that trap only exists while the component is on default CD. A side effect of the getters: `visibleOptions` re-filters on every change detection pass and `scrollable` calls it a second time, so each facet does two full passes per cycle — invisible at these list lengths, and what `computed()` would remove for free.

**Encoding consistency in `filtersToQuery`.** `creatorId` and `type` are the only two of thirteen fields interpolated without `encodeURIComponent`. In practice both carry constrained values — a UUID and an enum name — and the lines moved over verbatim, so this is a consistency nit inside a new helper rather than a defect: worth the one call the eleven fields around them already make, nothing more. Note for whoever touches it that the UUID shape is a convention, not an enforced one — `iot-hub-browse.component.ts:192` now reads `creatorId` straight off the query string with no validator between the URL and the request.

**One thing the description promises that I could not verify here:** `sortProperty=relevance` is sent by all four surfaces as stated, but it is only meaningful once thingsboard/iot-hub#153 is deployed. Against a Hub without it, the four surfaces default to a sort key the backend does not know — worth confirming the backend ignores an unknown `sortProperty` rather than rejecting the request, since the popup's `catchError` would turn that into a permanently empty panel.

---

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.ts
```diff
@@ -1,34 +1,45 @@
...
+/** Sort property served by relevance ranking. With no text the backend substitutes the
+ *  install count, so it is a safe default in both states. */
+const RELEVANCE = 'relevance';
+
+/**
+ * Item types the Type facet offers, in the order every other cross-type surface uses (the hero
+ * popup's sections, the site's). Fixed rather than sorted by count: the counts move with the
+ * catalogue and a facet whose options reshuffle under the pointer is hard to use.
+ */
+const FACET_ITEM_TYPES: ItemType[] = [
```
vvlladd28: This is byte-identical to `SEARCH_GROUP_TYPES` in `iot-hub-home.component.ts:51` — same six values, same order — and this PR touches both (renaming one from `TYPE_ORDER`, the other from `SEARCH_GROUP_ORDER`). Two hand-maintained copies of the catalogue's type list with no link between them.

The trigger: a seventh item type ships. Update the facet list and not the popup's and the type becomes filterable on the search page but its section is silently dropped from the popup by `toResultGroups`' own filter; update the popup and not this one and it never appears in the Type facet. Neither fails loudly. One exported const next to `itemTypeTranslations` in `iot-hub-item.models.ts` would serve both — the two comments explaining each copy's ordering rationale could then be one.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.ts
```diff
@@ -159,33 +349,9 @@ export class TbIotHubSearchComponent implements OnInit, OnDestroy {
   }
 
   // Type helpers
```
vvlladd28: Leftover from the removed grouping code: this section header now has four blank lines and no members under it — `isCompactType`, `getTypeLabel(type)`, `getTypeRoute` and `navigateToType` all went, and the replacement `getTypeLabel` arrow lives up in the Filters section. Worth deleting the heading and the gap.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.ts
```diff
@@ -37,32 +48,64 @@ const TYPE_ORDER: ItemType[] = [
...
+  @ViewChild('cardGridProbe', { static: true }) cardGridProbe!: ElementRef<HTMLElement>;
+  cols = 5;
+  pageSize = MIN_PAGE_SIZE;
+
+  get pageSizeOptions(): number[] {
+    const base = Math.max(MIN_PAGE_SIZE, this.cols * ROWS_PER_PAGE);
+    return [base, base * 2, base * 4];
+  }
+
+  /** Filter panel state. Empty sets mean "no filter", which is what the query object expects. */
```
vvlladd28: Two `/** */` blocks stacked on one member. TypeScript and IDE tooling only attach the second, so this first one — which describes the group of fields below rather than `filtersOpen` — is silently dropped everywhere it would be read. A plain `//` comment above the group, or folding both into one block, keeps it.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.html
```diff
@@ -85,37 +85,54 @@ <h1 class="tb-iot-hub-hero-title">{{ 'iot-hub.home-title' | translate }}</h1>
...
+                <span class="tb-search-popup-group-label">{{ getSearchGroupLabel(group.type) }}</span>
+                @if (group.remaining > 0) {
+                  <span class="tb-search-popup-group-more">
+                    {{ 'iot-hub.section-more' | translate:{ count: group.remaining } }}
+                    <mat-icon>chevron_right</mat-icon>
+                  </span>
+                }
+              </span>
+              <ng-container>
```
vvlladd28: This `<ng-container>` carries no directive and no structural binding, so it wraps the `@for` in nothing. Removing it and de-indenting takes a level out of an already deeply nested block.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-facet-list.component.html
```diff
@@ -0,0 +1,31 @@
...
+      @if (search) {
+        <button mat-icon-button class="tb-iot-hub-facet-search-clear" (click)="search = ''">
+          <mat-icon>close</mat-icon>
+        </button>
+      }
+    </div>
+  }
+
+  <div class="tb-iot-hub-facet-items" [class.scrollable]="scrollable">
+    @for (option of visibleOptions; track trackByKey($index, option)) {
```
vvlladd28: `@for`'s `track` takes an expression, not a `TrackByFunction` — `track option.key` does exactly this, and lets the `trackByKey` method at `iot-hub-facet-list.component.ts:80` go. The method signature (leading `_index`) is the `*ngFor` idiom, which the new control flow left behind.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.ts
```diff
@@ -35,13 +40,24 @@ interface HeroTypeConfig {
 interface SearchResultGroup {
   type: ItemType;
   items: MpItemVersionView[];
+  /** Rows of this type behind the answer, from the section's total. */
```
vvlladd28: `total` is written by `toResultGroups` but never read — the template uses only `group.remaining`, and `remaining` is already derived from the section's total at mapping time. Dropping the field (and this comment) leaves `SearchResultGroup` saying exactly what the popup uses.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.scss
```diff
@@ -143,69 +151,149 @@
...
+.tb-search-facets-clear {
+  align-self: flex-start;
+  margin-left: calc(var(--tb-facet-indent, 16px) - 16px);
+}
+
 // Responsive
 
 @media #{$mat-gt-xxl} {
   .tb-search-card-grid {
     grid-template-columns: repeat(6, 1fr);
```
vvlladd28, line 256: Removing the `-compact` rule left a stray blank line before the closing brace here, and the same in the three media queries below (`lt-xxl`, `lt-lg`, `lt-sm` — each now a single declaration followed by an empty line). Worth tidying in the same pass.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.html
```diff
@@ -76,7 +88,59 @@
...
+          <span>{{ 'iot-hub.filter' | translate }}</span>
+          <button mat-icon-button (click)="filtersOpen = false">
+            <mat-icon>close</mat-icon>
+          </button>
+        </div>
+        <ng-container *ngTemplateOutlet="facetPanel"></ng-container>
+      </aside>
+    }
 
   <div class="tb-results-container">
```
vvlladd28: This block and its closing `</div>` kept their old indentation after being wrapped in `.tb-search-layout`, so it reads as a sibling of the new wrapper rather than a child of it — and the two closing tags at the end now line up at the same depth while closing different elements. In a template this long that's the kind of thing the next editor closes the wrong tag over.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.ts
ShvaykaD: Fixed in e5a4a7e489. Each request now carries `catchError(() => of(null))`, so a failing type costs that type's categories and use cases and nothing else, and `typeOptions` is assigned regardless since it needs no server data.

The `error: () => {}` branch went with it: once every inner request is caught the outer stream can no longer error, so keeping it would have left a second unreachable branch next to the guards you flagged.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.html
ShvaykaD: Not changed here, deliberately. The footer immediately below carries the same `role="button" tabindex="0"` shape with the same problem, and it predates this PR — patching only the header would fix half of one pattern and leave the other half broken and out of sight.

Raising it as its own ticket covering the popup as a whole: both headers and footer, and the `role="button"` inside a `listbox` shape, which is the part that wants a decision rather than a patch. Your point that this PR makes the header load-bearing is what moves it from cosmetic to scheduled, so it goes in the ticket.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-facet-list.component.ts
ShvaykaD: Fixed in e5a4a7e489, taking the first of your two options: the host returns a new Set per toggle (`toggled()`), and `clearFilters` assigns fresh Sets rather than calling `.clear()`.

Kept the child's input as a `Set` rather than moving to `string[]` — the host would then have to spread in the template, allocating a new array on every change detection pass. That trades a latent trap for constant churn.

ui-ngx/src/app/modules/home/components/dashboard-page/dashboard-widget-select.component.ts
ShvaykaD: Fixed in e5a4a7e489. `SortOption`, `IOT_HUB_SORT_OPTIONS` and `RELEVANCE_SORT_PROPERTY` now live in `@shared/models/iot-hub/iot-hub-item.models.ts`, and all three components read the one array.

All three local interfaces were file-local, unexported, with exactly one use each and no mention in any template, so nothing outside had to move. The comments you flagged are gone with the duplication rather than rewritten.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.ts
ShvaykaD: Fixed in e5a4a7e489 as `CROSS_TYPE_ITEM_TYPES` — but not as one constant, and that part is worth flagging, because "one exported const would serve both" would have changed the UI.

`iot-hub-item.models.ts` already held a third copy of the same six types: `CREATOR_VISIBLE_ITEM_TYPES`, in a different order (WIDGET first), driving the type tabs. `SEARCH_GROUP_TYPES` is pure membership — `toResultGroups` only calls `.includes()` and the section order comes from the server — but `FACET_ITEM_TYPES` is display order with DEVICE first. Folding all three together would have silently reordered the Type facet.

So there are two constants with distinct meanings, and each says on itself why the other exists.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.ts
ShvaykaD: Fixed in e5a4a7e489.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.ts
ShvaykaD: Fixed in e5a4a7e489 — the group note is a plain `//` above the block and `filtersOpen` keeps its own `/** */`.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.html
ShvaykaD: Fixed in 2d72735070.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-facet-list.component.html
ShvaykaD: Fixed in e5a4a7e489 — `track option.key`, and `trackByKey` is gone with it.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.ts
ShvaykaD: Fixed in e5a4a7e489.

Worth recording why this diverges from the sibling site PR, which kept its equivalent field: there the field gained a reader, because the hero popup's analytics event now reports the real match count as the sum of section totals rather than the cards on screen. The platform has no analytics layer at all — no `iot_hub_query`, no equivalent anywhere in `ui-ngx` — so here there is no second reader and the field is simply unread. The difference between the two PRs is deliberate, not an oversight.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.scss
ShvaykaD: Fixed in 2d72735070 — four of them, the one here and the three media queries below.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.html
ShvaykaD: Fixed in 2d72735070.

ShvaykaD:
Review addressed in e5a4a7e489, 2d72735070 and edf81221ad. Eleven of the twelve inline findings are fixed and answered in their threads; the accessibility one is answered there too, as a ticket rather than a patch. `tsc --noEmit` is clean over the app sources and eslint is clean on every changed file.

**Component conventions.** `.claude/rules/thingsboard-ui.md` does not exist — not in the working tree, not in `git log --all`, not anywhere on disk. Since the rule can't be read, I'm treating the counter-evidence you gathered as the whole picture: none of the 17 existing `iot-hub` components use `OnPush` or signals, `ngModel` appears in 5, and the new file matches its neighbours exactly. Leaving it as is. The `selected` Set trap is fixed regardless — it was a real trap whether or not `OnPush` ever arrives.

**Comment length.** Most of the 212 lines were the ones apologising for the duplication, and they went with it: the 11-line block on the widget picker, its 9-line twin on browse, both type-list rationales, and the `/** total - items.length, floored at 0. */` docstring. Also removed "each holding at most four rows" from `getPublishedVersionsGrouped` — that is a server-side cap this code deliberately does not know, so it would have quietly become false the day the cap moved to three. Kept the ones you named as earning their place, plus the `MpItemVersionGroupedQuery` note: it is what stops the next person adding a `PageLink` the endpoint does not read.

**Encoding.** Changed in edf81221ad, and it turned out to be more than a consistency nit. `creatorId` is not an internal value: the creator profile takes it from the route path and binds it straight through to the query builder with no validator in between, so a crafted path segment could close the parameter and append its own. `type` encoded in the same pass. (Edited: this originally also cited `iot-hub-browse.component.ts:178` as a second reader. That read turned out to be unreachable and is removed in dab4ec72f9 — see the note below.)

**`sortProperty=relevance` against a Hub without #153.** Checked, and it is worse than "ignores it". `JpaMpItemVersionViewDao.findByQueryV2` on master hands the page link's sort to a native query, so Spring appends `ORDER BY relevance` against a view with no such column and the request fails outright. That is all four surfaces, not only the popup, since relevance is the default on each. The deploy order is understood and the risk accepted, but the PR description understates it as "degrades to an empty popup" and I'll correct that line.

**On the default itself**, since it borders on your AC question: Most Relevant being the default in both states does contradict TBIOH-33 SCOPE-IN 2, SCOPE-IN 4 and AC-4 as written. That is deliberate and agreed with the product owner, recorded on TBIOH-33 as comment 296473 together with a request to rewrite those three points so QA does not test superseded text. The hero popup was the one surface still switching on the field state; it now sends relevance unconditionally like the rest. Against the backend that is a no-op — `groupedSortKey` maps both spellings to the install count with no text, and `demoteBuiltIn` is true either way — but it makes the code say what the decision says.

**One thing outside your list**, found in the same pass and worth naming since it is the class of bug the branch already has a commit against: `toResultGroups` dereferenced `section.items` on an unvalidated cast of a new endpoint, inside a `subscribe` callback where the upstream `catchError` cannot reach it. A section arriving without `items` would have left the panel on its spinner rather than recovering on the next keystroke. Defaulted in e5a4a7e489.

ShvaykaD:
Two more commits after the ones above.

**dab4ec72f9 — a query parameter nothing produces.** The type page adopted `?creatorId=` from the URL whenever its input was unbound. Cross-checking readers against producers across `iot-hub`: the only navigations carrying query parameters send `search` and `itemType`, and the creator profile scopes its list by binding the component's input rather than by navigating. Nothing builds a URL with that parameter, so the branch was reachable only by typing it in by hand — where it rescoped a Hub-wide type page with nothing on screen to say why the counts had changed. Both live callers, the creator profile and the add-item dialog, bind the input that the removed code deferred to anyway.

That check was worth running over the whole branch rather than just this one spot, since the surfaces here changed shape more than once. The rest came back clean: the four new translation keys are all consumed, every new CSS class is used by a template, no declared member across the seven changed TypeScript files is unreferenced, and every `@Input`/`@Output` is bound. Five unused fields in `iot-hub-version.models.ts` predate the branch and are left alone.

**6aae7899bf — citations a reader here cannot follow.** Two comments I added explaining the sort default pointed at an internal ticket, one of them at three numbered sections of it that are being rewritten precisely because the code no longer matches them. The reasoning is now stated inline, including the cost that was previously only implied: with an empty field the menu reads "Most relevant" over an install-ordered list.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.html
ShvaykaD: Fixed in c5a3bea6df, all of it, rather than as a follow-up ticket.

**The header and the footer are options now**, not spans carrying `role="button"`. They take part in arrow navigation, they read as part of the list, and the `mat-optgroup` wrapper goes with them.

That wrapper turned out to be doing more harm than the ARIA-shape note suggested. Its template projects plain content into the label element, which carries `role="presentation"` and is the target of the group's own `aria-labelledby` — so the section's accessible name was being computed from the whole header, `"Devices +12 more"`, around a control nobody could operate. On `lts-4.2` that name was just `"Devices"`. Dropping the wrapper removes the distortion; the header option carries the name in a role the listbox actually admits.

**Every row acts on `(onSelectionChange)` guarded by `isUserInput`** instead of `(click)`. This was the bigger find and it predates the PR: Enter on a highlighted result calls `_selectViaInteraction()`, which is not a DOM click, so `(click)` never fired and the card never opened from the keyboard — the panel just closed. One path now serves pointer and keyboard both.

**`onSearch()` gained a guard**, which the above made necessary: the trigger acts on `keydown` and that handler runs on `keyup`, which the trigger's `preventDefault` does not stop. Without it one Enter would open a result and then navigate away from it.

For the record, the website popup has none of these problems, and the reason is worth keeping in mind for this component: there the section header and every card are ordinary `<a href>`, so focus, Enter and the accessible role come for free, and the panel closes only on Escape or an outside click rather than on blur.

`ng build` and eslint clean. Two things worth a manual pass, since this is keyboard behaviour I reasoned out rather than observed: arrow down through a full panel — headers, rows, footer, in that order — and Enter on each of the three kinds.

ShvaykaD: Correcting my earlier reply on this thread: it is fixed here after all, in c5a3bea6, rather than deferred to its own ticket.

Both halves of your finding are addressed, and the footer went with it — which was the reason I gave for deferring, so leaving it would have been fixing half a pattern:

- The group header and the footer were `role="button" tabindex="0"` elements sitting inside the listbox. Both are `mat-option` now, so they join the panel's active-descendant key manager and the keyboard reaches every row of the popup, the header included. That restores the only route from the popup into a type page, which this PR had made load-bearing.
- The ARIA shape issue goes with it: there is no `role="button"` among the options any more, so assistive tech sees a listbox of options rather than a button mixed into a group.
- Activation is one handler for both input methods — `(onSelectionChange)="$event.isUserInput && ..."` fires for a click and for Enter on the highlighted option, instead of a separate `(keyup.enter)` that could never fire.

One guard came with it: Enter in the field opens the search page, but the trigger has already acted on a highlighted option during keydown, and `keyup` still arrives afterwards. Without checking `searchAutoTrigger.activeOption` first, a single Enter would open a result and then navigate away from it.

The `role="button"` elements left in this file are the hero keyword chips and the category cards, which are outside the popup and unrelated to this finding.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.ts
```diff
@@ -227,7 +239,16 @@ export class TbIotHubHomeComponent implements OnInit, OnDestroy {
...
   }
 
+  /**
+   * Enter in the field opens the search page - unless the panel has an option highlighted, which
+   * the trigger has already acted on during keydown. This runs on keyup, which preventDefault on
+   * the keydown does not stop, so without the guard one Enter would both open a result and
+   * navigate away from it.
+   */
   onSearch(): void {
+    if (this.searchAutoTrigger?.activeOption) {
```
vvlladd28: This guard reads `activeOption` on `keyup`, but the trigger has already cleared it during `keydown` — so it never fires, and the double navigation the comment above describes is still live.

`MatAutocompleteTrigger._handleKeydown` in `@angular/material@20.2.14` (the version in `ui-ngx/package.json`) does:

```js
if (this.activeOption && keyCode === ENTER && this.panelOpen && !hasModifier) {
  this.activeOption._selectViaInteraction();
  this._resetActiveItem();   // <-- 
  event.preventDefault();
}
```

and `_resetActiveItem()` falls to `setActiveItem(-1)` because `<mat-autocomplete #searchAuto>` does not set `autoActiveFirstOption`. `ListKeyManager` then resolves index `-1` to `null`, so by the time `(keyup.enter)` reaches `onSearch()`, `activeOption` is `null`, the guard falls through and `seeAllResults()` runs.

Concretely: arrow down to a result, press Enter — `openItemDetail(item)` navigates to the item, then `seeAllResults()` navigates to `/iot-hub/search` and supersedes it. Same for a highlighted section header and `navigateToSection`. Mouse clicks are unaffected (no `keyup` on the input), which is likely why it survived the manual pass.

The fragile part is reading state the trigger owns and resets inside the same keystroke. A one-shot flag set by the selection handlers themselves is order-independent:

```ts
private selectionHandled = false;

onOptionSelected(): void { this.selectionHandled = true; }

onSearch(): void {
  if (this.selectionHandled) {
    this.selectionHandled = false;
    return;
  }
  this.seeAllResults();
}
```

This is exactly the arrow-down-then-Enter path you called out as wanting a browser pass — worth running it against whichever shape you land on.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.scss
```diff
@@ -220,33 +220,55 @@
...
     0 178px 50px 0 rgba(0, 0, 0, 0);
 
-  // Section group — 12px gap between groups
-  .mat-mdc-optgroup + .mat-mdc-optgroup .mat-mdc-optgroup-label {
+  // Section header — an option like every other row, not a button beside them, so the arrow keys
+  // reach it and a screen reader reads it in the list. It IS the way into the type page: title
+  // left, how much is not shown plus a chevron right. There is deliberately no per-section
+  // "see all" row under the items; six of those on one panel is what this variant avoids.
+  // Design: 14px Medium, 0.25 tracking, black-tertiary, 4px V / 16px H, 12px above each section.
+  .mat-mdc-option.tb-search-popup-group-header {
```
vvlladd28, line 249: `padding: 4px 16px` and `min-height: 28px` here never take effect — the item-row rule at line 270 overrides both.

Both selectors resolve to the same specificity: `.tb-search-popup-panel.mat-mdc-autocomplete-panel .mat-mdc-option.tb-search-popup-group-header` and `… .mat-mdc-option:not(.mat-mdc-option-multiple)` are each (0,4,0), since `:not()` contributes the specificity of its argument. On a tie, source order decides, and the item-row rule is declared later — so the header actually ships with `padding: 8px 16px`, `min-height: auto`, `height: auto`.

This is new with the optgroup removal rather than pre-existing: the item-row selector used to be `.mat-mdc-optgroup .mat-mdc-option:not(…)`, which could not match a `<span>` header. Now the header *is* an option, so it matches. The visible cost is ~8px of extra height on every section header, against the `4px V` the comment right above documents — up to six of them inside a 424px panel, which is the vertical space the sectioned layout exists to buy.

Either move this block below the item-row rule, or exclude the header from it:

```scss
.mat-mdc-option:not(.mat-mdc-option-multiple):not(.tb-search-popup-group-header) { … }
```

Worth noting the footer has the same collision and only wins because `.mat-mdc-option.tb-search-popup-footer` happens to be declared after the item-row rule — so it works today, but by source order rather than by intent.

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.scss
```diff
@@ -143,69 +151,146 @@
...
+  // each facet — padding here would inset the rules too — and matches the page's own margin, so
+  // the words line up with the heading above them.
+  --tb-facet-indent: var(--tb-search-side-pad, 16px);
+
+  align-self: stretch;
+  height: 100%;
+  overflow-y: auto;
+  padding-top: 20px;
+
+  &::-webkit-scrollbar {
```
vvlladd28: This 26-line scrollbar block is byte-identical to the one this PR also adds at `iot-hub-facet-list.component.scss:48`, and to two that already exist — `iot-hub-browse.component.scss:232` and `dashboard-widget-select.component.scss:543`. That makes four hand-maintained copies of one treatment, two of them added here.

The drift is already in the tree rather than hypothetical: `iot-hub-browse.component.scss:69` and `dashboard-widget-select.component.scss:474` carry an *inverted* variant of the same idea (visible by default, hidden while `.filter-items-hovered`), and `theme.scss:206` sets a third, app-wide, with a permanently visible `rgba(203, 203, 203, 0.8)` thumb. Changing the facet scrollbar now means finding four places and working out which of three treatments each one meant. A `@mixin tb-hover-scrollbar` next to the global rule in `theme.scss` collapses them; even sharing it between just the two files this PR owns would stop it growing.

Separately, on overriding the platform scrollbar at all — worth one decision for the whole IoT Hub filter UI rather than a change confined to this PR. The thumb is `transparent` until the container is hovered, so on a touch device, and for anyone driving the page from the keyboard, a scrollable facet rail gives no indication that it scrolls; `theme.scss` keeps its thumb visible for exactly that reason. `::-webkit-scrollbar` is also WebKit/Blink-only, so Firefox users get none of this styling in any of the four copies — `scrollbar-width` / `scrollbar-color` are the standardised pair, and this repo already pairs them in `mobile-app-configuration-dialog.component.scss:34` and `device-check-connectivity-dialog.component.scss:116`. Since the hover-only treatment predates this PR in these panels, the question is whether it should keep spreading, not whether this diff introduced it.

vvlladd28 (commented):
## Re-review summary

Re-reviewed [IoT Hub: group the hero search popup, and give the search page filters](https://github.com/thingsboard/thingsboard/pull/16148) — verified 12 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/16148#pullrequestreview-5219934131).

| Status | Count |
|--------|-------|
| ✅ Resolved | 12 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 0 |

All twelve previous findings are addressed in code. Also found 3 new issues in the fix commits, commented inline — two of them in `c5a3bea6`, the commit that converted the popup's header and footer to `mat-option`, which is the most intricate of the fixes and the one you flagged as reasoned-out rather than observed.

### Finding details

- ✅ **iot-hub-search.component.ts** — `forkJoin` over six `getFilterInfo` calls was all-or-nothing, making the `i?.categories` guards unreachable — *Fixed in code: per-request `catchError(() => of(null))`, and the now-unreachable `error: () => {}` removed with it.*
- ✅ **iot-hub-home.component.html** — the section header was mouse-only inside the autocomplete panel, and `role="button"` sat inside a `listbox` — *Fixed in code: header and footer are `mat-option`s driven by `(onSelectionChange)` + `isUserInput`, the `mat-optgroup` wrapper is gone, and one handler now serves pointer and keyboard. See the two new findings below — the `onSearch()` guard that came with it does not fire, and the header's own padding is overridden.*
- ✅ **iot-hub-facet-list.component.ts** — the `selected` Set was mutated in place, a trap for the first person to put the component on OnPush — *Fixed in code: `toggled()` returns a new Set and `clearFilters` assigns fresh ones.*
- ✅ **dashboard-widget-select.component.ts** — the sort option list was maintained in three places behind three identical interfaces — *Fixed in code: one `SortOption` / `IOT_HUB_SORT_OPTIONS` / `RELEVANCE_SORT_PROPERTY` in `iot-hub-item.models.ts`, read by all three.*
- ✅ **iot-hub-search.component.ts** — `FACET_ITEM_TYPES` was byte-identical to `SEARCH_GROUP_TYPES` — *Fixed in code as `CROSS_TYPE_ITEM_TYPES`. Keeping it separate from `CREATOR_VISIBLE_ITEM_TYPES` is the right call: folding them would have reordered the Type facet, and each constant now says why the other exists.*
- ✅ **iot-hub-search.component.ts** — leftover `// Type helpers` heading with no members — *Fixed in code.*
- ✅ **iot-hub-search.component.ts** — two `/** */` blocks stacked on one member, the first silently dropped by tooling — *Fixed in code.*
- ✅ **iot-hub-home.component.html** — `<ng-container>` wrapping the `@for` in nothing — *Fixed in code, removed with the optgroup restructure.*
- ✅ **iot-hub-facet-list.component.html** — `track trackByKey($index, option)` where `track option.key` does the same — *Fixed in code, and `trackByKey` is gone.*
- ✅ **iot-hub-home.component.ts** — `SearchResultGroup.total` written but never read — *Fixed in code. The note on why this diverges from the sibling site PR, which does have a second reader, is the right thing to have recorded.*
- ✅ **iot-hub-search.component.scss** — stray blank lines left by the removed `-compact` rule and in three media queries — *Fixed in code, all four.*
- ✅ **iot-hub-search.component.html** — the results block kept its old indentation after being wrapped in `.tb-search-layout` — *Fixed in code.*

---

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.ts
ShvaykaD: Confirmed and fixed in 14e3025a24. `_handleKeydown` calls `_selectViaInteraction()` and then `_resetActiveItem()`, which with no `autoActiveFirstOption` on `#searchAuto` takes the `setActiveItem(-1)` branch, so `activeOption` reads null by the time keyup arrives.

Worth stating how bad it was: this is worse than the behaviour it replaced. Before the option conversion `(click)` never ran on Enter, so a highlighted row did one thing — navigate to the search page. With the dead guard it opened the row's detail dialog and then navigated away from it.

Took the one-shot flag, through a gate the template already had a slot for — `$event.isUserInput && …` became `rowSelectedByUser($event) && …`, so the row that acts is the thing that records it. One wrinkle your sketch does not cover: a pointer selection sets the flag too and is not followed by a keyup, so the flag would sit there and swallow the next Enter in the field. `onSearchInput` clears it.

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.scss
ShvaykaD: Confirmed and fixed in 14e3025a24. Both selectors weigh (0,4,0) under `::ng-deep .tb-search-popup-panel.mat-mdc-autocomplete-panel`, and the catch-all is declared later, so the header shipped at `padding: 8px 16px; min-height: auto`.

Took neither of the two options as written. Reordering would have left the next edit to re-break it, and adding `:not(.tb-search-popup-group-header)` would have grown a chain each time a row kind is added. Instead the parts every row shares — the flex line, the highlight, the icon reset — are one `.mat-mdc-option` rule, and each kind names itself: `.tb-search-popup-group-header`, `.tb-search-popup-item` (new class on the item options), `.tb-search-popup-footer`. No tie left to lose, and the footer stops winning by accident.

`.tb-search-popup-no-results` was taking its horizontal padding from that catch-all, so it got its own.

Verified in the emitted CSS rather than by reading the cascade again:

```
.tb-search-popup-panel.mat-mdc-autocomplete-panel .mat-mdc-option.tb-search-popup-group-header { padding: 4px 16px; min-height: 28px; margin-top: 12px; }
.tb-search-popup-panel.mat-mdc-autocomplete-panel .mat-mdc-option.tb-search-popup-item { padding: 8px 16px; min-height: auto; height: auto; }
```

ui-ngx/src/app/modules/home/components/iot-hub/iot-hub-search.component.scss
ShvaykaD: Split this one. The duplication is fixed in 14e3025a24; the treatment is not, deliberately.

The two copies this branch adds are byte-identical — same hash once indentation is normalised — and both files already import `_iot-hub-common.scss`, which exists for exactly this kind of shared treatment. They now call `tb-iot-hub-hover-scrollbar` from there.

What the mixin does not do is settle the behaviour, and its comment says so: whether these panels should keep a thumb that stays invisible until hover, and whether to add the standardised `scrollbar-width` / `scrollbar-color` that Firefox reads, is one decision for the IoT Hub filter UI rather than something to land inside a PR about search relevance. `iot-hub-browse` and `dashboard-widget-select` carry the same treatment and predate this branch, and `theme.scss` deliberately keeps its own thumb visible — so changing it here would leave three behaviours instead of two. Your point about touch and keyboard users is the substance of that decision and belongs in the ticket, not in this diff.

Extending the mixin to those two files would be a mechanical follow-up once the behaviour is decided, and it is cheaper to do once than twice.

vvlladd28 (commented):
## Re-review summary

Re-reviewed [IoT Hub: group the hero search popup, and give the search page filters](https://github.com/thingsboard/thingsboard/pull/16148) — verified 3 finding(s) from [previous review](https://github.com/thingsboard/thingsboard/pull/16148#pullrequestreview-5233629429).

| Status | Count |
|--------|-------|
| ✅ Resolved | 3 |
| 💬 Acknowledged | 0 |
| ❌ Unresolved | 0 |

Also found 1 new issue in the fix commits, commented inline: the one-shot flag that replaced the dead `activeOption` guard is not scoped to the keystroke, so a pointer selection leaves it set.

### Finding details

- ✅ **iot-hub-home.component.ts:`onSearch`** — the `activeOption` guard read state the trigger had already cleared during the same keydown, so it never fired and one Enter both opened a result and navigated away from it — *Fixed in code: `rowSelectedByUser($event)` gates every row's `onSelectionChange` and records the selection itself, so the check no longer depends on the order the two handlers run in. Verified: the Enter path is now correct — keydown selects the row, sets the flag, and `onSearch()` on keyup consumes it. See the inline comment for the pointer path, which the same flag leaves stranded.*
- ✅ **iot-hub-home.component.scss:`.tb-search-popup-group-header`** — the header's `padding: 4px 16px` / `min-height: 28px` lost a specificity tie to the later `.mat-mdc-option:not(.mat-mdc-option-multiple)` catch-all — *Fixed in code, and by the better of the three options: what every row shares is one unclassed `.mat-mdc-option` rule that sets no box at all, and each kind carries its own class — `.tb-search-popup-group-header`, `.tb-search-popup-item` (new), `.tb-search-popup-footer`, plus explicit padding on `.tb-search-popup-no-results`, which had been taking it from the old catch-all. Checked every option kind in the template against the new rules: no tie is left, and the footer no longer wins by source order.*
- ✅ **iot-hub-search.component.scss:`.tb-search-facets-desktop`** — a 26-line scrollbar block byte-identical to the one added in `iot-hub-facet-list.component.scss`, and to two already in the tree — *Fixed in code for the two copies this branch owns: `tb-iot-hub-hover-scrollbar` in `_iot-hub-common.scss`, which both files already `@import` (legacy `@import`, so the unprefixed `@include` resolves). The second half — whether the hover-only treatment should survive at all, and whether to add the standardised `scrollbar-width` / `scrollbar-color` — is deliberately deferred to a decision covering `iot-hub-browse`, `dashboard-widget-select` and `theme.scss` together, which is the right scope for it; the mixin's own comment records that, so the question is findable from the code.*

---

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.ts
```diff
@@ -227,7 +242,34 @@ export class TbIotHubHomeComponent implements OnInit, OnDestroy {
...
+   * handler on keyup, and preventDefault on the first does not stop the second. The trigger's
+   * activeOption cannot tell them apart either: it clears it inside that same keydown, so by
+   * keyup it reads empty whether or not a row acted. The row that acted says so instead, which
+   * does not depend on the order the two handlers run in.
+   */
+  rowSelectedByUser(event: MatOptionSelectionChange): boolean {
+    if (!event.isUserInput) {
+      return false;
+    }
+    this.enterHandledByRow = true;
```
vvlladd28: The flag is set for a pointer selection too, and outside `onSearch()` the only thing that clears it is `onSearchInput()` — which needs a keystroke in the field. Your reply anticipated the pointer case, but the mitigation only reaches it when the user types next, and one of the three row actions doesn't navigate away: `openItemDetail` opens a dialog, so the component is still alive when it closes, with the flag still set and nothing queued to consume it.

Concretely: click any result row in the popup, close the detail dialog, then click the magnifier button — `onSearch()` takes the early return and the search page never opens. The second click works. The magnifier is the airtight repro because it calls `onSearch()` regardless of focus; pressing Enter in the field without editing the text is the same thing if focus lands back in the input, which is what `MatDialog`'s default `restoreFocus` does.

`onSearchFocus()` looks like the place to clear it as well, and it discriminates the two paths for free: a pointer selection ends in `_setValueAndClose`, which calls `focus()` on the input, and the dialog blurs and restores it on top of that — so the flag is dropped on whichever of those fires first. On Enter the input never loses focus, so `focus()` on an already-focused element dispatches no event and the flag survives to `keyup` as intended. `clearSearch()` is the other entry point that resets the query without going through `onSearchInput()`.

vvlladd28 (dismissed):

ui-ngx/src/app/modules/home/pages/iot-hub/iot-hub-home.component.ts
ShvaykaD: Confirmed and fixed in c410558dca. The magnifier repro holds: line 63 binds `(click)="$event.stopPropagation(); onSearch()"`, so it reaches `onSearch()` with no keystroke anywhere in the path, and `_setValueAndClose` calls `focus()` on the input at `autocomplete.mjs:860`, which is what puts Enter in the same position once the dialog restores focus.

Took the `onSearchFocus()` clear, and `clearSearch()` with it. But not on its own — on the pointer path it depends on the focus event arriving *after* `rowSelectedByUser` has set the flag, and those are two subscribers to the same emitter, so the ordering is not something the code states.

So the flag is armed from `activeOption` instead, which is exactly the signal that distinguishes them and is readable only at that instant: `_handleKeydown` holds it across `_selectViaInteraction()` and clears it on the very next line, and a pointer never sets it at all. The pointer path now never arms the flag, whichever order the two subscribers run in.

The two together leave nothing: arming covers the ordinary pointer case by never setting it, and the focus clear covers the one selection that can still arm it by pointer — arrowing to a row and then clicking a different one, where `activeOption` is genuinely set at click time.

The keyboard path is untouched by the clears, for the reason you gave: `focus()` on an already-focused element dispatches no event, and Enter never moves focus out of the field.

vvlladd28 (approved):

PR #16155, opened 2026-09-16
fix: do not clear edge activity state when session moved to another node

volodymyr-babak:
## Problem

`EdgeGrpcService.onEdgeDisconnect` persists `active=false` based only on **this node's** in-memory session map. In a multi-node TB-Core cluster:

3. Node A eventually notices its dead stream, finds the session still current in its own local map, and writes `active=false`.

## Fix

Guard the write with the cluster-wide service-id cache and skip it only when **another** node is the recorded owner:

```java
if (!isOwnedByAnotherNode(edgeId)) {
    save(tenantId, edgeId, ACTIVITY_STATE, false);
}
```

The predicate deliberately does **not** reuse the existing `isOwnedByThisNode`. An absent cache entry (expired, or never written) means ownership is *unknown*, not that another node owns it — inverting through `isOwnedByThisNode` would skip the write in that case and leave a genuinely disconnected edge marked active forever. `isOwnedByAnotherNode` fails safe toward the current behaviour.

## Deliberately out of scope

- The `saveEdgeEvent` drop policy itself is untouched. Making assignment events queue while an edge is offline is a separate, platform-wide behaviour change and does not belong in an LTS patch.

## Testing

Three cases added to the existing `EdgeGrpcServiceTest`, driving the real `onEdgeDisconnect` path and asserting on the actual `saveAttributes` call rather than the predicate in isolation:

- owned by another node → inactive state **not** persisted
- owned by this node → persisted
- no cache entry → persisted (the fail-safe)

Verified red before green: with the guard reverted, only `givenEdgeReconnectedToAnotherNode_whenDisconnect_thenInactiveStateNotPersisted` fails while the two fail-safe cases still pass.

```
EdgeGrpcServiceTest:  Tests run: 11, Failures: 0, Errors: 0, Skipped: 0
Edge*Test (module):   Tests run: 78, Failures: 0, Errors: 0, Skipped: 1
```

## Follow-up

Same change is needed on `lts-4.2` (which has the service-id cache but not the `isOwnedBy*` helpers) and on the corresponding PE branches.


application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -580,6 +586,13 @@ private boolean isOwnedByThisNode(EdgeId edgeId) {
         return wrapper != null && serviceInfoProvider.getServiceId().equals(wrapper.get());
     }
 
+    // The edge's service-id cache entry points at a different node, i.e. another node is the recorded owner.
+    // An absent entry means ownership is unknown, not that another node owns it.
+    private boolean isOwnedByAnotherNode(EdgeId edgeId) {
+        TbCacheValueWrapper<String> wrapper = edgeIdServiceIdCache.get(edgeId);
+        return wrapper != null && !serviceInfoProvider.getServiceId().equals(wrapper.get());
```
volodymyr-babak: This predicate treats "cache entry names some other service id" as "that node is live and owns the edge", but nothing guarantees the recorded owner is still alive. `CACHE_SPECS_EDGE_SESSIONS_TTL` defaults to `0` in `thingsboard.yml` (persistent for the Redis impl), and the only evict point is `evictServiceIdCacheIfOwnedByThisNode`, which only the owning node ever reaches.

So consider: edge E is on A (cache=A, `active=true`); E flaps to B, which writes `active=true` and cache=B while A still holds a stale session; B is then SIGKILLed / OOM-killed, so it never evicts and never writes `active=false`; E stays down. When A finally notices its stale session, `isOwnedByAnotherNode` returns true and A skips the write, so `ACTIVITY_STATE` stays `true` forever — `onEdgeConnect` and `onEdgeDisconnect` here are the only writers of that attribute, and there's no reconciler to correct it. Pre-PR that same sequence happened to self-heal because A wrote `false`.

Since failing safe toward `false` is explicitly the design goal for the unknown-ownership case, is it worth applying the same reasoning to the dead-owner case — e.g. also requiring `wrapper.get()` to be in `partitionService.getAllServiceIds(ServiceType.TB_CORE)` (`DefaultTbClusterService.broadcastEdgeNotification` already uses that), or giving the edge sessions cache a non-zero TTL so a crashed owner's claim expires?

volodymyr-babak (commented):
## Review summary

Reviewed 2 changed files in [fix: do not clear edge activity state when session moved to another node](https://github.com/thingsboard/thingsboard/pull/16155). Left 1 comment inline.

The guard is sound for the flapping-edge case it targets and the three added tests pin it properly. The main point worth a look is that `isOwnedByAnotherNode` treats a cache entry naming another service id as "that node owns a live session", with no liveness check — if the recorded owner dies without evicting, the attribute can stay `active=true` indefinitely, which is the mirror image of the bug being fixed. Two adjacent race windows in `onEdgeConnect` / `onEdgeDisconnect` are noted below.

## Additional findings

- **application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java:381** — The signal the new guard relies on is published later than the state it is meant to protect: `onEdgeConnect` does `save(ACTIVITY_STATE, true)` at line 378 and only puts the cache entry at line 381. So node B can already have submitted `active=true` while the cache still says A (or nothing), and if A's `onEdgeDisconnect` reads the cache in that window it sees "not another node" and submits `active=false` afterwards — the exact end state this PR is trying to prevent. The stale-session case the PR targets has a much wider window, so the fix does handle the common case, but publishing ownership first (moving the `edgeIdServiceIdCache.put` above the `ACTIVITY_STATE` write) would close most of the remaining hole for free. Worth noting the check-then-write still isn't atomic across nodes even then, since the two `saveAttributes` calls are async and issued from different nodes with no ordering guarantee.
- **application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java:529** — The new guard covers the cross-node variant of "stale session clobbers live state", but the same-node variant a few lines above is still open and the guard can't catch it, because there the cache legitimately points at this node. `sessions.get(edgeId)` at line 527 and `sessions.remove(edgeId)` at line 529 are not atomic: if the old session's close callback (one gRPC stream thread) reads `current == session1` and passes the id check, and `onEdgeConnect` for session2 (another gRPC stream thread) does its `sessions.put` in between, the unconditional remove drops the *new* live session2 out of the map, `destroySession(toRemove)` tears down session2's stream, and `ACTIVITY_STATE` is set to false while the edge is connected to this very node — the identical symptom, including the silent `saveEdgeEvent` drops. `sessions.remove(edgeId, current)` (returning early when it returns false) would make this branch claim-once the same way `fireDelayedDisconnectNotification`'s identity-keyed remove already does. Pre-existing rather than introduced here, but it's the same failure mode the PR is closing, so it may be worth folding in.

---

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
volodymyr-babak: Good catch — this is a regression I introduced, and the scenario is exactly right. Verified the premises: `CACHE_SPECS_EDGE_SESSIONS_TTL` is `0` (thingsboard.yml:733), and `evictServiceIdCacheIfOwnedByThisNode` is the only evict point in the codebase, reachable only by the owning node. So a SIGKILLed owner's claim is permanent, and since `onEdgeConnect`/`onEdgeDisconnect` are the only writers of `ACTIVITY_STATE`, nothing would ever correct it.

Fixed by requiring the recorded owner to still be a live TB-Core service, per your first suggestion:

```java
return partitionService.getAllServiceIds(ServiceType.TB_CORE).contains(ownerServiceId);
```

Preferred that over a cache TTL because the TTL would also change routing behaviour in `DefaultTbClusterService.processEdgeNotification`, which falls back to a broadcast when the entry is missing — a wider blast radius than this fix warrants on an LTS branch.

One residual trade-off worth naming: during a discovery blip the live owner could briefly be absent from `getAllServiceIds`, and we would then clear the flag — i.e. the original bug, transiently. I took that deliberately, on the grounds that it requires discovery instability *and* a concurrent stale-session disconnect, whereas the dead-owner case needs only a single `kill -9`. Failing toward `false` also keeps the pre-PR outcome in every ambiguous case, which was the design goal you quoted.

Covered by `givenOwnerNodeNoLongerInCluster_whenDisconnect_thenInactiveStateIsPersisted`; confirmed it fails when the liveness check is removed.

volodymyr-babak:
Thanks — all three findings were valid and are now addressed in bcf914d. Verified each against the code before acting rather than taking them on faith; notes below.

### 1. Dead owner honoured indefinitely (inline)

Valid, and a regression introduced by this PR. Answered in detail on the inline thread. Now requires the recorded owner to still be in `partitionService.getAllServiceIds(ServiceType.TB_CORE)`; chose that over a cache TTL because a TTL would also alter the routing fallback in `DefaultTbClusterService.processEdgeNotification`.

### 2. Ownership published after the state it protects (line 381)

Valid. Confirmed the ordering — `save(ACTIVITY_STATE, true)` at 378, `edgeIdServiceIdCache.put` at 381 — so the guard could read a stale owner in exactly the window it is meant to cover. The `put` now happens first:

```java
save(tenantId, edgeId, ACTIVITY_STATE, true);
```

Safe to move earlier because `sessions.put` already happened above it, so this node is a valid routing target by that point. Agreed the check-then-write still is not atomic across nodes — two async `saveAttributes` from different nodes have no ordering guarantee. This narrows the window rather than closing it, and a genuine fix would need the write to be conditional at the storage layer. Not attempting that on an LTS branch.

### 3. Same-node `sessions.get` / `sessions.remove` race (line 529)

Valid, and I have fixed it here despite it being pre-existing. It produces the identical customer-visible symptom — live session torn down, `active=false` while connected, assignment events silently dropped — and the cross-node guard provably cannot catch it, since the cache legitimately points at this node. Leaving it would have meant shipping a fix whose stated purpose is "a stale session must not clobber live state" while one variant of exactly that remained open.

Used the claim-once idiom already in the file:

```java
if (current != null && current.getSessionId().equals(sessionId) && sessions.remove(edgeId, current)) {
```

Losing the CAS falls through to the stale-session branch, which removes by session id and tears down the stale session — the correct handling for a callback that no longer owns anything.

### Tests

Two added, both verified red against their own fix before going green:

- `givenOwnerNodeNoLongerInCluster_whenDisconnect_thenInactiveStateIsPersisted`
- `givenSessionReplacedDuringDisconnect_whenStaleCallbackRuns_thenLiveSessionKept` — drives the interleaving deterministically by swapping the map entry from within the `getSessionId()` stub, so no concurrency harness is needed

```
EdgeGrpcServiceTest:  Tests run: 13, Failures: 0, Errors: 0, Skipped: 0
Edge*Test (module):   Tests run: 80, Failures: 0, Errors: 0, Skipped: 1
```

volodymyr-babak:

Rebuilt so the fix propagates along the real merge paths instead of four independent implementations. Each downstream branch carries the upstream fix commit as a genuine ancestor:

```
CE lts-4.2  thingsboard#16156          <- the change itself
  |
  +-> CE lts-4.3  thingsboard#16155     (merge + 4.3 adaptation)
  |     |
  |     +-> PE lts-4.3  thingsboard-pe#5130   (clean merge from CE)
  |
  +-> PE lts-4.2  thingsboard-pe#5131   (clean merge from CE)
```

**Please merge in this order:** CE `lts-4.2` -> CE `lts-4.3` -> PE `lts-4.2` -> PE `lts-4.3`.

Verified after rebuilding: CE `lts-4.2` is already an ancestor of CE `lts-4.3`, and the CE branches are already ancestors of their PE counterparts, so each of those merges is a fast-forward no-op. PE `lts-4.2` -> PE `lts-4.3` was trial-merged and resolves with zero conflicts.

The earlier revisions of these branches were independent patches per line, which conflicted in `EdgeGrpcService.java` and add/add-conflicted on `EdgeGrpcServiceTest.java` the moment `lts-4.2` was merged into `lts-4.3`. Force-pushed; file contents are byte-identical to what was reviewed.

volodymyr-babak:

| # | Finding | Action |
|---|---|---|

```
                    \-> PE lts-4.2 (pe#5131)
```

### Tests

| Branch | Tests |
|---|---|
| CE lts-4.3 | 17 |
| PE lts-4.2 | 11 |
| PE lts-4.3 | 17 |
| CE / PE lts-4.4 | 9 each |

PR #16156, opened 2026-09-17
Do not clear edge activity state when session moved to another node

volodymyr-babak:
`lts-4.2` backport of thingsboard/thingsboard#16155. Please review that PR first for the rationale; this branch needed more than a cherry-pick, as noted below.

## Problem

`onEdgeDisconnect` persists `active=false` based only on **this node's** in-memory session map. In a multi-node TB-Core cluster:

3. Node A eventually notices its dead stream, finds the session still current in its own local map, and writes `active=false`.

## Why this is not a straight cherry-pick

`lts-4.3` gained a disconnect-notification debounce, and with it `isOwnedByThisNode` / `isConnectedClusterWide` / `evictServiceIdCacheIfOwnedByThisNode`. None of that exists on `lts-4.2`, which has the service-id cache but no helpers around it. So this PR introduces the ownership helpers rather than reusing them, and there is no `EdgeGrpcServiceTest` on this branch — it is added here.

It also carries **one change the 4.3 PR did not need**: on `lts-4.2` the service-id cache eviction at the end of `onEdgeDisconnect` is unconditional. Left as-is, a stale callback would skip the `active=false` write correctly and then wipe the live owner's claim on its way out, defeating the guard for any later disconnect. The eviction is therefore guarded the same way 4.3 already guards it.

## Fix

1. **Guard the write.** Skip `active=false` only when another node is the recorded owner *and* that node is still a live `TB_CORE` service. The predicate deliberately does not invert `isOwnedByThisNode`: an absent cache entry means ownership is *unknown*, not claimed elsewhere, and inverting would skip the write in that case and leave a genuinely disconnected edge marked active forever. The liveness requirement covers the mirror-image failure — the edge sessions cache has no TTL by default (`CACHE_SPECS_EDGE_SESSIONS_TTL=0`) and only the owner ever evicts, so a node killed without evicting would otherwise hold its claim indefinitely.
2. **Publish ownership first.** `onEdgeConnect` claimed the cache entry *after* writing `active=true`, so a disconnect callback on the node the edge just left could read a stale owner and clear a state that was already set.
3. **Guard the eviction**, per the paragraph above.
4. **Atomic session removal.** `sessions.get()` then `sessions.remove()` were not atomic. A session replaced between the two was dropped from the map, torn down, and marked inactive by the stale callback even though it was the live one — the same symptom on a single node, where the ownership guard cannot help because the cache legitimately points here.

## Deliberately out of scope

- The `saveEdgeEvent` drop policy itself is untouched. Making assignment events queue while an edge is offline is a platform-wide behaviour change and does not belong in an LTS patch.
- No attempt to backport the 4.3 disconnect-notification debounce.

## Testing

New `EdgeGrpcServiceTest` drives the real `onEdgeDisconnect` path and asserts on the actual `saveAttributes` / `evict` calls rather than on the predicates in isolation:

- owned by another live node → inactive state **not** persisted; cache entry kept
- owned by this node → persisted; cache entry evicted
- no cache entry → persisted (fail-safe)
- recorded owner no longer in the cluster → persisted (dead-owner fail-safe)
- session replaced mid-disconnect → live session kept, state not cleared

Verified red before green: with the guards reverted, the discriminating assertions fail.

```
EdgeGrpcServiceTest:  Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
Edge*Test (module):   Tests run: 54, Failures: 0, Errors: 0, Skipped: 1
```


volodymyr-babak:

Rebuilt so the fix propagates along the real merge paths instead of four independent implementations. Each downstream branch carries the upstream fix commit as a genuine ancestor:

```
CE lts-4.2  thingsboard#16156          <- the change itself
  |
  +-> CE lts-4.3  thingsboard#16155     (merge + 4.3 adaptation)
  |     |
  |     +-> PE lts-4.3  thingsboard-pe#5130   (clean merge from CE)
  |
  +-> PE lts-4.2  thingsboard-pe#5131   (clean merge from CE)
```

**Please merge in this order:** CE `lts-4.2` -> CE `lts-4.3` -> PE `lts-4.2` -> PE `lts-4.3`.

Verified after rebuilding: CE `lts-4.2` is already an ancestor of CE `lts-4.3`, and the CE branches are already ancestors of their PE counterparts, so each of those merges is a fast-forward no-op. PE `lts-4.2` -> PE `lts-4.3` was trial-merged and resolves with zero conflicts.

The earlier revisions of these branches were independent patches per line, which conflicted in `EdgeGrpcService.java` and add/add-conflicted on `EdgeGrpcServiceTest.java` the moment `lts-4.2` was merged into `lts-4.3`. Force-pushed; file contents are byte-identical to what was reviewed.

volodymyr-babak:

| # | Finding | Action |
|---|---|---|

```
                    \-> PE lts-4.2 (pe#5131)
```

### Tests

| Branch | Tests |
|---|---|
| CE lts-4.3 | 17 |
| PE lts-4.2 | 11 |
| PE lts-4.3 | 17 |
| CE / PE lts-4.4 | 9 each |

PR #16160, opened 2026-09-17
fix: do not clear edge activity state when session moved to another node

volodymyr-babak:
## Why this PR exists

The `lts-4.3` -> `lts-4.4` merge dropped this fix. `EdgeGrpcService` had been moved from `service/edge/rpc/` to `service/edge/rpc/service/` in 4.4's refactor, so git presented the incoming change as delete-vs-modify with no rename detection, and the conflict was resolved by accepting 4.4's deletion. `EdgeGrpcServiceTest` went the same way.

The result is worse than a missed backport: both fix commits **are** ancestors of `lts-4.4` now, because `lts-4.3` was merged in wholesale. Git considers the change merged while none of its content is present, so no future `lts-4.3` -> `lts-4.4` merge will re-apply it. It has to be redone here, which is what this PR does.

Nothing in `lts-4.4` currently carries any part of it: no `isOwnedByAnotherNode`, no `PartitionService`, `save(..., ACTIVITY_STATE, false)` unguarded, the cache claim still published after the `active=true` write, and no `EdgeGrpcServiceTest` anywhere in the branch.

## Problem

`onEdgeDisconnect` persists `active=false` based only on **this node's** session holder. In a multi-node TB-Core cluster:

3. Node A eventually notices its dead stream, finds the session still current in its own holder, and writes `active=false`.

## Fix

1. **Guard the write.** Skip `active=false` only when another node is the recorded owner *and* that node is still a live `TB_CORE` service. The predicate deliberately does not invert the existing `isOwnedByThisNode`: an absent cache entry means ownership is *unknown*, not claimed elsewhere, and inverting would skip the write in that case and leave a genuinely disconnected edge marked active forever. The liveness requirement covers the mirror-image failure — the edge sessions cache has no TTL by default (`CACHE_SPECS_EDGE_SESSIONS_TTL=0`) and only the owner ever evicts, so a node killed without evicting would otherwise hold its claim indefinitely.
2. **Publish ownership first.** `onEdgeConnect` claimed the cache entry *after* writing `active=true`, so a disconnect callback on the node the edge just left could read a stale owner and clear a state that was already set.
3. **Atomic session removal.** The holder lookup and removal were not atomic, so a session replaced between the two was dropped from the holder, torn down, and marked inactive by the stale callback even though it was the live one — the same symptom on a single node, where the ownership guard cannot help because the cache legitimately points here.

### New in the 4.4 port

`sessions` is now an `EdgeSessionsHolder` rather than a raw `ConcurrentMap`, so the compare-and-remove needed a home. Rather than reach through the holder, this adds `EdgeSessionsHolder.removeByEdgeIdIfCurrent(EdgeId, EdgeGrpcSessionManager)`, keeping the CAS inside the class that encapsulates the session maps. The now-dead `toRemove != null` check that followed the old unconditional `removeByEdgeId` is dropped, since `current` is already known non-null.

## Deliberately out of scope

- The `saveEdgeEvent` drop policy itself is untouched. Making assignment events queue while an edge is offline is a platform-wide behaviour change and does not belong in an LTS patch.

## Testing

`EdgeGrpcServiceTest` recreated at the new package, driving the real `onEdgeDisconnect` path against a **real** `EdgeSessionsHolder` (not a mock) so the claim-once removal is genuinely exercised, and asserting on the actual `saveAttributes` call rather than on the predicate in isolation:

- owned by another live node → inactive state **not** persisted
- owned by this node → persisted
- no cache entry → persisted (fail-safe)
- recorded owner no longer in the cluster → persisted (dead-owner fail-safe)
- session replaced mid-disconnect → live session kept, state not cleared

Verified red before green: with the guard and the CAS reverted, the two discriminating tests fail.

```
EdgeGrpcServiceTest:  Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
```

> **Note on building `lts-4.4` locally:** this branch needs JDK 25 (`<release>25</release>`), not 17. Also, the `application` module's test sources do not currently compile on either edition of `lts-4.4` for reasons unrelated to this change — CE has an ambiguous-overload error in `NotificationApiClientTest`, and PE's whole `org.thingsboard.server.client` test package fails to resolve `org.thingsboard.client.api.ThingsboardApi.*Args`. Both files are byte-identical to the branch tip here. I ran the suite above with those pre-existing failures set aside locally; nothing in this PR touches them, but they will block CI on `lts-4.4` until fixed separately.


volodymyr-babak:

These two continue the chain from the 4.2/4.3 PRs. CE `lts-4.4` is the source; PE `lts-4.4` is a clean merge of it.

```
CE lts-4.2  thingsboard#16156
  |
  +-> CE lts-4.3  thingsboard#16155
  |     |
  |     +-> PE lts-4.3  thingsboard-pe#5130
  |
  +-> PE lts-4.2  thingsboard-pe#5131

CE lts-4.4  thingsboard#16160        <- re-applied after the 4.3->4.4 merge dropped it
  |
  +-> PE lts-4.4  thingsboard-pe#5132
```

**Merge CE `lts-4.4` before PE `lts-4.4`.** Verified: CE `lts-4.4` is already an ancestor of the PE branch, so that merge is a fast-forward no-op.

These are independent of the 4.2/4.3 set — because 4.4 already treats those commits as merged, this is a fresh change against the refactored `service/edge/rpc/service/` layout rather than a propagation of them.

application/src/main/java/org/thingsboard/server/service/edge/rpc/service/EdgeGrpcService.java
```diff
@@ -226,14 +232,22 @@ private void onEdgeDisconnect(Edge edge, UUID sessionId) {
...
+                && sessions.removeByEdgeIdIfCurrent(edgeId, current)) {
+            current.onEdgeDisconnect();
+            current.destroyAndMarkAsZombieIfFailed();
             sessions.removeBySessionId(sessionId);
-            save(tenantId, edgeId, ACTIVITY_STATE, false);
+            // A newer session on another node may have already claimed this edge and set active=true.
+            // Clearing the flag here would leave the edge marked inactive while it is in fact connected,
+            // which silently drops every downlink event that is not queued for offline edges.
+            // When ownership is unknown (no cache entry) fall through and persist the inactive state.
+            if (!isOwnedByAnotherNode(edgeId)) {
```
volodymyr-babak: The guard only covers the case where the new session landed on a *different* node — it doesn't cover a replacement on this node, which is the same interleaving the new compare-and-remove was added for. Sequence: the stale callback wins the CAS at t0 and removes its own session; `current.onEdgeDisconnect()` / `destroyAndMarkAsZombieIfFailed()` then run (for the Kafka session manager this stops a consumer and is not instantaneous); meanwhile at t1 the edge reconnects to this same node, so `onEdgeConnect` does `edgeIdServiceIdCache.put(edgeId, thisNode)` and `save(ACTIVITY_STATE, true)`; at t2 the stale callback reaches this line, `isOwnedByAnotherNode` sees its own service id and returns false, and it persists `active=false` after the live session already persisted `active=true`. That is exactly the stale-flag state the PR is fixing — assignments then get dropped by `BaseEdgeProcessor.saveEdgeEvent` until a full sync.

Worth noting the asymmetry: the eviction a few lines below is already guarded with `if (!sessions.hasByEdgeId(edgeId))` for precisely this "a newer local session owns it now" case. Would it make sense to use the same check here, i.e. `if (!sessions.hasByEdgeId(edgeId) && !isOwnedByAnotherNode(edgeId))`? It's cheap, keeps the two guards consistent, and the test `givenSessionReplacedDuringDisconnect_whenStaleCallbackRuns_thenLiveSessionKept` currently only pins the ordering where the CAS loses, not this one.

application/src/main/java/org/thingsboard/server/service/edge/rpc/service/EdgeGrpcService.java
```diff
@@ -208,10 +211,13 @@ private void onEdgeConnect(EdgeId edgeId, EdgeGrpcSessionManager edgeSession) {
             }
         }
         sessions.put(edgeSession);
+        // Claim ownership before writing the active state, not after: onEdgeDisconnect on the node this edge
+        // just left reads the cache to decide whether to clear the flag, so publishing the claim first keeps
+        // it from seeing a stale owner and clearing a state we have already set.
+        edgeIdServiceIdCache.put(edgeId, serviceInfoProvider.getServiceId());
```
volodymyr-babak, line 217: The ownership claim moving ahead of the `active=true` write is load-bearing — it is what stops the previous node's `onEdgeDisconnect` from seeing a stale owner — but nothing pins it; the only thing holding the order is the comment above it. The disconnect-side guard is well covered by the new `EdgeGrpcServiceTest`, so if a future merge or a refactor that groups the three `save(...)` calls together moves this `put` back below them, all of those tests still pass and the cross-node race is silently back. That is exactly the failure mode this PR is re-applying a fix for: the `lts-4.3` -> `lts-4.4` merge dropped it and nothing failed.

Would it be worth adding an `InOrder` assertion in `EdgeGrpcServiceTest` — invoke `onEdgeConnect` the same reflective way `onEdgeDisconnect` is invoked, then verify `edgeIdServiceIdCache.put` before `tsSubService.saveAttributes` for the `active=true` entry? The existing `setUp` already has both mocks, so it should be a cheap addition.

volodymyr-babak (commented):
## Review summary

Reviewed 3 changed files in [fix: do not clear edge activity state when session moved to another node](https://github.com/thingsboard/thingsboard/pull/16160). Left 2 comment(s) inline.

The one worth looking at first is in `onEdgeDisconnect`: the new ownership guard covers a reconnect that lands on a *different* node, but not a replacement on this same node, so the stale `active=false` write this PR exists to prevent can still land in that interleaving. The second comment is about the `onEdgeConnect` ordering change being unpinned by any test.

## Additional findings

- **application/src/main/java/org/thingsboard/server/service/edge/rpc/session/DefaultZombieSessionCleanupService.java:63** — This is the other caller of the unconditional `removeByEdgeId` and it has the same lookup-then-remove hazard the new `removeByEdgeIdIfCurrent` was introduced to close. The zombie list is built from a weakly-consistent snapshot of `sessions.values()`, then each entry is destroyed (`toRemove.destroy()`, which for the Kafka manager stops a consumer and can take a while) and only afterwards removed by edge id. If the edge reconnects in that gap, `onEdgeConnect` has already torn down the old session, removed it by session id and put the new live session under the same edge id — and this then removes the live one. The live session stays in `sessionsById`, so it isn't torn down and `forEach` keeps iterating it, but `sessions.getByEdgeId` now returns null for a connected edge, so `onEdgeEventUpdate`, `onEdgeHighPriorityEvent`, `updateEdge` and `startSyncProcess` all silently no-op until the next reconnect. Since this PR already adds the compare-and-remove primitive, switching this callback to `removeByEdgeIdIfCurrent(edgeId, toRemove)` would close it in one line. (Pre-existing, not introduced here, but it undercuts the invariant the PR is establishing.)

---

application/src/main/java/org/thingsboard/server/service/edge/rpc/service/EdgeGrpcService.java
volodymyr-babak: Confirmed and fixed. This was a real gap: the guard only covered the cross-node case, and your sequence is exactly right — the stale callback wins the CAS, `destroyAndMarkAsZombieIfFailed()` takes its time, the edge reconnects to this same node, and `isOwnedByAnotherNode` then sees our own service id and returns false.

Taking your suggestion, and the asymmetry you pointed at was the deciding argument — the eviction below already tests for a newer local session, so the two guards should agree:

```java
if (!sessions.hasByEdgeId(edgeId) && !isOwnedByAnotherNode(edgeId)) {
```

Pinned by `givenSessionReplacedOnThisNodeAfterRemoval_whenStaleCallbackRuns_thenNoDisconnectSideEffects`, which drives the interleaving deterministically by putting the live session into the holder from inside a `doAnswer` on `destroyAndMarkAsZombieIfFailed()`. Verified it fails with just the ownership check in place.

Also fixed on lts-4.2 and lts-4.3, since both carry the same gap, and propagated to the PE branches along the merge chain.

application/src/main/java/org/thingsboard/server/service/edge/rpc/service/EdgeGrpcService.java
volodymyr-babak: Agreed, and this one landed. The ordering was held only by a comment, and "a merge or tidy-up silently undoes the fix and no build goes red" is precisely how this PR came to exist — so an unpinned load-bearing ordering was not defensible.

Added `givenConnect_whenClaimingOwnership_thenCachePutHappensBeforeActiveStateWrite`, which drives `onEdgeConnect` reflectively the same way the disconnect tests do and asserts with `inOrder(edgeIdServiceIdCache, tsSubService)`. It needed a little more scaffolding than expected — `scheduleEdgeEventsCheck` at the tail of `onEdgeConnect` dereferences `ctx.getEdgeEventStorageSettings()` and puts the returned future into a `ConcurrentHashMap`, so the executor mock has to return a real `ScheduledFuture` rather than null. Verified it fails when the `put` is moved back below the two `save(...)` calls.

Added on all five branches.

volodymyr-babak:

| # | Finding | Action |
|---|---|---|

```
                    \-> PE lts-4.2 (pe#5131)
```

### Tests

| Branch | Tests |
|---|---|
| CE lts-4.3 | 17 |
| PE lts-4.2 | 11 |
| PE lts-4.3 | 17 |
| CE / PE lts-4.4 | 9 each |

volodymyr-babak:
Closing: for 4.4 we go with a single ThingsBoard — PE 4.4 is merged into `lts-4.4` (= master), so the CE-side PR is no longer needed. The change lives on in the PE PR: https://github.com/thingsboard/thingsboard-pe/pull/5132

The 4.2 and 4.3 PRs are unaffected and still have separate CE/PE pairs.

Issue #5131, opened 2021-08-23
[Question] RPC and HTTP integration to control devices using the Knob control


tobbelin:
**Component**

 * Rule Engine

**Description**
We have started to build a demu using the TB platform, and now trying to use the Knob control to control our devices.
Our devices are connected via a third party platform that we integrate with using http platform integration. Streaming telemetry works flawlessly. However using RPC from the knob control is a bit of a problem.
We can transmit messages created by setUpdate rpc to our devices using the rest api call in the rule engine. Problem is that we get a message in the knob control saying that the message timed out.
Preferably we would like to reply to the getValue rpc call in the rule engine adding a value from the devices attribute instead of going all the way out to the device. We stream the value anyway as part of our telemetry data.

Is there any guide how to do this, or any best practices? Do we have to make a custom integration, can we make this in the rule engine, or any other option?

Thanks

**Environment**
ThingsBoard version 3.3.0 PE
Minikube

dhaval2017:
@tobbelin 
Your device supports thingsboard RPC command??
If not then using custom widget will be the best way

Thanks

tobbelin:
It does not support anything Thingsboard specific.
I hoped we could customize rest api calls and doing so opening up to use all the features of the TB platform.
If I could only loop back the RPC requests as a response I would be able to append attribute values. Also if I could just translate the outgoing message in the rule engine to a http rest request and ack the rpc call I would be all done. :-)


> @tobbelin
> Your device supports thingsboard RPC command??
> If not then using custom widget will be the best way
> 
> Thanks

tobbelin:
> @tobbelin
> Your device supports thingsboard RPC command??
> If not then using custom widget will be the best way
> 
> Thanks

Created a simple widget. Thanks.

Issue #5130, opened 2021-08-23
[Bug] 


Sam-Hamma:
**Describe the bug**
Client is connected but can't see telemetry data.

**Your Server Environment**
* own setup
  * AWS EC2 1cpu, 1gb Ram
  * ThingsBoard Version 3.3
  * OS Name and Version ubuntu 20.04

**Your Device**

* Connectivity
  * MQTTS with self signed certificates (but also with token / username)
* Device vendor and model
  * Simulated device (node)
  * Simulated device (mosquitto_pub)

**To Reproduce**
1. Install thingsboard following [the guide](https://thingsboard.io/docs/user-guide/install/aws/) for AWS EC2.
2. Configure device (credentials, rule chain etc.)
3. Launch device
4. Can't see any errors on logs, device seems to be connected (in server logs and on device logs)

**Expected behavior**
See telemetry data.

**Screenshots**
![tcpdump (port 8883)](https://user-images.githubusercontent.com/84706100/130427250-2340d659-496a-450b-b540-095b2b8070c9.png)
![Device is connected](https://user-images.githubusercontent.com/84706100/130427398-9bfe70d8-b74a-4115-aba1-bfa54536f0c8.png)
![Postgres ts_kv table](https://user-images.githubusercontent.com/84706100/130428077-3b31bd9c-bce1-4fb7-99e9-783811253da9.png)

**Additional context**
- Using v3.3.deb package (no docker, no nginx proxy), with a local postgres install.
- Necessary ports are open
- Tried using a simple token auth (port 1883)
- Tried launching the device locally on the EC2 instance
- Tried logging _*input*_ rule chain node directly: no logs (on a local machine where everything works as expected, I can see the logs)

Sam-Hamma:
Here's the output of `tcpdump -nnSX port 1883 -i eth0` on the EC2 instance (using tokens instead of TLS: message is not encrypted).
```
12:27:27.744867 IP x.x.x.x.1883 > x.x.x.x.39360: Flags [.], ack 3974905548, win 442, options [nop,nop,TS val 921129801 ecr 2390786782], length 0
        0x0000:  4500 0034 de57 4000 4006 f5d0 ac1f 25e5  E..4.W@.@.....%.
        0x0010:  c500 cf96 075b 99c0 4b4e af63 ecec 3ecc  .....[..KN.c..>.
        0x0020:  8010 01ba 66c2 0000 0101 080a 36e7 5349  ....f.......6.SI
[...]

PR #16162, opened 2026-09-17
fix: harden edge activity-state guard - same-node replacement, disconnect side effects, zombie removal

volodymyr-babak:
Follow-up to the activity-state guard (CE #16156 / #16155, PE #5131 / #5130), which merged before these review findings were addressed. The same fixes are in the `lts-4.4` PRs (CE #16160, PE #5132); this brings 4.2 and 4.3 in line so the branches do not diverge.

All findings came from review of the original change and were verified in code before acting.

## 1. The guard missed a replacement on *this* node

The ownership check only covered a reconnect landing on a **different** node. If the replacement lands on this node while the stale callback is still tearing its own session down, `isOwnedByAnotherNode` sees our own service id, returns false, and the stale `active=false` is written anyway — exactly the state the guard exists to prevent.

The eviction a few lines below already tested for a newer local session, so the two guards now agree:

```java
if (!sessions.containsKey(edgeId) && !isOwnedByAnotherNode(edgeId)) {
```

## 2. `LAST_DISCONNECT_TIME` and `DISCONNECT_EVENT` ran unconditionally

This one overturns a call made in the original PR description, which put both "deliberately out of scope" as accurate statements about the session.

That rationale covered the diagnostic value of the timestamps and missed what the event's own payload does. `pushRuleEngineMessage` emits `{"active": false, "lastDisconnectTime": ...}` with `SCOPE=SERVER_SCOPE` plus `edgeName`/`edgeType` in the metadata — precisely the shape a **Save Attributes** node persists. The stock root rule chain routes Connect/Disconnect to `Log Other`, so a default install is unaffected, but any tenant wiring Disconnect Event → Save Attributes writes `active=false` straight back and re-creates the dropped-event behaviour through the rule engine.

An edge-scoped `lastDisconnectTime` newer than a live session's `lastConnectTime` is also wrong at the edge level, even though it is true of the session — and that contradiction is what made the original report hard to diagnose.

Both now sit inside the same guard. `scheduleDisconnectNotification` stays outside: `fireDelayedDisconnectNotification` re-verifies liveness against the local sessions and the cluster-wide cache before it fires.

## 3. The `onEdgeConnect` publish order was unpinned

Moving the cache claim ahead of the `active=true` write is half the fix, and nothing but a comment held it. A merge or a tidy-up that regrouped the `save(...)` calls would leave the whole suite green with the cross-node race silently restored — which is exactly how the `lts-4.3` → `lts-4.4` merge lost this change in the first place.

Pinned with an `inOrder(edgeIdServiceIdCache, tsSubService)` assertion over a single reflective `onEdgeConnect` call.

## 4. The dead-owner comment overclaimed

`getAllServiceIds` reflects the last discovery event, so a node restarting under the **same** service id — the normal case with a fixed `TB_SERVICE_ID` or a StatefulSet hostname — still resolves as live and its stale claim is honoured. That errs toward queueing edge events rather than dropping them, which is the safe direction, but the comment claimed the dead-owner case was covered. It now says what it does and does not cover, and that the guard narrows rather than closes the race, since the attribute is plain last-writer-wins across nodes.

Not taken: re-asserting ownership inside the save callback. It adds a write on every disconnect and a second read-then-write with the same weakness; the durable fix would be a version or owner column on the row.

## 5. The other unconditional remove-by-edge-id (pre-existing)

`cleanupZombieSessions` removed by edge id unconditionally after `destroy()`, which for the Kafka session stops a consumer and is not instantaneous. If the edge reconnected in that gap, the **live** session was dropped from `sessions` while remaining in `sessionsById` — so it was never torn down, but `getByEdgeId` returned null for a connected edge and every downlink lookup silently no-opped.

Now uses a compare-and-remove. `removeFunc` becomes a `Consumer`, since its return value was already discarded.

Outside the original diff, but it undercuts the invariant the guard establishes and is a one-liner with the primitive already in place. Happy to split it out if preferred.

## Testing

New cases: the same-node replacement, the `DISCONNECT_EVENT` suppression, its positive counterpart, and the connect ordering. Each verified red against its own fix before going green.

Worth flagging: the suppression assertion was initially **vacuous** — an unstubbed `ctx.getRuleProcessor()` NPE'd inside the try/catch before the push could happen, so `never()` passed with or without the guard. The positive counterpart test plus a lenient stub is what makes it meaningful.

```
EdgeGrpcServiceTest:  Tests run: 11, Failures: 0, Errors: 0, Skipped: 0
```


volodymyr-babak:

Built along the real merge paths, so each downstream branch carries the upstream commit as a genuine ancestor:

```
CE lts-4.2  thingsboard#16162        <- the change itself
  |
  +-> CE lts-4.3  thingsboard#16163
  |     |
  |     +-> PE lts-4.3  thingsboard-pe#5140
  |
  +-> PE lts-4.2  thingsboard-pe#5139
```

**Merge order:** CE `lts-4.2` → CE `lts-4.3` → PE `lts-4.2` → PE `lts-4.3`.

Verified after pushing: CE `lts-4.2` is already an ancestor of CE `lts-4.3`, and both CE branches are already ancestors of their PE counterparts, so each of those merges is a fast-forward no-op.

The equivalent fixes for `lts-4.4` are in CE #16160 / PE #5132, which are still open and already carry them. 4.4 is a separate lineage — it treats the earlier commits as merged even though the refactor dropped their content.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -545,16 +544,22 @@ private void onEdgeDisconnect(Edge edge, UUID sessionId) {
...
+            if (!sessions.containsKey(edgeId) && !isOwnedByAnotherNode(edgeId)) {
                 save(tenantId, edgeId, ACTIVITY_STATE, false);
+                long lastDisconnectTs = System.currentTimeMillis();
+                save(tenantId, edgeId, LAST_DISCONNECT_TIME, lastDisconnectTs);
+                pushRuleEngineMessage(toRemove.getEdge().getTenantId(), edge, lastDisconnectTs, TbMsgType.DISCONNECT_EVENT);
             }
-            long lastDisconnectTs = System.currentTimeMillis();
-            save(tenantId, edgeId, LAST_DISCONNECT_TIME, lastDisconnectTs);
-            pushRuleEngineMessage(toRemove.getEdge().getTenantId(), edge, lastDisconnectTs, TbMsgType.DISCONNECT_EVENT);
             cancelScheduleEdgeEventsCheck(edgeId);
```
volodymyr-babak: The new same-node guard stops before `cancelScheduleEdgeEventsCheck(edgeId)`, which still runs unconditionally right here. Take the exact window this PR targets: the edge reconnects to this node while `destroySession(toRemove)` is running. By then the replacement's `onEdgeConnect` has already run `cancelScheduleEdgeEventsCheck(edgeId)` and then `scheduleEdgeEventsCheck(newSession)`, so the live session's future sits in `sessionEdgeEventChecks` under the same `edgeId`. When the stale callback gets here, it skips the attribute writes but still cancels and removes that pending check. The check loop only reschedules itself from inside the task, so nothing restarts it. The connected edge stops receiving edge events until it reconnects again. That is the same silent downlink stall this PR is closing. Could the cancel move inside the `!sessions.containsKey(edgeId)` part of the guard, or be keyed per session? `givenSessionReplacedOnThisNodeAfterRemoval_…` would catch this if it also asserted that the replacement's scheduled check isn't cancelled.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -545,16 +544,22 @@ private void onEdgeDisconnect(Edge edge, UUID sessionId) {
...
+            //   - the DISCONNECT_EVENT payload is {"active": false, ...} with SERVER_SCOPE in the metadata,
+            //     so a rule chain wiring Disconnect Event -> Save Attributes writes active=false straight
+            //     back, re-creating the dropped-event behaviour through the rule engine.
+            // Ownership being unknown (no cache entry) still counts as disconnected, so a genuinely offline
+            // edge is always marked inactive.
+            if (!sessions.containsKey(edgeId) && !isOwnedByAnotherNode(edgeId)) {
                 save(tenantId, edgeId, ACTIVITY_STATE, false);
+                long lastDisconnectTs = System.currentTimeMillis();
+                save(tenantId, edgeId, LAST_DISCONNECT_TIME, lastDisconnectTs);
+                pushRuleEngineMessage(toRemove.getEdge().getTenantId(), edge, lastDisconnectTs, TbMsgType.DISCONNECT_EVENT);
```
volodymyr-babak: The description says `scheduleDisconnectNotification` stays outside the guard because `fireDelayedDisconnectNotification` re-checks liveness, but neither method exists on lts-4.2. On this branch the EDGE_CONNECTION (disconnected) notification is fired by `ctx.getRuleProcessor().process(EdgeConnectionTrigger…)` inside `pushRuleEngineMessage`. So moving that call into the guard also suppresses the user-facing disconnect notification. For a live replacement that is probably the right outcome, but this check-then-act guard alone now decides whether the notification fires, and no later liveness check backs it up. The new `isOwnedByAnotherNode` comment explicitly accepts one case: a stale claim from a node that restarted under the same service id still resolves as live. In that case a truly offline edge now gets no DISCONNECT_EVENT and no disconnect notification, not just a stale `active` flag. Was this intended for the backport? If so, the description's "stays outside" reasoning should be updated to match what lts-4.2 actually does.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
```diff
@@ -722,7 +732,10 @@ private void pushRuleEngineMessage(TenantId tenantId, Edge edge, long ts, TbMsgT
 
     private void cleanupZombieSessions() {
         try {
-            tryToDestroyZombieSessions(getZombieSessions(sessions.values()), s -> sessions.remove(s.getEdge().getId()));
+            // Compare-and-remove: destroy() above is not instantaneous, so the edge may have reconnected and
+            // put a live session under this id in the meantime. Removing it unconditionally would leave a
+            // connected edge with no entry in 'sessions', silently no-opping every downlink lookup.
+            tryToDestroyZombieSessions(getZombieSessions(sessions.values()), s -> sessions.remove(s.getEdge().getId(), s));
```
volodymyr-babak: The switch to compare-and-remove is the one behavioural change in this PR with no test: nothing in `EdgeGrpcServiceTest` calls `cleanupZombieSessions`. `remove(key)` and `remove(key, value)` differ by one argument. A merge or tidy-up that drops `, s` would bring back the bug the comment describes (a connected edge with no `sessions` entry, so downlinks silently do nothing), and the suite would stay green. The `givenSessionReplacedOnThisNodeAfterRemoval` pattern, which swaps in a replacement from a `destroy()` answer, would carry over. Stub a zombie whose `destroy()` puts a fresh session under the same edgeId, invoke `cleanupZombieSessions` via `ReflectionTestUtils`, and assert the fresh session is still in `sessions`.

application/src/test/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcServiceTest.java
```diff
@@ -165,6 +187,82 @@ public void givenSessionReplacedDuringDisconnect_whenStaleCallbackRuns_thenLiveS
...
+        when(stale.destroy()).thenAnswer(invocation -> {
+            sessions().put(edgeId, mock(EdgeGrpcSession.class));
+            return true;
+        });
+        lenient().when(ctx.getRuleProcessor()).thenReturn(ruleProcessor);
+        sessions().put(edgeId, stale);
+
+        onEdgeDisconnect(sessionId);
+
+        verify(tsSubService, never()).saveAttributes(argThat(inactiveStateWrite()));
```
volodymyr-babak: This test's comment says none of the disconnect side effects may run: not the flag, the timestamp, or the DISCONNECT_EVENT push. But nothing here, or elsewhere in the class, checks the `LAST_DISCONNECT_TIME` write; `inactiveStateWrite()` only matches `ACTIVITY_STATE=false`. So one of the side effects this PR moves inside the guard has no test. A merge that puts `save(tenantId, edgeId, LAST_DISCONNECT_TIME, …)` back outside the `if` would still pass. Could you add a matcher for the `LAST_DISCONNECT_TIME` write? Assert `never()` in the suppression cases and `times(1)` in the genuine-disconnect counterpart, so the check can actually fail.

volodymyr-babak (commented):
## Review summary

Reviewed 2 changed files in [fix: harden edge activity-state guard - same-node replacement, disconnect side effects, zombie removal](https://github.com/thingsboard/thingsboard/pull/16162). Left 4 comment(s) inline.

The one to look at first: `EdgeGrpcService.onEdgeDisconnect` still calls `cancelScheduleEdgeEventsCheck(edgeId)` unconditionally after the new guard. In the same-node replacement window this PR targets, that cancels the *live* session's edge-events check, and nothing restarts it, so the reconnected edge stops receiving events. Separately, on lts-4.2 the disconnect notification is fired from inside `pushRuleEngineMessage`, so moving that call into the guard also suppresses the notification. That contradicts the description's "`scheduleDisconnectNotification` stays outside" rationale, which only applies to lts-4.3+. Two of the moved behaviours (zombie compare-and-remove, the `LAST_DISCONNECT_TIME` write) have no test pinning them.

---

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
volodymyr-babak: Agreed, and this is the important one — thanks. Confirmed the whole path: `sessionEdgeEventChecks` is keyed by `edgeId`, `onEdgeConnect` puts the replacement's future under that same key, and the loop only ever reschedules from *inside* the task, so a cancelled task is gone for good.

Fixed by gating the cancel on the same "no live session on this node" answer the guard already computed:

```java
boolean replacedOnThisNode = sessions.containsKey(edgeId);
if (!replacedOnThisNode && !isOwnedByAnotherNode(edgeId)) {
    ... disconnect side effects ...
}
if (!replacedOnThisNode) {
    cancelScheduleEdgeEventsCheck(edgeId);
}
```

Deliberately `!replacedOnThisNode` and not the full guard: the check is node-local, so when another node owns the edge there is nothing live here to protect and the cancel should still run.

Two tests added: `givenSessionReplacedOnThisNode_whenStaleCallbackRuns_thenReplacementEdgeEventsCheckNotCancelled` (asserts the replacement's future is neither cancelled nor removed from the map) and `givenEdgeGenuinelyDisconnected_whenDisconnect_thenEdgeEventsCheckCancelled` as the positive counterpart, so the first one can't pass vacuously. Both verified red against the previous code.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
volodymyr-babak: You're right on the facts, and the PR description was wrong for this branch — corrected.

`fireDelayedDisconnectNotification` / `scheduleDisconnectNotification` don't exist on lts-4.2; here the EDGE_CONNECTION (disconnected) notification is raised by `ctx.getRuleProcessor().process(EdgeConnectionTrigger…)` inside `pushRuleEngineMessage`, so moving that call into the guard suppresses the notification along with the event.

Keeping that behaviour, intentionally: a connected edge must not produce a "disconnected" notification, and unlike lts-4.3+ this branch has no delayed re-verify that could withdraw one after the fact — so suppressing at the decision point is the only place it can be done.

On the residual case you name — a stale claim from a node that restarted under the same service id. It needs the cache to name a *different, live* node while this node holds the session that just dropped, and `onEdgeConnect` re-claims the entry on every connect, so the path that would produce it self-heals as soon as the edge attaches anywhere. It stays a real (if narrow) gap; the trade is a rare missed notification against a routine false one, and the comment in the code now says so explicitly rather than leaving it implied.

application/src/main/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcService.java
volodymyr-babak: Agreed — it was the one behavioural change with nothing pinning it. Added two tests driving the real `cleanupZombieSessions` through `ReflectionTestUtils`:

- `givenZombieSessionReplacedDuringDestroy_whenCleanupRuns_thenReplacementKept` — a `KafkaEdgeGrpcSession` stubbed to satisfy `isKafkaSessionAndZombie` (disconnected, consumer present and not stopped) whose `destroy()` swaps a fresh session in under the same `edgeId`; asserts the fresh session survives.
- `givenZombieSessionStillCurrent_whenCleanupRuns_thenItIsRemoved` — the counterpart, so the compare-and-remove can't be "fixed" by simply never removing anything.

Verified red: reverting `s -> sessions.remove(s.getEdge().getId(), s)` to the single-argument `remove(key)` fails the first test.

application/src/test/java/org/thingsboard/server/service/edge/rpc/EdgeGrpcServiceTest.java
volodymyr-babak: Agreed — the comment claimed more than the assertions checked. Added a `lastDisconnectTimeWrite()` matcher and wired it in alongside `inactiveStateWrite()`:

- `never()` in the suppression cases (`givenEdgeReconnectedToAnotherNode_…`, `givenSessionReplacedOnThisNodeAfterRemoval_…`)
- `times(1)` in the persist cases (`givenEdgeOwnedByThisNode_…`, `givenNoCacheEntry_…`)

Verified red by moving the `save(tenantId, edgeId, LAST_DISCONNECT_TIME, …)` call back outside the `if`.

volodymyr-babak:
All four review findings accepted and addressed — replies are on the individual threads. Summary:

| Finding | Outcome |
|---|---|
| `cancelScheduleEdgeEventsCheck` cancels the live replacement's check | Fixed — gated on `!sessions.containsKey(edgeId)`, plus a test and its positive counterpart |
| Disconnect notification not covered by the guard | Fixed on lts-4.3+ (moved inside the guard, zero-delay test); on lts-4.2 the notification is coupled to `pushRuleEngineMessage` and already suppressed — behaviour kept, code comment and PR description corrected |
| Zombie compare-and-remove untested | Two tests added driving the real `cleanupZombieSessions` |
| `LAST_DISCONNECT_TIME` write untested | `lastDisconnectTimeWrite()` matcher added, `never()` / `times(1)` across the four guard cases |

Every new assertion was verified red against the pre-fix code before being committed.

Tests: 15/15 on the lts-4.2 pair, 23/23 on the lts-4.3 pair.

Merge chain (order of merge): CE lts-4.2 → CE lts-4.3 → PE lts-4.2 → PE lts-4.3. Each hop is already an ancestor of the next, so every downstream merge is a fast-forward.

